PR #138, opened 2013-09-12
WIP: remove Rmath dependence

simonbyrne:
I've created a branch to remove the Rmath dependence. I've done some of the easy distributions, but we're going to need some more special functions before we can finish.
- [x] Beta
- [x] Binomial
- [x] Cauchy
- [x] Chisq
- [x] FDist
- [x] Gamma
- [x] Geometric
- [x] HyperGeometric
- [x] Logistic
- [x] LogNormal
- [x] NegativeBinomial
- [ ] NoncentralBeta
- [ ] NoncentralChisq
- [ ] NoncentralF
- [ ] NoncentralT
- [x] Normal
- [x] Poisson
- [x] TDist
- [x] Uniform
- [x] Weibull

Other tasks
- [ ] `HyperGeometric` still needs a proper `rand`, but I don't have access to the [Kachitvichyanukul & Schmeiser paper](http://www.tandfonline.com/doi/abs/10.1080/00949658508810839).
- [ ] Lots of log-precision functions are still missing. In particular, it would be really useful to have the log of `rcomp` and `brcomp`, which are used in a couple of pdfs.
- [ ] Make the NSWC special functions more usable, by renaming, throwing appropriate errors, and sorting out dispatch (I've made a start on this).
- [ ] The Noncentral distributions 
- [ ] More tests.


simonbyrne:
As an update on this, Caltech/JPL have now released their software library (based on the NSWF code) under an MIT-style licence: http://netlib.org/math/

StefanKarpinski:
It's very hard not to read that as NSFW. I was wondering if that was an oblique way of saying that it has a license that means you shouldn't look at the code.

ViralBShah:
Strange that people still release stuff on `netlib` even today!

matbesancon:
@simonbyrne maybe we can fix and merge this one and open another issue for the remaining R dependencies?

StefanKarpinski:
Why not merge what's already complete and then continue to try to eliminate the Rmath dependency?

simonbyrne:
Honestly, because I don't trust past self's code.

ViralBShah:
Should we make it a GSoC project to try and get high quality tests and make this happen?

simonbyrne:
Yes. We had a special functions one this year, but it didn't pan out.

matbesancon:
Fixes #770 (I think)
@simonbyrne if we don't trust your past self code, at least we can trust your past self tests I suppose

StefanKarpinski:
Aw, I trust past @simonbyrne. He's always seems like such a stand up guy!

ararslan:
I don't know what happened since then :trollface:

matbesancon:
ok so what should be done about this one? Closing for now and waiting for a GSoC project?

StefanKarpinski:
I have to say that I feel like this PR is a perfect example of why one doesn't make a single huge PR with an enormous checklist of things to finish. That kind of PR never gets done and always goes stale. Instead, it's much better to make incremental changes and actually merge them when they're ready. Sure, they may be imperfect but they'll get fixed if they're merged and used. It's fine to use a checklist to keep track of what remains to be done but work should merged when it's ready.

ararslan:
I don't think we should close this; closed PRs are not discoverable. Folks can also cherry pick bits and pieces from this to modernize and incorporate into the repo rather than attempting to rebase it all at once.

richardreeve:
I’ve fixed quite a few of the Rfunction issues (hopefully!) as part of making random number generation multithreading-compatible in #830.

richardreeve:
I have now taken all of the RNG code from here and put it into #830. Like this PR, it doesn't fix all of the RFunctions dependencies, but that isn't the purpose of the PR.

andreasnoack:
I believe almost all of this has been implemented in https://github.com/JuliaStats/StatsFuns.jl at this point so I will go ahead and close it.

ViralBShah:
Does this mean we no longer need to worry about `Rmath_jll` maintenance, since it isn't used by Distributions.jl anymore - which probably was its largest user?

PR #144, opened 2013-09-17
Add basic location-scale family support

johnmyleswhite:
This adds an initial draft of support for location-scale families of univariate distribution. This makes it possible to rescale and relocate univariate distributions using scalar arithmetic, giving things like `d1 = Normal(0, 1) + 3` and `d2 = 3 * Uniform(0, 1) + 11`.

The basic principle is simple: unless there is a clear way in which changing the location or scale of a distribution produces another distribution of the same type, we create an ad hoc `UnivariateLocationScaleFamily` type, which maintains a separate `location` and `scale` parameter. Thus, one has two ways of writing some distributions. for example, the two distributions below are meant to be identical:
- `UnivariateLocationScaleFamily(Normal(0, 1), 2.0, 2.0)`
- `Normal(2.0, 2.0)`

But special cases are handled as needed, so that `3 * Normal(0, 1)` gives `Normal(0, 3)` rather than the less efficient `UnivariateLocationScaleFamily(Normal(0, 1), 0.0, 3.0)`.

### DIFF
--- a/run_tests.jl
+++ b/run_tests.jl
@@ -14,7 +14,8 @@ tests = [
 	"kolmogorov",
 	"edgeworth",
 	"matrix",
-	"vonmisesfisher"]
+	"vonmisesfisher",
+	"locationscale"]
 
 println("Running tests:")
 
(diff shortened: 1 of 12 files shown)


matbesancon:
Closing this for #635 
@kleinschmidt same here :)

PR #148, opened 2013-09-17
Add regularized incomplete beta and gamma functions. Use them and their auxiliary functions as much possible.

andreasnoack:
This is a big one. The three major functions are the incomplete beta and gamma functions and the inverse of the last, but these bring some auxiliary functions which are of interest by their own rights. For example, some are useful for precise calculations of pdfs. Please have a look at the headers and see if it can be used anywhere else.

I have done a lot of testing of the functions agains Rmath and the Fortran version of NSWC, but there are many possible combinations of parameter values so I cannot guarantee that all error are gone. When the new implementation differed (relatively) from the old by more than 1e-14 I compared to Mathematica and in all cases both versions were inaccurate by more that they differed from each other.

I have also written fallback quantile functions. The continuous one is just Newton's method and the discrete is a normal approximation and a search. The two quantile functions also allow for a fallback random number method. I did this because I saw that a couple of the R's quantile functions and RNGs worked through "generic" methods. Please consider the choice of initial value in the continuous quantile function. I am not sure my choice is the right thing to do.

What remains is specialised RNG's for at least binomial and Poisson. The code R uses is not free enough for our needs.

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -69,6 +69,47 @@ logccdf(d::Distribution, q::Real) = log(ccdf(d,q))
 invlogccdf(d::Distribution, lp::Real) = quantile(d, -expm1(lp))
 invlogcdf(d::Distribution, lp::Real) = quantile(d, exp(lp))
 
+function quantile(d::ContinuousUnivariateDistribution, α::Real)
+
+    if α < 0 || α > 1 return NaN end
+    if α == 0 return 0.0 end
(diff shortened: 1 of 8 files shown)


lindahua:
@andreasnoack Can you close this and make a PR to the StatsFuns package?

andreasnoack:
I'll close her, but please keep the branch. I don't know when I'll prepare the new pull request.

PR #175, opened 2014-01-24
bugfix in `posterior_canon` for NormalWishart

sbos:
It seems from normalwishart.jl that corresponding code for the distribution is based on "Conjugate Bayesian analysis of the Gaussian distribution" paper by Murphy which contains some bugs, i.e. this one. 
For a proof, one may refer to wikipedia http://en.wikipedia.org/wiki/Conjugate_prior#Continuous_distributions or just see that both `NormalWishart` and `InverseNormalWishart` unlikely have same implementations of `posterior_canon` in what's regarding `Lam`.
I would also suggest to check other code directly implementing that paper.

### DIFF
--- a/test/conjugates_mvnormal.jl
+++ b/test/conjugates_mvnormal.jl
@@ -105,7 +105,7 @@ post = posterior(pri, MvNormal, X)
 @test_approx_eq post.mu (kappa0*mu0 + n*Xbar)./(kappa0 + n)
 @test_approx_eq post.kappa kappa0 + n
 @test_approx_eq post.nu nu0 + n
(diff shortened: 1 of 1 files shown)


sbos:
I still believe it is important to fix this bug and surprised it's survived for 4 years

matbesancon:
perfect, thanks, can you. I think these features have been moved to https://github.com/JuliaStats/ConjugatePriors.jl
Sorry this took so long, could you see if the bug has been fixed there and possibly re-do the PR there?

matbesancon:
(sorry for the inconvenience)

PR #267, opened 2014-07-26
Statistical tests for NormalGamma and NormalInverseGamma

anj1:
This pull request adds `mean` and `var` for `NormalGamma` and `NormalInverseGamma`. In addition, it uses those moments to verify that `rand` is working correctly for those distributions.


matbesancon:
NormalGamma has been moved to https://github.com/JuliaStats/ConjugatePriors.jl
Re-open there if necessary

Issue #264, opened 2014-07-25
Variance of a Truncated Gaussian returns a negative number


LaurenceA:
``` julia
julia> var(Truncated(Normal(1.8, 1.2), -Inf, 0))
-3.9721955851923996
```

I am on the latest release.

PR #276, opened 2014-08-14
Truncated normal variance

LaurenceA:
Fixes Issue #264.  There are three changes, all of which help us to compute the moments of a truncated Gaussian.
First, there was a bug in the implementation of the variance for Truncated Normal moments.
Second, the moment calculations often fail because the normalizing constant has a large error.  This has been fixed by making sure that we use ccdf if cdf(d, u) and cdf(d, l) are close to 1.
Third, tests have been added, one of which fails if the second alteration is not made.


cossio:
This should be closed? Superseded by https://github.com/JuliaStats/Distributions.jl/issues/292, https://github.com/JuliaStats/Distributions.jl/pull/295

andreasnoack:
Yes. Seem right.

PR #320, opened 2014-12-01
Zero VonMises distribution outside its support

getzdan:
Zero VonMises pdf outside support.
Fix CDF to be between 0.0 and 1.0 appropriately.


andreasnoack:
@simonbyrne What do you think should be done with `VonMises`?

simonbyrne:
The easiest would be to fix the range `[-pi,pi]`.

matbesancon:
@getzdan still around? This PR has been sleeping for quite a while :)
Could you fix the conflicts and see what convention would be preferable for this one?

simonbyrne:
Also, good to add a note to the docs that users can reduce to [-pi,pi] via `rem2pi(x,RoundNearest)`.

PR #328, opened 2014-12-30
bug fix in MvTDist (replaced GenericMvNormal with MvNormal)

afniedermayer:
This adjusts MvTDist to the refactoring GenericMvNormal->MvNormal.
I also fixed a bug in the function full in utils.jl that I discovered in the process.


lindahua:
Thanks.

Issue #333, opened 2015-01-13
bug in generating random variables from multivariate student's t distribution


afniedermayer:
```julia
julia> td=MvTDist(3.,[.0,.0],PDMat([1. 0.;0. 1.]))
GenericMvTDist{PDMat}(
df: 3.0
dim: 2
zeromean: false
μ: [0.0,0.0]
Σ: PDMat(2,2x2 Array{Float64,2}:
 1.0  0.0
 0.0  1.0,Cholesky{Float64} with factor:
2x2 Triangular{Float64,Array{Float64,2},:U,false}:
 1.0  0.0
 0.0  1.0)
)


julia> rand(td)
ERROR: GenericMvNormal not defined
 in _rand! at /home/niedermayer/.julia/v0.3/Distributions/src/multivariate/mvtdist.jl:142
 in rand at /home/niedermayer/.julia/v0.3/Distributions/src/multivariates.jl:18
```

This is fixed by pull request #328

afniedermayer:
It seems that pull request #328 only fixes the typo introduced by a refactoring from GenericMvNormal to MvNormal. But it seems that the original algorithm already contained a bug. Here's an example in which I generate 1 million random draws and do not get back the original covariance matrix:

```julia
using Distributions, PDMats
td=MvTDist(3.,[0.,0.],PDMat([1. 0.;0. 1.]))
x=rand(td,1_000_000);
println(cov(td))
println(cov(x'))
```
which returns
```
[3.0 0.0
 0.0 3.0]
[0.4397361101526044 -0.0013065885945805072
 -0.0013065885945805072 1.6215411206684638]
```

For the univariate t-distribution, random number generation seems to be OK:
```julia
td1=TDist(3.)
y=[rand(td1,1_000_000) rand(td1,1_000_000)]';
println(var(td1))
println(cov(y'))
```
outputs
```
3.0
[2.9357832915809703 -0.00216158423511952
 -0.00216158423511952 2.9300057612313912]
```

kmsquire:
Closed by #334.

PR #334, opened 2015-01-14
Fixed bug in _rand! functions of MvTDist

fairbrot:
- Fixed _rand! functions so that only one Chisq random variable is sampled for each MvTDist sample (see http://journal.r-project.org/archive/2013-2/hofert.pdf for more details)
  - Removed unnecessary MvNormal distributions

### DIFF
--- a/src/multivariate/mvtdist.jl
+++ b/src/multivariate/mvtdist.jl
@@ -138,31 +138,26 @@ end
 # Sampling (for GenericMvTDist)
 
 function _rand!{T<:Real}(d::GenericMvTDist, x::DenseVector{T})
-    normdim = d.dim
-    normd = MvNormal(ZeroVector(Float64, normdim), d.Σ)
     chisqd = Chisq(d.df)
-    y = Array(Float64, d.dim)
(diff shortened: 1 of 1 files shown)


afniedermayer:
Does this fix issue #333 (see second comment)?

fairbrot:
Yes, that's right. I came across the bug in the same way as you.

I'm not why the Travis CI test failed; it doesn't appear to have anything to do with my changes.

afniedermayer:
Thanks! This also fixes the second part of the bug reported in #333. It now does what it's supposed to do:

```julia
using Distributions, PDMats
td=MvTDist(3.,[0.,0.],PDMat([1. 0.;0. 1.]))
x=rand(td,1_000_000);
println(cov(td))
println(cov(x'))
```
outputs
```
[3.0 0.0
 0.0 3.0]
[3.010441887572456 0.041348097303148854
 0.041348097303148854 2.9292425549152403]
```

I hope someone with write access can merge this, the bug has been there for more than a month now.

kmsquire:
Thanks!

Issue #343, opened 2015-02-04
Inaccurate sampling from TruncatedNormal in tail regions


jdrugo:
The current quantile-based sampling method is inaccurate for TruncatedNormal when sampling from the tails of the Normal:
```Julia
rand(TruncatedNormal(0,1,35,Inf))
```
always returns ```Inf```.
```Julia
rand(TruncatedNormal(0,1,-Inf,-35))
```
always returns ```-Inf```. This problem can be fixed by re-enabling the Robert (1995) sampler that is currently commented out.

lindahua:
I suppose this was fixed by that commit.

jdrugo:
The TruncatedNormal distribution still features this bug. My pull request that fixes it has yet to be merged.

PR #344, opened 2015-02-04
Fix JuliaStats/Distributions.jl#343

jdrugo:
Uncomments the truncated Normal sampler by Robert (1995).

### DIFF
--- a/src/truncated/normal.jl
+++ b/src/truncated/normal.jl
@@ -61,67 +61,67 @@ end
 
 ### sampling
 
-## Benchmarks doesn't seem to show that this specialized
-## sampler is faster than the generic quantile-based method
+## Use specialized sampler, as quantile-based method is inaccurate in
+## tail regions of the Normal, issue #343
 
-# function rand(d::Truncated{Normal,Continuous})
-#     d0 = d.untruncated
-#     μ = mean(d0)
-#     σ = std(d0)
-#     a = (d.lower - μ) / σ
-#     b = (d.upper - μ) / σ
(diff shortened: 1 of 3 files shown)


johnmyleswhite:
Before merging this, I'd like to understand why it was commented out before. Was it purely for performance reasons?

simonbyrne:
It appears to be this commit 8637e3abf8af5535f1470719dc11c1655fef7d4e: @lindahua ?

jiahao:
bump

jdrugo:
Even though issue #343 was closed, the issue itself is not yet fixed.

In addition to this pull request, I would suggest adding.
```
TruncatedNormal(0,1,35,Inf)
TruncatedNormal(0,1,-Inf,-35)
```
to `Distributions.jl/test/continuous_test.lst`, to make sure that the tails are tested. I did not do this, as it requires re-generating `Distributions.jl/test/continuous_test.json`.

jmxpearson:
Given that the code was commented for negligible performance changes but that this sampler is needed for correctness, any objections to merging? 

Would close #343 #493

jdrugo:
I now rebased my branch, and added a separate test for sampling the tails of TruncatedNormal, directly addressing the cases in #343.

ararslan:
LGTM

jmxpearson:
Me, too.

simonbyrne:
Thanks, sorry this languished for so long.

nalimilan:
FWIW, it would be nice to use more descriptive commit titles in the future.

PR #346, opened 2015-02-19
PoissonBinomial distribution

chrodan:
This is an implementation of the Poisson-Binomial distribution (sum of independent but not necessarily identically distributed Bernoulli random variables). 
I am not sure about the following design choices of the code, though:

1. When an instance of PoissonBinomial is created the probability mass function is computed and stored in the type. It seemed way too wasteful to compute the (entire) pmf again for each evaluation of pdf / cdf / logpdf etc.

2. Since the pmf is readily available, the distribution is basically a shifted categorical distribution. Many functionals (entropy, median, modes) are therefore implemented by creating a Categorical type and using its implementation. 

Any comments or suggestions are welcome.


StefanKarpinski:
Significant drop in test coverage – definitely to be fixed before merging.

chrodan:
sorry, I don't understand. I thought the test coverage _increased_ by 0.55%?

StefanKarpinski:
Oh, that's weird. That's not what it said in the email I got that linked here. Weird. Carry on.

src/univariate/discrete/poissonbinomial.jl
```diff
@@ -0,0 +1,113 @@
...
+    ω = 2. * pi / (n + 1)
+
+    x = Array(Complex{Float64}, n+1)
+    lmax = int(ceil(n/2))
+    x[1] = 1./(n+1)
+    for l=1:lmax
+        logz = 0.
+        argz = 0.
+        for j=1:n
+            zjl = 1 - p[j] + p[j] * cos(ω*l) + im * p[j] * sin(ω * l)
```
simonbyrne: Rather than multiply `ω` by `pi`, you can use `cospi`/`sinpi` here: it will be more accurate for values like `cos(pi / 2)`.

src/univariate/discrete/poissonbinomial.jl
```diff
@@ -0,0 +1,113 @@
...
+            logz += log(abs(zjl))
+            argz += atan2(imag(zjl), real(zjl))
```
simonbyrne: This code is essentially taking a product, but in log space: do you know why they do this, as opposed to just using complex multiplication? The paper doesn't really explain why.

src/testutils.jl
```diff
@@ -293,7 +293,7 @@ function test_range_evaluation(d::DiscreteUnivariateDistribution)
     rmin = int(islowerbounded(d) ? vmin : quantile(d, 0.001))::Int
     rmax = int(isupperbounded(d) ? vmax : quantile(d, 0.999))::Int
 
-    p0 = pdf(d, [rmin:rmax])
+    p0 = pdf(d, [rmin:rmax;])
```
simonbyrne: Any reason for this?

garborg: `[itr]` no longer collects on 0.4. The idiomatic way is now `collect(itr)`, but `[itr;]` (which I believe is meant for cases like `[a; b; itr; c;]`) works, too.

simonbyrne: ah, makes sense.

src/univariate/discrete/poissonbinomial.jl
chrodan: I assumed its for numerical stability, similar to working in log-space when dealing with products of many probabilities. I have not tested if it is actually more stable, though.

src/testutils.jl
chrodan: thanks for the hint. I changed it to collect().

chrodan:
The code now uses collect() instead of [... ;] and cospi / sinpi instead of cos/sin.

simonbyrne:
Sorry I haven't got around to merging this yet. The only problem is the test json file: this is actually generated by a [python scipt](https://github.com/JuliaStats/Distributions.jl/blob/bb1e2e3c228b72400e1a9db30080043fc7713189/test/prepdref.py), so will be lost next time it is run.

Otherwise, looks good!

chrodan:
I removed the manually added parts from the json and instead do the tests manually in a separate julia file.

simonbyrne:
Great, thanks!

PR #351, opened 2015-03-09
[WIP] Fix mean(NoncentralT) when df is Inf

johnmyleswhite:
This changes the definition of mean(NoncentralT) so that a non-central T distribution with zero non-centrality parameter produces the same mean as the central T distribution would.

In practice, this reduces to changing the code to pass the following tests:

```
using Distributions
using Base.Test

@test mean(TDist(Inf)) == 0.0

@test mean(NoncentralT(Inf, 0)) == 0.0
```

### DIFF
--- a/src/univariate/continuous/noncentralt.jl
+++ b/src/univariate/continuous/noncentralt.jl
@@ -11,7 +11,18 @@ end
 
 @distr_support NoncentralT -Inf Inf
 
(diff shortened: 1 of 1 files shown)


johnmyleswhite:
Any objections to this?

simonbyrne:
No, I was trying to figure out the problem with the test, but it seems to be unrelated.

PR #356, opened 2015-03-16
tvpack.jl: Fix loop synthax

julian-gehring:
The existing syntax breaks current julia-0.4* when trying to load the package:

    using Distributions

yields the error

    ERROR: LoadError: LoadError: syntax: missing separator in tuple
    in include at /usr/local/Cellar/julia/HEAD/lib/julia/sys.dylib
    in include_from_node1 at /usr/local/Cellar/julia/HEAD/lib/julia/sys.dylib
    in include at /usr/local/Cellar/julia/HEAD/lib/julia/sys.dylib
    in include_from_node1 at /usr/local/Cellar/julia/HEAD/lib/julia/sys.dylib
    in reload_path at /usr/local/Cellar/julia/HEAD/lib/julia/sys.dylib
    in _require at /usr/local/Cellar/julia/HEAD/lib/julia/sys.dylib
    in require at /usr/local/Cellar/julia/HEAD/lib/julia/sys.dylib
    while loading /Users/julian/.julia/v0.4/Distributions/src/tvpack.jl, in expression starting on line 379
    while loading /Users/julian/.julia/v0.4/Distributions/src/Distributions.jl, in expression starting on line 245

Removing the obsolete parentheses fixes this.

### DIFF
--- a/src/tvpack.jl
+++ b/src/tvpack.jl
@@ -376,7 +376,7 @@ function bvnuppercdf(dh::Float64, dk::Float64, r::Float64)
 	      	c = (4.0 - hk) * 0.125
 	      	d = (12.0 - hk) * 0.0625
 	      	asr = -(bs / as + hk) * 0.5
-	      	if ( asr .gt. -100 ) 
+	      	if asr .gt. -100
 	      		bvn = a * exp(asr) * (1.0 - c * (bs - as) * (1.0 - d * bs / 5.0) / 3.0 + c * d * as * as / 5.0)
 	      	end
 	      	if -hk < 100


kmsquire:
Actually, I'm surprised that this parsed at all in Julia.  `.gt.` is Fortran, and it looks like @andreasnoack simply forgot to change it to `>`.

(It would also seem that this code path hasn't ever been tested, since that change is from Jan 2013!)

kmsquire:
See #357.

andreasnoack:
Yes it's really odd. @julian-gehring did you only call `using Distributions`? It would be great with some tests of these paths.

wildart:
@andreasnoack Yes, it crashed on ``using Distributions``.

julian-gehring:
@andreasnoack  As @wildart already pointed out, `using Distributions` with recent julia devels causes the package loading error. Interestingly, removing the parenthesis avoids the error, although the problem is the `.gt.` here.

PR #357, opened 2015-03-16
tvtcdf: cleanup and tests

wildart:


kmsquire:
Thanks!  Would you be able to add a quick test that exercised this code path?

wildart:
It crashed when `using Distributions` was called. I guess something happened in nightly build.

kmsquire:
Thanks!  @andreasnoack should be the one to pull the trigger here.

andreasnoack:
@wildart Thanks. I'm still confused why this has only surfaced now.

julian-gehring:
Thanks for the quick fix! Can 50d4a9c590660a41b0edf164dda48c3c9ee8d17d be tagged as a new version?

andreasnoack:
Yes. Since distributions actually does load right now it is probably a good idea to tag a version. I'll do that.

simonbyrne:
Actually, can we wait until the tests pass? We should also tag a new StatsBase, and bump the requirements.

andreasnoack:
Yes that is the right way to proceed. I think the deprecation warnings are making tests hang. There has been something weird things going on with `Tasks` lately. I'll update StatsBase with `@compat`s to get rid of the warnings and hopefully that will make the tests pass.

andreasnoack:
@simonbyrne Oh. You just did that.

simonbyrne:
Yep, I'm just about to open a PR.

simonbyrne:
I can update the REQUIRE and tag Distributions.jl as well.

andreasnoack:
Great. Then I'll just return to other duties.

simonbyrne:
Okay, should be done now, sorry about the delay.

PR #359, opened 2015-03-26
add continuous power law distribution

afternone:
add continuous power law distribution, implement find_xmin, fit_mle, pvalue


johnmyleswhite:
This is cool. A few suggestions:

* It's not obvious to me we want to add Optim as a dependency for a single distribution. Is there no way to hardcode a custom MLE fitting routine?
* You've reverted some of my recent work on NoncentralT. You probably want to make sure you rebase against the most recent master.
* I don't think we want to either define or export `pvalue` here since that clobbers the HypothesisTests package.

simonbyrne:
What's the difference between PowerLaw and a Pareto distribution?

PR #361, opened 2015-04-21
Use `ifelse` for faster `rand(Laplace())`

kshramt:
Elapsed times for 400 trials of 2,000,000 iterations:

p

```julia
import Distributions


function go(r, n)
    for _ in 1:n
        rand(r)
    end
end


function main()
    r = Distributions.Laplace()
    n = 2000000
    rand(r)
    for _ in 1:400
        @time go(r, n)
    end
end

main()
```

### DIFF
--- a/src/univariate/continuous/laplace.jl
+++ b/src/univariate/continuous/laplace.jl
@@ -74,10 +74,7 @@ end
 
 #### Sampling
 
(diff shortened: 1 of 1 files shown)


lindahua:
Thanks!

PR #363, opened 2015-04-21
Fix fit(::Cauchy, x)

jiahao:
Due to an apparent change in the output of iqr(x), this fit method has
been broken for awhile and threw a BoundsError because it expected iqr()
to return two numbers.

This PR restores the functionality of fit(::Cauchy, x), by calling the
quantile function directly. Doing so has the added benefit of getting
the median in the same computation. I've also generalized this fit
method to accept as input x::AbstractArray, not just x::Array.
Finally, I've also added a simple test for this fit method.

### DIFF
--- a/src/univariate/continuous/cauchy.jl
+++ b/src/univariate/continuous/cauchy.jl
@@ -69,7 +69,7 @@ cf(d::Cauchy, t::Real) = exp(im * (t * d.μ) - d.β * abs(t))
 #### Fitting
 
 # Note: this is not a Maximum Likelihood estimator
-function fit{T <: Real}(::Type{Cauchy}, x::Array{T})
-    l, u = iqr(x)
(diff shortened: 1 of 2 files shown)


lindahua:
Travis is complaining. May you take a look?

jiahao:
Oops, I checked in a typo.

IainNZ:
Someone with more advanced powers than I might want to disable Coveralls comments

andreasnoack:
@IainNZ Can you explain how to do that?

IainNZ:
On this page: https://coveralls.io/r/JuliaStats/Distributions.jl
Should be checkbox and a save button at the bottom - not sure what permission level is necessary, but I'm sure you've got it.

IainNZ:
(make sure you are logged in!)

andreasnoack:
Thanks. Done.

PR #368, opened 2015-04-30
removed deprecated inf and nan

zenna:
I was getting many warnings when using `quantile` due to

Inf(Float32/64)
nan(Float32/64)

### DIFF
--- a/src/specialfuns.jl
+++ b/src/specialfuns.jl
@@ -116,16 +116,16 @@ for (fn,arg) in ((:Φinv,:p),(:logΦinv,:logp))
             else
                 if $(fn == :Φinv)
                     if p <= 0f0
-                        return p == 0f0 ? -inf(Float32) : nan(Float32)
+                        return p == 0f0 ? -Inf32 : NaN32
                     elseif p >= 1f0 
-                        return p == 1f0 ? inf(Float32) : nan(Float32)
+                        return p == 1f0 ? inf(Float32) : NaN32
                     end
                     r = sqrt(q < 0f0 ? -log(p) : -log1p(-p))
                 else
                     if logp == -Inf
-                        return -inf(Float32)
+                        return -Inf32
                     elseif logp >= 0f0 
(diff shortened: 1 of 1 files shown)


src/specialfuns.jl
```diff
@@ -116,16 +116,16 @@ for (fn,arg) in ((:Φinv,:p),(:logΦinv,:logp))
             else
                 if $(fn == :Φinv)
                     if p <= 0f0
-                        return p == 0f0 ? -inf(Float32) : nan(Float32)
+                        return p == 0f0 ? -Inf32 : NaN32
                     elseif p >= 1f0 
-                        return p == 1f0 ? inf(Float32) : nan(Float32)
+                        return p == 1f0 ? inf(Float32) : NaN32
```
simonster: `inf(Float32) -> Inf32` here as well?

src/specialfuns.jl
```diff
@@ -116,16 +116,16 @@ for (fn,arg) in ((:Φinv,:p),(:logΦinv,:logp))
...
                     end
                     r = sqrt(q < 0f0 ? -log(p) : -log1p(-p))
                 else
                     if logp == -Inf
-                        return -inf(Float32)
+                        return -Inf32
                     elseif logp >= 0f0 
-                        return logp == 0f0 ? inf(Float32) : nan(Float32)
+                        return logp == 0f0 ? inf(Float32) : NaN32
```
simonster: and here

zenna:
This isn't failing because of me, I don't think.  Some library is using Range1 which it seems was recently removed from Julia.

simonster:
Merged. But we should really fix that Range1 issue...

simonster:
This is actually passing for me locally, so I suspect that one of the dependencies needs to be bumped in METADATA.

zenna:
It might be ArrayViews

simonster:
I think you're right, or at least ArrayViews definitely needed to be bumped, and I just did that. We'll see if it builds now.

simonster:
Also StatsBase.

PR #370, opened 2015-05-11
Add Maxwell-Boltzmann distribution.

awpxii:
I based it on the Rayleigh and Chi distributions. The sampling, cdf, and logcdf rely on the Chisq distribution. I've tested it against the existing Chi distribution and the Scipy implementation of Maxwell. The tests are here: https://gist.github.com/awpxii/c2d2724ce7ca5f961139

### DIFF
--- a/src/constants.jl
+++ b/src/constants.jl
@@ -15,3 +15,5 @@ import Base.@math_const
 @math_const loghalf -0.6931471805599453094 log(big(0.5))
 
 @math_const sqrthalfπ 1.2533141373155002512 sqrt(big(0.5)*π)
+@math_const sqrt2onπ 0.7978845608028653559 sqrt(big(2.)/π)
+@math_const logsqrt2onπ -0.225791352644727432 log(sqrt(big(2.)/π))
--- a/src/univariate/continuous/maxwellboltzmann.jl
+++ b/src/univariate/continuous/maxwellboltzmann.jl
@@ -1,0 +1,67 @@
+immutable MaxwellBoltzmann <: ContinuousUnivariateDistribution
+    a::Float64 # Scale factor
+    chisqd::Chisq
+
+    function MaxwellBoltzmann(a::Real)
+        a > zero(a) || error("MaxwellBoltzmann: a must be positive")
+        @compat new(Float64(a), Chisq(3.0))
+    end
+
+    function MaxwellBoltzmann(m::Real, T::Real) # mass and temperature of particles
+        m > zero(m) && T > zero(T) || error("MaxwellBoltzmann: m and T must be positive")
+        @compat new(Float64(sqrt(1.3806488e-23*T/m)), Chisq(3.0))
+    end
+    MaxwellBoltzmann() = new(1.0,Chisq(3.0))
+end
+
+@distr_support MaxwellBoltzmann 0.0 Inf
+
+#### Parameters
+
+scale(d::MaxwellBoltzmann) = d.a
+params(d::MaxwellBoltzmann) = (d.a,)
+
+#### Statistics
+
+mean(d::MaxwellBoltzmann) = 1.5957691216057308 * d.a # 2*sqrt(2/π)
+median(d::MaxwellBoltzmann) = 1.5381722544550522 * d.a # quantile(d, 0.5)
+mode(d::MaxwellBoltzmann) = sqrt2 * d.a # sqrt(2) = 1.4142135623730951
+
+var(d::MaxwellBoltzmann) = 0.4535209105296745 * d.a^2 # (3.0π - 8.0)/π
+std(d::MaxwellBoltzmann) = 0.6734396116428515 * d.a
+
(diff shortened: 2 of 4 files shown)


src/constants.jl
```diff
@@ -15,3 +15,5 @@ import Base.@math_const
 @math_const loghalf -0.6931471805599453094 log(big(0.5))
 
 @math_const sqrthalfπ 1.2533141373155002512 sqrt(big(0.5)*π)
+@math_const sqrt2onπ 0.7978845608028653559 sqrt(big(2.)/π)
```
johnmyleswhite: Maybe call this sqrt2overpi? On pi sounds a bit strange to an American English speaker.

johnmyleswhite:
I think this generally looks good. My one concern is to minimize the usage of magic numbers, which makes the code seem hard to reason about. I would also be interested to know if the code is numerically stable given the presence of a `1e-23` term.

src/univariate/continuous/maxwellboltzmann.jl
```diff
@@ -0,0 +1,61 @@
...
+std(d::MaxwellBoltzmann) = 0.6734396116428515 * d.a
+
+skewness(d::MaxwellBoltzmann) = 0.4856928280495921 # 2.0 * sqrt(2.0) * (16.0 - 5.0π)/(3π - 8)^(3/2)
+kurtosis(d::MaxwellBoltzmann) = 0.10816384281628826 # 4.0 * (-96.0 + 40.0π - 3π^2)/(3.0π - 8.0)^2
+
+entropy(d::MaxwellBoltzmann) = 0.9961541981062054 + log(d.a) # log(sqrt(2π)) + γ - 0.5
+
+#### Evaluation
+
+pdf(d::MaxwellBoltzmann, x::Float64) = exp(logpdf(d, x))
```
simonbyrne: Is this better than appropriately modifying the `pdf` of a `Chisq`?

src/univariate/continuous/maxwellboltzmann.jl
```diff
@@ -0,0 +1,61 @@
...
+@distr_support MaxwellBoltzmann 0.0 Inf
+
+#### Parameters
+
+scale(d::MaxwellBoltzmann) = d.a
+params(d::MaxwellBoltzmann) = (d.a,)
+
+#### Statistics
+
+mean(d::MaxwellBoltzmann) = 1.5957691216057308 * d.a # 2*sqrt(2/π)
```
andreasnoack: No reason to write out the computed value since these will be computed at compile time.

src/univariate/continuous/maxwellboltzmann.jl
```diff
@@ -0,0 +1,61 @@
...
+median(d::MaxwellBoltzmann) = 1.5381722544550522 * d.a # quantile(d, 0.5)
```
andreasnoack: `quantile` is probably too hard for the compiler to handle so here the value is required if speed is a high priority for the `median` function.

src/univariate/continuous/maxwellboltzmann.jl
```diff
@@ -0,0 +1,61 @@
...
+mode(d::MaxwellBoltzmann) = sqrt2 * d.a # sqrt(2) = 1.4142135623730951
```
andreasnoack: `sqrt(2)` is sufficient.

src/univariate/continuous/maxwellboltzmann.jl
```diff
@@ -0,0 +1,61 @@
...
+
+var(d::MaxwellBoltzmann) = 0.4535209105296745 * d.a^2 # (3.0π - 8.0)/π
+std(d::MaxwellBoltzmann) = 0.6734396116428515 * d.a
+
+skewness(d::MaxwellBoltzmann) = 0.4856928280495921 # 2.0 * sqrt(2.0) * (16.0 - 5.0π)/(3π - 8)^(3/2)
```
andreasnoack: If you avoid the floating point power here it will also get folded, i.e. `2.0 * sqrt(2.0) * (16.0 - 5.0π)/sqrt(3π - 8)^3`

src/univariate/continuous/maxwellboltzmann.jl
```diff
@@ -0,0 +1,61 @@
...
+kurtosis(d::MaxwellBoltzmann) = 0.10816384281628826 # 4.0 * (-96.0 + 40.0π - 3π^2)/(3.0π - 8.0)^2
+
+entropy(d::MaxwellBoltzmann) = 0.9961541981062054 + log(d.a) # log(sqrt(2π)) + γ - 0.5
```
andreasnoack: Unfortunately, we cannot fold `log` yet so here the value is also required (if this function really has to be that fast).

matbesancon:
@awpxii there seems to be some work left before merging this + some conflicts to fix. What's the status here?

bojohnson02:
So I've been looking at adding the Maxwell-Boltzmann distribution to Distributions.jl. It looks like this has stalled a bit and I'd like to help out. Is there a way for me to pick this up? Or should I just start a new pull request?

matbesancon:
this sadly looks stall indeed, you can start a fresh PR

PR #374, opened 2015-05-27
Use generic methods for support functions instead of macro-generated ones

### LINKED ISSUE #372: insupport fallback method
The [documentation](http://distributionsjl.readthedocs.org/en/latest/extends.html#create-a-univariate-distribution) indicates that ''a generic fallback as ``insupport(d, x) = minimum(d) <= x <= maximum(d)``'' is provided for user-created distributions.  However, I get an error if ``insupport`` is not explicitly defined for a new distribution (see example below).

Is the fallback still implemented?  If not, would you be willing to add it in a future version?   Thanks.

```julia
using Distributions

type NewUnivarDist <: ContinuousUnivariateDistribution
  mu::Float64
  sigma::Float64
end

minimum(d::NewUnivarDist) = -Inf
maximum(d::NewUnivarDist) = Inf

d = NewUnivarDist(1.0, 2.0)
minimum(d)         ## -Inf
maximum(d)         ## Inf
insupport(d, 1.0)  ## error: no matching method
```

simonbyrne:
Fixes #372, and avoids calling `eval` inside a macro.

Although less specific, LLVM is still able to optimise away most constant calls, e.g.

```
julia> @code_llvm isbounded(Binomial())

define i1 @julia_isbounded_20509(%Binomial) {
top:
  ret i1 true, !dbg !1623
}
```

### DIFF
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -11,7 +11,11 @@ minimum(r::RealInterval) = r.lb
 maximum(r::RealInterval) = r.ub
 @compat in(x::Real, r::RealInterval) = (r.lb <= Float64(x) <= r.ub)
 
-isbounded(d::UnivariateDistribution) = isupperbounded(d) && islowerbounded(d)
+isbounded{D<:UnivariateDistribution}(d::Union(D,Type{D})) = isupperbounded(d) && islowerbounded(d)
+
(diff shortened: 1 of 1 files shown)


simonbyrne:
@lindahua Is this okay by you?

PR #375, opened 2015-05-27
Generalize truncated distributions to discrete	distributions

bdeonovic:


src/truncate.jl
```diff
@@ -17,11 +17,20 @@ function Truncated(d::ContinuousUnivariateDistribution, l::Float64, u::Float64)
...
 
+function Truncated(d::DiscreteUnivariateDistribution, l::Float64, u::Float64)
+    l < u || error("lower bound should be less than upper bound.")
+    lcdf = isinf(l) ? 0.0 : cdf(d, l)
+    ucdf = isinf(u) ? 1.0 : cdf(d, u)
+    tp = ucdf - lcdf
+    Truncated{typeof(d),Discrete}(d, l, u, lcdf, ucdf, tp, log(tp))
+end
+
+@compat Truncated(d::DiscreteUnivariateDistribution, l::Real, u::Real) = Truncated(d, Float64(l), Float64(u))
```
simonbyrne: It should be possible to use parametric dispatch to use the same methods for both Continuous and Discrete

src/truncate.jl
```diff
@@ -61,6 +70,30 @@ logccdf(d::Truncated, x::Float64) = x <= d.lower ? 0.0 :
...
+                              (d.ucdf - cdf(d.untruncated, x)) / d.tp
+
+logccdf(d::Truncated, x::Int64) = x <= d.lower ? 0.0 :
+                                 x >= d.upper ? -Inf :
+                                 log(d.ucdf - cdf(d.untruncated, x)) - d.logtp 
+
+
+quantile(d::Truncated, p::Int64) = quantile(d.untruncated, d.lcdf + p * d.tp)
+
+
```
simonbyrne: Can you combine these using `Real` signatures?

src/truncate.jl
```diff
@@ -61,6 +60,29 @@ logccdf(d::Truncated, x::Float64) = x <= d.lower ? 0.0 :
...
+                              x >= d.upper ? 0.0 :
+                              (d.ucdf - cdf(d.untruncated, x)) / d.tp
+
+logccdf(d::Truncated, x::Int64) = x <= d.lower ? 0.0 :
+                                 x >= d.upper ? -Inf :
+                                 log(d.ucdf - cdf(d.untruncated, x)) - d.logtp 
+
+
+quantile(d::Truncated, p::Int64) = quantile(d.untruncated, d.lcdf + p * d.tp)
+
```
bdeonovic: When I try to change these to `Real` I get a complaint about ambiguity with definitions in `univariates.jl`


```julia
julia> using Distributions
Warning: New definition 
    pdf(Truncated{S<:ValueSupport,D<:Distribution{Univariate,S<:ValueSupport}},Real) at /Users/bdeonovic/.julia/v0.3/Distributions/src/truncate.jl:39
is ambiguous with: 
    pdf(Distribution{Univariate,Discrete},Int64) at /Users/bdeonovic/.julia/v0.3/Distributions/src/univariates.jl:127.
To fix, define 
    pdf(Truncated{Discrete,D<:Distribution{Univariate,S<:ValueSupport}},Int64)
before the new definition.
Warning: New definition 
    pdf(Truncated{S<:ValueSupport,D<:Distribution{Univariate,S<:ValueSupport}},Real) at /Users/bdeonovic/.julia/v0.3/Distributions/src/truncate.jl:39
is ambiguous with: 
    pdf(Distribution{Univariate,Discrete},Integer) at /Users/bdeonovic/.julia/v0.3/Distributions/src/univariates.jl:128.
To fix, define 
    pdf(Truncated{Discrete,D<:Distribution{Univariate,S<:ValueSupport}},Integer)
before the new definition.
...
```

simonbyrne: Ah, I see. What about if you made them a union type instead? If that still causes problems, you can leave them as they are.

Also, you would probably want `Int`, instead of `Int64` (that way it will work on 32bit machines). And you don't need the quantile methods (since integer arguments don't really make sense, the usual promotion machinery for them is fine)

src/truncate.jl
```diff
@@ -60,6 +59,25 @@ logccdf(d::Truncated, x::Float64) = x <= d.lower ? 0.0 :
...
+                                x >= d.upper ? 0.0 :
+                                log(cdf(d.untruncated, x) - d.lcdf) - d.logtp
+
+ccdf(d::Truncated, x::Int) = x <= d.lower ? 1.0 : 
+                              x >= d.upper ? 0.0 :
+                              (d.ucdf - cdf(d.untruncated, x)) / d.tp
+
+logccdf(d::Truncated, x::Int) = x <= d.lower ? 0.0 :
+                                 x >= d.upper ? -Inf :
+                                 log(d.ucdf - cdf(d.untruncated, x)) - d.logtp
```
bdeonovic: The `Union(Float64,Int)` had the same ambiguity problem. I have changed from `Int64` to `Int` ready to pull?

simonbyrne:
Looks great, thanks for sorting it out!

PR #378, opened 2015-06-16
Add documentation for Poisson Binomial distribution

nignatiadis:
Was thinking of implementing the Poisson Binomial distribution myself when I noticed it had already been done :).

### DIFF
--- a/doc/source/univariate.rst
+++ b/doc/source/univariate.rst
@@ -253,6 +253,7 @@ List of Distributions
     - :ref:`geometric`
     - :ref:`hypergeometric`
     - :ref:`negativebinomial`
+    - :ref:`poissonbinomial`
     - :ref:`poisson`
     - :ref:`skellam`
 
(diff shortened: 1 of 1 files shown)


johnmyleswhite:
LGTM. Is the code tested enough that we're comfortable making it public?

nignatiadis:
Hmm, @chrodan implemented it (see #346), so he would be most suited to answer this question. The case where the Poisson Binomial just reduces to the Binomial distribution is already tested thoroughly though.

It would also be easy to add some more tests, e.g. by comparing the CDF calculated by the Fourier transform (implemented here) to the CDF calculated by the (numerically unstable) recursive method for small K.

chrodan:
In #383 I added tests for cases where the Poisson-Binomial distribution does not reduce to the Binomial distribution. The tests rely on the Binomial distribution but I guess that is acceptable.

johnmyleswhite:
Thanks for doing this!

PR #379, opened 2015-06-18
Remove subtype constraints in support methods

brian-j-smith:
The subtype constraints on the new fallback support methods may not be needed.  If not, the changes in this pull will get rid of the current warnings about ambiguous method definitions (shown below).

P.S.  Do you even need the Type{*} parts of the Unions?

```julia
Warning: New definition
    insupport(Truncated{D<:Distribution{Univariate,S<:ValueSupport},S<:ValueSupp
ort},Real) at C:\Users\bjsmith\.julia\v0.3\Distributions\src\truncate.jl:33
is ambiguous with:
    insupport(Union(D<:Distribution{Univariate,Continuous},Type{D<:Distribution{
Univariate,Continuous}}),Real) at C:\Users\bjsmith\.julia\v0.3\Distributions\src
\univariates.jl:34.
To fix, define
    insupport(_<:Truncated{D<:Distribution{Univariate,S<:ValueSupport},Continuou
s},Real)
before the new definition.
Warning: New definition
    insupport(Truncated{D<:Distribution{Univariate,S<:ValueSupport},S<:ValueSupp
ort},Real) at C:\Users\bjsmith\.julia\v0.3\Distributions\src\truncate.jl:33
is ambiguous with:
    insupport(Union(Type{D<:Distribution{Univariate,Discrete}},D<:Distribution{U
nivariate,Discrete}),Real) at C:\Users\bjsmith\.julia\v0.3\Distributions\src\uni
variates.jl:35.
To fix, define
    insupport(_<:Truncated{D<:Distribution{Univariate,S<:ValueSupport},Discrete}
,Real)
before the new definition.
```

### DIFF
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -11,15 +11,15 @@ minimum(r::RealInterval) = r.lb
 maximum(r::RealInterval) = r.ub
 @compat in(x::Real, r::RealInterval) = (r.lb <= Float64(x) <= r.ub)
 
-isbounded{D<:UnivariateDistribution}(d::Union(D,Type{D})) = isupperbounded(d) && islowerbounded(d)
+isbounded(d::Union(UnivariateDistribution,Type{UnivariateDistribution})) = isupperbounded(d) && islowerbounded(d)
 
-islowerbounded{D<:UnivariateDistribution}(d::Union(D,Type{D})) = minimum(d) > -Inf
-isupperbounded{D<:UnivariateDistribution}(d::Union(D,Type{D})) = maximum(d) < +Inf
+islowerbounded(d::Union(UnivariateDistribution,Type{UnivariateDistribution})) = minimum(d) > -Inf
+isupperbounded(d::Union(UnivariateDistribution,Type{UnivariateDistribution})) = maximum(d) < +Inf
 
 hasfinitesupport(d::DiscreteUnivariateDistribution) = isbounded(d)
 hasfinitesupport(d::ContinuousUnivariateDistribution) = false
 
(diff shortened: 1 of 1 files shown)


src/univariates.jl
```diff
@@ -11,15 +11,15 @@ minimum(r::RealInterval) = r.lb
 maximum(r::RealInterval) = r.ub
 @compat in(x::Real, r::RealInterval) = (r.lb <= Float64(x) <= r.ub)
 
-isbounded{D<:UnivariateDistribution}(d::Union(D,Type{D})) = isupperbounded(d) && islowerbounded(d)
+isbounded(d::Union(UnivariateDistribution,Type{UnivariateDistribution})) = isupperbounded(d) && islowerbounded(d)
```
andreasnoack: This is not the same. See e.g.
```jl
julia> foo(::Type{Number}) = "Doesn't work"
foo (generic function with 1 method)

julia> foo(Float64)
ERROR: MethodError: `foo` has no method matching foo(::Type{Float64})

julia> foo{T<:Number}(::Type{T}) = "Works!"
foo (generic function with 2 methods)

julia> foo(Float64)
"Works!"
```

brian-j-smith:
I see.  So Type is included in the Union to allow users to get the support of distribution types, in addition to instances of distributions, such as:

```julia
using Distributions

julia> support(Beta)
RealInterval(0.0,1.0)
```
Makes sense.  In that case, the subtype constraints are needed and the pull changes won't work.

I'll close this and look for another solution to the warning messages.

PR #383, opened 2015-06-26
added PoissonBinomial test for non-Binomial case

chrodan:
PoissonBinomial distribution has only been tested in the special case where it reduces to the Binomial distribution. These tests also cover the case where the individual success probabilities differ (sum of three Binomial random variables).


johnmyleswhite:
LGTM? Squash, then merge?

chrodan:
Squashed the commits.

johnmyleswhite:
Thanks!

Issue #384, opened 2015-06-26
pdf of Binomial(n, 1.)


chrodan:
There is a bug in the full pdf of the binomial distribution with success probability 1:
```julia
using Distributions
d = Binomial(5, 1.)
pdf(d)
```
gives
```
6-element Array{Float64,1}:
   0.0
 NaN  
 NaN  
 NaN  
 NaN  
 NaN 
```

The issue is the recursive evaluator in binomial.jl
```RecursiveBinomProbEvaluator(d::Binomial) = RecursiveBinomProbEvaluator(d.n, d.p / (1.0 - d.p))```
The problem occurs only in the range-version of the pdf, single queries work.

johnmyleswhite:
Would be great to have a pull request for this.

lindahua:
Looks like it has been fixed by #386.

PR #386, opened 2015-06-28
Fix for Issue #384 (pdf of Binomial)

chrodan:
This patch should fix issue #384  and adds a quick test case to check consistency of `pdf(Binomial(n,p))` and `pdf(Binomial(n,p), x)`. It also removes a bit of duplicate code in `univariates.jl` by moving it in a helper function `_pdf_fill_outside!`.


chrodan:
updated the PR; fixed a minor bug and added a small test for consistency of different pdf calls of the Binomial distribution. Now all tests pass.

kmsquire:
Thanks!  Sorry this sat for so long.

PR #389, opened 2015-06-30
 Updating constants.jl - @irrational instead of @math_const

ranjanan:


johnmyleswhite:
Thanks for doing this. Let's see what Travis says (and hope Travis isn't wrong).

ranjanan:
It has failed the v0.3.10 test I think, but passed the nightly test. This was expected right? @spencerlyon2

spencerlyon2:
Yes, I think this should be expected because `@irrational` is only defined on 0.4dev.

Getting something into Compat would be the best way to resolve the problem. That entails defining the `@irrational` as an unexported member of `Base` (from within Compat), making a release of Compat, then leveraging the new compat from within distributions.

simonster:
I took a different route in JuliaLang/Compat.jl#109 and defined `Compat.@irrational` instead of mucking with Base. If you change this code to `import Compat.@irrational` it should work once that's merged.

ranjanan:
Cool stuff @simonster . Let's hope Travis passes both tests.

IainNZ:
@simonster can a new version be tagged?

simonster:
Tagged.

PR #390, opened 2015-07-07
Added ZeroInflated type that mirrors the Truncated type

slundberg:
I found myself needing a zero inflated version of a distribution and thought it would be good to have a general type that supported this.

As it stands right now this pull request does not yet have unit tests since I am not sure about the conventions you are using for those.

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -121,6 +121,7 @@ export
     Triweight,
     Truncated,
     TruncatedNormal,
+    ZeroInflated,
     Uniform,
     UnivariateGMM,
     VonMises,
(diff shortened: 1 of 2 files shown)


johnmyleswhite:
This is really cool. To what extent should this be subsumed into a redesign of mixture models?

slundberg:
Haha, well I should have considered the mixture models first, since my need was for a zero-inflated negative binomial, and I now see that can be represented as:

```julia
MixtureModel([DiscreteUniform(0,0), NegativeBinomial(r,p)], [z,1-z])
```

This does not work for continuous distributions since `Uniform(0,0)` is not supported. If it was then it should work for both.

My suggestion would be to scrap this pull request, but add a zero-inflated poisson and negative binomial example to the mixture distributions docs, just so searches for any of those terms will point people in the right direction. If at some point there is a need for it, the Uniform distribution might also be extended to allow dirac distributions.

johnmyleswhite:
All of those suggestions sound good to me. I think we'd want a PointValue distribution instead of `Uniform(0, 0)`.

Issue #391, opened 2015-07-13
Ambiguity warnings on master


jiahao:
Loading the master version of Distributions on julia 0.4 master now throws an ambiguity warning:

```
Warning: New definition 
    insupport(Distributions.Truncated{#D<:Distributions.Distribution{Distributions.Univariate, S<:Distributions.ValueSupport}, #S<:Distributions.ValueSupport}, Real) at /Users/jiahao/.julia/v0.4/Distributions/src/truncate.jl:33
is ambiguous with: 
    insupport(Union{Type{#D<:Distributions.Distribution{Distributions.Univariate, Distributions.Continuous}}, #D<:Distributions.Distribution{Distributions.Univariate, Distributions.Continuous}}, Real) at /Users/jiahao/.julia/v0.4/Distributions/src/univariates.jl:34.
To fix, define 
    insupport(_<:Distributions.Truncated{#D<:Distributions.Distribution{Distributions.Univariate, S<:Distributions.ValueSupport}, Distributions.Continuous}, Real)
before the new definition.
Warning: New definition 
    insupport(Distributions.Truncated{#D<:Distributions.Distribution{Distributions.Univariate, S<:Distributions.ValueSupport}, #S<:Distributions.ValueSupport}, Real) at /Users/jiahao/.julia/v0.4/Distributions/src/truncate.jl:33
is ambiguous with: 
    insupport(Union{Type{#D<:Distributions.Distribution{Distributions.Univariate, Distributions.Discrete}}, #D<:Distributions.Distribution{Distributions.Univariate, Distributions.Discrete}}, Real) at /Users/jiahao/.julia/v0.4/Distributions/src/univariates.jl:35.
To fix, define 
    insupport(_<:Distributions.Truncated{#D<:Distributions.Distribution{Distributions.Univariate, S<:Distributions.ValueSupport}, Distributions.Discrete}, Real)
before the new definition.
```

One of the set of ambiguous methods was introduced in #382; the other in #374.

cc: @brian-j-smith @simonbyrne - can we fix this?

brian-j-smith:
I think this can be fixed by changing the ``insupport`` function in [truncated.jl](https://github.com/JuliaStats/Distributions.jl/blob/master/src/truncate.jl) to

```julia
insupport{D<:UnivariateDistribution}(d::Truncated{D,Union(Discrete,Continuous)}, x::Real) =
    d.lower <= x <= d.upper && insupport(d.untruncated, x)
```

[...]

PR #382, opened 2015-06-23
Support method warnings fix and added type arguments

brian-j-smith:
Changes to address the issues described in #380.

PR #392, opened 2015-07-13
Fix ambiguous Truncated insupport method for julia 0.4

brian-j-smith:
This is a fix to eliminate the ambiguous definition warnings for the Truncated insupport method, noted in #391 to be occurring now in julia 0.4.

The warnings started with pull #374 and were fixed for julia 0.3 with pull #382. Apparently, that previous fix is not working for julia 0.4, and so this one is being proposed to finish the job.  It has been tested on julia 0.3 and 0.4 and seems to eliminate the ambiguous warnings on both.


simonbyrne:
Thanks!

PR #394, opened 2015-07-28
Use Cholesky when inverting random Wishart matrices. Fixes #393

### LINKED ISSUE #393: InverseWishart rand is not always posdef
It is typically the case that `x` does not equal `100` in the following code.

```Julia
using Distributions
d = InverseWishart(2, eye(2))
x = false
for i = 1:100
	x = x + isposdef(rand(d))
end
```

It might be nice if the output from random draws from the InverseWishart distribution was guaranteed to be some kind of `PDMat`.

andreasnoack:
The present version didn't assume anything about the Wishart matrix so the inverse was calculated with the LU factorization and consequently it was numerically unsymmetric.


johnmyleswhite:
Thanks!

Issue #283, opened 2014-09-16
Boundary parameter handling


simonbyrne:
How should we handle parameters which lie on the boundary? e.g.
- `Poisson(0)` ([see original discussion](https://groups.google.com/d/msg/julia-users/ZvT0ariDWg8/Ia_JIxqbks0J)) FIXED!
- ~~`Normal(0,0)`~~
- ~~`Beta(a,0)` and `Beta(0,b)` for `a,b > 0`.~~

In most cases the limits end up being Dirac measures, though in some cases there can be ambiguity (e.g. `Beta(0,0)`).

If we do include these, we also need to decide how to handle `skewness`/`kurtosis`: presumably either `NaN` or throw an error.

Update:
Don't allow this for continuous distributions. Distributions that need update:

- [x] `Poisson`
- [ ] `Geometric`
- [ ] `NegativeBinomial`

richardreeve:
I've submitted pull request #398 to fix the Poisson(0) issue.

richardreeve:
The Poisson(0) issue is now fixed since pull requests #398 and #401 have been merged.

simonbyrne:
I would be keen to allow `Normal` with std dev = 0, as it comes up fairly often.

itsdebartha:
I would very much like to incorporate `Geometric` with the success parameter `p=1`. I came across a situation in a simulation study of a response-adaptive treatment allocation and encountered some error relating to `zero(p) < p < one(p)` when an allocation probability becomes `1`. Moreover, I think including `p=1` will be generalising this distribution a bit more.

PR #398, opened 2015-07-31
Fixing Boundary parameter handling #283 for Poisson(0)

richardreeve:
As spaceLem, I need a rate of 0 for the Poisson distribution. So I've gone through the distribution code and test scripts, and added in support (and tests) for a rate of 0. This has a v0.9 milestone, and has been sitting around unfixed for almost a year. I hope it's useful to just provide it, otherwise it seems like it may hang around for ever.

Cheers,

Richard.

### DIFF
--- a/test/discrete_test.json
+++ b/test/discrete_test.json
@@ -1376,6 +1376,33 @@
     }
   ], 
   [
+    "Poisson(0.0)", 
+    {
+      "dtype": "Poisson", 
+        "entropy": 0.0,
+      "maximum": 0, 
+      "mean": 0.0, 
+      "median": 0.0, 
+      "minimum": 0, 
+      "params": {
+        "rate": 0.0
+      }, 
+      "points": [
+        {
+          "cdf": 1.0, 
+          "logpdf": 0.0, 
+          "x": 0
+        }
+      ], 
+      "q10": 0.0, 
+      "q25": 0.0, 
+      "q50": 0.0, 
+      "q75": 0.0, 
+      "q90": 0.0, 
+      "var": 0.0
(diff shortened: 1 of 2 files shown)


andreasnoack:
LGTM.

richardreeve:
I'm not sure how the coverage can have fallen when I added in tests for every line of code I wrote?

richardreeve:
Merged in latest update to constructor by @lindahua - fe4cba0efdf63c82f85e82fc8885d27d0d5dc24f

test/discrete_test.json
```diff
@@ -1376,6 +1376,33 @@
     }
   ], 
   [
+    "Poisson(0.0)",
```
lindahua: The JSON file is generated automatically, and it will be overwritten next time the generation script is run to update the file.

richardreeve: In that case I don't understand how to add in a new test? I can only find there, and test/discrete_test.lst, but that gives no information on what the values should be.

lindahua: It is to add a line to `test/discrete_test.lst`. You should see things there as:
```
Poisson()
Poisson(0.5)
Poisson(2.0)
Poisson(10.0)
Poisson(80.0)
```
Then you may add `Poisson(0.0)` over there, and re-run `predref.py` to update the json file.

richardreeve: Okay, since it tests values against scipy, which does not behave correctly with Poisson(0.0), I've had to fix predref.py as well as adding it to discrete_test.lst. I've added a new pull request as I don't seem to be able to reopen this.

Issue #395, opened 2015-07-29
Fieldname naming convention


lindahua:
There have been some debates about naming conventions of fieldnames (*e.g.* #340). Actually, looking at the codes, we can find inconsistencies in naming. 

I am opening this issue, for the purpose of establishing a naming convention that we should follow to name parameter fields in distribution types. 

Below is a tentative proposal:
```julia
α   # \alpha: shape parameters (e.g. Gamma)
β   # \beta: scale parameters (e.g. Gamma, Exponential)
μ   # \mu: mean parameter (e.g. Normal, Poisson)
ν   # \nu: degree of freedom (e.g. Chisq)
σ   # \sigma: standard deviation (e.g. Normal)
λ   # \lambda: non-centrality
p   # probability, success-rate (e.g. Bernoulli, Binomial, Geometric)
n   # the number of trials (e.g. Binomial, Multinomial)
```
For other cases, the general guideline is to follow what is being used in Wikipedia.

Any comments? I may try to go through the distributions and implement this rule (possibly this week or next), and also put the naming convention to the document.

---

EDIT: Now a modification is adopted, according to @simonbyrne's suggestion:

```julia
μ, σ: # location and scale parameter in location-scale families (e.g. Normal, LogNormal, etc)
θ:  # \theta: scale parameter, in scale-only families (e.g. Exponential, and Gamma)
```

Also, according to the current convention, there's another rule:

```julia
(a, b):  # indicate two bounds of a bounded support (e.g. Arcsine, Uniform, Triangular)
```

andreasnoack:
It would be great eith consistency in the field names so thanks for opening this. I support your suggested naming convention. Should we also decide on a localtion/non-centrality parameter name?

lindahua:
I think non-centrality may use `ν (\nu)`, as in Wikipedia (Noncentral T). 

For location, if that parameter represents the mean (e.g. Normal), we should just use `mu`. For others, we may refer to Wikipedia. I will bring up the issue if I come across names that I am not sure of.

lindahua:
We can use `loc` for location parameter, if there's no conventional name for that specific parameter in certain distributions.

andreasnoack:
On Wikipedia, It appears that the non-central beta, chi-square and F distributions use λ. I think I've also seen ν used for the degrees of freedom in some places, so maybe we should use λ for location/non-centrality.

lindahua:
Sounds good, then let's use `λ` for noncentrality. 

[...]

PR #400, opened 2015-08-02
Enforce consistency of parameter naming

lindahua:
This PR is to enforce the rules in #395.


lindahua:
Merge this, as it seems that we have consensus in #395 and Travis gives a green light.

PR #402, opened 2015-08-08
Allow precompiling Distributions

johnmyleswhite:
Title says it all.

This implies precompiling the dependencies for Distributions. We should probably explicit flag all of them as safe for precompilation.


lindahua:
Great!

johnmyleswhite:
As compilation is a file-level property, this was moved outside of the `module`.

PR #403, opened 2015-08-11
Generalized Pareto distribution

woest:
First pull request on github :).

Needs a decent review, especially the test stuff. I wasn't sure where I needed to add tests, so I added it in two places (continuous.jl en prepdref.py/continuous_test.lst/continuous_test.json).

Also if I added all the required methods to the generalized pareto distribution.


woest:
Also, I did not special case for the exponential distribution/pareto distribution: https://en.wikipedia.org/wiki/Generalized_Pareto_distribution#Special_cases

woest:
Added some documentation.

woest:
FYI, Just noticed this: 
https://github.com/sammorris81/ExtremeValueDistributions.jl

woest:
squashed commits.

PR #405, opened 2015-08-20
Implement fit_mle for pareto distribution

davidanthoff:


johnmyleswhite:
Very nice work

Seems ready to merge once CI finishes up

johnmyleswhite:
This failed on the nightlies when checking the shape parameter. Any thoughts?

davidanthoff:
I can't reproduce locally because the tests crash on something else with the current Windows 0.4 nightly...

But I suspect the sample size is simply to small. ``n0`` is a 100, so in the old test I sampled a 100 values from the Pareto and then fit a new one to that sample. I just pushed a commit that uses 10^5 samples, let's see whether that fixes the issue.

johnmyleswhite:
Can we check if a CI (based on the analytic SE) for the parameters covers the true parameters? In the other tests, we're generating a lot of data (I think closer to 10^9) and they still fail every once in a while.

davidanthoff:
That would be nice. But I was not sure where the CI/SE would actually show up in the current API design. The ``fit`` nor ``fit_mle`` method doesn't really have any way to return CIs/SEs, right?

johnmyleswhite:
No, they don't.

To be honest, what I had in mind isn't really a CI. I was being sloppy when I used that term. What I mean is that when you fix a value of n, the Wikipedia page gives you an analytic SE for the MLE parameters. Assuming the sampling distribution is as advertised, you should be able to compute a range of viable MLE fitted values that you'd see with probability > 0.99999. Would be good to make sure fitted parameters live in that range.

davidanthoff:
Wikipedia only has the SE for one of the two parameters, and I would like to read up on that one before I use it, but don't have the time right now. Could we just merge as is, and if someone wants to improve later, great? Or has this been done for all the other tests? Looking through them it seems not.

johnmyleswhite:
Let's merge this since the tests pass now.

davidanthoff:
Thanks!

johnmyleswhite:
Thank you for doing all the actual work!

PR #409, opened 2015-08-28
Change logpdf(LogNormal...) to return -Inf if evaluated outside support

davidanthoff:
Currently it crashes with a DomainError.


andreasnoack:
LGTM

Issue #408, opened 2015-08-28
NegativeBinomial(0.5,0.5) gives InexactError()


augustinyiptong:
Hello,

I am trying to create a negative binomial distribution with a float for the r parameter. The comments at the top of the negativebinomial.jl file lead me to believe this should be possible?

I am using Julia 0.3.11 and Distributions 0.8.3

Thanks, Gus

johnmyleswhite:
https://github.com/JuliaStats/Distributions.jl/blob/a4568d75baa797498983ded6a1e8f0442e326b7a/src/univariate/discrete/negativebinomial.jl#L8 makes it impossible to use a non-integer value of `r`. Perhaps the code would work with that restriction relaxed, but that's an empirical question someone needs to look into.

andreasnoack:
Fixed by b140bf08e9d967bae06f60731981c30cda8252f8

PR #411, opened 2015-08-31
Change r parameter from Int to Float64

augustinyiptong:
First time github user, so i would appreciate any feedback.

I need to use a negative binomial distribution with a float for parameter r. https://github.com/JuliaStats/Distributions.jl/issues/408

The pmf looks same as one i computed using the wikipedia formula. 

The rand(d) function also works as intended. 

Pkg.test("Distributions") pass

Thanks, Gus


lindahua:
You probably want to add testing cases where `r` is non-integer.

augustinyiptong:
Hello, i have added a test case for non-integer r parameter. Thanks, Gus

augustinyiptong:
Squashed. Thanks

andreasnoack:
Great. Thanks for the contribution.

PR #414, opened 2015-09-06
Expand the testing coverage using R as the reference

lindahua:
Currently, we use Python's Scipy package as a reference to verify the correctness of our implementation of distributions. 

However, Scipy only provides a limited coverage of what we currently have. This PR aims to expand the coverage of our testing by switching to using R as our reference. R, together with the packages in CRAN, can cover most (if not all) of the distributions.


johnmyleswhite:
Good call

ararslan:
Thanks for all of your work here, @lindahua, this is great. Glad to see you around these parts again. 🙂

lindahua:
This PR is ready, and passed all tests in my machine.

Now almost all distributions have been covered by this R-based reference framework, including many distributions that were not available in scipy, e.g. Generalized Extreme Value, Beta Binomial, Noncentral Beta/Chi squared/F/T, Noncentral Hypergeometric, Normal Inverse Gaussian, and Von Mises, etc.

Also, with the new test cases, several bugs have been found and fixed.

A [readme describing this testing framework](https://github.com/JuliaStats/Distributions.jl/blob/055b065d51acb6b0f969bec3b7e7e7bc76cdaa94/test/ref/readme.md) was provided.

src/univariate/continuous/triangular.jl
```diff
@@ -82,12 +82,12 @@ end
...
 function skewness(d::TriangularDist)
     (a, b, c) = params(d)
-    sqrt2 * (a + b - 2c) * (2a - b - c) * (a - 2b + c) / (5 * _pretvar(a, b, c)^3//2)
+    sqrt2 * (a + b - 2c) * (2a - b - c) * (a - 2b + c) / (5 * _pretvar(a, b, c)^1.5)
 end
 
 kurtosis{T<:Real}(d::TriangularDist{T}) = T(-3)/5
 
-entropy(d::TriangularDist) = 1//2 + log((d.b - d.a) / 2)
+entropy(d::TriangularDist) = 0.5 + log((d.b - d.a) / 2)
```
andreasnoack: Shouldn't use `Float64` literal here to avoid promoting e.g. `Float32` input. Maybe `one(d.b)/2`.

src/univariate/continuous/triangular.jl
```diff
@@ -82,12 +82,12 @@ end
 
 function skewness(d::TriangularDist)
     (a, b, c) = params(d)
-    sqrt2 * (a + b - 2c) * (2a - b - c) * (a - 2b + c) / (5 * _pretvar(a, b, c)^3//2)
+    sqrt2 * (a + b - 2c) * (2a - b - c) * (a - 2b + c) / (5 * _pretvar(a, b, c)^1.5)
```
andreasnoack: Same here

lindahua:
@andreasnoack Fix per your comments.

andreasnoack:
The failures on master are because of https://github.com/JuliaLang/julia/issues/20344

PR #416, opened 2015-09-11
Add Normal-inverse Gaussian distribution

davidanthoff:
This needs tests, but I'm at a loss what to do with the new test framework. Is there some explanation how to hook into that?

### DIFF
--- a/doc/source/univariate.rst
+++ b/doc/source/univariate.rst
@@ -285,6 +285,7 @@ List of Distributions
     - :ref:`logistic`
     - :ref:`lognormal`
     - :ref:`normal`
+    - :ref:`normalinversegaussian`
     - :ref:`pareto`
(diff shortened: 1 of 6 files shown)


lindahua:
I am working on a major overhaul of the tests. 

You can just write a separate test at this point. I will try to incorporate it after the work is done.

davidanthoff:
I added tests. The ones for the moment are a bit clumsy, I computed the "correct" value in R by sampling and then computing the moments based on that sample, which seems not ideal. But I couldn't find any other function in R that would compute things like the mean of the distribution, and I did want to use code that I hadn't written, just avoid errors that I might repeat.

I also rebased and squashed, so if this passes CI I think this could be merged.

johnmyleswhite:
Looked good to me. Thanks for this!

PR #422, opened 2015-10-04
Add BetaBinomial Distribution

rawls238:
@lindahua @simonbyrne and any other contributors to the project who want to chime in...

This adds the BetaBinomial distribution. I'm still working on this and have to add a few more functions as well as validate the outputs but I wanted to get this up in case someone had any objections to add this to package.


.gitignore
```diff
@@ -1 +1,2 @@
+.DS_store
```
johnmyleswhite: You can set this in your personal `.gitignore_global` file

src/univariate/discrete/betabinomial.jl
```diff
@@ -0,0 +1,64 @@
...
+#### Properties
+
+mean(d::BetaBinomial) = (d.n * d.α) / (d.α + d.β)
+function var(d::BetaBinomial)
+    n, α, β = d.n, d.α, d.β
+    numerator = n*α*β * (α + β + n)
+    denominator = (α+β)^2.0 * (α+β+1.0)
+    numerator / denominator
+end
+std(d::BetaBinomial) = sqrt(var(d))
```
johnmyleswhite: There's a generic fallback that should provide this behavior automatically.

johnmyleswhite:
This seems like a really good start. Thanks for working on this.

src/univariate/discrete/betabinomial.jl
```diff
@@ -0,0 +1,64 @@
...
+end
+#### Parameters
+
+params(d::BetaBinomial) = (d.n, d.α, d.β)
+#### Properties
+
+mean(d::BetaBinomial) = (d.n * d.α) / (d.α + d.β)
+function var(d::BetaBinomial)
+    n, α, β = d.n, d.α, d.β
+    numerator = n*α*β * (α + β + n)
```
johnmyleswhite: Minor nit: might be nice to use consistent spacing between the arguments to infix operators. Not worth spending much time on, but it's always nice to have whitespace used systematically.

src/univariate/discrete/betabinomial.jl
rawls238: :+1:

src/univariate/discrete/betabinomial.jl
rawls238: yes it probably would be much nicer to read if there was a space between arguments to operators, will make that change

.gitignore
rawls238: I usually like to have it set in the project's gitignore so that people don't accidentally end up committing it when they don't have it set globally, but either is fine with me :).

rawls238:
@johnmyleswhite sorry about the super long delay on this...I had forgotten I put this PR up. It should be ready to go now, let me know if you have any comments

src/univariate/discrete/betabinomial.jl
```diff
@@ -0,0 +1,87 @@
...
+
+function cdf(d::BetaBinomial, x::Int)
+  c = 0.0
+  for y = minimum(d):min(floor(Int,x), maximum(d))
+    c += pdf(d, y)
+  end
+
+  # this is needed due to the fact that pdf(d, y) can return irrational values and so due to the impl
+  # above we can end up with cdf values that are basically 1 but are either slightly greater or smaller
+  if abs(c - 1.0) < 1.0e-10
```
rawls238: this makes me sad but this impl should do the trick for now

src/univariate/discrete/betabinomial.jl
```diff
@@ -0,0 +1,87 @@
...
+entropy(d::BetaBinomial) = entropy(Categorical(pdf(d)))
+median(d::BetaBinomial) = median(Categorical(pdf(d))) - 1
+mode(d::BetaBinomial) = indmax(pdf(d)) - 1
+modes(d::BetaBinomial) = [x - 1 for x in modes(Categorical(pdf(d)))]
+
+quantile(d::BetaBinomial, p::Float64) = quantile(Categorical(pdf(d)), p) - 1
+
+function cdf(d::BetaBinomial, x::Int)
+  c = 0.0
+  for y = minimum(d):min(floor(Int,x), maximum(d))
```
kmsquire: Since `x` is an `Int`, `floor` doesn't seem to be necessary.  Is it here with the expectation that the type of `x` will be relaxed at some point?

src/univariate/discrete/betabinomial.jl
```diff
@@ -0,0 +1,87 @@
...
+    denominator = beta(α, β)
+    return choose * (numerator / denominator)
+end
+
+function pdf(d::BetaBinomial)
+    n, α, β = d.n, d.α, d.β
+    k = 0:n
+    binoms = Float64[binomial(n, i) for i in k]
+    fixed_beta = beta(α, β)
+    return binoms .* beta(k .+ α, n .- k .+ β) ./ fixed_beta
```
kmsquire: Doesn't need to hold up this PR, but `.+` and `.-` are not necessary here.

src/univariate/discrete/betabinomial.jl
kmsquire: Never mind, I see where you [copied it from the original](https://github.com/JuliaStats/Distributions.jl/blob/4303d591f5c53a067a46c94298e78604a13754c2/src/univariates.jl#L112).

src/univariate/discrete/betabinomial.jl
kmsquire: Is this testable?  Can you give an example where the return value of `pdf(d,y)` is irrational?

kmsquire:
Can you squash the commits?  After that, I'll merge if there are no further comments, although I there is a better solution to rounding the `cdf` to `1.0` if it's close.

src/univariate/discrete/betabinomial.jl
rawls238: yea these methods were taken from the implementation for binomial

rawls238:
sure, will squash the commits tonight. and address the PR comments tonight. Thanks for looking it over

src/univariate/discrete/betabinomial.jl
rawls238: testing it again it looks like it's not a matter of irrational numbers but there's some degree of slight imprecision at higher decimal places which I suspect comes from the `beta` function. For instance, take this example instance of the distribution: ```BetaBinomial(2, 0.2, 0.25)```.

Results of the pdf from Wolfram Alpha which sums to 1:
```
n = 0 : 0.36781609195402293 
n = 1:  0.15325670498084293 
n = 2 : 0.47892720306513414
```

Results of the pdf from this implementation which sums to 1.00000000000000027:
```
n = 0 : 0.47892720306513414
n = 1 : 0.15325670498084298
n = 2 : 0.36781609195402315
```

Let me know if you have any ideas about anything you see that may cause this slight imprecision apart from the `beta` function or some thing to do with floating point arithmetic

rawls238:
@kmsquire addressed all the comments - let me know what if you have any ideas around the slight imprecision around the pdf values.

src/univariate/discrete/betabinomial.jl
rawls238: looks like the underlying issue is that beta seems to be returning truncated results:

invocation of `beta(0.2, 0.25)` in Julia: `8.45704009347458`
invocation of `beta(0.2, 0.25)` in Wolfram Alpha: `8.4570400934745837`

src/univariate/discrete/betabinomial.jl
```diff
@@ -0,0 +1,87 @@
...
+  left = numerator / denominator
+  right = (alpha_beta_sum) * (alpha_beta_sum - 1 + 6*n) + 3 * alpha_beta_product * (n - 2) + 6 * n^2
+  right -= (3 * alpha_beta_product * n * (6 - n)) / alpha_beta_sum
+  right -= (18 * alpha_beta_product * n^2) / (alpha_beta_sum)^2
+  return (left * right) - 3
+end
+
+function pdf(d::BetaBinomial, k::Int)
+    n, α, β = d.n, d.α, d.β
+    choose = Float64(binomial(n, k))
```
kmsquire: This line needs a `@compat` in order to work on Julia v0.3.  (Currently, the tests on v0.3 are failing here.)

src/univariate/discrete/betabinomial.jl
kmsquire: Digging a little further, the `beta` function in Julia is defined [here] (https://github.com/JuliaLang/julia/blob/86cd5ea3c61426ebddd1d3809ffa869d6726fb30/base/special/gamma.jl#L407-L412) as

```julia
function beta(x::Number, w::Number)
    yx, sx = lgamma_r(x)
    yw, sw = lgamma_r(w)
    yxw, sxw = lgamma_r(x+w)
    return exp(yx + yw - yxw) * (sx*sw*sxw)
end
```

The `lgamma_r` function here is a call to libm (which is usually openlibm from Julia).  If you compare the values returned from this function call with those produced by the log gamma function on Wolfram Alpha, you'll see a small discrepancy in the last few digits (I'd post more details, but I need to head to bed).

Would you be able to open an issue on Julia or openlibm (or both) describing the problem?  That might be the easiest way to get to a better solution here.  (There are other paths that could be taken--e.g., calculating these values using higher precision--but hopefully that would lead to a better solution.)

And that shouldn't hold up this PR.

rawls238:
@kmsqure filed the issue here: https://github.com/JuliaLang/julia/issues/14620

Added an additional commit noting that this should be removed when the issue is resolved. Everything else in the PR should be good to go, thanks for the comments

src/univariate/discrete/betabinomial.jl
andreasnoack: I think the `x` argument shouldn't be restricted to `Int` here. It's usually how CDFs are defined. The restriction was changed [here](https://github.com/JuliaStats/Distributions.jl/commit/fa922d8bc9679f59445bfb633253d99e2cc5f237) without reason so I think it was an error.

src/univariate/discrete/betabinomial.jl
```diff
@@ -0,0 +1,88 @@
...
+function cdf(d::BetaBinomial, x::Int)
+  c = 0.0
+  for y = minimum(d):min(floor(Int,x), maximum(d))
+    c += pdf(d, y)
+  end
+
+  # TODO: remove when https://github.com/JuliaLang/julia/issues/14620 is resolved
+  # this is needed due to the fact that pdf(d, y) can return values that are imprecise at higher decimal places (due to https://github.com/JuliaLang/julia/issues/14620)
+  # and so due to the impl above we can end up with cdf values that are basically 1 but are either slightly greater or smaller
+  if abs(c - 1.0) < 1.0e-10
```
andreasnoack: I don't think you can blame the implementation in Base here. This is a consequence of floating point numbers. E.g. you easily get
```jl
julia> randn(10) |> t -> norm(t/norm(t))
0.9999999999999999
```
Even though that computation is much simpler than the beta function.

What about calculating the CDF from the right when `x > div(d.n,2)`? I guess that would fix the issue unless the distribution is extremely skewed.

src/univariate/discrete/betabinomial.jl
kmsquire: Actually, looking a little closer: the non-int version was [factored out](https://github.com/JuliaStats/Distributions.jl/blob/4303d591f5c53a067a46c94298e78604a13754c2/src/univariates.jl#L120).  So that should probably happen here as well.

(And `floor` should be removed in both versions).

I'll do this.

andreasnoack: I don't see the point in having two separate method definitions. One should be sufficient.

kmsquire: (Not necessary to reimplement the `Real` version--the factored out version I pointed out will call this one.)

kmsquire:
@rawls238, thanks for submitting that.  I have a separate PR which cleans up the `cdf` code, which I'll submit separately in a few minutes.  But since this PR works fine (and that one might require some review), I'm merging this.  Thanks for the contribution!

kmsquire:
`cdf` cleanup in #440

rawls238:
thanks @kmsquire could you let me know when you publish the new version of the package with the distribution?

kmsquire:
@rawls238, done!  https://github.com/JuliaLang/METADATA.jl/pull/4389

PR #426, opened 2015-10-08
bugfix for utils.jl

evanmason:


jiahao:
Thanks!

Could you include a test for this behavior? (Obviously no one's tried to use this function...)

dmbates:
I believe that the `any` and `all` methods now short-circuit so `allzeros`should be the same performance as
```julia
!any(Bool,x)
```

simonbyrne:
thanks!

PR #427, opened 2015-10-20
Allow calling logpdf() for univariate normal so as to work with ForwardDiff

scidom:


src/univariate/continuous/normal.jl
```diff
@@ -35,6 +35,11 @@ entropy(d::Normal) = 0.5 * (log2π + 1.0) + log(d.σ)
 
 @_delegate_statsfuns Normal norm μ σ
 
+### Added by scidom: start
+logpdf(d::Normal, x::Real) = -(log(2*π)/2+log(d.σ)+abs2((x-d.μ)/d.σ)/2)
+logpdf(d::Normal, x::Number) = -(log(2*π)/2+log(d.σ)+abs2((x-d.μ)/d.σ)/2)
```
mlubin: Why two copies of the same function?

scidom: The compiler complained I should define the case of real input argument first... and then I can't define the second function by invoking the first, as x must be passed as a number. How could I deal with this?

mlubin: The solution is to fix the source of the ambiguity [here](https://github.com/JuliaStats/Distributions.jl/blob/245cfe3c5546636ac2a204951d775f5b377f44e2/src/univariates.jl#L108). We need to figure out how to make this fallback work for ``Number``.

andreasnoack: Do we really want all these fallback promotions to `Float64` when it is not really required?

mlubin: Probably not. The promotions should really happen at the distribution level if necessary

andreasnoack: I think we should see what would happen if we only promoted when it was required by the function e.g. a call to Rmath.

mlubin: Right, so replace
```
logpdf(d::ContinuousUnivariateDistribution, x::Float64) = log(pdf(d, x))
@compat logpdf(d::ContinuousUnivariateDistribution, x::Real) = logpdf(d, Float64(x))
```
with
```
logpdf(d::ContinuousUnivariateDistribution, x::Number) = log(pdf(d, x))
```
This also subsumes this PR

scidom:
Yes, I have seen the lines you pointed to @mlubin, and tried to avoid touching them (due to having in mind Rmath integration). The solution you propose is cleaner, but not sure how well it will work with Rmath, which requires Float64. This is sth that @johnmyleswhite and @lindahua will know better.

mlubin:
It clearly will throw an error if the function is only implemented for ``Float64`` in Rmath, but just lower down the chain. Now that there's an abstraction between Distributions and StatsFuns, I don't think Distributions needs to be in the business of deciding which number types are supported.

scidom:
I will close this PR now, given that the newer one (#427) supersedes it.

PR #428, opened 2015-10-21
don't automatically promote to Float64 for ContinuousUnivariateDistributions

mlubin:
Combined with https://github.com/JuliaStats/StatsFuns.jl/pull/5, closes #427. I think the automatic promotion was out of place anyway, since some of the implementations in StatsFuns already supported non-Float64 types. This now works:
```
julia> logpdf(Normal(0,1),big(1))
-1.418938533204672741780329736405617639861397473637783412817151540482765695927251

julia> using ForwardDiff

julia> f(x::Number) = logpdf(Normal(1., 0.15), x);

julia> derivative(f, 2.5)
-66.66666666666667
```

Added a couple unit tests, not sure if they're in the right format or the right place. (The tests will fail until https://github.com/JuliaStats/StatsFuns.jl/pull/5 is merged and tagged.)

CC @scidom @jrevels

### DIFF
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -95,8 +95,7 @@ pdf(d::DiscreteUnivariateDistribution, x::Int) = throw(MethodError(pdf, (d, x)))
 pdf(d::DiscreteUnivariateDistribution, x::Integer) = pdf(d, round(Int, x))
 pdf(d::DiscreteUnivariateDistribution, x::Real) = isinteger(x) ? pdf(d, round(Int, x)) : 0.0
 
-pdf(d::ContinuousUnivariateDistribution, x::Float64) = throw(MethodError(pdf, (d, x)))
-@compat pdf(d::ContinuousUnivariateDistribution, x::Real) = pdf(d, Float64(x))
+pdf(d::ContinuousUnivariateDistribution, x::Number) = throw(MethodError(pdf, (d, x)))
 
 # logpdf
 
@@ -104,8 +103,7 @@ logpdf(d::DiscreteUnivariateDistribution, x::Int) = log(pdf(d, x))
 logpdf(d::DiscreteUnivariateDistribution, x::Integer) = logpdf(d, round( Int, x))
 logpdf(d::DiscreteUnivariateDistribution, x::Real) = isinteger(x) ? logpdf(d, round(Int, x)) : -Inf
 
-logpdf(d::ContinuousUnivariateDistribution, x::Float64) = log(pdf(d, x))
-@compat logpdf(d::ContinuousUnivariateDistribution, x::Real) = logpdf(d, Float64(x))
+logpdf(d::ContinuousUnivariateDistribution, x::Number) = log(pdf(d, x))
 
 # cdf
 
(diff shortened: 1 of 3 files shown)


test/continuous.jl
```diff
@@ -74,3 +74,8 @@ end
 #     println("    testing $(distr)")
 #     test_samples(distr, n_tsamples)
 # end
+
+# Test for non-Float64 input
+using DualNumbers
+@test string(logpdf(Normal(0,1),big(1))) == "-1.418938533204672741780329736405617639861397473637783412817151540482765695927251"
```
mlubin: This doesn't seem like a safe test either, open to suggestions

scidom:
Apart from the example

```
using Distributions
using ForwardDiff

f(x::Number) = logpdf(Normal(1., 0.15), x)
derivative(f, 2.5)
```

I also tried using the multivariate normal using the example below:

```
using Distributions
using ForwardDiff

f(x::AbstractVector) = logpdf(MvNormal([1., 3.5], [1. 0.25; 0.25 1.]), x)
gradient(f, [2.5, 7.3])
```

The multivariate works, and by the way this happens out of the box, prior to coding changes. Shall we press on with merging this PR? @mlubin, @jrevels

test/continuous.jl
scidom: How about sth like this for a test:

```
@test derivative(logpdf(Normal(1., 0.15), x), 2.5) == gradlogpdf(Normal(1., 0.15), 2.5)
 ```

Comment (mlubin) [inline test/continuous.jl]: Does that test pass? Equality might be too strict a requirement if the exact order of operations isn't consistent.

Comment (scidom) [inline test/continuous.jl]: `@test_approx_eq_eps` can come to a rescue :)

**Comment by mlubin** [comment]
https://github.com/JuliaStats/StatsFuns.jl/pull/5 needs to be merged and tagged before we merge this

**Comment by scidom** [comment]
:+1: 

Awesome, once this is all merged, I will write a couple of toy scipts/stats examples for autodiff.

**Comment by andreasnoack** [comment]
https://github.com/JuliaLang/METADATA.jl/pull/3854

File: test/continuous.jl (hunk 1 above)
Comment (andreasnoack) [inline test/continuous.jl]: Tests are passing now except for this on 0.3.

File: src/univariates.jl [hunk 2]
```diff
@@ -95,8 +95,7 @@ pdf(d::DiscreteUnivariateDistribution, x::Int) = throw(MethodError(pdf, (d, x)))
 pdf(d::DiscreteUnivariateDistribution, x::Integer) = pdf(d, round(Int, x))
 pdf(d::DiscreteUnivariateDistribution, x::Real) = isinteger(x) ? pdf(d, round(Int, x)) : 0.0
 
-pdf(d::ContinuousUnivariateDistribution, x::Float64) = throw(MethodError(pdf, (d, x)))
-@compat pdf(d::ContinuousUnivariateDistribution, x::Real) = pdf(d, Float64(x))
+pdf(d::ContinuousUnivariateDistribution, x::Number) = throw(MethodError(pdf, (d, x)))
```
kmsquire: Probably want `Number` => `Real`, here and below.

kmsquire:
Bump.

I made a similar change in #440, but just for `cdf` functions.  I'll remove those changes.

This needs to depend on `StatsFuns` v0.2.0.

src/univariates.jl
kmsquire: Oh, I see--presumably this is to support `Dual` numbers.

andreasnoack:
I think this should be restricted to `Real` instead of `Number` now that `ForwardDiff.jl` doesn't need `Number` anymore. @mlubin would you mind making that change and a rebase?

mlubin:
I'm a bit buried with work at the moment. I'll get to this eventually but consider it up for grabs.

andreasnoack:
I'll do it.

jmxpearson:
If I understand correctly, this can now be closed?

andreasnoack:
I think so

PR #429, opened 2015-10-22
New Shifted distribution type

slundberg:
I ran into the need for this when working with mixture distributions. Sometimes a shifted distribution is needed when creating mixture distributions. For example:

``` julia
# model where 10% of the time a constant is added to a Poisson
MixtureModel([Poisson(4), Shifted(Poisson(4), 2)], [0.9,0.1])
```

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -120,6 +120,7 @@ export
     PoissonBinomial,
     QQPair,
     Rayleigh,
+    Shifted,
     Skellam,
     SymTriangularDist,
     TDist,
@@ -274,6 +275,7 @@ include("samplers.jl")
 
 # others
 include("truncate.jl")
+include("shifted.jl")
 include("conversion.jl")
 include("qq.jl")
 include("estimators.jl")
--- a/src/shifted.jl
+++ b/src/shifted.jl
@@ -1,0 +1,67 @@
+immutable Shifted{D<:UnivariateDistribution, S<:ValueSupport, T} <: UnivariateDistribution{S}
+    unshifted::D        # the original distribution (unshifted)
+    offset::T           # the shift offset
+end
+
+### Constructors
+
+function Shifted(d::UnivariateDistribution, offset::Real)
+    Shifted{typeof(d),value_support(typeof(d)), typeof(offset)}(d, offset)
+end
+
+### range and support
+
+islowerbounded(d::Shifted) = islowerbounded(d.unshifted)
+isupperbounded(d::Shifted) = isupperbounded(d.unshifted)
+
+minimum(d::Shifted) = minimum(d.unshifted)+d.offset
+maximum(d::Shifted) = maximum(d.unshifted)+d.offset
+
+@compat insupport{D<:UnivariateDistribution}(d::Shifted{D,Union{Discrete,Continuous}}, x::Real) = insupport(d.unshifted, x-d.offset)
+
+
+### evaluation
+
+pdf(d::Shifted, x::Float64) = pdf(d.unshifted, x-d.offset)
+
+logpdf(d::Shifted, x::Float64) = logpdf(d.unshifted, x-d.offset)
+
+cdf(d::Shifted, x::Float64) = cdf(d.unshifted, x-d.offset)
+
+logcdf(d::Shifted, x::Float64) = logcdf(d.unshifted, x-d.offset)
+
+ccdf(d::Shifted, x::Float64) = ccdf(d.unshifted, x-d.offset)
+
+logccdf(d::Shifted, x::Float64) = logccdf(d.unshifted, x-d.offset)
+
+
+quantile(d::Shifted, p::Float64) = quantile(d.unshifted, p) + s
+
+pdf(d::Shifted, x::Int) = pdf(d.unshifted, x-d.offset)
+
+logpdf(d::Shifted, x::Int) = logpdf(d.unshifted, x-d.offset)
+
+cdf(d::Shifted, x::Int) = cdf(d.unshifted, x-d.offset)
+
+logcdf(d::Shifted, x::Int) = logcdf(d.unshifted, x-d.offset)
+
+ccdf(d::Shifted, x::Int) = ccdf(d.unshifted, x-d.offset)
+
+logccdf(d::Shifted, x::Int) = logccdf(d.unshifted, x-d.offset)
+
+rand(d::Shifted) = rand(d.unshifted) + d.offset
+
+
+## show
+
+function show(io::IO, d::Shifted)
+    print(io, "Shifted(")
+    d0 = d.unshifted
+    uml, namevals = _use_multline_show(d0)
+    uml ? show_multline(io, d0, namevals) :
+          show_oneline(io, d0, namevals)
+    print(io, ", offset=$(d.offset))")
+    uml && println(io)
+end
+
+_use_multline_show(d::Shifted) = _use_multline_show(d.unshifted)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -21,7 +21,8 @@ tests = [
     "conversion",
     "mixture",
     "gradlogpdf",
-    "truncate"]
+    "truncate",
+    "shifted"]
 
 print_with_color(:blue, "Running tests:\n")
 
--- a/test/shifted.jl
+++ b/test/shifted.jl
@@ -1,0 +1,108 @@
+# Testing discrete univariate distributions
+
+using Distributions
+import JSON
+using Base.Test
+using Compat
+
+
+function verify_and_test_drive(jsonfile, selected, n_tsamples::Int, offset::Int)
+    R = JSON.parsefile(jsonfile)
+
+    for (ex, dct) in R
+        dsym = symbol(dct["dtype"])
+        dname = string(dsym)
+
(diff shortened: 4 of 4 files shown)


johnmyleswhite:
Nice work.

I think this is a special case of the location-scale family work we've tried to do and never finished. If it's not too much trouble, I'd prefer merging the general case rather than this more specific variant.

kmsquire:
Perfect being the enemy of the good and all that, it's sometimes nice to have something imperfect but usable, rather than perfect but a year late. Does anyone have concrete plans to complete the general case anytime soon?

slundberg:
Thanks. I hadn't seen that thread on the location-scale work. That would certainly address this need.

As to the idea behind the more general case, is the goal to have

```julia
UnivariateLocationScaleFamily(dist, location, scale)
```

instead of

```julia
Scaled(Shifted(dist, location), scale)
```

?

I'll currently ignore the 0.3 travis issues until I know what we would be to do with this PR.

StefanKarpinski:
For what it's worth, the composition of `Scaled` and `Shifted` seems much better to me.

johnmyleswhite:
Ok, here's my stance: we should start to pull in functionality for the location-scale concept incrementally to address Kevin's "perfect as enemy" concern. That would let us merge this PR sooner rather than later.

But we should also make sure that the framework we're using for all the incremental work is consistent with the framework we want to end up with, given that there is an enormous literature on location-scale families. Getting the nomenclature right from the start means that we avoid an ugly period of deprecations and it may even encourage people to add missing functionality.

Part of the benefit of going to the literature is that it encourages us to provide even shorter syntax than `Shifted(d, x)`: we can instead work with `d + x` since location changes are exactly equivalent to the addition of a fixed constant to a random variable. Our distribution objects are exactly random variables, so we can eventually support both the addition of random variables to constants and the addition of random variables to other random variables.

slundberg:
@johnmyleswhite Sounds good. I am happy to do whatever here, and I agree that the operator notation is best. If I understand correctly, merging this PR would mean that eventually we want:

```julia
+(d::UnivariateDistribution, offset::Real) = Shifted(d, offset)
*(d::UnivariateDistribution, scale::Real) = Scaled(d, scale)
```

...along with distribution-specific shortcuts for those distributions who's parametric form contains the location-scale family.

Let me know what you prefer the end interface to be. We could even not export the `Shifted` symbol and force people to use the addition operator to minimize the chance of deprecation.

StefanKarpinski:
Wouldn't the main thing to do here be provide special cases for `+` and `*` that construct, e.g. a new `Normal` with the right parameters when the distribution argument is `Normal`? Then the `Shifted` and `Scaled` constructions could be used as a generic fallback for distributions that aren't closed under shifting and/or scaling?

johnmyleswhite:
Yes, I think that's equivalent to what @slundberg is suggesting, unless I'm confused.

The main things I'm concerned about before merging all live in type hierarchy. What exactly are the types we're adding? What are the names? How are they parameterized?

For example, things get tricky when adding non-integer values to discrete distributions. Our current type hierarchy sometimes couples integer-valued and discreteness as concepts. We should make sure we get that stuff right. We need, at a minimum, to figure out how to address univariate vs multivariate distinctions as well as integer vs non-integer distinctions.

As such, we probably need to end up with distinctions like `Shifted{Univariate, Discrete}` or something to that effect.

slundberg:
Yes, I think we are saying the same thing. As it stands right now the types are defined the same way as for the Truncated type:

```julia
immutable Shifted{D<:UnivariateDistribution, S<:ValueSupport, T} <: UnivariateDistribution{S}
    unshifted::D        # the original distribution (unshifted)
    offset::T           # the shift offset
end

### Constructors

function Shifted(d::UnivariateDistribution, offset::Real)
    Shifted{typeof(d),value_support(typeof(d)), typeof(offset)}(d, offset)
end
```

So I believe this maintains the distinction in types between univariate and multi-variate (only implementing univariate). An alternative would be to create an `AbstractShifted` type and follow the pattern set by the `MixtureModel` type.

johnmyleswhite:
Yeah, the current parameterization is close. I'm mostly worried about the offset invalidating assumptions about the discreteness of the distribution.

slundberg:
That's fair. I'll update it here soon to address that and you can see what you think.

johnmyleswhite:
Can we try to handle rescaling at the same time? I think it will be better to have both operations addressed at once. Rather than nest shifting and scaling. Imagine doing something like:

```jl
d1 = Poisson(4)
d2 = d1 + 1
d3 = d2 * 3
d4 = d3 - 1
```

I'd like to make sure all of the `d` after `d1` have the same type.

slundberg:
Sure. You already have a partial implementation of this somewhere?

johnmyleswhite:
Here's a basic example: https://github.com/JuliaStats/Distributions.jl/pull/144

I'm pretty sure my code doesn't solve these problems properly, so exercise caution with taking ideas from it.

StefanKarpinski:
Can't you just make sure that shifting composed with scaling collapse correctly and still always express this with two nested objects? I guess having a ShiftedAndScaled type would work, but it seems ugly compared to having two separate types. The main issue that I can think of is that if you standardize on having `Scaled` on the inside and `Shifted` on the outside, then you don't really want to call the constructors explicitly since doing `Scaled(Shifted(d, b), a)` would give you `Shifted(Scaled(d, a), a*b)` which is obviously not of type `Scaled`. Maybe the operator versions should handle this kind of simplification while the constructors would just construct?

johnmyleswhite:
I don't really agree. The idea of location scale families doesn't derive from a generically composable set of operations: the math only works out well for exactly these two operations.

StefanKarpinski:
Is the truncation of a shifted distribution not the shift of a truncated distribution? Ditto for scaling, censoring, etc.

johnmyleswhite:
I don't think that's true unless your definition of truncation accounts for the location shift by also shifting the location of the truncation point.

Example: Start with `Uniform(0, 1)`. Then truncate at `0.5`. Now shift up by `+0.5`. You get `Uniform(0.5, 1.0)`. But if you instead start with `Uniform(0, 1)`, then shift up by `+0.5` to get `Uniform(0.5, 1.5)`. Then truncate at `0.5`. You get `Uniform(0.5, 1.5)`. You would need to work with a shifted truncation point to get `Uniform(0.5, 1.0)`.

I really think location-scale operations are special in the same way that conjugate priors are special. They're not universal, but they have a rich theory of closed operations that doesn't extend to things like truncation or censoring.

slundberg:
I am not particularly attached either way, but it seems that the order of application of truncation vs. shift matters in a similar way that the order of shift vs. scale matters. So perhaps I missed the distinction you are getting at.

I think a practical question is: Could there ever be any reason we would want to define scaling or shifting for a distribution but not both? Perhaps for optimization purposes?

johnmyleswhite:
You're right: I was hung up on the invertibility of location-scale operations, which makes those swaps easier to compute.

I doubt there's ever a case when you don't need to account for both location and scale.

StefanKarpinski:
> I don't think that's true unless your definition of truncation accounts for the location shift by also shifting the location of the truncation point.

Yes, of course you would have to do that. You can normalize `Shifted(Truncate(d, a), b)` to `Truncated(Shifted(d, b), a+b)`. Similarly, you can normalize `Scaled(Truncated(d, a), b)` to `Truncated(Scaled(d, b), a*b)`. The point is that if you pick a standard precedence for all of these distribution operators you can normalize everything and all you need to do is define how to pass each construction through all the constructions with higher precedence. Each of these pass-throughs is simple – but the more complex you make the building blocks, the harder it gets. The reductio ad absurdum of combining Scaled and Shifted into a single object is combining all of these kinds of operations into a single "god operation" – `ShiftedScaledTruncatedCensoredDistribution`.

StefanKarpinski:
One argument for keeping Shifted and Scaled separate is if there are basic distribution families that are closed under one but not the other. I'm not sure if that's a realistic consideration, but it's a thought.

johnmyleswhite:
Here is a list of distributions and the operations they're closed under: http://www.math.uah.edu/stat/special/LocationScale.html

This seems like a more compelling argument to keep the properties separate.

StefanKarpinski:
Right, so there are some distributions that are closed under scaling only, which having separate Shifted and Scaled types would handle nicely. I.e. you could write `a*Normal(μ, σ) + b` and get back `Normal(μ+b, a*σ)` as an object, but if you wrote `a*Exponential(λ) + b` you would get back `Shifted(Exponential(a*λ), b)` instead of another `Exponential` object. The fact that there are scale-closed families but not location-closed families makes me think that you want shift to come outside of scale, so that you can do `a*(Exponential(λ)+b)` → `a*Shifted(Exponential(λ), b)` → `Shifted(a*Exponential(λ), a*b)` → `Shifted(Exponential(a*λ), a*b)` and minimize the number of objects required to construct this distribution.

johnmyleswhite:
I'm convinced. Let's keep them separate.

slundberg:
Sounds good. Other than fixing any v0.3 related errors, what else should be integrated into this PR?

matbesancon:
Closed by #635 
@kleinschmidt could be interesting to check also

Issue #423, opened 2015-10-04
Amplify number system of Distributions to cope with arbitrary Number subtypes


scidom:
I am opening this issue as a follow up of a skype discussion that @mlubin, @jrevels and I had last Friday. One of the topics of our discussion has been how to integrate ForwardDiff with Distributions. This would allow to use ForwardDiff with a broad number of statistical problems.

To elaborate more, consider the simple example of computing the derivative of the log-pdf of a distribution. I deliberately keep the context simple at the moment so as to demonstrate the major issue. Once this is resolved, then more realistic statistical examples can be constructed.

So, consider

```
using Distributions

gradlogpdf(Normal(1., 0.15), 2.5)
```

It returns `-66.66666666666667`.

Now, let's say we would like to derive the gradient of the same log-pdf using automatic differentiation via ForwardDiff. The code would become

```
using Distributions
using ForwardDiff

f(x::Number) = logpdf(Normal(1., 0.15), x)
derivative(f, 2.5)
```

This does not work as intended, but instead returns the error

```
ERROR: MethodError: `logpdf` has no metho1 matching logpdf(::Distributions.Normal, ::ForwardDiff.GradientNumber{1,Float64,Tuple{Float64}})
Closest candidates are:
  logpdf(::Distributions.Normal, ::Float64)
  logpdf(::Distributions.Distribution{Distributions.Univariate,Distributions.Continuous}, ::Float64)
  logpdf(::Distributions.Distribution{Distributions.Univariate,Distributions.Continuous}, ::Real)
  ...
 in f at none:1
 in derivative at /Users/theodore/.julia/v0.4/ForwardDiff/src/api/derivative.jl:24
```

2 things went wrong here. ForwardDiff uses its own number types, which are all subtypes of `Number`. On the other hand, Distributions uses `Float64`, both as input arguments to methods such as `logpdf()` and as field types inside distribution definitions. To alleviate the problem, the following definition for `Normal` works:

```
using ForwardDiff

immutable Normal
    μ::Number
    σ::Number
end

logpdf(d::Normal, x::Number) = -0.5*(((x-d.μ)/d.σ)^2+log(2*π*(d.σ)^2))

f(x::Number) = logpdf(Normal(1., 0.15), x)
derivative(f, 2.5)
```

[...]

PR #430, opened 2015-10-23
Redefinition of Normal

scidom:


scidom:
@mlubin, I coun't resist my curiosity, tried to redefine `Normal`. If you run

```
using Distributions

Normal()
```

it works, but if you uncomment the external constructor I have commented out now for `Normal`, I get a stack overflow error. Why?!

Everyone, don't merge of course this PR, I created it to communicate, and it is WIP.

scidom:
I actually understand why it fails, as I changed the inner (commented out) constructor to an external one. Whould the best be to simply leave out the commented constructor or is it syntactically ok to move it inside the type again?

src/univariate/continuous/normal.jl
```diff
@@ -1,12 +1,18 @@
...
+    Normal(μ::T, σ::T) = new(μ, σ)
 end
 
+Normal{T<:Number}(μ::T, σ::T) = Normal{T}(μ, σ)
+Normal(μ::Number, σ::Number) = Normal(promote(μ, σ)...)
+Normal(μ::Number) = Normal(μ, one(μ))
+#Normal(μ::Real, σ::Real) = (@check_args(Normal, σ > zero(σ)); Normal(μ, σ))
+Normal(μ::Real) = Normal(μ, one(μ))
+Normal() = Normal(0.0, 1.0)
+Normal(μ::Complex, σ::Complex) = error("Normal with complex mean and variance not defined")
```
StefanKarpinski: If complex normals aren't defined, why allow any kind of number here?

mlubin: There's no abstract type which contains real numbers and AD (dual) numbers but not complex numbers

src/univariate/continuous/normal.jl
```diff
@@ -1,12 +1,18 @@
...
-    μ::Float64
-    σ::Float64
+immutable Normal{T<:Number} <: ContinuousUnivariateDistribution
+    μ::T
+    σ::T
 
-    Normal(μ::Real, σ::Real) = (@check_args(Normal, σ > zero(σ)); new(μ, σ))
-    Normal(μ::Real) = Normal(μ, 1.0)
-    Normal() = Normal(0.0, 1.0)
+    Normal(μ::T, σ::T) = new(μ, σ)
```
mlubin: Why remove the check on ``σ > zero(σ)``?

src/univariate/continuous/normal.jl
scidom: @StefanKarpinski, in case previous relevant discussions in Distributions haven't reached you, the main aim is to integrate probability distributions with automatic differentiation. ForwardDiff defines its own number types, so we want to loosen the restriction of Float64 distribution type fields in Distributions...

src/univariate/continuous/normal.jl
scidom: You are right, I don't want to remove it... I just wasn't sure if it is correct syntactically to retain more than one inner constructor in a situation like this (the `Rational` example in the official Julia doc for constructors defines exactly one inner constructor for `Rational` and sets the rest of them as outer constructors...

src/univariate/continuous/normal.jl
StefanKarpinski: Aren't dual numbers a special kind of real number? Or are they more general so that you can have complex dual numbers?

scidom: Indeed, the difficulty is that dual and related numbers are not subtypes of Real...

mlubin: Eh, mathematically they're a special kind of real number to the same extent that complex numbers are. There's an argument that they should drop-in replacements for real numbers though.

StefanKarpinski: That's really true – dual numbers are more like [surreal numbers](c97b2b36498eb6402a9b322b77cf1b786f72cf27) than anything else. And the whole premise of surreal numbers is that they behave logically just like real numbers but there are these infinitesimals and infinites as well.

scidom: The big side-plus of this chat is that I became aware of surreal numbers that I didn't know! I think the major difference between reals and duals is that they live in a different space (they have different dimensions)... beyond maths, our consideration is how duals fit in Julia's number type system.

StefanKarpinski: Sorry, I pasted the wrong think into that link. Fixed now.

mlubin: So is it kosher to say that dual numbers are a subtype of Real? CC @jrevels

StefanKarpinski: The hyperreals are probably a better place to start: https://en.wikipedia.org/wiki/Hyperreal_number.

StefanKarpinski: I'm not sure, but I've always thought of them as a funny kind of real. The real question is if there are generic methods for reals that wouldn't make sense for duals and it seems like there shouldn't be. By comparison there are definitely generic methods for reals that would be completely wrong for complex numbers.

mlubin: It's hard to think of a generic method for reals which would be completely wrong when evaluated using dual numbers, but I don't doubt that someone will come up with one if we let them. Maybe that's okay though

scidom: I am a thrilled by hyperreals. I have been reading their wikipedia link. I don't fully understand their math formality, but they seem an astonishingly puzzling concept. At the beginning I thought, complex or dual numbers live in a space of dimension greater than one. But if I get it right, hyperreals offer an alternative in that they "embed any extra dimensions (thus duals) by more densely occupying the real line". I wonder now if they play any role in understanding the continuum hypothesis... I got carried away now. On the computing side of things; are then big floats in a sense hyperreals too? How one represents digitally a hyperreal?

scidom: Hm, actually maybe my first impression was wrong... maybe hyperreals don't get constructed by more densely occupying the real line, but by expanding it beyond its edges infinitely. It is maybe the same way of reducing dimensionality as we do with matrices - we can easily turn them to vectors (so in a sense we stitch duals right after the real line...). Ah, I will ponder some more...

StefanKarpinski: You might find this interesting: http://math.stackexchange.com/questions/341535/is-the-theory-of-dual-numbers-strong-enough-to-develop-real-analysis-and-does-i. It seems like dual numbers serve as a first-order approximation of hyperreals, so the answer to your question above is that dual numbers are exactly how you represent hyperreals.

scidom: This was very interesting too... one more link (if it opens, page 21, section 1.4   https://books.google.co.uk/books?id=eo_u18h2REIC&pg=PA29&lpg=PA29&dq=hyperreal+computing&source=bl&ots=sDDyZK9cyE&sig=PAYurkn77hJdpe6VjPx1O7x_87w&hl=en&sa=X&ved=0CCIQ6AEwAmoVChMIn_KB5MLZyAIVBvEUCh3q9g6v#v=onepage&q=hyperreal%20computing&f=false). I will read this nice intro section of the book tomorrow... About computing, I presume the easiest thing to do is to let functions take Number arguments for AD purposes probably.

jrevels: > So is it kosher to say that dual numbers are a subtype of Real?

I don't think we can say `Dual <: Real`; dual numbers aren't a subset of the reals, they're ([a very special](http://math.stackexchange.com/questions/1196775/having-trouble-understanding-generalized-complex-numbers)) subset of the generalized complex numbers. 

However, I'm really growing on the idea that we should define `ForwardDiffNumber <: Real`. The purpose of  ForwardDiff is to do AD, not make algebraically correct statements about dual numbers - that should be the purview of the DualNumbers package. From a practical perspective, we already just drop in `ForwardDiffNumber`s in place of `Real`s - so, computationally, `ForwardDiffNumber` should lie under `Real` in the type hierarchy, even though these numbers are not algebraically a subset of the real numbers.

My main issue with `ForwardDiffNumber <: Real` - and the reason I haven't really advocated it - is that I'd still like to explore using ForwardDiff in complex spaces.

This is a case where either a traits-based approach or multiple inheritance could really save us, because then we could have something like

```julia
abstract ForwardDiffNumber <: Number
abstract ForwardDiffReal <: ForwardDiffNumber, Real # direct subtype of both ForwardDiffNumber and Real

immutable ForwardDiffComplex{F<:ForwardDiffReal} <: ForwardDiffNumber, Complex
    a::F
    b::F
end
```

I mean, I'm not sure that the way I defined `ForwardDiffComplex` makes any sense, but you get the point - the above would completely solve the type hierarchy issue. In fact, we could just do the whole `ForwardDiffReal` thing now, and then use Traits.jl to define `ForwardDiffNumber` if we ever get an implementation of `ForwardDiffComplex` going.

Sorry for the rant - I think this is compelling enough that I'm going to open up an issue in ForwardDiff, and we can continue discussion there.

All that being said, I don't think it should prevent you guys from doing what you're doing here. AFAIK, there's no reason *not* to allow arbitrary number types for `Normal`, especially since you can easily exclude the few known "nonsense" cases with inner constructors (e.g. `Complex`).

scidom: I would feel naturally a little bit sceptical to make ForwardDiffNumber a subtype of Real. I understand the practical benefits of such choice. My consideration comes from the fact that it seems to me naturally ok not to want to build a computer algebra system as this is not our goal indeed, but making mathematically inconsistent number hierarchies does not make me feel very comfortable.

```
julia> subtypes(Real)
4-element Array{Any,1}:
 AbstractFloat
 Integer
 Irrational{sym}
 Rational{T<:Integer}
```

All four subtypes of Real are consistent with the math universe... I wouldn't be very excited to break consistency, but if you are all for it, I will (slowly) adjust!

jiahao: One of the defining characteristics of real numbers is their ability to be ordered, i.e. for x and y real, one of a) x < y, b) x = y, c) x > y must be true. Note that floating point numbers do not strictly satisfy this criterion since NaNs are unordered.

scidom: Good point @jiahao, I haven't considered this. At the same time, looking at the complexity of `GradualNumber` (which has a `partials` field that can be a vector) induces another level of math definition violation as opposed to the somewhat simpler matter of `NaN` for `AbstractFloat`. To me it feels that calling a `GradualNumber` to be a `Real` is a stretch conceptually.

scidom: Yet another suggestion/thought struck me; how about we define the abstract `HyperReal` type such that Real <: HyperReal <: Number? This will have dual benefit; it will allow adhering to the math formality (i.e. allowing Julia to consistently address the matter in its core), while at the same retaining coding simplicity. Then Distributions and any new packages would use HyperReal instead of Real (and GradualNumber could be <: HyperReal). This avoids redefining the Normal as a parametric type by instead doing only this minor change:

```
immutable Normal <: ContinuousUnivariateDistribution
    μ::HyperReal
    σ::HyperReal

    Normal(μ::HyperReal, σ::HyperReal) = (@check_args(Normal, σ > zero(σ)); new(μ, σ))
    Normal(μ::HyperReal) = Normal(μ, 1.0)
    Normal() = Normal(0.0, 1.0)
end
```

jrevels: > To me it feels that calling a GradualNumber to be a Real is a stretch conceptually.

I agree with you that making `GradientNumber`, `HessianNumber`, etc. subtypes of `Real` may not be *algebraically* sound, but it is very *programmatically* sound.

> how about we define the abstract HyperReal type such that Real <: HyperReal <: Number?

While I'm not necessarily against this idea in general, it doesn't solve the problem here. We want functions over `Real`s to accept `ForwardDiffNumber`s. Our options are to:

1. go through everyone's code and submit PRs to change the way their types (we'd still have to do this if we defined `HyperReal`)
2. make our custom type a subtype of  `Real`, which will immediately allow ForwardDiff to be used with a slew of packages that it currently doesn't support

My central argument in favor of option 2 is that it can be done quickly, with no cost to performance or correctness (if we define correctness w.r.t. program behavior). If somebody needs a dual number implementation that resides in the algebraically correct level of the type hierarchy, other packages already exist that provide that.

jiahao: Such a substitution would not be programmatically sound on any user that did comparisons, for example multibranch functions

```
if x > 1
    f(x)
else
    g(x)
end
```

If `x` is now a new custom `Real` variable that is no longer ordered then you would violate the interface of a `Real` number.

scidom: I remember when I first coded `DualNumber` in par (in terms of style) with `Complex` 2 years ago, there was a brief discussion about adding them to Base. This didn't happen at that time, but autodiff has come a rather immediate (if not long) way in Julia since then. While Julia has been developing, there have been several and more central breakages than that of `HyperReal`. I think the central point of discussion is how vital and central as a concept we consider automatic differentiation (and thus associated numbers) to be. If we think that in fact AD is a rather important tool for derivative computations in scientific computing (such as optimizations and MCMC for ex),despite the fact that AD has perhaps not received the magnitude of attention it deserves in the wider scientific community, then we could seize this opportunity and give it a more glorious position as an innovative step. If on the other hand the consensus is that AD is not as crucial so as to promote derivative computations at a different scale, then yes, it is not an optimal choice to add an abstract HyperReal type in Base. Perhaps we could go with your solution, despite my view of how underrated AD is...

jrevels: > Such a substitution would not be programmatically sound...if x is now a new custom Real variable that is no longer ordered then you would violate the interface of a Real number.

This is a good point, but is actually in favor of `ForwardDiffNumber <: Real` - `ForwardDiffNumber`s *are* ordered. Now, I'm not sure whether our definition of ordering has a solid basis algebraically, but such a definition is computationally useful for differentiating programs with branches. 

I feel like the most compelling reason to not define `ForwardDiffNumber <: Real` is that such a definition might confuse a user. However, this point is easily countered by the fact `ForwardDiffNumber` types aren't exposed to the user, such that the average user would never need to know they even exist.

mlubin: @jiahao, comparison with dual numbers is defined, it just compares the real component.

@scidom, nobody except us knows what ``HyperReal`` means. It seems bizarre to ask users to use this abstraction in their code. The most we can ask is for users to write generic code that accepts ``Real`` or ``Number``.

scidom: ok, if you offer me free lunch, I won't say no, this means no work on my part to make AD work with Distributions, I can just write up the code for the stats example.

One more thing; does that mean we need to reverse the recent PR in StatsFuns, as we don't need to generalize to Number input arguments there anymore?

mlubin: @scidom, ``Normal`` still assumes ``Float64`` fields and https://github.com/JuliaStats/Distributions.jl/pull/428 is still needed to avoid automatic conversion to ``Float64``, so there's more work to be done.

scidom: Ah, yes, you are right actually.

scidom: So the bottom line is that I should try change the fields (parameters) of Normal to Real instead of Number as we had initially thought?

mlubin: After https://github.com/JuliaDiff/ForwardDiff.jl/issues/66 is implemented, yes

StefanKarpinski: The idea that dual numbers are an odd kind of complex number seems completely wrong to me. They have completely different algebraic properties: `im^2 == -1` whereas `ɛ^2 == 0` – there's no way to square that discrepancy (get it?). The fact that you can in some cases use complex numbers to do automatic differentiation of analytic functions is just an accident and doesn't mean that dual numbers are a variation on complex numbers.

In particular, the reason I think that treating dual numbers as a funny kind of real number with some extra information is that there's a field homomorphism from the dual numbers to the reals by dropping the coefficient of `ɛ`. The projection from the complex numbers to the real line by taking the real component, on the other hand, is not a homomorphism – it only preserves addition. For example:

```jl
julia> using DualNumbers

julia> real((1 + 2du)*(3 + 4du))
3

julia> real(1 + 2du)*real(3 + 4du)
3

julia> real((1 + 2im)*(3 + 4im))
-5

julia> real(1 + 2im)*real(3 + 4im)
3
```

This is because `(ℝ[ɛ]/ɛ²)/ɛ ≅ ℝ` whereas `(ℝ[i]/(i²+1))/i ≅ {0}` – i.e. you can mod out the dual numbers by ɛ and get back the real number line, whereas if you mod out the complex numbers by `i` you've collapsed 0 and -1, which forces all of the real number line to collapse to a single point.

scidom: That's a great intuitive (and abstract algebraic) justification!

jrevels: > The idea that dual numbers are an odd kind of complex number seems completely wrong to me. 

Dual numbers are a special case of the class of [**generalized** complex numbers](http://people.rit.edu/harkin/research/articles/generalized_complex_numbers.pdf) (as are the "normal" complex numbers, where i² = -1).

> The fact that you can, in some cases, use complex numbers to do automatic differentiation of analytic functions is just an accident and doesn't mean that dual numbers are a variation on complex numbers.

Agreed. I'm not sure whether you wrote this in response to me or not - but if you did, I'll mention that my idea involving `ForwardDiffComplex` was in regards to applying dual numbers for differentiating complex functions, not using complex numbers for AD.

> This is because (ℝ[ɛ]/ɛ²)/ɛ ≅ ℝ whereas (ℝ[i]/(i²+1))/i ≅ {0} – i.e. you can mod out the dual numbers by ɛ and get back the real number line, whereas if you mod out the complex numbers by i you've collapsed 0 and -1, which forces all of the real number line to a single point.

Interesting point

StefanKarpinski: I'm not entirely sure that saying that dual numbers are a special case of generalized complex numbers gives much insight into what to do here since the dual numbers are the only one of those generalizations that can be algebraically mapped back to the reals. It's pretty easy to see that if you're adjoining an element to ℝ and then identifying its square with some real value, the only value that allows you to recover ℝ by further modding is zero – i.e. the dual numbers.

goretkin:
@StefanKarpinski, I a-lot agree with what you wrote.

I mentioned elsewhere but I'll repeat it here. If `Dual{T<:Real}<:Real`, then I think it's only consistent to have also `real(x::Dual{T<:Real})=x`. There will then have to be a different accessor for the non-epsilon component of the Dual number.

However, note that you could then have Dual{Dual{Float64}}, for example. I think this cannot be made to behave like a hyperdual numbers (with four parts: `noneps`, `eps1`, `eps2`, and `eps1eps2`), so it's probably a nonsense type.

However, also note that you could then have `Complex{Dual}` (which makes sense if you ask me, and I think would work today) and `Quaternion{Dual}` (which I haven't played with). I don't know what to do about the fact that `Dual{Complex}` (which would require a definition `Dual{T<:Number}`  instead of `Dual{T<:Real}`) should be essentially the same as `Complex{Dual}`.

src/univariate/continuous/normal.jl
goretkin: @StefanKarpinski, I a-lot agree with what you wrote.

I mentioned elsewhere but I'll repeat it here. If `Dual{T<:Real}<:Real`, then I think it's only consistent to have also `real(x::Dual{T<:Real})=x`. There will then have to be a different accessor for the non-epsilon component of the Dual number.

However, note that you could then have Dual{Dual{Float64}}, for example. I think this cannot be made to behave like a hyperdual numbers (with four parts: `noneps`, `eps1`, `eps2`, and `eps1eps2`), so it's probably a nonsense type.

However, also note that you could then have `Complex{Dual}` (which makes sense if you ask me, and I think would work today) and `Quaternion{Dual}` (which I haven't played with). I don't know what to do about the fact that `Dual{Complex}` (which would require a definition `Dual{T<:Number}`  instead of `Dual{T<:Real}`) should be essentially the same as `Complex{Dual}`.

scidom: The algebraic insight that @StefanKarpinski gave makes me feel convinced that there is a reasonable way of viewing dual or forward diff numbers as some sort of real number. At the same time, the technical concerns that @goretkin raised make me wonder if we will end up producing erroneous parametric number types that are not so welcome.

I am not so sure myself what is the ideal solution here... I would only say, @jrevels, it may be better to wait a bit before making this drastic coding change to treat duals or forward diff numbers as a subtype of real till the discussion here matures and a consensus is reached.

jrevels: I was under the impression that nobody has any issues with making ForwardDiffNumbers <: Real. It seems like most of the concerned parties here are in favor of it?

> At the same time, the technical concerns that `@goretkin` raised make me wonder if we will end up producing erroneous parametric number types that are not so welcome.

If you're talking about nested ForwardDiffNumbers, we already allow this (it's how ForwardDiff supports nested differentiation). 

Anyway, this discussion really should move to https://github.com/JuliaDiff/ForwardDiff.jl/issues/66.

jmxpearson:
Is this idea (parameterize `Generic<dist>` and then typealias concrete definitions) still live?

Is it just waiting on work, or is there another strategy planned? I'm interested in variational inference applications, where #427 isn't enough, since I'd need derivatives of all sorts of expectations of distributions.

I can likely do the work, if a pull request is welcome.

mlubin:
It's still alive, just needs effort on pushing it forward. At this point we can restrict the parameters to be subtypes of ``Real``.

jmxpearson:
Great. I will post here if I can get some time cleared in the next week to take this up.

I assume the first phase is just to handle the univariates. MvNormal would require generalizing PDMats, which strikes me as more complex.

jmxpearson:
One question: 

My understand was that [this comment](https://github.com/JuliaStats/Distributions.jl/issues/423#issuecomment-145367232) was the plan, but that leaves open the problem that a line like

```
Normal(μ, σ)
```
with μ and σ not both Float64 runs into a conversion issue, since it just calls `GenericNormal{Float64}(μ, σ)`. This can be solved with an inner constructor

```
GenericNormal(μ::Real, σ::Real) = new(T(μ), T(σ))
```

but this will still fail when the parameters are, e.g., ForwardDiffNumber. And the constructor call `Normal(μ, σ)` can't result in a `GenericNormal{ForwardDiffNumber}` without destroying type stability, right? Plus there's the problem that we *want* some subtypes of Real (Int, Rational) to convert to Normal, whereas others should convert to GenericNormal of other types.

So it would seem that any code that wants to use autodiff would need to replace `Normal` with `GenericNormal` everywhere, right?

mlubin:
@jmxpearson, I think this issue can be resolved by using promotion rules. The parameters ``μ`` and ``σ`` should be promoted to the same type, e.g.,:
```
julia> promote(1.0,1)
(1.0,1.0)
```
and we'll implement the necessary promotions for ``ForwardDiffNumber`` if they don't already exist.

jmxpearson:
Let me propose another alternative: a separate small package that sits "between" ForwardDiff and Distributions. This requires only that `Normal`, etc. derive from an `AbstractNormal` (as `MvNormal` already does) and facilitates differentiation through constructor forwarding to a sibling type.

Proof of concept implementation [here](https://github.com/jmxpearson/DiffableDistributions.jl).

Seems to me this is a bit cleaner. Thoughts?

scidom:
My feeling is that parametrising the `Normal` (and possibly other) distributions inside the `Distributions` package is a more transparent and holistic solution because i) it avoids creating a separate package that may not offer functionality general enough so as to be hosted in a package of its own, ii) in principle provides a more generic solution from which other aspects of `Distributions` or of other packages may benefit (i.e. parametrising a distribution might be a step forward in terms of generalising the existing distribution types).

jmxpearson:
I totally get (i). There may be less of a role in the Julia package ecosystem for very focused small packages, though that has its advantages in composability in exactly cases like this where you're trying to make two packages work together.

But I think there are a few factors related to (ii) that make generalizing the types in Distributions less appealing:

1. The idea of defining `GenericNormal{T <: Real}` and then
    ```julia
    typealias Normal GenericNormal{Float64}
    ```
    means that you have to do some hackery to make generalizing work work for autodiff, since writing `Normal(μ, σ)` when `μ::ForwardDiffNumber` calls the *inner* constructor `GenericNormal{Float64}`, requiring something like
    ```julia
    function GenericNormal(μ, σ)
        S = promote_type(typeof(μ), typeof(σ))
        if S == T
            new(μ, σ)
        else
            GenericNormal{S}(μ, σ)
        end
    end
    ```
    Works, but ugly, IMO.

2. I think it's a little more confusing as a user to write `Normal(0, 2.5)` and get `GenericNormal{Float64}(μ = 0.0, σ = 2.5)` out at the REPL. Again, it's a matter of taste, but generalizing requires more work to maintain the simple interface that's already in place. Plus, now `Normal(0, 2)` gives a `GenericNormal{Int}` and there is the inevitable extra TypeVar running around in all the types that use distributions to make sure covariance is handled properly.

3. Is there another use case for this generalization other than compatibility with automatic differentiation (and here we're only talking ForwardDiff, after all)? Maybe if you really want `Float32`?

    I ask because making the distributions generic leads to all sorts of difficult corner cases in Distributions itself. What about all the mle fitting? Should that generalize? As far as I can tell, all sorts of small decisions have to be made about whether to generalize functionality that really only makes sense for the case that's already implemented.

    Plus, there's *a lot* more tests and code that need to be modified for generalization. My guess is several times as much work to overhaul Distributions for arbitrary `Real` subtypes than to simply implement the autodiff case in a small additional codebase. As I said above, that would require adding `AbstractNormal`, `AbstractPoisson`, etc. and generalizing the interface functions (`params`, `mean`), but that's a much milder change and doesn't require changing the tests already in place.

mlubin:
Wrapper packages really go against the spirit of automatic differentiation. Users shouldn't need to include any extra packages besides ForwardDiff in order to differentiate their code.

mlubin:
```julia
function GenericNormal(μ, σ)
    S = promote_type(typeof(μ), typeof(σ))
    if S == T
        new(μ, σ)
    else
        GenericNormal{S}(μ, σ)
    end
end
```

Not sure I understand the need for this, why not have
```julia
GenericNormal{S,T}(μ::S, σ::T) = GenericNormal(promote(μ, σ)...)
```

jmxpearson:
I'm assuming people are still calling `Normal(μ, σ)`, which gets translated to `GenericNormal{Float64}(μ, σ)`, the inner constructor. So if that's the desired behavior, the inner constructor needs to be able to "discard" the TypeVar information for the more "accurate" information from `promote`.

If Distributions maintainers are willing to require either that everyone write `Normal{T, S}` in their field type declarations or write `GenericNormal` in their diffable code, then there's no need for this.

jmxpearson:
In particular, I'm hoping @andreasnoack @johnmyleswhite @lindahua will also weigh in on the relative cost/benefit to JuliaStats of these proposed changes to the type structure along with @scidom. I can mock up a PR for Normal{T, S} as at the top (reflecting subsequent discussion here), but before I do anything larger, I want to be sure it'll get merged.

scidom:
From where I stand, I realise that such a major change in Distributions feels intimidating and might potentially affect a large number of Julia fellows, but I think it is long-term investment. For one thing, as @mlubin argued, we want to ForwardDiff to work seamlessly with Distributions (and more generally with any package) without wrappers or ad hoc solutions of any kind, and secondly I think that such parameterisation of distribution types is a natural generalisation that can be handy for other (now) unforeseen reasons.

I agree, before you embark on doing so much work @jmxpearson, it makes sense to have some reassurance it won't be in vain, i.e. that your PR will be accepted.

jmxpearson:
Just to be concrete, #475 is a minimal working example of the `Normal` case.

jmxpearson:
Any thoughts? I have some time and would really love to get started on this soon.

StefanKarpinski:
I haven't followed all of this conversation but making distribution types parametric strikes me as desirable assuming that it doesn't overly complicate all of the code. The other issue is whether generic functions on distributions can be defined accurately and reasonably efficiently in a general way. Can we compute PDFs for all distributions for arbitrary parameter types?

mlubin:
@StefanKarpinski, in principle yes, if the code to compute the PDF is pure Julia. The strategy is to do it in a piecemeal way, with Normal being the first step.

StefanKarpinski:
Seems like a good start to see how it goes.

jmxpearson:
I think it will have to work by ultimately converting subtypes of `Real` back to `Float64` (someone please correct me if I'm wrong). I think there are two alternatives here:

1. Parameterize everything in Distributions. This involves a little cruft with inner constructors to allow typealiases for `Normal`, etc. to work, but passes tests (see #475). This is the preferred option for ForwardDiff to work automatically.
1. Leave Distributions as-is but create a set of types (in a separate package) that handle subtypes of `ForwardDiffNumber` specially and have constructors in Distributions forward to these when called with ForwardDiff. This would require a separate package like [here](https://github.com/jmxpearson/DiffableDistributions.jl) that had to be imported. As @mlubin notes, this is ugly in its own way, but keeps the implementation of Distributions simpler.

My preference is 2, but Miles advocates strongly for 1. I want this badly, so will implement either that the Distributions team will accept.

StefanKarpinski:
If everything is going to have to be computed through Float64 anyway, it seems a bit weird to pretend to support any kind of Real. If the whole issue is just supporting automatic differentiation, then a more limited parametricity seems warranted.

jmxpearson:
Well, the basic interface (mean, var, etc.) is pure Julia and works fine.
My understanding is that other functions forward to StatsFuns, where there
is some lingering RMath dependency that needs Float64.

My bigger concern is with things like mle fitting and sufficient
statistics, which are trickier to generalize in their current form and I
didn't alter in the PR. It's those potentially tricky parts of the
interface that lead me to prefer the narrower option, which requires an
import for autodiff but only a minor change to Distributions.

mlubin:
Only the "legacy" stats functions need conversion to ``Float64``, almost everything with Normal works fine.
Ref https://github.com/JuliaStats/Distributions.jl/pull/428 https://github.com/JuliaStats/StatsFuns.jl/pull/5

johnmyleswhite:
I'm not sure when I'll have time to give a more comprehensive comment here, but I think the core question we need to answer is what we want from parametric distribution objects beyond AD support.

Some concerns I have that I think we need to decide on:

* Does our current generic code produce reasonable answers for `Float32` or not? If not, we should probably error out. Call this the "clearly wrong" category of issues.

* Should we fallback to `Float64` values if we don't have code that is custom built for `BigFloat`? Call this the "clearly suboptimal and maybe wrong" category of issues.

Otherwise, I agree with everything Miles has said.

jmxpearson:
Right. So *is* there a use case right now other than AD support?

johnmyleswhite:
I think we need to decide whether using the `Float64` code to process `BigFloat` is acceptable. If the answer is yes, then I think we do have another use case.

andreasnoack:
I don't think it's okay to use `Float64` when processing `BigFloat` input. I'm pretty sure we don't allow that in Base. However, I think that `cdf`, `pdf` and probably a few other functions would have proper `BigFloat` support.

scidom:
It is extra work for sure, yet in principle it sounds reasonable to want/wish to implement `BigFloat` support for some methods in `Distributions`, in which case AD numbers might not be the only anticipated subtype of `Real` in the context of `Distributions`.

What I don't know is whether parametric distribution types incur computational penalty (even minor) in comparison with the existing types.

StefanKarpinski:
If you use the `Normal{Float64}` concrete type, performance is just as good. In dispatch, writing `::Normal` ok fine too. What would become problematic, however, is using `Normal` with no type parameters as a field type or the element type of an array. For those you'd want to use `::Normal{Float64}` instead or `::Normal{T}` for whatever element type `T` you want.

StefanKarpinski:
I agree with @andreasnoack that having `Normal{BigFloat}` methods call the Float64 versions seems wrong – if I've bothered to construct a `Normal{BigFloat}` distribution, it's because I want my computations to have more precision. One expects that `pdf(::Normal{BigFloat})` to return a `BigFloat` and if that's really just a Float64 value upcast to BigFloat, then it's basically wrong.

mlubin:
``pdf(::Normal{BigFloat})`` should already essentially work without conversion to ``Float64`` because the code is in julia.

johnmyleswhite:
There could be a big gap between the code being written in Julia and the code working. Imagine a series expansion written in Julia that's truncated somewhere chosen for 64-bit precision. If we're going to let the current Julia code be used for arbitrary precision and assert that it works better than promoting a Float64 value, we need to test that it actually works. Hopefully it will, but that assertion substantially increases our workload.

jmxpearson:
Along the same lines: would mle fitting work for BigFloats?

johnmyleswhite:
I'm not sure.

StefanKarpinski:
I think this is the crux of the matter. It would be great if distributions were parametric and algorithms computing properties about distributions were generically correct. *But* that's a *lot* more work than just making sure that everything works for Float64: someone needs to implement all the generic methods and write sufficient tests for them so that we're confident that the generic methods are correct for a variety of parameter types. This would be a wonderful state to reach, but it's a massive amount of work.

andreasnoack:
> Along the same lines: would mle fitting work for BigFloats?

In general, I don't see why it wouldn't and specifically for `Normal` it's `mean` and `std` which almost trivially work for `BigFloat`. Did have any specific concern in mind?

> There could be a big gap between the code being written in Julia and the code working.

In Base, I don't think we define any `BigFloat` math functions unless the precision is handled. Many of there are in MPFR which handles these correctly. In `StatsFuns`, I think methods that are only precise for `Float64` are restricted to `Float64`. There might be cases where we are wrong but, in general, I think we are fine.

jmxpearson:
@StefanKarpinski This is why I proposed the (arguably less elegant) solution of just having a [simple little package](https://github.com/jmxpearson/DiffableDistributions.jl)  that creates "sibling" distributions that work with ForwardDiff.

Cons: 
- less general
- does require an extra import

Pros:
- a *lot* less work
- requires only a very minor change to Distributions right now
- leaves option open for Distributions to go parametric in future

jmxpearson:
@andreasnoack 

No specific concern. Just trying to figure out where the edge cases might be.

datnamer:
@simonbyrne  Would Unums help with this and how far are we from seeing them in base?

mlubin:
@jmxpearson, I don't want my objections to a wrapper package to block your progress. If it's the right solution for your needs and consensus here is slow to come by, I understand going with the quick and dirty approach so that you can move on to solving other problems (I've done it myself e.g. with https://github.com/mlubin/NaNMath.jl).

jmxpearson:
@mlubin 

That makes sense. I'm going to get a pull request in for the simple thing today, and this discussion can continue to hash out what the more general long-term strategy is.

StefanKarpinski:
This seems like the right approach: let's see how it pans out and that will inform our future strategy.

andreasnoack:
Superseded by #475

PR #435, opened 2015-11-26
added mle implementation for beta distribution fitting

amgad-naiem:
MLE fitting for beta distribution using dL/dalpha = 0 & dL/dbeta = 0 nonlinear system of equations solving (NLsolve)

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -4,3 +4,4 @@ PDMats 0.3.2
 StatsFuns 0.1.1
 StatsBase 0.7.0
 Compat 0.4.0
+NLsolve 0.5.0
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -7,6 +7,7 @@ using PDMats
 using StatsFuns
 using StatsBase
 using Compat
+using NLsolve
 
 import Base.Random
 import Base: size, eltype, length, full, convert, show, getindex, scale, rand, rand!
--- a/src/univariate/continuous/beta.jl
+++ b/src/univariate/continuous/beta.jl
@@ -80,7 +80,18 @@ rand(d::Beta) = StatsFuns.Rmath.betarand(d.α, d.β)
 
 #### Fit model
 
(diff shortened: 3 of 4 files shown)


kmsquire:
What are people's thoughts on adding a dependency on `NLsolve`?

andreasnoack:
I'd prefer that we factor out estimation/fitting into a separate package instead of extending the support here and add a new dependency. `Distributions.jl` is already quite heavy and estimation is a big area. It is also more likely that people will disagree about how to *fit* distributions.

simonbyrne:
Agreed, I think a separate package would be appropriate for now.

kmsquire:
Sounds good.  Any thoughts on the granularity/name?  `DistributionFitting.jl` or 
`DistributionMLE.jl`?

tkelman:
+1 for spelled out words like Fitting rather than the acronym MLE

johnmyleswhite:
+1 for DistributionFitting since we already have support for MAP estimation and I've often wished that we had method of moments fitting as well.

simonbyrne:
FWIW, it should be possible to do this via Newton's method or something similar, in which case I would be happy to include it (I just don't want to add NLSolve as a dependency).

andreasnoack:
I thought we reached consensus on delegating fitting methods to a separate package. To me, it is a good separation of tasks it also separates the dependency requirements.

simonbyrne:
I agree, but we already have some fitting methods in Distributions, so it would be useful to have some standard criteria. I think a useful one would be that it doesn't add any further dependencies.

andreasnoack:
It's not urgent but I think it would make sense to migrate the existing fitting methods here to the new package as well.

amgad-naiem:
@simonbyrne NLSolve is using Newton's method and I have not found any other package other that NLSolve for doing that.

johnmyleswhite:
I think the expectation is that you'll write the Newton's method code for hand, which is what we've done in the past for other distributions.

PR #439, opened 2016-01-05
Make wishart.jl support any Real type, not just Float64

maximsch2:
Make wishart.jl support any Real type, not just Float64


andreasnoack:
LGTM

crbinz:
I just ran into this problem and was about to change it... :+1:

kmsquire:
New version tagged.

PR #441, opened 2016-01-10
fixup the docs

rawls238:
@kmsquire 

Doesn't look like this renders as I expected in the docs


rawls238:
@kmsquire could you merge this in + republish the docs when you get a chance?

kmsquire:
Merged! AFAIK, the docs are built automatically. Post back here if that's not the case.

kmsquire:
And having just merged that: you could also just write α and β directly (e.g., they can be typed using `\alpha<tab>` and `\beta<tab>` at the julia REPL or julia or latex mode in many editors).

PR #442, opened 2016-01-11
Small grammatical doc change

DanielArndt:
Either is typically used for two options. More than two is informal, and sometimes confusing.

Sorry if this is tediously small, it just tripped me up a bit.


kmsquire:
Thanks!

PR #446, opened 2016-01-15
Implement the generalised extreme value distribution

dourouc05:
I could not get how to prepare tests for a new distribution, so this has code and documentation but no tests. 

(Includes a typo in the doc for Gumbel.)


src/univariate/continuous/generalizedextremevalue.jl
```diff
@@ -0,0 +1,211 @@
...
+#### Parameters
+
+shape(d::GeneralizedExtremeValue) = d.ξ
+scale(d::GeneralizedExtremeValue) = d.σ
+location(d::GeneralizedExtremeValue) = d.μ
+params(d::GeneralizedExtremeValue) = (d.μ, d.σ, d.ξ)
+
+
+#### Statistics
+g(d::GeneralizedExtremeValue, k::Real) = gamma(1 - k * d.ξ) # This is should not be exported.
```
andreasnoack: Remove *is* in comment

andreasnoack:
Thanks for a nicely prepared pull request. However, it needs test before we can merge. Have you tried to expend [this file](https://github.com/JuliaStats/Distributions.jl/blob/master/test/continuous_test.lst) and run the script as described [here](https://github.com/JuliaStats/Distributions.jl/blob/master/test/continuous.jl)(with discrete replaced by continuous)?

dourouc05:
The problem is, this script no more exists. I guess it was replaced by prepdref.py, but I'm having a hard time to run it (it seems it's still Python 2, with a dependency on SciPy, which is a nightmare to install on Windows).

dourouc05:
After more testing on my side, either there is a SciPy convention I am not aware of, or they don't compute the same probability distribution: I checked my results against Mathematica (https://reference.wolfram.com/language/ref/MaxStableDistribution.html) and R (https://cran.r-project.org/web/packages/fExtremes/), both agree with mine, but not SciPy (except on the first test case).

andreasnoack:
I spent some time on this yesterday as well. My guess is that the script was used for the first version of the jsons, but that all the values has been modified by hand afterwards. I also realized that the SciPy values are quite far off in some cases. It is also to be expected that R has better values for these.

I think you should simply populate `continuous_test.json` by hand and then the test script should automatically parse them.

andreasnoack:
Alternatively, you can just add a separate file with appropriate tests for the this distribution and then add the filename to `runtests.jl`.

dourouc05:
Here it is, with an external test file and the script Mathematica that generated it. I restored the JSON file to current master, as it is no more used; however, I let the Python part and commented the relevant parts in the list of tests to generate. This passes the tests locally.

andreasnoack:
Thanks for the persistence here. We are getting close now. It appears that there are changes to the JSON file which I don't think should be the case since you have added a new separate test file. The changes to the `.py` file and the list of tests shouldn't be necessary either. I think it is fine with the comment about the Matematica code in the tests, but I don't think it is necessary with the notebook. Finally, please squash your commits into one (or maybe two).

PR #448, opened 2016-01-21
fix type restriction on params function

jmxpearson:
`params` was untyped for geometric, resulting in fallback for other distributions for which params weren't defined. So something like

```
d = Dirichlet(5, 1)
params(d)
```
resulted in a call here rather than method not found.


andreasnoack:
Thanks for fixing this.

PR #450, opened 2016-01-21
added params to multivariate and matrix vars

jmxpearson:
- Also mixtures and truncated distributions
- Tests enforce contract `typeof(d)(params(d)...) == d` for `isa(d,
Distribution)`
- Similar contracts hold for mixtures and truncated distributions

### DIFF
--- a/src/matrix/inversewishart.jl
+++ b/src/matrix/inversewishart.jl
@@ -11,7 +11,7 @@ end
 
 #### Constructors
 
-function InverseWishart{ST<:AbstractPDMat}(df::Real, Ψ::ST)
+function InverseWishart{ST <: AbstractPDMat}(df::Real, Ψ::ST)
     p = dim(Ψ)
     df > p - 1 || error("df should be greater than dim - 1.")
     InverseWishart{ST}(df, Ψ, _invwishart_c0(df, Ψ))
@@ -35,7 +35,7 @@ insupport(d::InverseWishart, X::Matrix{Float64}) = size(X) == size(d) && isposde
 
 dim(d::InverseWishart) = dim(d.Ψ)
 size(d::InverseWishart) = (p = dim(d); (p, p))
-
+params(d::InverseWishart) = (d.df, d.Ψ, d.c0)
 
 #### Show
 
--- a/src/matrix/wishart.jl
+++ b/src/matrix/wishart.jl
@@ -11,7 +11,7 @@ end
 
 #### Constructors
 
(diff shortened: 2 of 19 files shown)


johnmyleswhite:
LGTM. I'll give it a day to see if others have comments. Otherwise, I'll merge tomorrow.

Would be helpful to squash down to one commit prior to merging.

jmxpearson:
Cool. Squashed!

jmxpearson:
Also not sure why coveralls is throwing a fit. It's the same set of changes as before.

jmxpearson:
Also, made one more quick change:

`Wishart` and `InverseWishart` now calculate `c0` in inner constructors, so that the normalization invariant is enforced. This means that the params invariant can be preserved *without* having to return it, and `params` only needs to return truly free parameters.

I think these are the only cases like this.

johnmyleswhite:
That change seems reasonable, but is more substantive. I'm not sure if anyone is optimizing for performance by avoiding that invariant check. Thoughts, @simonbyrne? I'm not sure if there's any around these days who wrote the original code for the Wishart distributions.

jmxpearson:
I can see that. Happy to revert, if needed.

My concern was for something like extracting parameters and unrolling to be fed to something like Optim.jl. Forgetting for the moment that the optimization is more complex for positive-definite matrices, special-casing this to separating out free from calculated parameters seemed a bit of a pain.

But again, I can easily undo.

johnmyleswhite:
I personally like this change, but want to get a little more feedback.

Thanks for doing all of this. I really appreciate your thoroughness and thoughtfulness.

jmxpearson:
It's a great package. Happy to help.

simonbyrne:
I'm not so sure about including `c0` in the inner constructor: there are some cases, such as `rand` where you don't need it, so could pass a dummy value. Not sure if we want to encourage that though.

The rest looks good though, thanks!

johnmyleswhite:
Let's remove it for now. My impression is that Dahua was the person most in favor of high-performance at the occasional expense of some invariants, so perhaps we should reconsider our design stance on how much we should enforce things that aren't strictly necessary for the correctness of a subset of functions on a Distribution object?

jmxpearson:
Makes sense! Changes reverted.

johnmyleswhite:
Thanks again for doing this!

jmxpearson:
Thanks to you guys for being so responsive!

PR #451, opened 2016-01-26
Fix median(Bernoulli(0.5)). 

andreasnoack:
Medians (or quantiles in general) for discrete distributions may not be unique as for Bernoulli(0.5), but it is better to pick a value in the support than outside. This also makes `median(Bernoulli(0.5)) ==quantile(Bernoulli(0.5),0.5)`


simonbyrne:
This seems reasonable.

PR #455, opened 2016-01-30
A multivariate lognormal using MvNormal internally.

KrisDM:
A port of my private MvLogNormal class to the Distributions package. Implemented the minimum interface described in the documentation and tested using similar tests as MvNormal.

Implements the same constructors as MvNormal.

No changes to the existing code from the Distributions package (other than includes and exports).

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -9,7 +9,7 @@ using StatsBase
 using Compat
 
 import Base.Random
-import Base: size, eltype, length, full, convert, show, getindex, scale, rand, rand!
+import Base: size, eltype, length, full, convert, show, getindex, scale, scale!, rand, rand!
 import Base: sum, mean, median, maximum, minimum, quantile, std, var, cov, cor
 import Base: +, -, .+, .-
 import Base.Math.@horner
@@ -99,6 +99,7 @@ export
     MixtureModel,
     Multinomial,
     MultivariateNormal,
+    MvLogNormal,
     MvNormal,
     MvNormalCanon,
     MvNormalKnownCov,
@@ -207,6 +208,7 @@ export
     sqmahal,            # squared Mahalanobis distance to Gaussian center
     sqmahal!,           # inplace evaluation of sqmahal
     location,           # get the location parameter
+    location!,          # provide storage for the location parameter (used in multivariate distribution mvlognormal)
     mean,               # mean of distribution
     meandir,            # mean direction (of a spherical distribution)
     meanform,           # convert a normal distribution from canonical form to mean form
@@ -221,6 +223,7 @@ export
     ncomponents,        # the number of components in a mixture model
     ntrials,            # the number of trials being performed in the experiment
     params,             # get the tuple of parameters
+    params!,            # provide storage space to calculate the tuple of parameters for a multivariate distribution like mvlognormal
     pdf,                # probability density function (ContinuousDistribution)
     pmf,                # probability mass function (DiscreteDistribution)
     probs,              # Get the vector of probabilities
@@ -230,6 +233,7 @@ export
     rate,               # get the rate parameter
     sampler,            # create a Sampler object for efficient samples
     scale,              # get the scale parameter
+    scale!,             # provide storage for the scale parameter (used in multivariate distribution mvlognormal)
     shape,              # get the shape parameter
     skewness,           # skewness of the distribution
     span,               # the span of the support, e.g. maximum(d) - minimum(d)
--- a/doc/source/multivariate.rst
+++ b/doc/source/multivariate.rst
@@ -47,7 +47,7 @@ Computation of statistics
 
 .. function:: entropy(d)
 
-    Return the entropy of distribution ``d``. 
+    Return the entropy of distribution ``d``.
 
 
 Probability evaluation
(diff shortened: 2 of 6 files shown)


KrisDM:
One squashed commit with code, tests, and documentation all in one. 

This is new functionality (a new multivariate distribution) that doesn't touch existing code.

@kmsquire or  @andreasnoack would you mind checking and merging?

src/Distributions.jl
```diff
@@ -207,6 +209,7 @@ export
     sqmahal,            # squared Mahalanobis distance to Gaussian center
     sqmahal!,           # inplace evaluation of sqmahal
     location,           # get the location parameter
+    location!,          # provide storage for the location parameter (used in multivariate distribution mvlognormal)
```
kmsquire: I'm a little hesitant to export this without support for more distributions, but it does follow the pattern of many other functions defined here.  Same for `params!` and `scale!` below.

Can you open an issue for defining these for other distributions?

src/Distributions.jl
```diff
@@ -230,6 +234,7 @@ export
     rate,               # get the rate parameter
     sampler,            # create a Sampler object for efficient samples
     scale,              # get the scale parameter
+    scale!,             # provide storage for the scale parameter (used in multivariate distribution mvlognormal)
```
kmsquire: Note that `scale` and `scale!` are both defined in base, but with a different meaning (that of scaling a vector or matrix, as opposed to extracting a vector).

It's probably okay here (especially since `scale` has been defined, but in general, having the same function refer to different concepts is discouraged in Julia.  

@StefanKarpinski, I often want to cite one of your talks or an issue when saying this, but I can't find a good one--any suggestions?  Assuming, of course, that I'm not misrepresenting the idea.

(@KrisDM, this probably has no direct impact on this PR, just things to keep in mind.)

kmsquire:
LGTM.  If there aren't any other comments, I'll merge in the next day or two.

src/Distributions.jl
KrisDM: @kmsquire I opened an issue (https://github.com/JuliaStats/Distributions.jl/issues/459). The in-place variants would only be relevant to multivariate distributions, and the other multivariates that have been implemented to date may not have a location or scale, but I'll double-check it. 

I also realized that location and scale should be defined for some univariate distributions that use that terminology, yet some are missing. I opened https://github.com/JuliaStats/Distributions.jl/issues/460 for that.

src/Distributions.jl
KrisDM: @kmsquire @StefanKarpinski True, but it's an occasion where Julia's out-of-class function definitions show a problem of - excuse the pun - scalability. "scale" has a different meaning for matrices as it has for distributions. In a C++ class for the lognormal distirbution, we wouldn't think twice about using the function name (even if the standard library would contain an entirely unrelated function "scale") because the class context would define the semantics of the word, 

I have to say that one thing that surprised me about Julia is that if a method like "scale" is defined in base, and I want to also define it in my module, I have to import and then export it again to avoid warnings (which happened the last time I tried this in 0.3 - haven't tried this anymore in 0.4 so apologies if already changed). A more scalable behaviour, in my view, would be to allow separate definition in the two namespaces, and if I import the two definitions into the same workspace, let the mulitple-dispatch figure out which one to call.

StefanKarpinski: This seems like a defensibly similar usage of "scale" to me, so I would just extend Base's versions of these.

StefanKarpinski: But if you don't feel so, don't listen to me – I'm not an expert on this domain.

src/Distributions.jl
```diff
@@ -44,6 +44,7 @@ export
     ContinuousMultivariateDistribution,
     ContinuousMatrixDistribution,
     SufficientStats,
+    AbstractMvLogNormal,
```
andreasnoack: I think this can wait until there'll be more subtypes.

andreasnoack:
@KrisDM Generally a good PR. The documentation and tests are great. I'm in doubt about the special `location!` and `scale!` methods. They seem very specific to the log-normal. I haven't done any work with the log-normal. Are these conversion methods frequently used in applications?

I also know it is a convention for log-normals, but it also a bit weird that `location` and `scale` are the location and scale parameters for all distributions in in `Distributions.jl` (for which a method is defined) *but* the log-normal distribution.

Finally, another time please remove whitespace in a separate request. It adds a lot of noise to the diff.

src/Distributions.jl
KrisDM: OK, will remove.

KrisDM:
@andreasnoack Thanks. 

I'll remove the whitespace, will resubmit later this afternoon (I blame Juno with its default setting of 2 :-) - I have changed it now to 4).

location! and scale!. As pattern it follows sqmahal!, which I think is also defined for one distribution only (MvNormal).

As for the conversions itself:
- For mean and cov, the conversions are commonly known (both wiki and MATLAB documentation define them), but I don't know how commonly they are used, given the lognormal itself has a rather small niche (mostly in ecology as I understand from the literature). 
- Centering the lognormal on the median (not on the location or mean) is what I've seen used in the MCMC ecology literature, so for that case we have to calculate the location from median and scale. 
- I decided to implement the mode as well, as I may want to try it at a later point.

An alternative solution to providing location! and scale! would be to provide additional constructors, and supply an optional symbol to signal how to interpret the parameters (e.g., :locscale as the default, :meancov, :meanscale, :medianscale and :modescale for the ones relying on the conversions). I would then leave the internal _location! and _scale! functions to do the actual calculations, not export the location! and scale! functions, and leave the location and scale functions exposed for use cases that need a conversion without creating a distribution. 

I agree it's ironic that location and scale were not defined for univariate LogNormal. Opened an issue to look into it.

KrisDM:
I cleaned up the indentation and the AbstractMvLogNormal. Ready to go in my view...

PR #461, opened 2016-02-08
Implement generalised extreme value distribution.

dourouc05:
Replaces https://github.com/JuliaStats/Distributions.jl/pull/446


andreasnoack:
Thanks for contributing this.

PR #463, opened 2016-02-09
Hypergeometric: allow `n` to take 0 and ns+nf values

alyst:
When `n=0` or `n=ns+nf`, the distribution is degenerated, but still defined.
Improves corner cases support in `FisherExactTest`.


andreasnoack:
:+1:

PR #464, opened 2016-02-09
Use syrk instead of gemm for computing covariance matrices

simonster:
This is roughly twice as fast. Ideally PDMats would wrap Symmetric/Hermitian matrices so the copytri! step would not be necessary, but for reasonably large numbers of data points its cost is negligible anyways.

### DIFF
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -257,7 +257,8 @@ function fit_mle(D::Type{FullNormal}, x::AbstractMatrix{Float64})
     n = size(x, 2)
     mu = vec(mean(x, 2))
     z = x .- mu
-    C = Base.LinAlg.BLAS.gemm('N', 'T', 1.0/n, z, z)
+    C = Base.LinAlg.BLAS.syrk('U', 'N', 1.0/n, z)
+    Base.LinAlg.copytri!(C, 'U')
     MvNormal(mu, PDMat(C))
 end
 
@@ -276,7 +277,8 @@ function fit_mle(D::Type{FullNormal}, x::AbstractMatrix{Float64}, w::AbstractVec
             @inbounds z[i,j] = (x[i,j] - mu[i]) * cj
         end
     end
-    C = Base.LinAlg.BLAS.gemm('N', 'T', inv_sw, z, z)
+    C = Base.LinAlg.BLAS.syrk('U', 'N', inv_sw, z)
+    Base.LinAlg.copytri!(C, 'U')
     MvNormal(mu, PDMat(C))
 end
 


dmbates:
@simonster Did you actually time syrk versus gemm?  Early on we timed them and found that gemm is actually faster than syrk for calculating such a product.  The author of OpenBLAS said that they really hadn't done much work on syrk whereas gemm was optimized to death.

simonster:
I did, albeit with threads disabled because I'm still having trouble with OpenBLAS threading, and it really does seem to be much faster:

```julia
julia> Base.disable_threaded_libs()

julia> x = randn(1000, 100000);

julia> @time LinAlg.copytri!(Base.LinAlg.BLAS.syrk('U', 'N', 1.0, x), 'U');
  3.437264 seconds (60.48 k allocations: 10.447 MB)

julia> @time LinAlg.copytri!(Base.LinAlg.BLAS.syrk('U', 'N', 1.0, x), 'U');
  3.335116 seconds (9 allocations: 7.630 MB, 0.30% gc time)

julia> @time LinAlg.copytri!(Base.LinAlg.BLAS.syrk('U', 'N', 1.0, x), 'U');
  3.313992 seconds (9 allocations: 7.630 MB)

julia> @time BLAS.gemm('N', 'T', 1.0, x, x);
  6.481310 seconds (17.60 k allocations: 8.464 MB, 0.50% gc time)

julia> @time BLAS.gemm('N', 'T', 1.0, x, x);
  6.289041 seconds (9 allocations: 7.630 MB)

julia> @time BLAS.gemm('N', 'T', 1.0, x, x);
  6.226366 seconds (9 allocations: 7.630 MB
```

StefanKarpinski:
Maybe @xianyi and co. got a chance to do some work on optimizing `syrk` since then?

kmsquire:
Or threading changes the timing?  (Answer, well, not really)

```julia
julia> x = randn(1000, 100000);

julia> @time LinAlg.copytri!(Base.LinAlg.BLAS.syrk('U', 'N', 1.0, x), 'U');
  1.334999 seconds (44.53 k allocations: 9.679 MB)

julia> @time LinAlg.copytri!(Base.LinAlg.BLAS.syrk('U', 'N', 1.0, x), 'U');
  1.183293 seconds (9 allocations: 7.630 MB, 0.97% gc time)

julia> @time LinAlg.copytri!(Base.LinAlg.BLAS.syrk('U', 'N', 1.0, x), 'U');
  1.179436 seconds (9 allocations: 7.630 MB)

julia> @time BLAS.gemm('N', 'T', 1.0, x, x);
  1.780006 seconds (27.39 k allocations: 8.824 MB)

julia> @time BLAS.gemm('N', 'T', 1.0, x, x);
  1.772514 seconds (9 allocations: 7.630 MB, 1.99% gc time)

julia> @time BLAS.gemm('N', 'T', 1.0, x, x);
  1.784444 seconds (9 allocations: 7.630 MB)
```

@simonster, just curious what is the problem you're having with threading?

PR #466, opened 2016-02-10
Update PDMats.jl dependency.

simonbyrne:


REQUIRE
```diff
@@ -1,6 +1,6 @@
 julia 0.3
 ArrayViews 0.4.12
-PDMats 0.3.2 0.4
+PDMats 0.4 0.5
```
andreasnoack: Shouldn't `PDMats 0.4` be sufficient here?

simonbyrne: I'm trying to continue the upper bound thing so that next time we break the api, we can upgrade the dependencies cleanly.

Issue #462, opened 2016-02-09
negative non-centrality parameter for NoncentralT


Ph0non:
If I'm not missing something, it's not neccessary for the noncentral-T-distribution to have only positiv noncentrality parameter.

`@check_args(NoncentralT, λ >= zero(λ))`

andreasnoack:
No. I agree with you here. This is probably an error. Have you checked R?

Ph0non:
In R it's well defined. Also Wikipedia states, that the noncentrality parameter may be negative. I can't see any consequences for the rest of NoncentralT for a negative parameter.

It should only be checked, if λ is real.

andreasnoack:
@Ph0non Would you mind opening a PR with a fix for this?

PR #470, opened 2016-02-20
fix noncentral-T

Ph0non:
fix noncentral-T for negative noncentrality parameter #462

### DIFF
--- a/src/univariate/continuous/noncentralt.jl
+++ b/src/univariate/continuous/noncentralt.jl
@@ -4,7 +4,7 @@ immutable NoncentralT <: ContinuousUnivariateDistribution
 
     function NoncentralT(ν::Real, λ::Real)
     	@check_args(NoncentralT, ν > zero(ν))
-        @check_args(NoncentralT, λ >= zero(λ))
+        @check_args(NoncentralT, isreal(λ))
         new(ν, λ)
     end
 end


johnmyleswhite:
This is a great change. Can we get some tests?

Ph0non:
I didn't understand. Is this irony? The change is really small and didn't lead to any unsuspected result, because in `noncentralt.jl` happens nothing spectacular with λ.

In my local branch it works as expected to calculate the UCL of an .95-percentile.

johnmyleswhite:
No, it's not irony. Please write a test.

jiahao:
Every change is appreciated. If you had an example which led you to discover this bug, it would be valuable to put that example in the test suite, so that future code changes won't break it by accident.

nalimilan:
Welcome, and don't worry, people are seldom sarcastic here. :-)

While you're at it, it would be great to have a more explicit commit message. Just saying what you mentioned in https://github.com/JuliaStats/Distributions.jl/issues/462 about the distribution's support would be good.

Ph0non:
I'm not very familiar with git. As I see, I need to amend this commit, to edit the message. Is this also possible, if I commit over github?

Would this be enough for a test?
```
using Distributions
using Base.Test

# non-central T-Distributions
ν = 50
λ = -5

d = NoncentralT(ν, λ)
@test_throws MethodError NoncentralT(ν, complex(λ))
@test d == typeof(d)(params(d)...)
@test_approx_eq_eps mean(d) -5.0766 1e-5

λ = 5
d = NoncentralT(ν, λ)
@test_approx_eq_eps mean(d) 5.0766 1e-5

λ = 0
d = NoncentralT(ν, λ)
@test var(d) == var(TDist(ν))
```

kleinschmidt:
That's fine, you'll just have to "git push --force" because you've overwritten a commit you already pushed. Use this power with great caution :)

PR #475, opened 2016-03-18
prototype of parameterized Normal distribution

jmxpearson:
- for concrete reference in discussion at #430

### DIFF
--- a/src/truncated/normal.jl
+++ b/src/truncated/normal.jl
@@ -8,20 +8,20 @@ TruncatedNormal(mu::Real, sigma::Real, a::Real, b::Real) =
 
 ### statistics
 
-minimum(d::Truncated{Normal,Continuous}) = d.lower
-maximum(d::Truncated{Normal,Continuous}) = d.upper
+minimum{T <: Real}(d::Truncated{Normal{T},Continuous}) = d.lower
+maximum{T <: Real}(d::Truncated{Normal{T},Continuous}) = d.upper
 
 
-function mode(d::Truncated{Normal,Continuous})
+function mode{T <: Real}(d::Truncated{Normal{T},Continuous})
     μ = mean(d.untruncated)
     d.upper < μ ? d.upper :
     d.lower > μ ? d.lower : μ
 end
 
-modes(d::Truncated{Normal,Continuous}) = [mode(d)]
+modes{T <: Real}(d::Truncated{Normal{T},Continuous}) = [mode(d)]
 
 
-function mean(d::Truncated{Normal,Continuous})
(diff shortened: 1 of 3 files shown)


mlubin:
This seems reasonable to me. Any opinions from those with commit access?

andreasnoack:
Why is it that we need to introduce `GenericNormal`? Wouldn't it be okay to make `Normal` parametric and define a constructor that defaults to `Float64` when the type is not specified?

jmxpearson:
Correct me if I'm wrong, but isn't it because otherwise 

```julia
type Foo
    a::Normal
    b::Float64
end
```
no longer has all its fields as leaf types?

johnmyleswhite:
That's true, but you just need to do:

```
type Foo{T}
    a::Normal{T}
    b::Float64
end
```

So we can affect downstream code with this change, but that code can get fixed relatively automatically (i.e. it's lintable).

jmxpearson:
Okay. Because there were comments from @lindahua in previous discussion that this sort of breakage was a no-go.

andreasnoack:
Technically, I don't think it's a breakage but only a performance regression and I think we are too young to have that strict rules.

jmxpearson:
Okay, cf. new changes in the revised PR.

andreasnoack:
Thanks. I like this. Maybe someone could try out AD on this branch and see if it works.

jmxpearson:
Just added similar prototype code for `MvNormal`. Won't work for differentiation on covariances until JuliaStats/PDMats.jl/pull/41 is merged, but should work for means now.

I'm happy to write some tests for this (I've checked out that `Normal` does work with ForwardDiff), but is there someplace they should live longer-term?

jmxpearson:
FYI, [these tests](https://gist.github.com/jmxpearson/15b447e36ecd4cacf1e1) all pass.

andreasnoack:
I've tagged a version for `PDMats` so soon it should be possible to update the version requirement here. Maybe you should split up this PR in the scalar and multivariate normal distributions. The scalar case can probably be merged right away. Then open a new PR with the multivariate change that also requires a higher version of `PDMats`

jmxpearson:
Done. I'll submit the other PR once this one gets merged so I can make the rebase easier on myself.

Also, moving forward: I'm hoping to start marching through the other distributions, generalizing them in the same way. Should I really open a separate PR for each, or is it possible to open a PR for, say, all the continuous univariates together?

andreasnoack:
Thanks. No, it shouldn't be necessary to open a separate one for each distribution. Here, it would just make it easier to progress since only on of the distributions required a dependency version change.

PR #479, opened 2016-03-31
fixed bug in betaprim constructor for single argument

jonalm:
Fixing bug

```
julia> BetaPrime(3)
ERROR: StackOverflowError:
 in call at /Users/jonalm/.julia/v0.4/Distributions/src/univariate/continuous/betaprime.jl:9 (repeats 80000 times)
```


johnmyleswhite:
Thanks for the fix!

PR #481, opened 2016-04-07
Improved generalised extreme values distribution

dourouc05:
Follows #461. Fixes one bug in a statistical moment (threw an exception), and improves test coverage.


andreasnoack:
:+1:

PR #486, opened 2016-04-19
added entropy functionality to multinomial.jl

esproff:
designed function to be as overflow resistant as possible.  The derivation can be found here:
http://stats.stackexchange.com/questions/207893/entropy-of-the-multinomial-distribution/208075#208075

### DIFF
--- a/src/multivariate/multinomial.jl
+++ b/src/multivariate/multinomial.jl
@@ -86,6 +86,17 @@ function cf(d::Multinomial, t::AbstractVector)
     return s^n
 end
 
+function entropy(d::Multinomial)
+	n, p = params(d)
+	s = -sum(log(2:n)) - n*sum(log(p.^p))
+	for pr in p
+		for x in 0:n
+			s += binomial(n, x) * pr^x * (1 - pr)^(n - x) * sum(log(2:x))
+		end
+	end
+	return s
+end
+
 
 # Evaluation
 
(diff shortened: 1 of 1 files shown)


src/multivariate/multinomial.jl
```diff
@@ -86,6 +86,17 @@ function cf(d::Multinomial, t::AbstractVector)
     return s^n
 end
 
+function entropy(d::Multinomial)
+    n, p = params(d)
+    s = -sum(log(2:n)) - n*sum(log(p.^p))
```
simonbyrne: You can use `lgamma` and `entropy(d.p)` here

src/multivariate/multinomial.jl
```diff
@@ -86,6 +86,17 @@ function cf(d::Multinomial, t::AbstractVector)
...
+    for pr in p
+        for x in 0:n
+            s += binomial(n, x) * pr^x * (1 - pr)^(n - x) * sum(log(2:x))
```
simonbyrne: It would be better to use the binomial pdf function directly: this is more numerically accurate than writing it out manually. You can also use `lgamma` for the last term.

simonbyrne:
Thanks for doing this. See my inline notes.

esproff:
Hi Simon,

I made the changes you suggested and merged the pull request (I believe that's the way I was supposed to do it, I'm new to this).

andreasnoack:
@simonbyrne Do you have further comments?

simonbyrne:
LGTM

PR #487, opened 2016-04-19
Update README.md (Add a badge for the license)

alexndrejoly:


KristofferC:
Does this provide any extra information of is it just adding a badge because badges are pretty? Anyone who is interested in the licence can just look at the `LICENSE.md` file.

ViralBShah:
I agree. I think a license badge is not really necessary.

ViralBShah:
I don't mean to discourage the contribution, but just echoing my thoughts here.

simonbyrne:
I don't really see what the licence badge adds.

PR #491, opened 2016-05-02
Change randbool to rand(Bool) in vonmises sampler

jkbest2:
As per JuliaLang/julia#9569, `randbool()` has been deprecated in favor of `rand(Bool)`.


johnmyleswhite:
LGTM. Will see if others comment or will merge tomorrow.

PR #503, opened 2016-06-01
logpdfs handle mixed eltype inputs

jmxpearson:
Companion work to JuliaStats/PDMats.jl#45 to allow for automatic differentiation of logpdfs, etc.

- Removes explicity Float64 casting in MvNormal
- defines a new utility function, `partype` that returns a type capable of holding all the element types of a distribution's parameters
- when `logpdf(d, X)`, etc. are called, the storage array passed to their in-place versions is of type `promote_type(partype(d), eltype(X))`

### DIFF
--- a/doc/source/multivariate.rst
+++ b/doc/source/multivariate.rst
@@ -144,12 +144,12 @@ We realize that the mean vector and the covariance often have special forms in p
 
 .. code-block:: julia
 
-    immutable MvNormal{Cov<:AbstractPDMat,Mean<:Union{Vector{Float64},ZeroVector{Float64}}} <: AbstractMvNormal
+    immutable MvNormal{Cov<:AbstractPDMat,Mean<:Union{Vector,ZeroVector}} <: AbstractMvNormal
         μ::Mean
         Σ::Cov
     end
 
-Here, the mean vector can be an instance of either ``Vector{Float64}`` or ``ZeroVector{Float64}``, where the latter is simply an empty type indicating a vector filled with zeros. The covariance can be of any subtype of ``AbstractPDMat``. Particularly, one can use ``PDMat`` for full covariance, ``PDiagMat`` for diagonal covariance, and ``ScalMat`` for the isotropic covariance -- those in the form of :math:`\sigma \mathbf{I}`. (See the Julia package `PDMats <https://github.com/lindahua/PDMats.jl>`_ for details).
+Here, the mean vector can be an instance of either ``Vector`` or ``ZeroVector``, where the latter is simply an empty type indicating a vector filled with zeros. The covariance can be of any subtype of ``AbstractPDMat``. Particularly, one can use ``PDMat`` for full covariance, ``PDiagMat`` for diagonal covariance, and ``ScalMat`` for the isotropic covariance -- those in the form of :math:`\sigma \mathbf{I}`. (See the Julia package `PDMats <https://github.com/lindahua/PDMats.jl>`_ for details).
 
 We also define a set of alias for the types using different combinations of mean vectors and covariance:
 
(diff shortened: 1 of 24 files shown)


jmxpearson:
The error here is that tests are timing out on nightly. I'm leaving this up for code review and then will try again to see if I can get the all-clear before merging.

doc/source/multivariate.rst
```diff
@@ -144,12 +144,12 @@ We realize that the mean vector and the covariance often have special forms in p
 
 .. code-block:: julia
 
-    immutable MvNormal{Cov<:AbstractPDMat,Mean<:Union{Vector{Float64},ZeroVector{Float64}}} <: AbstractMvNormal
+    immutable MvNormal{Cov<:AbstractPDMat,Mean<:Union{Vector,ZeroVector}} <: AbstractMvNormal
```
andreasnoack: It might be necessary to add the element type as a type parameter here if you want to use it for dispatch. This is similar to e.g. how `SubArray`s are parameterized.

src/multivariate/mvnormal.jl
```diff
@@ -36,20 +36,20 @@ entropy(d::AbstractMvNormal) = 0.5 * (length(d) * (Float64(log2π) + 1.0) + logd
 
 mvnormal_c0(g::AbstractMvNormal) = -0.5 * (length(g) * Float64(log2π) + logdetcov(g))
 
-sqmahal{T<:Real}(d::AbstractMvNormal, x::AbstractMatrix{T}) = sqmahal!(Array(Float64, size(x, 2)), d, x)
+sqmahal(d::AbstractMvNormal, x::AbstractMatrix) = sqmahal!(Array(Float64, size(x, 2)), d, x)
```
andreasnoack: Maybe avoid hard coded `Float64` conversions when we try to go generic.

src/multivariate/mvnormal.jl
```diff
@@ -36,20 +36,20 @@ entropy(d::AbstractMvNormal) = 0.5 * (length(d) * (Float64(log2π) + 1.0) + logd
 
 mvnormal_c0(g::AbstractMvNormal) = -0.5 * (length(g) * Float64(log2π) + logdetcov(g))
```
andreasnoack: Use `/2` instead of `*0.5` to avoid `Float64` promotion.

src/multivariate/mvnormal.jl
```diff
@@ -36,20 +36,20 @@ entropy(d::AbstractMvNormal) = 0.5 * (length(d) * (Float64(log2π) + 1.0) + logd
...
 
-sqmahal{T<:Real}(d::AbstractMvNormal, x::AbstractMatrix{T}) = sqmahal!(Array(Float64, size(x, 2)), d, x)
+sqmahal(d::AbstractMvNormal, x::AbstractMatrix) = sqmahal!(Array(Float64, size(x, 2)), d, x)
 
-_logpdf{T<:Real}(d::AbstractMvNormal, x::AbstractVector{T}) = mvnormal_c0(d) - 0.5 * sqmahal(d, x)
+_logpdf(d::AbstractMvNormal, x::AbstractVector) = mvnormal_c0(d) - 0.5 * sqmahal(d, x)
```
andreasnoack: Use `/2`.

andreasnoack:
Do the tests pass locally on master?

doc/source/multivariate.rst
jmxpearson: Just to make sure I understand, something like
```julia
immutable MvNormal{Cov<:AbstractPDMat, Mean<:Union{Vector,ZeroVector, T<:Real} <: AbstractMvNormal
```
and then outer constructors set `T = promote_type(eltype(μ), eltype(Σ))`?

Right now, μ and Σ can have different element types.

andreasnoack: Yes. Typically it's the first argument so maybe follow that convention here as well. I think we should restrict μ and Σ to have the same element type to simplify things. When would it be useful to allow for that?

src/multivariate/mvnormal.jl
jmxpearson: Thanks. I thought I'd caught most of these.

jmxpearson:
Yes, passes on my machine (recent OSX) on master.

Also, a heads up: I have a student (@Halmoni100) working on a PR with about 20 more distributions ready to go (most of the discrete and continuous univariates). Most things will be done the same way as here, but there are a couple of edge cases we've been going back and forth on that I'd like to catch you about next week, if possible.

andreasnoack:
Great with all that work being done. We should definitely find some time for discussing this next week.

doc/source/multivariate.rst
jmxpearson: I don't have a strong opinion, but the case I had in mind was one where you could save memory if you knew you only needed derivatives with respect to μ, say, and Σ was constant. If both get promoted to the same element type, it's full gradients across the board.

This may be too much worry for a case nobody cares about, though. I'm happy to simply add the type parameter.

andreasnoack: That's a good example. However, it would complicate the type significantly since it would require two element type parameters for full dispatch generality so, at least initially until a demand appears, I think it would be better to restrict the parameters.

jmxpearson: That took more work than I thought, but see the latest commit. Also fixed GenericMvTDist and MvNormalCanon so that eltypes are the same everywhere and eltype is the first type parameter.

src/multivariate/mvnormalcanon.jl
```diff
@@ -2,41 +2,45 @@
...
+immutable MvNormalCanon{T<:Real,P<:AbstractPDMat,V<:Union{Vector,ZeroVector}} <: AbstractMvNormal
     μ::V    # the mean vector
     h::V    # potential vector, i.e. inv(Σ) * μ
     J::P    # precision matrix, i.e. inv(Σ)
 end
 
-typealias FullNormalCanon MvNormalCanon{PDMat{Float64,Matrix{Float64}},Vector{Float64}}
-typealias DiagNormalCanon MvNormalCanon{PDiagMat{Float64,Vector{Float64}},Vector{Float64}}
-typealias IsoNormalCanon  MvNormalCanon{ScalMat{Float64},Vector{Float64}}
+typealias FullNormalCanon MvNormalCanon{Float64, PDMat{Float64,Matrix{Float64}},Vector{Float64}}
```
andreasnoack: Shouldn't these be parametric on the element type as well?

jmxpearson: Perhaps. I had worried that these concrete types were used in the mle fitting and I didn't want to wade too much into that at first. I suggest that we merge this and get that on another pass.

jmxpearson:
Note that this PR requires JuliaStats/PDMats.jl#46

src/multivariate/mvnormalcanon.jl
andreasnoack: Fine with me

andreasnoack:
Then a new version of `PDMats` should be tagged and added as a requirement here.

jmxpearson:
Can do.

jmxpearson:
I'll run tests locally on master tonight. If everything passes there, should I just go ahead and merge?

andreasnoack:
I'm not sure. It wouldn't be great to have the tests failing on master.
Please try with `Pkg.test("Distributions", coverage=true)` to have the same
setting as Travis, i.e. without inlining.

jmxpearson:
I'll check that. I'm also happy to hold off until the timing on the tests can be brought down.

PR #504, opened 2016-06-07
Replace deprecated call to hist

dmbates:
`hist` is deprecated in v0.5 so this line threw a lot of warnings in `Pkg.test("Distributions")`


jmxpearson:
Thanks!

Issue #505, opened 2016-06-14
Parallelized testing swallowing errors


jmxpearson:
I discovered today that when running tests in parallel, not all errors are shown. In fact, tests may pass on Travis run in parallel but not when run serially. I've been testing by disabling parallelization locally.

For an example, see the jp/partests branch. That branch fails tests run serially but passes (on my machine) in parallel.

PR #506, opened 2016-06-27
Import shape function, fix deprecation warnings

ranjanan:
Base has a `shape` function, so there would be a conflict. This fixes that.

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -8,7 +8,7 @@ using StatsBase
 using Compat
 
 import Base.Random
-import Base: size, eltype, length, full, convert, show, getindex, scale, scale!, rand, rand!
+import Base: size, eltype, length, full, convert, show, getindex, scale, scale!, rand, rand!, shape
 import Base: sum, mean, median, maximum, minimum, quantile, std, var, cov, cor
 import Base: +, -, .+, .-
 import Base.Math.@horner
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -234,7 +234,6 @@ export
     sampler,            # create a Sampler object for efficient samples
     scale,              # get the scale parameter
     scale!,             # provide storage for the scale parameter (used in multivariate distribution mvlognormal)
-    shape,              # get the shape parameter
     skewness,           # skewness of the distribution
     span,               # the span of the support, e.g. maximum(d) - minimum(d)
     std,                # standard deviation of distribution
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -8,7 +8,10 @@ using StatsBase
 using Compat
 
 import Base.Random
-import Base: size, eltype, length, full, convert, show, getindex, scale, scale!, rand, rand!, shape
+import Base: size, eltype, length, full, convert, show, getindex, scale, scale!, rand, rand!
+if VERSION > v"0.4"
+    import Base.shape
+end
 import Base: sum, mean, median, maximum, minimum, quantile, std, var, cov, cor
 import Base: +, -, .+, .-
 import Base.Math.@horner
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -7,6 +7,7 @@ using StatsFuns
 using StatsBase
 using Compat
 
+import Compat.view 
 import Base.Random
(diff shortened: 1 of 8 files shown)


ranjanan:
@simonbyrne any idea why coverage has decreased by 42% ( 😱 )?

jmxpearson:
Tests are currently failing on nightly because of timeout on Travis. It's on my to-do list, but any help would be appreciated. As far as I can tell, it's mostly due to deprecation warnings external to Distributions.

ranjanan:
@jmxpearson Which packages do you think the external deprecations are coming from? I'll try pushing to fix them.

jmxpearson:
Actually, looking at the Travis log, there appear to be a few instances of `slice` remaining (should be `view`). Is Compat currently handling that?

Also: WARNING: Method definition (::Type{ArrayViews.UnsafeContiguousView})(Ptr{#T<:Any}, Int64, Tuple{Vararg{Int64, #N<:Any}}) in module ArrayViews at /home/travis/.julia/v0.5/ArrayViews/src/arrviews.jl:18 overwritten at /home/travis/.julia/v0.5/ArrayViews/src/arrviews.jl:26.

Other than that, things look okay. Note also that #505 is also slowing you down, since one of the workers is quitting.

jmxpearson:
This is great. Thanks! I found another test error that is failing on my machine for nightly. Once I can get that fixed, I'll merge this, #505, and that fix together.

ranjanan:
I'm actually sending a Compat PR (https://github.com/JuliaLang/Compat.jl/pull/238) to fix the view issue so it's compatible on release. Once that is merged, I might have to add a commit here saying `import Compat.view

jmxpearson:
Great. I'll hold off. I'll probably also ask you to squash before I merge.

ranjanan:
@jmxpearson since the new Compat version has now been tagged and merged into METADATA, could you restart Travis please?

src/Distributions.jl
```diff
@@ -7,6 +7,7 @@ using StatsFuns
 using StatsBase
 using Compat
 
+import Compat.view 
 import Base.Random
 import Base: size, eltype, length, full, convert, show, getindex, scale, scale!, rand, rand!
 if VERSION > v"0.4"
```
jmxpearson: Don't you want something like:

```julia
if VERSION < v"0.5.0-"
    export shape
else
    import Base.shape
end
```

jmxpearson:
I could, but I don't think it will matter much until #507 is dealt with. My plan is to work on that this morning, then merge this and #508. After that, we should be good with 0.5.

src/Distributions.jl
ranjanan: If it's v0.4, it's going to pick from compat. Nothing happens in v0.5. So this would work.

andreasnoack: I guess it would be a problem that
```jl
julia> v"0.4.1"> v"0.4"
true
```

ranjanan: @andreasnoack you're right. Let me change that to ` < v"0.5.0-" ` instead.

jmxpearson:
Restarted Travis, since JuliaLang/julia#17185 merged. If all goes well, will close #507.

jmxpearson:
Looks like I jumped the gun. I may need to give nightly [a little time](https://status.julialang.org/) to percolate.

jmxpearson:
@ranjanan Master is throwing a real error now. Cf the Travis logs, but I think it's complaining about `shape(d::Erlang)`.

ranjanan:
The error log says `indicesbehavior(d::Erlang)`

jmxpearson:
Right, but the trace seems to relate it to shape. I don't get this error on the jp/partests branch, where I've made most of the same fixes. If you don't have time to look at it right now, I can include your changes from here in that PR  (#508).

ranjanan:
@jmxpearson I apologize, I'm actually travelling a fair bit right now, so I might not be as responsive as I want to be. Please feel free to take commits/whatever changes you need. Thanks!

jmxpearson:
No worries. I'll take a look.

jmxpearson:
Superseded by #508

PR #509, opened 2016-06-29
Refactor slice->view, plus some more v0.5 changes

davidavdav:
In julia-v0.5, `slice` and `sub` are covered by `view`.


jmxpearson:
Thanks, but I think this will be taken care of in #506.

PR #511, opened 2016-07-06
Dual number compatibility

Halmoni100:
Added compatibility for dual numbers while resolving type instabilities.

Contributed changes to univariate distributions and unit tests, while @jmxpearson contributed to other distributions and some other bug fixes.


jmxpearson:
Note: This is a big PR and I've reviewed in-house, but we'd appreciate another pair of eyes on this before I merge. 

Chris has done a lot of work to ensure that the basic univariate API remains type-stable for, e.g., Float32 and DualNumbers, though some special functions used in the package (`besseli`, `besselk`, etc.) ccall out to Float64-only versions in base, so not all type parameter choices will work for all distributions.

jmxpearson:
Oops. Inadvertently included old versions of `mvnormal.jl` and `mvnormalcanon.jl`. Will push corrections shortly.

src/multivariate/mvnormal.jl
```diff
@@ -26,30 +26,30 @@ abstract AbstractMvNormal <: ContinuousMultivariateDistribution
...
 
-insupport(d::AbstractMvNormal, x::AbstractVector) =
+insupport{T<:Real}(d::AbstractMvNormal, x::AbstractVector{T}) =
     length(d) == length(x) && allfinite(x)
 
 mode(d::AbstractMvNormal) = mean(d)
 modes(d::AbstractMvNormal) = [mean(d)]
 
-entropy(d::AbstractMvNormal) = (length(d) * (Float64(log2π) + 1) + logdetcov(d))/2
+entropy(d::AbstractMvNormal) = 0.5 * (length(d) * (Float64(log2π) + 1.0) + logdetcov(d))
```
andreasnoack: Avoid `Float64` literals since it will promote in cases where it is not necessary. You can just divide with `2`. LLVM is smart enough to optimize it to a multiplication when possible, e.g.
```jl


julia> f(x) = x/2

julia> @code_llvm f(2.0)

define double @julia_f_68028(double) #0 {
top:
  %1 = fmul double %0, 5.000000e-01
  ret double %1
}
```
 `log2π` is bit tricky so maybe start out with computing `ldcd = logdetcov(d)` and use `typeof(ldcd)(log2π)`.

src/multivariate/mvnormal.jl
```diff
@@ -26,30 +26,30 @@ abstract AbstractMvNormal <: ContinuousMultivariateDistribution
...
 
-mvnormal_c0(g::AbstractMvNormal) = -(length(g) * Float64(log2π) + logdetcov(g))/2
+mvnormal_c0(g::AbstractMvNormal) = -0.5 * (length(g) * Float64(log2π) + logdetcov(g))
```
andreasnoack: Same

src/multivariate/mvnormal.jl
```diff
@@ -26,30 +26,30 @@ abstract AbstractMvNormal <: ContinuousMultivariateDistribution
...
 
-sqmahal(d::AbstractMvNormal, x::AbstractMatrix) = sqmahal!(Array(promote_type(partype(d), eltype(x)), size(x, 2)), d, x)
+sqmahal{T<:Real}(d::AbstractMvNormal, x::AbstractMatrix{T}) = sqmahal!(Array(Float64, size(x, 2)), d, x)
```
andreasnoack: Why do you hard code `Float64` here?

src/multivariate/mvnormal.jl
```diff
@@ -26,30 +26,30 @@ abstract AbstractMvNormal <: ContinuousMultivariateDistribution
...
 
-_logpdf(d::AbstractMvNormal, x::AbstractVector) = mvnormal_c0(d) - sqmahal(d, x)/2
+_logpdf{T<:Real}(d::AbstractMvNormal, x::AbstractVector{T}) = mvnormal_c0(d) - 0.5 * sqmahal(d, x)
```
andreasnoack: Use `/2`

src/multivariate/mvnormal.jl
```diff
@@ -26,30 +26,30 @@ abstract AbstractMvNormal <: ContinuousMultivariateDistribution
...
 
-function _logpdf!(r::AbstractArray, d::AbstractMvNormal, x::AbstractMatrix)
+function _logpdf!{T<:Real}(r::AbstractArray, d::AbstractMvNormal, x::AbstractMatrix{T})
     sqmahal!(r, d, x)
-    c0 = mvnormal_c0(d)
+    c0::Float64 = mvnormal_c0(d)
```
andreasnoack: Why type assert?

src/multivariate/mvnormal.jl
```diff
@@ -26,30 +26,30 @@ abstract AbstractMvNormal <: ContinuousMultivariateDistribution
...
     for i = 1:size(x, 2)
-        @inbounds r[i] = c0 - r[i]/2
+        @inbounds r[i] = c0 - 0.5 * r[i]
```
andreasnoack: `/2`

src/multivariate/mvnormal.jl
```diff
@@ -59,42 +59,37 @@ _pdf!(r::AbstractArray, d::AbstractMvNormal, x::AbstractMatrix) = exp!(_logpdf!(
...
     μ::Mean
     Σ::Cov
 end
 
 const MultivariateNormal = MvNormal  # for the purpose of backward compatibility
 
-typealias IsoNormal  MvNormal{Float64,ScalMat{Float64},Vector{Float64}}
-typealias DiagNormal MvNormal{Float64,PDiagMat{Float64,Vector{Float64}},Vector{Float64}}
-typealias FullNormal MvNormal{Float64,PDMat{Float64,Matrix{Float64}},Vector{Float64}}
+typealias IsoNormal  MvNormal{ScalMat{Float64},Vector{Float64}}
```
andreasnoack: I think this might be reverting @jmxpearson's recent changes to `MvNormal`. I think it is necessary to have the element type as a parameter.

src/multivariate/mvnormal.jl
jmxpearson: Sorry, these are changes we'd already made on master. Will push correct versions up momentarily.

src/multivariate/mvnormal.jl
```diff
@@ -122,19 +117,19 @@ logdetcov(d::MvNormal) = logdet(d.Σ)
 
 ### Evaluation
 
-sqmahal(d::MvNormal, x::AbstractVector) = invquad(d.Σ, x - d.μ)
+sqmahal{T<:Real}(d::MvNormal, x::AbstractVector{T}) = invquad(d.Σ, x - d.μ)
```
andreasnoack: Avoid parameters when possible. The original version should be fine.

src/multivariate/mvnormal.jl
```diff
@@ -122,19 +117,19 @@ logdetcov(d::MvNormal) = logdet(d.Σ)
...
 
-sqmahal!(r::AbstractVector, d::MvNormal, x::AbstractMatrix) =
+sqmahal!{T<:Real}(r::AbstractVector, d::MvNormal, x::AbstractMatrix{T}) =
```
andreasnoack: Same

src/multivariate/mvnormal.jl
```diff
@@ -122,19 +117,19 @@ logdetcov(d::MvNormal) = logdet(d.Σ)
...
     invquad!(r, d.Σ, x .- d.μ)
 
-gradlogpdf(d::MvNormal, x::Vector) = -(d.Σ \ (x - d.μ))
+gradlogpdf(d::MvNormal, x::Vector{Float64}) = -(d.Σ \ (x - d.μ))
```
andreasnoack: Function body doesn't require this restriction

src/multivariate/mvnormal.jl
```diff
@@ -122,19 +117,19 @@ logdetcov(d::MvNormal) = logdet(d.Σ)
...
 
 # Sampling (for GenericMvNormal)
 
-_rand!(d::MvNormal, x::VecOrMat) = add!(unwhiten!(d.Σ, randn!(x)), d.μ)
+_rand!(d::MvNormal, x::VecOrMat{Float64}) = add!(unwhiten!(d.Σ, randn!(x)), d.μ)
```
andreasnoack: Same. Better to let it fail in `randn!` if element type is not supported. You'll never know when somebody extends with a new method.

src/multivariate/mvnormal.jl
```diff
@@ -122,19 +117,19 @@ logdetcov(d::MvNormal) = logdet(d.Σ)
...
 
 # Workaround: randn! only works for Array, but not generally for AbstractArray
-function _rand!(d::MvNormal, x::AbstractVecOrMat)
+function _rand!(d::MvNormal, x::AbstractVecOrMat{Float64})
```
andreasnoack: Is this needed for correctness of the body of the function?

src/multivariate/mvnormal.jl
```diff
@@ -232,8 +227,8 @@ function suffstats(D::Type{MvNormal}, x::AbstractMatrix{Float64}, w::Array{Float
     m = s * inv(tw)
     z = similar(x)
     for j = 1:n
-        xj = view(x,:,j)
-        zj = view(z,:,j)
+        xj = slice(x,:,j)
```
andreasnoack: Should remain a `view.

jmxpearson:
@andreasnoack Again, sorry for the mix-up: corrected versions of `mvnormal.jl` and `mvnormalcanon.jl` now pushed.

andreasnoack:
Ok. Makes sense.

src/quantilealgs.jl
```diff
@@ -30,93 +30,97 @@ quantile_bisect(d::ContinuousUnivariateDistribution, p::Float64) =
 #   Distribution, with Application to the Inverse Gaussian Distribution
 #   http://www.statsci.org/smyth/pubs/qinvgaussPreprint.pdf
 
-function quantile_newton(d::ContinuousUnivariateDistribution, p::Float64, xs::Float64=mode(d), tol::Float64=1e-12)
-    if 0.0 < p < 1.0
+function quantile_newton(d::ContinuousUnivariateDistribution, p::Real, xs::Real=mode(d), tol::Real=1e-12)
+    T = promote_type(partype(d), typeof(p), typeof(xs), typeof(tol))
```
andreasnoack: I don't think the tolerance needs to be part of the type determination. It can be convenient to specify the tolerance as a `Float64` without promoting all the other parameters.

src/quantilealgs.jl
```diff
@@ -30,93 +30,97 @@ quantile_bisect(d::ContinuousUnivariateDistribution, p::Float64) =
...
+    if 0 < p < 1
         while true
             x = xs + (p - cdf(d, xs)) / pdf(d, xs)
             abs(x-xs) >= max(abs(x),abs(xs))*tol || return x
             xs = x
```
andreasnoack: Might cause trouble to reuse `xs` here when it is allowed to be any `Real` number. E.g. an `Int` would probably make the type of `xs` change and therefore it will end up being boxed. Probably a good idea to check functino with `@code_warntype` for various input types.

src/quantilealgs.jl
```diff
@@ -30,93 +30,97 @@ quantile_bisect(d::ContinuousUnivariateDistribution, p::Float64) =
...
 
-function cquantile_newton(d::ContinuousUnivariateDistribution, p::Float64, xs::Float64=mode(d), tol::Float64=1e-12)
-    if 0.0 < p < 1.0
+function cquantile_newton(d::ContinuousUnivariateDistribution, p::Real, xs::Real=mode(d), tol::Real=1e-12)
+    T = promote_type(partype(d), typeof(p), typeof(xs), typeof(tol))
+    if 0 < p < 1
         while true
             x = xs + (ccdf(d, xs)-p) / pdf(d, xs)
             abs(x-xs) >= max(abs(x),abs(xs))*tol || return x
             xs = x
```
andreasnoack: Same

src/quantilealgs.jl
jmxpearson: We'll check `@code_warntype`. And yes, I'll remove `typeof(tol)` from `T`.

What about something like
```julia
if 0 < p < 1
   x0 = T(xs)
   while true
      x = x0 + (p - cdf(d, x0))/pdf(d, x0)
      abs(x-xs) >= max(abs(x),abs(xs))*tol || return x
      x0 = x
   end
```

ararslan:
Just a note regarding the use of special functions from Base: Those will eventually be migrated out of base. We've already begun the process in [JuliaMath/SpecialFunctions.jl](https://github.com/JuliaMath/SpecialFunctions.jl). So note that the use of such functions will eventually require a dependency on SpecialFunctions. (Not that that would be a big deal or anything, just wanted to give you all a heads up.)

jmxpearson:
Just pushed more code. I found out that `partype` was type unstable, which in addition to slowing things down, made quantilealgs unstable. That should all be fixed now.

`partype` now defaults to Float64 unless the distribution is parametric. This is hacky, but necessary even in cases like `Hypergeometric` with only integer parameters since we use the result of this call in a `promote_type` with `eltype(X)` to allocate storage for in-place calls to `pdf(d, X)`, etc. (Note that `Hypergeometric` is the only case with only integer parameters I know of. Some distributions like `Kolmogorov` have no parameters.)

src/quantilealgs.jl
andreasnoack: I think all the inputs could be `Int`s and then it would still be an issue. Maybe something like
```jl
if 0 < p < 1
    x = xs + (p - cdf(d, xs))/pdf(d, xs)
    x0 = typeof(x)(xs)
    while abs(x - xs) >= max(abs(x), abs(xs))
        x = x0 + (ccdf(d, x0)-p) / pdf(d, x0)
        x0 = x
    end
    return x
end
```

jmxpearson: That makes sense.

jmxpearson: cf latest push on this.

jmxpearson:
Okay. I think this really does cover it. Univariates are passing tests; I wrote a bunch of new tests; one or two distributions (Skellam, von Mises) that weren't being fully tested now are (moreso); and the multivariates we've changed now also have tests for weird edge cases. Things like
```julia
d = GenericMvTDist(1, [1.0f0, 2.0f0], PDMat([4. 2; 2 3]))
@test partype(d) == Float64
```
now pass.

I'm going to let this settle for a couple of days, and then if there are no objections, I'll merge.

src/parametric.jl
```diff
@@ -0,0 +1,64 @@
...
+    Triweight,
+    Uniform,
+    VonMises,
+    WalleniusNoncentralHypergeometric,
+    Weibull,
+    Wishart
+]
+
+for D in parametrics
+    ex = :(@inline partype{T<:Real}(d::$(D.name.primary){T}) = T)
```
andreasnoack: I think it would be better to just have a one liner, e.g. `partype{T}(TDist{T})=T`,  for each of the parametric distributions in the place where they are defined. Usually, we shouldn't hook into the type internals like you do here.

jmxpearson: Okay. Easy enough to fix.

jmxpearson:
Currently failing because of issues detailed in JuliaStats/StatsBase.jl#187

ararslan:
The recent JuliaStats fires seem to be under control at the moment. I've restarted the Travis build here to see how we fare. It would be great if you could resolve the branch conflicts. Looking forward to getting this merged! Thanks so much for all of your hard work here!

ararslan:
btw you'll probably have to rebase your branch from the current master since there's been some changes

jmxpearson:
Thanks. Taking a look now.

jmxpearson:
Looks like this is only failing on Mac nightly (not updated with RMath yet?).

andreasnoack:
I think there is something wrong with the nightly Mac binary.

jmxpearson:
Okay. No hurry to merge this, since I don't think we should tag until I've finished parameterizing the last several distributions (in separate PRs).

ararslan:
The nightly OS X issue is https://github.com/JuliaLang/julia/issues/17484

jmxpearson:
Okay. Looks like we're all green again. Any objections to merging this?

andreasnoack:
Let me try to get through the PR this afternoon.

jmxpearson:
Sounds good.

src/univariate/continuous/arcsine.jl
```diff
@@ -22,49 +22,67 @@ External links
...
 median(d::Arcsine) = mean(d)
 mode(d::Arcsine) = d.a
 modes(d::Arcsine) = [d.a, d.b]
 
-var(d::Arcsine) = 0.125 * abs2(d.b - d.a)
-skewness(d::Arcsine) = 0.0
-kurtosis(d::Arcsine) = -1.5
+var(d::Arcsine) = abs2(d.b - d.a) / 8
+skewness{T<:Real}(d::Arcsine{T}) = zero(T)
+kurtosis{T<:Real}(d::Arcsine{T}) = -3/2*one(T)
```
andreasnoack: This one is tricky. It evaluates as `(3/2)*one(T)` which will be `Float64` even when `T=Float32`. It should just be `T(3/2)`.

src/univariate/continuous/arcsine.jl
```diff
@@ -22,49 +22,67 @@ External links
...
 ### Evaluation
 
-pdf(d::Arcsine, x::Float64) = insupport(d, x) ? 1.0 / (π * sqrt((x - d.a) * (d.b - x))) : 0.0
+function pdf(d::Arcsine, x::Real)
+    insupport(d, x) ? one(d.a) / (π * sqrt((x - d.a) * (d.b - x))) : zero(d.a)
+end
 
-logpdf(d::Arcsine, x::Float64) = insupport(d, x) ? -(logπ + 0.5 * log((x - d.a) * (d.b - x))) : -Inf
+function logpdf{T<:Real}(d::Arcsine{T}, x::Real)
+    insupport(d, x) ? -(logπ + 1/2*log((x - d.a) * (d.b - x))) : -T(Inf)
```
andreasnoack: Divide with 2 in the end to avoid `1/2` which becomes the `Float64` value `0.5`.

src/univariate/continuous/arcsine.jl
```diff
@@ -22,49 +22,67 @@ External links
...
+end
 
-cdf(d::Arcsine, x::Float64) = x < d.a ? 0.0 :
-                              x > d.b ? 1.0 :
+cdf{T<:Real}(d::Arcsine{T}, x::Real) = x < d.a ? zero(T) :
+                              x > d.b ? one(T) :
                               0.636619772367581343 * asin(sqrt((x - d.a) / (d.b - d.a)))
```
andreasnoack: `0.636619772367581343` is a `Float64` literal. Wolfram Alpha tells me that we can write `asin(sqrt((x - d.a) / (d.b - d.a)))/π`

src/univariate/continuous/biweight.jl
```diff
@@ -1,58 +1,63 @@
...
 mean(d::Biweight) = d.μ
 median(d::Biweight) = d.μ
 mode(d::Biweight) = d.μ
 
-var(d::Biweight) = d.σ^2 / 7.0
-skewness(d::Biweight) = 0.0
-kurtosis(d::Biweight) = -2.9523809523809526  # = 1/21-3
+var(d::Biweight) = d.σ^2 / 7
+skewness{T<:Real}(d::Biweight{T}) = zero(T)
+kurtosis{T<:Real}(d::Biweight{T}) = -2.9523809523809526*one(T)  # = 1/21-3
```
andreasnoack: `T(1)/21-3`, I think

src/univariate/continuous/biweight.jl
```diff
@@ -1,58 +1,63 @@
...
 
 ## Functions
-function pdf(d::Biweight, x::Float64)
+function pdf{T<:Real}(d::Biweight{T}, x::Real)
     u = abs(x - d.μ) / d.σ
-    u >= 1.0 ? 0.0 : 0.9375 * (1 - u^2)^2 / d.σ
+    u >= 1 ? zero(T) : 0.9375 * (1 - u^2)^2 / d.σ
```
andreasnoack: `Float64` literal

src/univariate/continuous/biweight.jl
```diff
@@ -1,58 +1,63 @@
...
 
-function cdf(d::Biweight, x::Float64)
+function cdf{T<:Real}(d::Biweight{T}, x::Real)
     u = (x - d.μ) / d.σ
-    u <= -1.0 ? 0.0 :
-    u >= 1.0 ? 1.0 :
-    0.0625 * (u + 1.0)^3 * @horner(u,8.0,-9.0,3.0)
+    u <= -1 ? zero(T) :
+    u >= 1 ? one(T) :
+    0.0625(u + 1)^3 * @horner(u,8,-9,3)
```
andreasnoack: `Float64` literal

src/univariate/continuous/biweight.jl
```diff
@@ -1,58 +1,63 @@
...
 
-function ccdf(d::Biweight, x::Float64)
+function ccdf{T<:Real}(d::Biweight{T}, x::Real)
     u = (d.μ - x) / d.σ
-    u <= -1.0 ? 1.0 :
-    u >= 1.0 ? 0.0 :
-    0.0625 * (u + 1.0)^3 * @horner(u,8.0,-9.0,3.0)
+    u <= -1 ? one(T) :
+    u >= 1 ? zero(T) :
+    0.0625(u + 1)^3 * @horner(u,8,-9,3)
```
andreasnoack: `Float64` literal

src/univariate/continuous/chi.jl
```diff
@@ -50,37 +59,37 @@ function kurtosis(d::Chi)
     μ = mean(d)
     σ = sqrt(d.ν - μ^2)
     γ = _chi_skewness(μ, σ)
-    (2.0 / σ^2) * (1 - μ * σ * γ - σ^2)
+    (2/σ^2) * (1 - μ * σ * γ - σ^2)
 end
 
 entropy(d::Chi) = (ν = d.ν;
-    lgamma(ν / 2.0) - 0.5 * logtwo - ((ν - 1.0) / 2.0) * digamma(ν / 2.0) + ν / 2.0)
+    lgamma(ν/2) - 1//2*logtwo - ((ν - 1)/2) * digamma(ν/2) + ν/2)
```
andreasnoack: `1//2*logtwo` will end up as a `Float64`

src/univariate/continuous/cosine.jl
```diff
@@ -33,32 +45,34 @@ mode(d::Cosine) = d.μ
 
 var(d::Cosine) = d.σ^2 * 0.13069096604865779  # 0.130... = 1/3 - 2 / π^2
 
-skewness(d::Cosine) = 0.0
+skewness{T<:Real}(d::Cosine{T}) = zero(T)
 
-kurtosis(d::Cosine) = -0.59376287559828102362
+kurtosis{T<:Real}(d::Cosine{T}) = -0.59376287559828102362*one(T)
```
andreasnoack: `6*(90-T(pi))/(5*(T(π)^2-6)^2)` if `T` is a floating point type.

src/univariate/continuous/cosine.jl
```diff
@@ -33,32 +45,34 @@ mode(d::Cosine) = d.μ
 
 var(d::Cosine) = d.σ^2 * 0.13069096604865779  # 0.130... = 1/3 - 2 / π^2
```
andreasnoack: `Float64` literal

src/univariate/continuous/cosine.jl
```diff
@@ -33,32 +45,34 @@ mode(d::Cosine) = d.μ
...
 
-function pdf(d::Cosine, x::Float64)
+function pdf{T<:Real}(d::Cosine{T}, x::Real)
     if insupport(d, x)
         z = (x - d.μ) / d.σ
-        return (1.0 + cospi(z)) / (2 * d.σ)
+        return (1 + cospi(z)) / (2d.σ)
     else
-        return 0.0
+        return zero(T)
```
andreasnoack: I realize that some of these will be type unstable for integers and some will fail when a conversion to integer is not possible. I'm not sure how important these cases will be.

src/univariate/continuous/epanechnikov.jl
```diff
@@ -1,59 +1,73 @@
...
-kurtosis(d::Epanechnikov) = -2.914285714285714  # 3/35-3
+skewness{T<:Real}(d::Epanechnikov{T}) = zero(T)
+kurtosis{T<:Real}(d::Epanechnikov{T}) = -2.914285714285714*one(T)  # 3/35-3
 
 ## Functions
-function pdf(d::Epanechnikov, x::Float64)
+function pdf{T<:Real}(d::Epanechnikov{T}, x::Real)
     u = abs(x - d.μ) / d.σ
-    u >= 1 ? 0.0 : 0.75 * (1 - u^2) / d.σ
+    u >= 1 ? zero(T) : (3/4) * (1 - u^2) / d.σ
```
andreasnoack: Divide with `4` in the end.

src/univariate/continuous/exponential.jl
```diff
@@ -19,59 +19,69 @@ External links
...
 
 mean(d::Exponential) = d.θ
 median(d::Exponential) = logtwo * d.θ
-mode(d::Exponential) = 0.0
+mode{T<:Real}(d::Exponential{T}) = zero(T)
 
 var(d::Exponential) = d.θ^2
-skewness(d::Exponential) = 2.0
-kurtosis(d::Exponential) = 6.0
+skewness{T<:Real}(d::Exponential{T}) = 2*one(T)
```
andreasnoack: Cleaner with `T(2)` here

src/univariate/continuous/cosine.jl
ararslan: Could you not just do `T(-0.59376287559828102362)`?

andreasnoack: Wouldn't be correct for `BigFloat`s. On the other hand, my suggestion fails for integers. Generic coding can be tricky.

jmxpearson: Note that we are everywhere promoting integer arguments to Floats when the
parameter is continuous. There are a lot of type instabilities for Ints
lurking in here that we simply couldn't resolve without weird promotion
hijinks everywhere. So given that, Andreas's suggestion is the one to go
with here.

src/univariate/continuous/gamma.jl
```diff
@@ -98,8 +100,8 @@ immutable GammaStats <: SufficientStats
 end
 
 function suffstats{T<:Real}(::Type{Gamma}, x::AbstractArray{T})
-    sx = 0.
-    slogx = 0.
+    sx = 0
+    slogx = 0
```
andreasnoack: Type unstable. You'll have to initialize these more carefully

src/univariate/continuous/gamma.jl
```diff
@@ -113,9 +115,9 @@ function suffstats{T<:Real}(::Type{Gamma}, x::AbstractArray{T}, w::AbstractArray
         throw(ArgumentError("Inconsistent argument dimensions."))
     end
 
-    sx = 0.
-    slogx = 0.
-    tw = 0.
+    sx = 0
```
andreasnoack: Same

src/univariate/continuous/generalizedextremevalue.jl
```diff
@@ -51,199 +68,202 @@ shape(d::GeneralizedExtremeValue) = d.ξ
...
     end
 end
 
-function mean(d::GeneralizedExtremeValue)
+function mean{T<:Real}(d::GeneralizedExtremeValue{T})
     (μ, σ, ξ) = params(d)
-    
-    if abs(ξ) < eps() # ξ == 0.0
+
+    if abs(ξ) < eps() # ξ == 0
```
andreasnoack: `abs(ξ) < eps(one(ξ))`

src/univariate/continuous/generalizedextremevalue.jl
```diff
@@ -51,199 +68,202 @@ shape(d::GeneralizedExtremeValue) = d.ξ
...
+        return T(Inf)
     end
 end
 
 function mode(d::GeneralizedExtremeValue)
     (μ, σ, ξ) = params(d)
-    
-    if abs(ξ) < eps() # ξ == 0.0
+
+    if abs(ξ) < eps() # ξ == 0
```
andreasnoack: same

src/univariate/continuous/generalizedextremevalue.jl
```diff
@@ -51,199 +68,202 @@ shape(d::GeneralizedExtremeValue) = d.ξ
...
-function kurtosis(d::GeneralizedExtremeValue)
+function kurtosis{T<:Real}(d::GeneralizedExtremeValue{T})
     (μ, σ, ξ) = params(d)
-    
-    if abs(ξ) < eps() # ξ == 0.0
-        return 12.0 / 5.0
-    elseif ξ < 1.0 / 4.0
+
+    if abs(ξ) < eps() # ξ == 0
+        return 12 / 5 * one(T)
```
andreasnoack: T(12)/5

src/univariate/continuous/generalizedpareto.jl
```diff
@@ -30,132 +30,147 @@ External links
...
+scale(d::GeneralizedPareto) = d.σ
+shape(d::GeneralizedPareto) = d.ξ
+params(d::GeneralizedPareto) = (d.μ, d.σ, d.ξ)
+@inline partype{T<:Real}(d::GeneralizedPareto{T}) = T
 
 
 #### Statistics
 
-median(d::GeneralizedPareto) = d.μ + d.σ * expm1(d.ξ * log(2.0)) / d.ξ
+median(d::GeneralizedPareto) = d.ξ == 0 ? d.μ + d.σ * log(2) : d.μ + d.σ * expm1(d.ξ * log(2)) / d.ξ
```
andreasnoack: `log(2)` is `Float64`.

src/univariate/continuous/gumbel.jl
```diff
@@ -51,34 +60,34 @@ mode(d::Gumbel) = d.μ
 
 var(d::Gumbel) = 1.6449340668482264 * d.θ^2
```
andreasnoack: Maybe `(π*d.θ)^2/6`

src/univariate/continuous/gumbel.jl
```diff
@@ -51,34 +60,34 @@ mode(d::Gumbel) = d.μ
...
 
-skewness(d::Gumbel) = 1.13954709940464866
+skewness{T<:Real}(d::Gumbel{T}) = 1.13954709940464866*one(T)
```
andreasnoack: `12*sqrt(T(6))*zeta(T(3))/pi/pi/pi` should work but is pretty ugly.

src/univariate/continuous/gumbel.jl
```diff
@@ -51,34 +60,34 @@ mode(d::Gumbel) = d.μ
...
 
-kurtosis(d::Gumbel) = 2.4
+kurtosis{T<:Real}(d::Gumbel{T}) = 12/5*one(T)
 
 entropy(d::Gumbel) = 1.57721566490153286 + log(d.θ)
```
andreasnoack: `Float64` literal

src/univariate/continuous/gumbel.jl
```diff
@@ -51,34 +60,34 @@ mode(d::Gumbel) = d.μ
 
 var(d::Gumbel) = 1.6449340668482264 * d.θ^2
 
-skewness(d::Gumbel) = 1.13954709940464866
+skewness{T<:Real}(d::Gumbel{T}) = 1.13954709940464866*one(T)
 
-kurtosis(d::Gumbel) = 2.4
+kurtosis{T<:Real}(d::Gumbel{T}) = 12/5*one(T)
```
andreasnoack: `T(12)/5`

src/univariate/continuous/inversegaussian.jl
```diff
@@ -48,82 +61,82 @@ mean(d::InverseGaussian) = d.μ
...
+skewness(d::InverseGaussian) = 3sqrt(d.μ / d.λ)
 
-kurtosis(d::InverseGaussian) = 15.0 * d.μ / d.λ
+kurtosis(d::InverseGaussian) = 15d.μ / d.λ
 
 function mode(d::InverseGaussian)
     μ, λ = params(d)
     r = μ / λ
-    μ * (sqrt(1.0 + 2.25 * r^2) - 1.5 * r)
+    μ * (sqrt(1 + (9/4)*r^2) - (3/2)*r)
```
andreasnoack: Maybe `(3*r/2)^2`

src/univariate/continuous/inversegaussian.jl
```diff
@@ -48,82 +61,82 @@ mean(d::InverseGaussian) = d.μ
...
     end
 end
 
-function logpdf(d::InverseGaussian, x::Float64)
-    if x > 0.0
+function logpdf{T<:Real}(d::InverseGaussian{T}, x::Real)
+    if x > 0
         μ, λ = params(d)
-        return 0.5 * (log(λ) - (log2π + 3.0 * log(x)) - λ * (x - μ)^2 / (μ^2 * x))
+        return 1/2*(log(λ) - (log2π + 3log(x)) - λ * (x - μ)^2 / (μ^2 * x))
```
andreasnoack: divide with two

src/univariate/continuous/kolmogorov.jl
```diff
@@ -49,44 +49,44 @@ function ccdf_raw(d::Kolmogorov, x::Float64)
...
-        return 0.0
-    elseif x <= 1.0
-        c = π/(2.0*x)        
-        s = 0.0
+function pdf(d::Kolmogorov,x::Real)
+    if x <= 0
+        return 0
+    elseif x <= 1
+        c = π/(2*x)
+        s = 0
```
andreasnoack: Type unstable. Will become a float in the loop

src/univariate/continuous/kolmogorov.jl
```diff
@@ -49,44 +49,44 @@ function ccdf_raw(d::Kolmogorov, x::Float64)
...
 # TODO: figure out how best to truncate series
-function pdf(d::Kolmogorov,x::Float64)
-    if x <= 0.0
-        return 0.0
-    elseif x <= 1.0
-        c = π/(2.0*x)        
-        s = 0.0
+function pdf(d::Kolmogorov,x::Real)
+    if x <= 0
+        return 0
```
andreasnoack: Type unstable. Should be converted

src/univariate/continuous/kolmogorov.jl
```diff
@@ -105,12 +105,12 @@ function rand(d::Kolmogorov)
         while true
             g = rand_trunc_gamma()
 
-            x = pi/sqrt(8.0*g)
-            w = 0.0
-            z = 1/(2*g)
+            x = pi/sqrt(8g)
+            w = 0
```
andreasnoack: type unstable

src/univariate/continuous/kolmogorov.jl
```diff
@@ -105,12 +105,12 @@ function rand(d::Kolmogorov)
...
+            z = 1/(2g)
             p = exp(-g)
             n = 1
-            q = 1.0
+            q = 1
```
andreasnoack: type unstable

src/univariate/continuous/kolmogorov.jl
```diff
@@ -128,7 +128,7 @@ function rand(d::Kolmogorov)
             e = randexp()
             u = rand()
             x = sqrt(t*t+e/2)
-            w = 0.0
+            w = 0
```
andreasnoack: type unstable. Use `zero` function

src/univariate/continuous/kolmogorov.jl
```diff
@@ -145,12 +145,12 @@ function rand(d::Kolmogorov)
 end
 
 # equivalent to
-# rand(Truncated(Gamma(1.5,1.0),tp,Inf))
+# rand(Truncated(Gamma(1.5,1),tp,Inf))
 function rand_trunc_gamma()
-    tp = 2.193245422464302 #pi^2/(8*t^2)    
+    tp = 2.193245422464302 #pi^2/(8*t^2)
```
andreasnoack: probably `(pi/(2t))^2/2`

src/univariate/continuous/ksdist.jl
```diff
@@ -16,52 +16,52 @@ function cdf(d::KSDist,x::Float64)
     n = d.n
     b = x*n
     # known exact values
-    if b <= 0.5
-        return 0.0
-    elseif b <= 1.0
+    if b <= 1/2
+        return 0
```
andreasnoack: type unstable

src/univariate/continuous/ksdist.jl
```diff
@@ -16,52 +16,52 @@ function cdf(d::KSDist,x::Float64)
...
+    elseif b <= 1
         # accuracy could be improved
-        return exp(lfact(n)+n*(log(2.0*b-1.0)-log(n)))
-    elseif x >= 1.0
-        return 1.0
+        return exp(lfact(n)+n*(log(2*b-1)-log(n)))
+    elseif x >= 1
+        return 1
```
andreasnoack: type unstable

src/univariate/continuous/ksdist.jl
```diff
@@ -16,52 +16,52 @@ function cdf(d::KSDist,x::Float64)
...
     elseif b >= n-1
-        return 1.0 - 2.0*(1.0-x)^n
+        return 1 - 2*(1-x)^n
     end
 
     a = b*x
-    if a >= 18.0
-        return 1.0
+    if a >= 18
+        return 1
```
andreasnoack: type unstable

src/univariate/continuous/ksdist.jl
```diff
@@ -16,52 +16,52 @@ function cdf(d::KSDist,x::Float64)
...
     n = d.n
     b = x*n
     # Ruben and Gambino (1982) known exact values
     if b <= 0.5
-        return 1.0
-    elseif b <= 1.0
-        return 1.0-exp(lfact(n)+n*(log(2.0*b-1.0)-log(n)))
-    elseif x >= 1.0
-        return 0.0
+        return 1
```
andreasnoack: unstable

src/univariate/continuous/ksdist.jl
```diff
@@ -16,52 +16,52 @@ function cdf(d::KSDist,x::Float64)
...
+    elseif b <= 1
+        return 1-exp(lfact(n)+n*(log(2*b-1)-log(n)))
+    elseif x >= 1
+        return 0
```
andreasnoack: unstable

src/univariate/continuous/ksdist.jl
```diff
@@ -16,52 +16,52 @@ function cdf(d::KSDist,x::Float64)
...
-        return 2.0*(1.0-x)^n
+        return 2*(1-x)^n
     end
 
     a = b*x
-    if a >= 370.0
-        return 0.0
-    elseif a >= 4.0 || (n > 140 && a >= 2.2)
+    if a >= 370
+        return 0
```
andreasnoack: unstable

src/univariate/continuous/kolmogorov.jl
jmxpearson: Correct. A few distributions (ksdist, kolmogorov, and maybe one or two others) didn't have any parameters to make parametric.

src/univariate/continuous/ksdist.jl
jmxpearson: See above. This distribution has no parameters, so we didn't have any type instabilities to worry about.

src/univariate/continuous/ksdist.jl
```diff
@@ -73,12 +73,12 @@ function cdf_durbin(d::KSDist,x::Float64)
...
     m = 2*k-1
-    H = Array(Float64,m,m)    
+    H = Array(Float64,m,m)
     for i = 1:m, j = 1:m
-        H[i,j] = i-j+1 >= 0 ? 1.0 : 0.0        
+        H[i,j] = i-j+1 >= 0 ? 1 : 0
     end
-    r = 1.0
-    for i = 1:m 
+    r = 1
```
andreasnoack: Since `h` is a the same type as `x` this will get promoted to a `Float64` in the loop so it should be initialized differently. Should this function be parametric?

src/univariate/continuous/ksdist.jl
```diff
@@ -120,16 +120,16 @@ function ceil_rems_mult(n,x)
 end
 
 # n!*(e/n)^n
-function stirling(n)    
+function stirling(n)
     if n < 500
-        s = 1.0
+        s = 1
```
andreasnoack: unstable. Can we use the `stirling` approximation function in `StatsFuns`?

src/univariate/continuous/ksdist.jl
jmxpearson: We didn't parameterize any distribution with only integer arguments. One could perhaps want an Int32 here, but the original motivation, automatic differentiation, wouldn't apply for these cases with only discrete parameters.

src/univariate/continuous/ksonesided.jl
```diff
@@ -11,19 +11,18 @@ end
 
 # formula of Birnbaum and Tingey (1951)
 function ccdf(d::KSOneSided, x::Float64)
-    if x >= 1.0
-        return 0.0
-    elseif x <= 0.0
-        return 1.0
+    if x >= 1
+        return 0
```
andreasnoack: unstable

src/univariate/continuous/ksonesided.jl
```diff
@@ -11,19 +11,18 @@ end
...
+    elseif x <= 0
+        return 1
     end
     n = d.n
-    s = 0.0
+    s = 0
```
andreasnoack: will get promoted in the loop

src/univariate/continuous/ksonesided.jl
```diff
@@ -11,19 +11,18 @@ end
...
     for j = 0:floor(Int,n-n*x)
         p = x+j/n
         s += pdf(Binomial(n,p),j) / p
     end
     s*x
 end
 
-cdf(d::KSOneSided, x::Float64) = 1.0 - ccdf(d,x)
-
+cdf(d::KSOneSided, x::Float64) = 1 - ccdf(d,x)
```
andreasnoack: `Real`?

src/univariate/continuous/ksdist.jl
jmxpearson: I can take a look, but that seems like it might be out of scope for this PR.

I do appreciate the very thorough review, though. I will get all these small things fixed soon.

src/univariate/continuous/ksonesided.jl
jmxpearson: Again, this distribution not parameterized.

src/univariate/continuous/laplace.jl
```diff
@@ -47,42 +61,42 @@ mean(d::Laplace) = d.μ
...
+pdf(d::Laplace, x::Real) = exp(-abs(zval(d, x))) / 2scale(d)
+logpdf(d::Laplace, x::Real) = - (abs(zval(d, x)) + log(2scale(d)))
 
-cdf(d::Laplace, x::Float64) = (z = zval(d, x); z < 0.0 ? 0.5 * exp(z) : 1.0 - 0.5 * exp(-z))
-ccdf(d::Laplace, x::Float64) = (z = zval(d, x); z > 0.0 ? 0.5 * exp(-z) : 1.0 - 0.5 * exp(z))
-logcdf(d::Laplace, x::Float64) = (z = zval(d, x); z < 0.0 ? loghalf + z : loghalf + log2mexp(-z))
-logccdf(d::Laplace, x::Float64) = (z = zval(d, x); z > 0.0 ? loghalf - z : loghalf + log2mexp(z))
+cdf(d::Laplace, x::Real) = (z = zval(d, x); z < 0 ? exp(z)/2 : 1 - exp(-z)/2)
+ccdf(d::Laplace, x::Real) = (z = zval(d, x); z > 0 ? exp(-z)/2 : 1 - exp(z)/2)
+logcdf(d::Laplace, x::Real) = (z = zval(d, x); z < 0 ? loghalf + z : loghalf + log2mexp(-z))
```
andreasnoack: is `loghalf` a problem here? Might be `Float64`.

src/univariate/continuous/laplace.jl
```diff
@@ -47,42 +61,42 @@ mean(d::Laplace) = d.μ
...
+logccdf(d::Laplace, x::Real) = (z = zval(d, x); z > 0 ? loghalf - z : loghalf + log2mexp(z))
 
-quantile(d::Laplace, p::Float64) = p < 0.5 ? xval(d, log(2.0 * p)) : xval(d, -log(2.0 * (1.0 - p)))
-cquantile(d::Laplace, p::Float64) = p > 0.5 ? xval(d, log(2.0 * (1.0 - p))) : xval(d, -log(2.0 * p))
-invlogcdf(d::Laplace, lp::Float64) = lp < loghalf ? xval(d, logtwo + lp) : xval(d, -(logtwo + log1mexp(lp)))
-invlogccdf(d::Laplace, lp::Float64) = lp > loghalf ? xval(d, logtwo + log1mexp(lp)) : xval(d, -(logtwo + lp))
+quantile(d::Laplace, p::Real) = p < 1/2 ? xval(d, log(2p)) : xval(d, -log(2(1 - p)))
+cquantile(d::Laplace, p::Real) = p > 1/2 ? xval(d, log(2(1 - p))) : xval(d, -log(2p))
+invlogcdf(d::Laplace, lp::Real) = lp < loghalf ? xval(d, logtwo + lp) : xval(d, -(logtwo + log1mexp(lp)))
```
andreasnoack: same with `logtwo`

src/univariate/continuous/levy.jl
```diff
@@ -19,63 +19,72 @@ External links
...
+kurtosis{T<:Real}(d::Levy{T}) = T(NaN)
 
-mode(d::Levy) = d.σ / 3.0 + d.μ
+mode(d::Levy) = d.σ / 3 + d.μ
 
-entropy(d::Levy) = (1.0 - 3.0 * digamma(1.0) + log(16π * d.σ^2)) / 2.0
+entropy(d::Levy) = (1 - 3digamma(1) + log(16π * d.σ^2)) / 2
 
-median(d::Levy) = d.μ + d.σ / 0.4549364231195728  # 0.454... = (2.0 * erfcinv(0.5)^2)
+median(d::Levy) = d.μ + d.σ / 0.4549364231195728  # 0.454... = (2 * erfcinv(0.5)^2)
```
andreasnoack: `Float64` literal

src/univariate/continuous/levy.jl
```diff
@@ -19,63 +19,72 @@ External links
...
+mean{T<:Real}(d::Levy{T}) = T(Inf)
+var{T<:Real}(d::Levy{T}) = T(Inf)
+skewness{T<:Real}(d::Levy{T}) = T(NaN)
+kurtosis{T<:Real}(d::Levy{T}) = T(NaN)
 
-mode(d::Levy) = d.σ / 3.0 + d.μ
+mode(d::Levy) = d.σ / 3 + d.μ
 
-entropy(d::Levy) = (1.0 - 3.0 * digamma(1.0) + log(16π * d.σ^2)) / 2.0
+entropy(d::Levy) = (1 - 3digamma(1) + log(16π * d.σ^2)) / 2
```
andreasnoack: Need to split up `16` and `π`. Maybe `16 * d.σ^2 * π`

src/univariate/continuous/levy.jl
```diff
@@ -19,63 +19,72 @@ External links
...
 
 
 #### Evaluation
 
-function pdf(d::Levy, x::Float64)
+function pdf(d::Levy, x::Real)
     μ, σ = params(d)
     z = x - μ
-    (sqrt(σ) / sqrt2π) * exp((-σ) / (2.0 * z)) / z^1.5
+    (sqrt(σ) / sqrt2π) * exp((-σ) / (2z)) / z^(3//2)
```
andreasnoack: `sqrt2π` is probably `Float64` here

src/univariate/continuous/levy.jl
```diff
@@ -19,63 +19,72 @@ External links
...
 end
 
-function logpdf(d::Levy, x::Float64)
+function logpdf(d::Levy, x::Real)
     μ, σ = params(d)
     z = x - μ
-    0.5 * (log(σ) - log2π - σ / z - 3.0 * log(z))
+    1/2 * (log(σ) - log2π - σ / z - 3log(z))
```
andreasnoack: Divide with two instead

src/univariate/continuous/logistic.jl
```diff
@@ -48,34 +61,34 @@ median(d::Logistic) = d.μ
 mode(d::Logistic) = d.μ
 
 std(d::Logistic) = π * d.θ / sqrt3
-var(d::Logistic) = (π * d.θ)^2 / 3.0
-skewness(d::Logistic) = 0.0
-kurtosis(d::Logistic) = 1.2
+var(d::Logistic) = (π * d.θ)^2 / 3
+skewness{T<:Real}(d::Logistic{T}) = zero(T)
+kurtosis{T<:Real}(d::Logistic{T}) = 6/5*one(T)
```
andreasnoack: `T(6)/5`

src/univariate/continuous/lognormal.jl
```diff
@@ -23,42 +23,50 @@ External links
...
+@inline partype{T<:Real}(d::LogNormal{T}) = T
 
 #### Statistics
 
 meanlogx(d::LogNormal) = d.μ
 varlogx(d::LogNormal) = abs2(d.σ)
 stdlogx(d::LogNormal) = d.σ
 
-mean(d::LogNormal) = ((μ, σ) = params(d); exp(μ + 0.5 * σ^2))
+mean(d::LogNormal) = ((μ, σ) = params(d); exp(μ + 1/2*σ^2))
```
andreasnoack: Divide with two

src/univariate/continuous/lognormal.jl
```diff
@@ -67,40 +75,40 @@ function kurtosis(d::LogNormal)
...
     e3 = e2 * e
     e4 = e3 * e
-    e4 + 2.0 * e3 + 3.0 * e2 - 6.0
+    e4 + 2*e3 + 3*e2 - 6
 end
 
 function entropy(d::LogNormal)
     (μ, σ) = params(d)
-    0.5 * (1.0 + log(twoπ * σ^2)) + μ
+    1/2*(1 + log(twoπ * σ^2)) + μ
```
andreasnoack: Divide with two. Careful with `twoπ`

src/univariate/continuous/normalcanon.jl
```diff
@@ -37,32 +40,32 @@ mean(d::NormalCanon) = d.μ
...
-kurtosis(d::NormalCanon) = 0.0
+skewness{T<:Real}(d::NormalCanon{T}) = zero(T)
+kurtosis{T<:Real}(d::NormalCanon{T}) = zero(T)
 
-var(d::NormalCanon) = 1.0 / d.λ
+var(d::NormalCanon) = 1 / d.λ
 std(d::NormalCanon) = sqrt(var(d))
 
-entropy(d::NormalCanon) = 0.5 * (log2π + 1.0 - log(d.λ))
+entropy(d::NormalCanon) = (log2π + 1 - log(d.λ)) / 2
```
andreasnoack: Should be `-log(d.λ) + log2π + 1`.

src/univariate/continuous/normalcanon.jl
```diff
@@ -37,32 +40,32 @@ mean(d::NormalCanon) = d.μ
...
 
 
 #### Evaluation
 
-pdf(d::NormalCanon, x::Float64) = (sqrt(d.λ) / sqrt2π) * exp(-0.5 * d.λ * abs2(x - d.μ))
-logpdf(d::NormalCanon, x::Float64) = 0.5 * (log(d.λ) - log2π - d.λ * abs2(x - d.μ))
+pdf(d::NormalCanon, x::Real) = (sqrt(d.λ) / sqrt2π) * exp(-1/2* d.λ * abs2(x - d.μ))
```
andreasnoack: Not `1/2`

src/univariate/continuous/normalcanon.jl
```diff
@@ -37,32 +40,32 @@ mean(d::NormalCanon) = d.μ
...
+logpdf(d::NormalCanon, x::Real) = 1/2 * (log(d.λ) - log2π - d.λ * abs2(x - d.μ))
```
andreasnoack: same

src/univariate/continuous/ksdist.jl
andreasnoack: ...but `exp(lfact(n)+n*(log(2*b-1)-log(n)))` below will be `Float64` so the return type depends on the value of `b`.

jmxpearson: Ugh. Yes, I see. Sorry.

src/univariate/continuous/ksdist.jl
andreasnoack: Ok but then `r` should probably remain `1.0`.

src/univariate/continuous/ksdist.jl
andreasnoack: Yes. I thought there might be more issue in the function and therefore that using a different implementation would solve it but this issue should be easy to fix.

src/univariate/continuous/ksonesided.jl
andreasnoack: ok

andreasnoack:
Ok. I'll have to take a break now. Will return to this tonight or tomorrow.

jmxpearson:
Thanks. As I said, I appreciate the careful review. We tried to be careful,
but there are a lot of floats lurking in there.

andreasnoack:
I fully understand. It's a lot of code and it can be difficult to see the issues.

src/univariate/continuous/rayleigh.jl
```diff
@@ -20,58 +20,67 @@ External links
...
 scale(d::Rayleigh) = d.σ
 params(d::Rayleigh) = (d.σ,)
+@inline partype{T<:Real}(d::Rayleigh{T}) = T
 
 
 #### Statistics
 
 mean(d::Rayleigh) = sqrthalfπ * d.σ
-median(d::Rayleigh) = 1.177410022515474691 * d.σ   # sqrt(log(4.0)) = 1.177410022515474691
+median(d::Rayleigh) = 1.177410022515474691 * d.σ   # sqrt(log(4)) = 1.177410022515474691
```
andreasnoack: Floating point literal

src/univariate/continuous/symtriangular.jl
```diff
@@ -44,76 +57,78 @@ mean(d::SymTriangularDist) = d.μ
 median(d::SymTriangularDist) = d.μ
 mode(d::SymTriangularDist) = d.μ
 
-var(d::SymTriangularDist) = d.σ^2 / 6.0
-skewness(d::SymTriangularDist) = 0.0
-kurtosis(d::SymTriangularDist) = -0.6
+var(d::SymTriangularDist) = d.σ^2 / 6
+skewness{T<:Real}(d::SymTriangularDist{T}) = zero(T)
+kurtosis{T<:Real}(d::SymTriangularDist{T}) = -3/5*one(T)
```
andreasnoack: T(-3/5)

src/univariate/continuous/symtriangular.jl
```diff
@@ -44,76 +57,78 @@ mean(d::SymTriangularDist) = d.μ
...
+end
 
-function cdf(d::SymTriangularDist, x::Float64)
+function cdf{T<:Real}(d::SymTriangularDist{T}, x::Real)
     (μ, σ) = params(d)
-    x <= μ - σ ? 0.0 :
-    x <= μ ? 0.5 * (1.0 + zval(d, x))^2 :
-    x < μ + σ ? 1.0 - 0.5 * (1.0 - zval(d, x))^2 : 1.0
+    x <= μ - σ ? zero(T) :
+    x <= μ ? 1/2 * (1 + zval(d, x))^2 :
```
andreasnoack: not `1/2` (also next line)

src/univariate/continuous/symtriangular.jl
```diff
@@ -44,76 +57,78 @@ mean(d::SymTriangularDist) = d.μ
...
 end
 
-function ccdf(d::SymTriangularDist, x::Float64)
+function ccdf{T<:Real}(d::SymTriangularDist{T}, x::Real)
     (μ, σ) = params(d)
-    x <= μ - σ ? 1.0 :
-    x <= μ ? 1.0 - 0.5 * (1.0 + zval(d, x))^2 :
-    x < μ + σ ? 0.5 * (1.0 - zval(d, x))^2 : 0.0
+    x <= μ - σ ? one(T) :
+    x <= μ ? 1 - 1/2 * (1 + zval(d, x))^2 :
```
andreasnoack: same

src/univariate/continuous/symtriangular.jl
```diff
@@ -44,76 +57,78 @@ mean(d::SymTriangularDist) = d.μ
...
 
-invlogcdf(d::SymTriangularDist, lp::Float64) = lp < loghalf ? xval(d, expm1(0.5*(lp - loghalf))) :
-                                                              xval(d, 1.0 - sqrt(-2.0 * expm1(lp)))
+invlogcdf(d::SymTriangularDist, lp::Real) = lp < loghalf ? xval(d, expm1(1/2*(lp - loghalf))) :
+                                                              xval(d, 1 - sqrt(-2expm1(lp)))
 
-invlogccdf(d::SymTriangularDist, lp::Float64) = lp > loghalf ? xval(d, sqrt(-2.0 * expm1(lp)) - 1.0) :
-                                                               xval(d, -(expm1(0.5 * (lp - loghalf))))
+invlogccdf(d::SymTriangularDist, lp::Real) = lp > loghalf ? xval(d, sqrt(-2*expm1(lp)) - 1) :
+                                                               xval(d, -(expm1(1/2*(lp - loghalf))))
```
andreasnoack: and here

src/univariate/continuous/tdist.jl
```diff
@@ -18,45 +18,52 @@ External links
...
+    ν > 2 ? T(Inf) : T(NaN)
 end
 
 function entropy(d::TDist)
-    h = 0.5 * d.ν
-    h1 = h + 0.5
-    h1 * (digamma(h1) - digamma(h)) + 0.5 * log(d.ν) + lbeta(h, 0.5)
+    h = d.ν/2
+    h1 = h + 1//2
+    h1 * (digamma(h1) - digamma(h)) + log(d.ν)/2 + lbeta(h, 0.5)
```
andreasnoack: `Flaot64` literal

src/univariate/continuous/triangular.jl
```diff
@@ -25,107 +25,118 @@ External links
...
 end
 
 function skewness(d::TriangularDist)
     (a, b, c) = params(d)
-    sqrt2 * (a + b - 2.0c) * (2.0a - b - c) * (a - 2.0b + c) / (5.0 * _pretvar(a, b, c)^1.5)
+    sqrt2 * (a + b - 2c) * (2a - b - c) * (a - 2b + c) / (5 * _pretvar(a, b, c)^3//2)
 end
 
-kurtosis(d::TriangularDist) = -0.6
+kurtosis{T<:Real}(d::TriangularDist{T}) = -3/5 * one(T)
```
andreasnoack: `T(-3/5)`

src/univariate/continuous/triweight.jl
```diff
@@ -1,57 +1,66 @@
...
 mean(d::Triweight) = d.μ
 median(d::Triweight) = d.μ
 mode(d::Triweight) = d.μ
 
-var(d::Triweight) = d.σ^2 / 9.0
-skewness(d::Triweight) = 0.0
-kurtosis(d::Triweight) = -2.9696969696969697  # 1/33-3
+var(d::Triweight) = d.σ^2 / 9
+skewness{T<:Real}(d::Triweight{T}) = zero(T)
+kurtosis{T<:Real}(d::Triweight{T}) = T(-2.9696969696969697)  # 1/33-3
```
andreasnoack: `T(1)/33-3`

src/univariate/discrete/binomial.jl
```diff
@@ -164,8 +179,8 @@ function suffstats{T<:Integer}(::Type{Binomial}, n::Integer, x::AbstractArray{T}
 end
 
 function suffstats{T<:Integer}(::Type{Binomial}, n::Integer, x::AbstractArray{T}, w::AbstractArray{Float64})
-    ns = 0.
-    ne = 0.
+    ns = 0
+    ne = 0
```
andreasnoack: These initializations with be unstable

src/univariate/discrete/categorical.jl
```diff
@@ -65,21 +66,21 @@ mean(d::Categorical) = categorical_mean(d.p)
 function median(d::Categorical)
     k = ncategories(d)
     p = probs(d)
-    cp = 0.
+    cp = 0
```
andreasnoack: This will change type in the loop

src/univariate/discrete/categorical.jl
```diff
@@ -89,8 +90,8 @@ function skewness(d::Categorical)
     k = ncategories(d)
     p = probs(d)
     m = categorical_mean(p)
-    s = 0.0
-    for i = 1 : k
+    s = 0
```
andreasnoack: again

src/univariate/discrete/categorical.jl
```diff
@@ -101,29 +102,29 @@ function kurtosis(d::Categorical)
...
```
andreasnoack: and here

src/univariate/discrete/geometric.jl
```diff
@@ -146,8 +158,8 @@ function suffstats{T<:Integer}(::Type{Geometric}, x::AbstractArray{T}, w::Abstra
     if length(w) != n
         throw(ArgumentError("Inconsistent argument dimensions."))
     end
-    sx = 0.
-    tw = 0.
+    sx = 0
+    tw = 0
```
andreasnoack: These will change types in the loop

src/univariate/discrete/noncentralhypergeometric.jl
```diff
@@ -13,17 +13,17 @@ end
...
         range = support(d)
-        if q > 0.5
-            q = 1-q
+        if q > 1/2
+            q = 1 - q
             range = reverse(range)
         end
 
-        qsum, i = 0.0, 0
+        qsum, i = 0, 0
```
andreasnoack: `qsum` is not an integer in the loop

src/univariate/discrete/poisson.jl
```diff
@@ -57,18 +59,18 @@ skewness(d::Poisson) = one(typeof(d.λ)) / sqrt(d.λ)
...
+        λk = one(T)
         for k = 1:100
             λk *= λ
-            s += λk * lgamma(k + 1.0) / gamma(k + 1.0)
+            s += λk * lgamma(k + 1) / gamma(k + 1)
         end
-        return λ * (1.0 - log(λ)) + exp(-λ) * s
+        return λ * (1 - log(λ)) + exp(-λ) * s
     else
         return 0.5 * log(2 * pi * e * λ) -
```
andreasnoack: `Float64` literal

src/univariate/discrete/poissonbinomial.jl
```diff
@@ -38,39 +38,51 @@ immutable PoissonBinomial <: DiscreteUnivariateDistribution
...
 #### Properties
 
 mean(d::PoissonBinomial) = sum(succprob(d))
 var(d::PoissonBinomial) = sum(succprob(d) .* failprob(d))
 
 function skewness(d::PoissonBinomial)
-    v = 0.
-    s = 0.
+    v = 0
+    s = 0
```
andreasnoack: Get promoted in the loop

src/univariate/discrete/poissonbinomial.jl
```diff
@@ -38,39 +38,51 @@ immutable PoissonBinomial <: DiscreteUnivariateDistribution
...
+        s += p[i] * (1 - p[i]) * (1 - 2 * p[i])
     end
     s / sqrt(v) / v
 end
 
 function kurtosis(d::PoissonBinomial)
-    v = 0.
-    s = 0.
+    v = 0
+    s = 0
```
andreasnoack: Same

src/univariate/discrete/poissonbinomial.jl
```diff
@@ -108,27 +122,27 @@ pdf(d::PoissonBinomial) = copy(d.pmf)
...
 
     x = Array(Complex{Float64}, n+1)
     lmax = ceil(Int, n/2)
-    x[1] = 1./(n+1)
+    x[1] = 1/(n + 1)
     for l=1:lmax
-        logz = 0.
-        argz = 0.
+        logz = 0
+        argz = 0
```
andreasnoack: Same

test/continuous.jl
```diff
@@ -58,9 +58,11 @@ for distr in [
     NoncentralF(2,2,2),
     NoncentralF(8,10,5),
     NoncentralT(2,2),
-    NoncentralT(10,2),
+    NoncentralT(10.,2),
```
andreasnoack: Why is this required?

test/univariates.jl
```diff
@@ -81,18 +115,20 @@ function verify_and_test(d::UnivariateDistribution, dct::Dict, n_tsamples::Int)
...
-        Base.Test.test_approx_eq(cdf(d, x), cf, "cdf(d, $x)", "cf")
+        if !isa(d, Skellam)
+            Base.Test.test_approx_eq(cdf(d, x), cf, "cdf(d, $x)", "cf")
+        end
     end
 
     try
         m = mgf(d,0.0)
-        @test m == 1.0
+        Base.Test.test_approx_eq(m, 1.0, "mgf(d, 0.0)", "1.0")
```
andreasnoack: Why is this necessary and why not use the macro?

test/univariates.jl
```diff
@@ -81,18 +115,20 @@ function verify_and_test(d::UnivariateDistribution, dct::Dict, n_tsamples::Int)
...
     catch e
         isa(e, MethodError) || throw(e)
     end
     try
         c = cf(d,0.0)
-        @test c == 1.0
+        Base.Test.test_approx_eq(c, 1.0, "cf(d, 0.0)", "1.0")
```
andreasnoack: same

andreasnoack:
Made it through. This is great work. Thanks for pushing this forward. One question, what its the reason for all the changes to `continuous_test.json`. Has it been regenerated?

I still concerned about the consequences of this PR. It's a huge PR so and there might still be many places where performance could be affected. The ideal solution would be to set up performance testing for the package. It might require a log of work to set up the tests but have you benchmarked the existing test suite with this PR? I guess the most relevant thing is to test the second run after everything has been compiled.

src/univariate/continuous/gumbel.jl
jmxpearson: Why not `12*sqrt(T(6))*zeta(T(3))/pi^3` ?

andreasnoack: `pi^3` is a `Float64`.

src/univariate/continuous/ksdist.jl
jmxpearson: Yes. I just put `s = 1.0` back. I had asked Chris to change Float64 literals in the code, but I think he got carried away in some places like this.

src/univariate/continuous/laplace.jl
jmxpearson:
```julia
julia> typeof(loghalf)
Irrational{:loghalf}
```

src/univariate/continuous/laplace.jl
jmxpearson: Also `Irrational`.

src/univariate/continuous/levy.jl
jmxpearson: Also Irrational.

src/univariate/continuous/normalcanon.jl
jmxpearson: You just mean divide by 2, right? Because I know that the natural parameter is `1/(2σ^2)`, but I think `λ` here is just supposed to be the precision.

andreasnoack: Yes. Sorry for being too brief. Divide with two instead of multiplying with `1/2==0.5`

src/univariate/discrete/binomial.jl
jmxpearson: I put them back. I plan to go through and fix the`suffstats` routines in a later PR. For now, we've left the fitting alone. Again, this is overzealous `.` removal.

test/continuous.jl
jmxpearson: I can revert. I think this was a holdover from before I had code to test all the promoting constructors.

test/univariates.jl
jmxpearson: I have no idea why this got changed. Reverting this and the previous.

jmxpearson: Seems the problem was that `@test` failed for `Skellam` here, so it was switched to `test_approx_eq`. I reordered the definition expressions to be more numerically stable, so the original test should now pass.

jmxpearson:
Okay. All the fixes to the notes you made should be in.

Yes, I regenerated the json file. This is partly because a couple of people (including me) had been editing this by hand for new distributions and tests, which is really brittle. I moved one distribution that had been hand-coded out of the discrete tests and into its own file.

The other reason for regenerating the json was that I was able to add tests for the `Skellam

jmxpearson:
On `master`:
```julia
julia> @time include("runtests.jl")
55.143075 seconds (225.02 k allocations: 80.569 MB, 0.04% gc time)
```

This PR:
```julia
julia> @time include("runtests.jl")
55.618851 seconds (234.07 k allocations: 81.447 MB, 0.04% gc time)
```

Note that there are some extra allocations. I think part of this may have to do with the fact that `MvLogNormal` has a field `normal::MvNormal`, which is no longer a concrete type. This is on my list to fix. There are also additional tests for every univariate added by this PR, so that may be some of it as well.

PR #512, opened 2016-07-14
Replace references to Rmath with StatsFuns

ararslan:
This bumps the StatsFuns dependency to 0.3.0, which provides Rmath stuff. Fixes the libRmath errors on 0.5 and avoids a direct dependency on Rmath.jl.

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -10,11 +10,6 @@ using Compat
 import Compat.view
 import Base.Random
 import Base: size, eltype, length, full, convert, show, getindex, scale, scale!, rand, rand!
-if VERSION < v"0.5.0-"
-    export shape
-else
-    import Base.shape
-end
 import Base: sum, mean, median, maximum, minimum, quantile, std, var, cov, cor
 import Base: +, -, .+, .-
 import Base.Math.@horner
@@ -241,6 +236,7 @@ export
     sampler,            # create a Sampler object for efficient samples
     scale,              # get the scale parameter
     scale!,             # provide storage for the scale parameter (used in multivariate distribution mvlognormal)
+    shape,              # shape parameter for applicable distributions
     skewness,           # skewness of the distribution
     span,               # the span of the support, e.g. maximum(d) - minimum(d)
     std,                # standard deviation of distribution
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.4
 PDMats 0.4.2 0.5
-StatsFuns 0.1.1
+StatsFuns 0.3.0
 Calculus
-StatsBase 0.7.0
+StatsBase 0.8.3
 Compat 0.8.4
--- a/src/samplers/binomial.jl
+++ b/src/samplers/binomial.jl
@@ -4,8 +4,7 @@ immutable BinomialRmathSampler <: Sampleable{Univariate,Discrete}
     prob::Float64
 end
 
-rand(s::BinomialRmathSampler) =
-    round(Int, ccall((:rbinom, "libRmath-julia"), Float64, (Float64, Float64), s.n, s.prob))
+rand(s::BinomialRmathSampler) = round(Int, StatsFuns.RFunctions.binomrand(s.n, s.prob))
 
 
 # compute probability vector of a Binomial distribution
--- a/src/samplers/gamma.jl
+++ b/src/samplers/gamma.jl
@@ -3,8 +3,7 @@ immutable GammaRmathSampler <: Sampleable{Univariate,Continuous}
     d::Gamma
 end
 
-rand(s::GammaRmathSampler) =
-    ccall((:rgamma, "libRmath-julia"), Float64, (Float64, Float64), shape(s.d), scale(s.d))
+rand(s::GammaRmathSampler) = StatsFuns.RFunctions.gammarand(shape(s.d), scale(s.d))
 
 
 # "Generating gamma variates by a modified rejection technique"
--- a/src/samplers/poisson.jl
+++ b/src/samplers/poisson.jl
@@ -3,8 +3,7 @@ immutable PoissonRmathSampler <: Sampleable{Univariate,Discrete}
     mu::Float64
 end
 
-rand(s::PoissonRmathSampler) =
-    round(Int, ccall((:rpois, "libRmath-julia"), Float64, (Float64,), s.mu))
+rand(s::PoissonRmathSampler) = round(Int, StatsFuns.RFunctions.poisrand(s.mu))
 
 
 function poissonpvec(μ::Float64, n::Int)
--- a/src/univariate/continuous/beta.jl
+++ b/src/univariate/continuous/beta.jl
@@ -101,7 +101,7 @@ gradlogpdf(d::Beta, x::Float64) =
 
 #### Sampling
 
-rand(d::Beta) = StatsFuns.Rmath.betarand(d.α, d.β)
+rand(d::Beta) = StatsFuns.RFunctions.betarand(d.α, d.β)
 
 
 #### Fit model
--- a/src/univariate/continuous/chisq.jl
+++ b/src/univariate/continuous/chisq.jl
@@ -73,5 +73,5 @@ gradlogpdf(d::Chisq, x::Float64) =  x > 0.0 ? (d.ν * 0.5 - 1) / x - 0.5 : 0.0
 
 #### Sampling
 
-_chisq_rand(ν::Float64) = StatsFuns.Rmath.chisqrand(ν)
+_chisq_rand(ν::Float64) = StatsFuns.RFunctions.chisqrand(ν)
 rand(d::Chisq) = _chisq_rand(d.ν)
--- a/src/univariate/continuous/erlang.jl
+++ b/src/univariate/continuous/erlang.jl
@@ -60,4 +60,4 @@ cf(d::Erlang, t::Real)  = (1.0 - im * t * d.θ)^(-d.α)
 
 @_delegate_statsfuns Erlang gamma α θ
 
-rand(d::Erlang) = StatsFuns.Rmath.gammarand(d.α, d.θ)
+rand(d::Erlang) = StatsFuns.RFunctions.gammarand(d.α, d.θ)
--- a/src/univariate/continuous/fdist.jl
+++ b/src/univariate/continuous/fdist.jl
@@ -85,4 +85,4 @@ end
 
 @_delegate_statsfuns FDist fdist ν1 ν2
 
-rand(d::FDist) = StatsFuns.Rmath.fdistrand(d.ν1, d.ν2)
+rand(d::FDist) = StatsFuns.RFunctions.fdistrand(d.ν1, d.ν2)
--- a/src/univariate/continuous/gamma.jl
+++ b/src/univariate/continuous/gamma.jl
@@ -84,7 +84,7 @@ cf(d::Gamma, t::Real) = (1.0 - im * t * d.θ)^(-d.α)
 gradlogpdf(d::Gamma, x::Float64) =
     insupport(Gamma, x) ? (d.α - 1.0) / x - 1.0 / d.θ : 0.0
 
-rand(d::Gamma) = StatsFuns.Rmath.gammarand(d.α, d.θ)
+rand(d::Gamma) = StatsFuns.RFunctions.gammarand(d.α, d.θ)
 
 
 #### Fit model
--- a/src/univariate/continuous/noncentralchisq.jl
+++ b/src/univariate/continuous/noncentralchisq.jl
@@ -35,4 +35,4 @@ end
 
 @_delegate_statsfuns NoncentralChisq nchisq ν λ
 
-rand(d::NoncentralChisq) = StatsFuns.Rmath.nchisqrand(d.ν, d.λ)
+rand(d::NoncentralChisq) = StatsFuns.RFunctions.nchisqrand(d.ν, d.λ)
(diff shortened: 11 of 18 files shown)


src/Distributions.jl
```diff
@@ -10,11 +10,6 @@ using Compat
 import Compat.view
 import Base.Random
 import Base: size, eltype, length, full, convert, show, getindex, scale, scale!, rand, rand!
-if VERSION < v"0.5.0-"
-    export shape
-else
-    import Base.shape
```
tkelman: maybe it did temporarily? ref https://github.com/JuliaStats/Distributions.jl/pull/506, cc @ranjanan when you do things that depend on `VERSION` please be as specific as you can - or better, use `isdefined`

ararslan:
Ugh, the 0.5 tests are timing out.

ararslan:
The long build times on 0.5 are probably because it's building libRmath for the StatsFuns dependency. Not sure there's much we can do about that.

tkelman:
It's not a very big library, the build shouldn't take all that long? Does it pass for you locally? May be either the deprecation warnings causing a big slowdown, or some bug causing the tests to actually hang.

ararslan:
Yeah it works for me locally, and a couple of the builds have gotten to the very last test before timing out.

jmxpearson:
Two things to note:
1. There are a lot of warnings, which I'm given to believe can slow things down. Some of them at least seem to be due to methods in Rmath overwriting other methods in Rmath.
2. Tests on nightly have been taking twice as long as v0.4. The cap is 49 minutes, and on #511, they've been taking 41-42 minutes. Even a small delay *might* be screwing it up?

@ararslan Thanks for all this. I had already had to fix the `if VERSION` on my branch, since `Base.shape` *was* a thing for a few days on master.

ararslan:
It looks like most (all?) of the warnings are coming from Rmath itself rather from Distributions. Since this works locally and works up until timeout on Travis, I'm inclined to say we should go ahead and merge it. In the meantime, I think the unflappable @tkelman is working on Rmath binaries which should speed things up, hopefully at least enough for the Travis builds to finish. What do you think?

jmxpearson:
I'd like to get this merged for reasons of my own, but it's not good if the tests are basically crippled. Thought I guess they currently are anyway with the failure. 

I'll defer to @andreasnoack and @tkelman.

ararslan:
Yeah, I'm not sure what we can do about the 0.5 tests at the moment. Hopefully Andreas or Tony will have some insight.

tkelman:
if your tests really take 40 minutes you should be splitting them up into a more fine grained test matrix, or splitting this package up into smaller components if it really needs tests that take longer to run than the test of the base language itself.

ararslan:
The weird thing is that the 0.4 tests take less than 20 minutes but the 0.5 tests time out at 50. Any idea why they would be that different?

tkelman:
if this is leveraging rmath's random number generation, then this is one of the benchmarking things I was hoping @simonbyrne could do before @StefanKarpinski prematurely deleted the version from base instead of scrambling to fix things that are also accidentally broken on 0.4.

ararslan:
Yes, I believe all of the sampling functions here are using Rmath's random stuff.

jmxpearson:
There was some work done to revert part of the inlining penalty we get for turning code coverage on, but there are more inlines in v0.5 than v0.4 in the random sampling code. And we call those functions tens of millions of times. The last fix brought us down from something like 2.5 hours to 40 min (on my machine).

jmxpearson:
Also, if you run the tests locally with `noinline`, the vast majority of the time is taken up testing the multivariate lognormal and normal distributions (maybe also the multivariate t-dist). Everything else is pretty snappy, but those tests are calling `randn` a *lot*.

ararslan:
`randn` is from Base rather than Rmath though, isn't it?

jmxpearson:
Yes. I think in v0.4 there are 2 inlined calls; in v0.5 there are 3. Not entirely an explanation, but I think that's some of it.

andreasnoack:
I tried to restart the build after https://github.com/JuliaLang/METADATA.jl/pull/5627 has been merged but it still times out. Last successful build from two weeks ago only took 32 minutes so it would be great to find out what has happened since then.

tkelman:
Please do profile. We may need to paste more of dsfmt into Rmath-julia now that it's built outside of Julia.

ararslan:
I wonder if syncing libRmath with the most recent R version would help? Looks like that hasn't been done for a couple years.

andreasnoack:
I'd like to try timing the tests right before the Rmath changes to Base. Actually, we don't know if it is Rmath which is causing the slowdown since latest successful Travis build is two weeks old.

andreasnoack:
Locally on
```jl
julia> versioninfo()
Julia Version 0.5.0-dev+5181
Commit 853317d* (2016-07-06 14:00 UTC)
Platform Info:
  System: Linux (x86_64-redhat-linux)
  CPU: Intel(R) Xeon(R) CPU E5-2697 v2 @ 2.70GHz
  WORD_SIZE: 64
  BLAS: libmkl_rt
  LAPACK: libmkl_rt
  LIBM: libopenlibm
  LLVM: libLLVM-3.7.1 (ORCJIT, ivybridge)
```
I get
```jl
        From worker 3:      testing Truncated(TriangularDist(1, 3, 2),3,5)
        From worker 3:      testing Truncated(TruncatedNormal(0, 1, -2, 2),3,5)
        From worker 3:      testing Truncated(TruncatedNormal(3, 10, 7, 8),3,5)
        From worker 3:      testing Truncated(TruncatedNormal(27, 3, 0, Inf),3,5)
        From worker 3:      testing Truncated(TruncatedNormal(-5, 1, -Inf, -10),3,5)
        From worker 3:      testing Truncated(TruncatedNormal(1.8, 1.2, -Inf, 0),3,5)
        From worker 3:      testing Truncated(Uniform(0.0, 1.0),3,5)
        From worker 3:      testing Truncated(Uniform(3.0, 17.0),3,5)
        From worker 3:      testing Truncated(Uniform(3.0, 3.1),3,5)
        From worker 3:      testing Truncated(Weibull(),3,5)
        From worker 3:      testing Truncated(Weibull(0.5),3,5)
        From worker 3:      testing Truncated(Weibull(5.0),3,5)
        From worker 3:      testing Truncated(Weibull(20.0, 1.0),3,5)
        From worker 3:      testing Truncated(Weibull(1.0, 2.0),3,5)
        From worker 3:      testing Truncated(Weibull(5.0, 2.0),3,5)
        From worker 3:
Error During Test
  Expression evaluated to non-Boolean
  Expression: include("runtests.jl")
       Value: nothing
 in record(::Base.Test.FallbackTestSet, ::Base.Test.Error) at ./test.jl:397
 in do_test(::Base.Test.Returned, ::Expr) at ./test.jl:281
 in eval(::Module, ::Any) at ./boot.jl:234
 in macro expansion at ./REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:46
```
which is exactly the place where Travis hangs so maybe this is the issue. I'll investigate further.

ararslan:
What's weird is that Travis has made it past that point a couple of times.

ararslan:
Wow, could I really be that stupid? I think I was looking at the 0.4 logs when I said it got past that point, and when I said it works for me locally I was running it on 0.4. *slams face against desk*

andreasnoack:
I think the tests hang because of a segfault in the `univariates` tests. So the good news is that there might not be a slowdown. It's still unclear what is causing the segfault.

ararslan:
Maybe I'm not crazy after all... I just checked out this branch, did `Pkg.build("Distributions")` and `Pkg.test("Distributions")`, and not only did the tests pass on my computer, but the whole ordeal took just about 5 minutes.

For reference:
```julia
julia> versioninfo()
Julia Version 0.5.0-dev+5453
Commit 1fd440e (2016-07-15 23:33 UTC)
Platform Info:
  System: Darwin (x86_64-apple-darwin14.5.0)
  CPU: Intel(R) Core(TM) i7-2820QM CPU @ 2.30GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Sandybridge)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.7.1 (ORCJIT, sandybridge)
```

tkelman:
you can contact travis support and ask for 24 hr ssh access into a worker, if that would help reproduce

ararslan:
That sounds like it could be a useful thing to do. Would we be able to see anything other than what we just get in the build log though?

andreasnoack:
You'll have to run the tests with `Pkg.test("Distributions",coverage=true)` to be comparable with Travis. It turns inlining off which is the reason why the tests run so slowly. You might also want to launch julia with `JULIA_NUM_THREADS=2` to be comparable to Travis.

ararslan:
Ah perfect, thanks @andreasnoack. I'll try that now.

tkelman:
you'll be able to catch the segfault in lldb or gdb and get a backtrace, inspect state etc, which isn't really doable if you can't reproduce locally

ararslan:
Oh, gotcha. My knowledge of GDB et al. is pretty shaky so I should probably defer the Travis SSH stuff to someone else.

andreasnoack:
I saw the segfault on a version of Julia pre the Rmath changes but @yuyichao told me that the issue had been fixed. I haven't been able to get a segfault on this PR. I've just run the test without inlining and they take ~37 minutes on my laptop which is comparable to what @jmxpearson reported earlier.

ararslan:
Alright, so when emulating the Travis conditions, the tests take an hour on my system and end with this:
```julia
WARNING: Forcibly interrupting busy workers
InterruptException(InterruptException())InterruptException(WARNING: Unable to terminate all workers
InterruptException()InterruptException(INFO: Distributions tests passed
)InterruptException(InterruptException(InterruptException(INFO: No packages to install, update or remove
```

andreasnoack:
I've done some more profiling and bisecting. It noticed that the multivariate normal tests were way slower on 0.5 and saw that it spent much time in `broadcast!`. It appears that `c68a2cd9919cdf4a23243b313670ad8dc41a9c27` made

```jl
x = randn(10^6)
y = zeros(x)
@time broadcast!(+, x, x, y);
```
five times slower when inlining is off. It doesn't seem to have an effect with inlining on. I think this alone could cause the timeout but will look more into it.

It would really be great if we could test with inlining on. Anybody knows how bad the coverage reports become?

andreasnoack:
We should probably merge this. It fixes the Rmath issue and I only think there is a slowdown then julia is run without inlining. However, there is an issue on 0.4. I restarted the tests and got

https://travis-ci.org/JuliaStats/Distributions.jl/jobs/144873119

Eventually, we should find a solution such that we can get green lights again either by running the tests with inlining or figuring out if Base can be changed to be faster again but that is separate from this PR. To show how fast the tests run with inlining I did

https://travis-ci.org/JuliaStats/Distributions.jl/builds/145361103

ararslan:
I bet the 0.4 packages issue is the same thing we're seeing in https://github.com/JuliaLang/METADATA.jl/pull/5625. It must be because StatsBase requires StatsFuns 0.3.0 and allows Julia 0.4 but StatsFuns 0.3.0 was temporarily set to Julia 0.5 only to avoid breaking Distributions on 0.4 for everyone. I have a hunch that if we set StatsFuns 0.3.0 back to allowing 0.4, Travis would pass both here and https://github.com/JuliaLang/METADATA.jl/pull/5625.

andreasnoack:
I've added a commit which reduces the number of iterations for the `MvNormal` and `MvLogNormal`. This is a huge speedup so the tests now complete well in time. Now the only problems are https://github.com/JuliaLang/METADATA.jl/pull/5613#issuecomment-233195922 and broken nightlies for mac.

ararslan:
Could not load libgit2? https://travis-ci.org/JuliaStats/Distributions.jl/jobs/145424861#L146 😕

ararslan:
Looks like everything is golden save for this very strange and (hopefully) unrelated OS X issue.

tkelman:
https://github.com/JuliaLang/julia/issues/17463, should hopefully be resolved when the next binaries make it up

ararslan:
Looks like https://github.com/JuliaLang/julia/pull/17468 fixed that. Whose night defines "nightly"?

tkelman:
See https://status.julialang.org/ for how old the nightlies are, and https://build.julialang.org/builders for how they get rebuilt. At the moment the way it works is: every commit to master triggers a set of "build" jobs that build and run the tests on a few platforms, and if those all succeed, then the same commit is sent to the set of "package" jobs that build binaries for a few more platforms. We will probably combine that at some point, there are a few issues preventing the tests from properly working under buildbot on windows, and I'm not sure if ARM is passing all its tests (right now it doesn't build on master but hopefully that'll get fixed).

ararslan:
Oh nice, thanks for the links and the explanation! That's good to know (and very interesting).

I guess the new nightlies should be up pretty soon then. 🎉 Then we can restart Travis and hopefully get this bad boy merged!

tkelman:
should maybe find a central "how the sausage gets made" place to put this kind of information so people can find out without having to ask me. JuliaLang/julia/DISTRIBUTING.md definitely needs a rewrite to account for this stuff.

ararslan:
Sorry for asking you things all the time...

tkelman:
believe me, the more people that know these things the better

ararslan:
Okay, rebased and built Julia from master then updated, rebuilt, and tested Distributions from this branch on OS X. Everything looks good. 👍

ararslan:
That green Travis badge looks great on the README. :smile: Thanks so much for your help, everyone!

ararslan:
We should tag a new version now that this is merged. I can do that in several hours if no one beats me to it.

PR #521, opened 2016-08-09
changed binomial to accept generic first argument

jmxpearson:
This addresses the issue raised in JuliaGraphs/LightGraphs.jl#407. It's entirely reasonable for users to be able to write
```julia
Binomial(1., 0.75)
```
Previously, the inner constructor handled the implicit conversion of the first argument to Int. This just restores that behavior. 
```julia
Binomial(1.5, 0.5)
```
will throw an `InexactError` as before.

### DIFF
--- a/src/univariate/discrete/betabinomial.jl
+++ b/src/univariate/discrete/betabinomial.jl
@@ -30,7 +30,6 @@ end
 
 BetaBinomial{T<:Real}(n::Integer, α::T, β::T) = BetaBinomial{T}(n, α, β)
 BetaBinomial(n::Integer, α::Real, β::Real) = BetaBinomial(n, promote(α, β)...)
-BetaBinomial(n::Real, α::Real, β::Real) = Base.depwarn("BetaBinomial(n::Real, α, β) is deprecated. Please use BetaBinomial(n::Integer, α, β) instead.", :BetaBinomial)
 BetaBinomial(n::Integer, α::Integer, β::Integer) = BetaBinomial(n, Float64(α), Float64(β))
 
 @distr_support BetaBinomial 0 d.n
--- a/src/univariate/discrete/betabinomial.jl
+++ b/src/univariate/discrete/betabinomial.jl
@@ -28,9 +28,10 @@ immutable BetaBinomial{T<:Real} <: DiscreteUnivariateDistribution
     end
 end
 
-BetaBinomial{T<:Real}(n::Int, α::T, β::T) = BetaBinomial{T}(n, α, β)
-BetaBinomial(n::Int, α::Real, β::Real) = BetaBinomial(n, promote(α, β)...)
-BetaBinomial(n::Int, α::Integer, β::Integer) = BetaBinomial(n, Float64(α), Float64(β))
+BetaBinomial{T<:Real}(n::Integer, α::T, β::T) = BetaBinomial{T}(n, α, β)
+BetaBinomial(n::Integer, α::Real, β::Real) = BetaBinomial(n, promote(α, β)...)
+BetaBinomial(n::Real, α::Real, β::Real) = Base.depwarn("BetaBinomial(n::Real, α, β) is deprecated. Please use BetaBinomial(n::Integer, α, β) instead.", :BetaBinomial)
+BetaBinomial(n::Integer, α::Integer, β::Integer) = BetaBinomial(n, Float64(α), Float64(β))
 
 @distr_support BetaBinomial 0 d.n
 insupport(d::BetaBinomial, x::Real) = 0 <= x <= d.n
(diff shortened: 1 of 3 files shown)


andreasnoack:
This shouldn't be a problem as mentioned in 

https://github.com/JuliaStats/Distributions.jl/commit/c3892d96ba6fb9c1c22d311d0a87778359c37b97#commitcomment-18573874

I'm wondering if it's more helpful with the tighter signature. This discussion is quite similar to the discussion about allowed types for indices in `getindex`. Typically, you'd always get the index or, in this case, length of the `Binomial` as a result of integer operations. So the better fix is to use integer division instead of converting back and forth between integers and floating point values.

jmxpearson:
Makes sense.

sbromberger:
For the record, I'm not opposed to having `Int` as the first argument. My main concern is with a larger issue of API stability with highly-dependent packages and the fact that we got reports that LG was suddenly failing tests on a Sunday evening, which is no fun.

andreasnoack:
@sbromberger We might be far away from having a way of running the package evaluator before tagging a new version. That would be really useful and would have been able to fix LG before we tagged a new version of `Distributions`. However, we are not there yet so would it help if we didn't tag during weekends?

For the reasons mention earlier, I think we should keep the integer restriction here, but it should probably be `Integer` instead of `Int`.

sbromberger:
No - please don't change / delay whatever you're doing. This is an issue for the larger Julia community to solve and is not unique to Distributions by any stretch! We'll muddle through until a solution presents itself, and I'm fully on board with keeping this as `Integer` - we've made the necessary corrections in LG.

tkelman:
LightGraphs wasn't the only package broken by #511. BayesNets, FixedEffectModels, GLM, MixedModels, and StatsBase also had issues that look related to the tag of Distributions.

jmxpearson:
@tkelman: Do you have a sense of what kinds of errors these are? This particular one was an oversight, but in general, the API was supposed to be preserved. That's why it was tagged as a minor release. 

If there's a hole in the test suite that accounts for a common source of these new errors, I'm happy to plug it. Coverage did increase with that PR by a nontrivial amount, but it would be good to know what we're missing.

tkelman:
See the logs linked from http://pkg.julialang.org/pulse.html. They will get overwritten in a little over 12 hours when PkgEval next finished, but you can get the old versions by looking at older revisions of e.g. https://github.com/JuliaCI/pkg.julialang.org/blob/gh-pages/logs/LightGraphs_0.4.log.

The rewrite was tagged as a patch release which I think was a mistake, but we've now superseded that patch release with one that was branched from before the merge.

jmxpearson:
My error. Yes, patch. 

Looking at the logs, there is another instance of the error related to this PR. There is also an ambiguity warning I'll fix tomorrow that is not causing tests to fail. The modeling packages appear to be erroring on an unrelated DataFrames issue. Am I missing something else?

tkelman:
That looks like it. BlsData has flaky tests, don't mind it. I forgot that DataFrames also had a tag yesterday, so maybe it shares some blame for breaking a few of these. cc @araslan

jmxpearson:
Ambiguity warning: #523.

andreasnoack:
The right solution here is probably to deprecate the non-integer version. Then we'll not break code but still encourage people to write integer arithmetic with integers.

jmxpearson:
So I just pushed a fix, but I might have done the wrong thing. I used `Base.depwarn` but did not return anything. Better to warn and also return?

src/univariate/discrete/betabinomial.jl
```diff
@@ -28,9 +28,10 @@ immutable BetaBinomial{T<:Real} <: DiscreteUnivariateDistribution
     end
 end
 
-BetaBinomial{T<:Real}(n::Int, α::T, β::T) = BetaBinomial{T}(n, α, β)
-BetaBinomial(n::Int, α::Real, β::Real) = BetaBinomial(n, promote(α, β)...)
-BetaBinomial(n::Int, α::Integer, β::Integer) = BetaBinomial(n, Float64(α), Float64(β))
+BetaBinomial{T<:Real}(n::Integer, α::T, β::T) = BetaBinomial{T}(n, α, β)
+BetaBinomial(n::Integer, α::Real, β::Real) = BetaBinomial(n, promote(α, β)...)
+BetaBinomial(n::Real, α::Real, β::Real) = Base.depwarn("BetaBinomial(n::Real, α, β) is deprecated. Please use BetaBinomial(n::Integer, α, β) instead.", :BetaBinomial)
```
ararslan: These should go in src/deprecates.jl

jmxpearson: I get putting them in deprecates, but I thought `@deprecate` deprecated the entire symbol?

jmxpearson:
I think this is ready to merge. Any objections?

ararslan:
None from me; looks good. Thanks!

PR #522, opened 2016-08-09
more parametric distributions

jmxpearson:
This PR encompasses two sets of changes:
- fleshing out conversions (with tests) for multivariate and matrix-variate distributions that had already been parameterized
- adding the last unparameterized multivariates:
  - `MvLogNormal`
  - `Multinomial`
  - `VonMisesFisher`

On the roadmap to fully parametric, there remains:

1. Fixing remaining hardcoded `Float64`s. In particular:
  - aliases like `ZeroMeanIsoNormal` in the multivariates
  - code for computing MLEs, where type stability is trickier to think through
1. Truncated distributions
1. Mixtures (though I was under the impression that these were moving out; should this be a priority?)

### DIFF
--- a/src/matrix/inversewishart.jl
+++ b/src/matrix/inversewishart.jl
@@ -29,7 +29,7 @@ InverseWishart(df::Real, Ψ::Cholesky) = InverseWishart(df, PDMat(Ψ))
 function _invwishart_c0(df::Real, Ψ::AbstractPDMat)
     h_df = df / 2
     p = dim(Ψ)
-    h_df * (p * logtwo - logdet(Ψ)) + logmvgamma(p, h_df)
+    h_df * (p * typeof(df)(logtwo) - logdet(Ψ)) + logmvgamma(p, h_df)
 end
 
 
@@ -42,6 +42,16 @@ dim(d::InverseWishart) = dim(d.Ψ)
 size(d::InverseWishart) = (p = dim(d); (p, p))
 params(d::InverseWishart) = (d.df, d.Ψ, d.c0)
 @inline partype{T<:Real}(d::InverseWishart{T}) = T
+
+### Conversion
+function convert{T<:Real}(::Type{InverseWishart{T}}, d::InverseWishart)
+    P = convert_eltype(T, d.Ψ)
+    InverseWishart{T, typeof(P)}(T(d.df), P, T(d.c0))
+end
+function convert{T<:Real}(::Type{InverseWishart{T}}, df, Ψ::AbstractPDMat, c0)
+    P = convert_eltype(T, Ψ)
+    InverseWishart{T, typeof(P)}(T(df), P, T(c0))
+end
 
 #### Show
 
--- a/src/multivariate/multinomial.jl
+++ b/src/multivariate/multinomial.jl
@@ -1,8 +1,8 @@
-immutable Multinomial <: DiscreteMultivariateDistribution
+immutable Multinomial{T<:Real} <: DiscreteMultivariateDistribution
     n::Int
-    p::Vector{Float64}
+    p::Vector{T}
 
-    function Multinomial(n::Integer, p::Vector{Float64})
+    function Multinomial(n::Integer, p::Vector{T})
         if n <= 0
             throw(ArgumentError("n must be a positive integer."))
         end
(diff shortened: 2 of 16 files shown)


src/matrix/inversewishart.jl
```diff
@@ -43,6 +43,16 @@ size(d::InverseWishart) = (p = dim(d); (p, p))
 params(d::InverseWishart) = (d.df, d.Ψ, d.c0)
 @inline partype{T<:Real}(d::InverseWishart{T}) = T
 
+### Conversion
+function convert{T<:Real}(::Type{InverseWishart{T}}, d::InverseWishart)
+    P = convert_eltype(T, d.Ψ)
```
andreasnoack: I didn't notice that you'd defined these. The idiom for this in base is
```jl
convert(AbstractMatrix{T},d.Ψ)
```

src/multivariate/multinomial.jl
```diff
@@ -118,16 +122,18 @@ end
...
+    s = R(lgamma(n + 1))
+    t = zero(T)
     for i = 1:length(p)
         @inbounds xi = x[i]
-        @inbounds pi = p[i]
+        @inbounds p_i = p[i]
         t += xi
-        s -= lgamma(xi + 1.0)
-        @inbounds s += xi * log(pi)
+        s -= R(lgamma(xi + 1))
```
andreasnoack: Is `lgamma` promoting too much here?

andreasnoack:
Great work.

src/multivariate/multinomial.jl
jmxpearson: `lgamma(Float32(1) + 1)` is `Float32`
`lgamma(Float16(1) + 1)` is `Float32`

If `T` is `Int` and `S` is `Float32`, then `R` is `Float32` but `lgamma(xi + 1)` is `Float64`.
That's a bit of a weird edge case, but it could come up.

andreasnoack: I get
```julia
julia> typeof(lgamma(Float32(1.0) + 1))
Float32
```

src/matrix/inversewishart.jl
jmxpearson: Cool. I knew there had to be some way.

These (`convert_eltype`) are new to this PR and aren't needed if there are other facilities for handling it. For instance, I can easily define `convert(ZeroVector{T}, Z)`.

Here, though, we'd want `convert(AbstractPDMat{T}, d.Ψ)` to avoid calculating the Cholesky over again, right? Should I submit a PR in PDMats for that?

src/multivariate/multinomial.jl
jmxpearson: Right. But if `xi = 1`, then 
```julia
julia> typeof(lgamma(xi + 1))
Float64
```
and if `typeof(d)` is `Multinomial{Float32}`, we need this to be `Float32`.

src/matrix/inversewishart.jl
jmxpearson: One more question related to that: it's trivial to define the `convert` for a `ZeroVector`, but since `ZeroVector <: Any`, I can't just `convert(AbstractArray{T}, d.μ)`. Any ideas for a clean way to do that using dispatch while still getting rid of my little helper function?

andreasnoack: The way it's done in Base is that each special matrix defines
```jl
convert(::Type{AbstractMatrix{T},A::SpecialMatrix) = convert(SpecialMatrix{T}, A)
```
However, your second comment is quite relevant. We might want to decouple conversion of the element type from abstract types. Many different possibly unrelated types could have elements for which it would be relevant to convert them. So it's worth considering if Base should adopt something like `convert_eltype` instead of opposite.

jmxpearson: Well, if we leave in `convert_eltype`, I'll simply alias to `convert` for `ZeroVector` and `AbstractArray`. If we update `PDMats`, then that can be aliased, too. 

Happy to go either way you think is best.

jmxpearson: So I tweaked the code. `convert_eltype` still seems to me the cleanest solution for now, but it's essentially just aliasing `convert` methods of the type you suggested, so it should be easy to swap out for an alternative in Base later, if available.

andreasnoack: I agree. Let's keep it like this and merge when tests pass.

src/multivariate/multinomial.jl
```diff
@@ -118,16 +122,18 @@ end
 function _logpdf{T<:Real}(d::Multinomial, x::AbstractVector{T})
     p = probs(d)
     n = ntrials(d)
-    s = lgamma(n + 1.0)
-    t = 0
+    S = eltype(p)
+    R = promote_type(T, S)
+    s = R(lgamma(n + 1))
```
andreasnoack: Couldn't `R` end up being `Int` if both `T` and `S` are `Int`s. The this would throw an inexact error.

jmxpearson: Except `p` must be a probability vector, so we can't have `S <: Integer`.

andreasnoack: Sorry. I read it as `S = eltype(d)`.

PR #523, opened 2016-08-10
Fix Skellam cdf ambiguity warning

jmxpearson:
Adding cdf(d::Skellam, x::Int)
- avoids ambiguity warning in v0.4


tkelman:
On 0.5 there's now a `detect_ambiguities` function, might be worth testing that out and seeing if it would have caught this and trying to add it to the tests.

jmxpearson:
Thanks! I've added this to the tests. It caught an ambiguity that wasn't previously showing up in the warnings.

PR #529, opened 2016-09-01
rename scale parameter from beta to theta

jonalm:
Minor change for consistency with wikipedia. beta is inverse of theta and this and it is therefore confusing to call beta the scale.

### DIFF
--- a/src/univariate/continuous/gamma.jl
+++ b/src/univariate/continuous/gamma.jl
@@ -4,7 +4,7 @@ doc"""
 The *Gamma distribution* with shape parameter `α` and scale `θ` has probability density
 function
 
-$f(x; \alpha, \beta) = \frac{x^{\alpha-1} e^{-x/\beta}}{\Gamma(\alpha) \beta^\alpha},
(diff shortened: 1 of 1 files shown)


jonalm:
maybe also change \alpha to k?

nalimilan:
I suspect the current naming for chosen for consistency between distributions, which in some cases implies diverging from Wikipedia. Can you check similar distributions with shape and scale?

Cc: @lindahua

jmxpearson:
LGTM. Note that the actual fields in the type are `α` and `θ`, so the old docstring simply didn't match the code.

I don't really see a case for changing `α` to `k`, since this would either (a) make the docstring inconsistent with the type's fields (if we change the doc) or (b) break code (if we changed the field name).

andreasnoack:
See also https://github.com/JuliaStats/Distributions.jl/issues/395

matbesancon:
@andreasnoack appart from this issue, should this be in the documentation? In the "developing new dists" section

matbesancon:
This looks fixed now, closing

PR #531, opened 2016-09-09
composite distribution

eford:
Some algorithms require a Distribution type.  For models with several parameters, its common for different parameters to come from different distributions.  The CompositeDist type allows users to easily create a Distribution type in which different variables are drawn from different types.  
I'll be using this for my own work.  I'm happy to share if it's useful to others.  If not, that's ok, too.

### DIFF
--- a/src/multivariate/composite.jl
+++ b/src/multivariate/composite.jl
@@ -1,0 +1,188 @@
+# Composite Multivariate distribution
+
+## Generic Composite multivariate distribution class
+
+module CompositeDistributions
+
+if !isdefined(:Distributions) using Distributions end
+
+import Base.length, Base.mean, Base.show
+import Distributions.mean, Distributions.mode, Distributions.var, Distributions.cov
+import Distributions.entropy, Distributions.insupport
+import Distributions._logpdf, Distributions._pdf!
+import Distributions._rand!
+
+export AbstractCompositeContinuousDist, ContinuousMultivariateDistribution
+export GenericCompositeContinuousDist, CompositeDist
+export length, params, mean, mode, var, cov, entropy
+export insupport, _logpdf, gradlogpdf, _rand!
+
+
+abstract AbstractCompositeContinuousDist <: ContinuousMultivariateDistribution
+#abstract AbstractCompositeDiscreteDist <: DiscreteMultivariateDistribution  # An idea, but not implemented yet
+
+immutable GenericCompositeContinuousDist <: AbstractCompositeContinuousDist
+    dist::Vector{ContinuousDistribution}
+    indices::Vector{UnitRange{Int64}}
+
+    function GenericCompositeContinuousDist(dist::Vector{ContinuousDistribution})
+      length(dist) > 0 || error("number of distributions must be positive")
+      #@compat new(dist)
+      indices = Array(UnitRange{Int64},length(dist))
+      idx_start = 1; idx_stop = 0
+      for i in 1:length(dist)
+         idx_stop += length(dist[i])
+         indices[i] = idx_start:idx_stop
+         idx_start += length(dist[i])
+      end
+      new(dist,indices)
+    end
+end
+
+typealias CompositeDist  GenericCompositeContinuousDist
+
+### Parameters
+length(d::GenericCompositeContinuousDist, idx::Integer) = length(d.dist[idx])
+length(d::GenericCompositeContinuousDist)  = d.indices[end].stop
+
+params(d::GenericCompositeContinuousDist, idx::Integer) = params(d.dist[idx])
+function params(d::GenericCompositeContinuousDist) 
+  n = length(d)
+  p = Array(Float64,n)
+  for i in 1:length(d.dist)
+     p[d.indices[i]] = params(d,i)
+  end
+  return p
+end
+
+#=  Seems like a good idea, but would require set_params for all distributions used
+function set_params(d::GenericCompositeContinuousDist, x::DenseVector{T}) 
+  n = length(d)
+  for i in 1:length(d.dist)
+     set_params(d.dist[i], params(d,i) )
+  end
+end
+=#
+
+### Basic statistics
+
+mean(d::GenericCompositeContinuousDist, idx::Integer) = mean(d.dist[idx])
+function mean(d::GenericCompositeContinuousDist) 
+  n = length(d)
+  mu = Array(Float64,n)  
+  for i in 1:length(d.dist)
+     mu[d.indices[i]] = mean(d,i)
+  end
+  return mu
+end
+
+mode(d::GenericCompositeContinuousDist, idx::Integer) = mode(d.dist[idx])
+function mode(d::GenericCompositeContinuousDist) 
+  n = length(d)
+  mo = Array(Float64,n)
+  for i in 1:length(d.dist)
+     mo[d.indices[i]] = mode(d,i)
+  end
+  return mo
+end
+
+var(d::GenericCompositeContinuousDist, idx::Integer) = var(d.dist[idx])
+function var(d::GenericCompositeContinuousDist) 
+  n = length(d)
+  v = Array(Float64,n)
+  for i in 1:length(d.dist)
+     v[d.indices[i]] = var(d,i)
+  end
+  return v
+end
+
+cov(d::GenericCompositeContinuousDist, idx::Integer) = cov(d.dist[idx])
+function cov(d::GenericCompositeContinuousDist) 
+  n = length(d)
+  covar = zeros(Float64,(n,n) )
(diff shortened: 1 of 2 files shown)


johnmyleswhite:
I think something like this would be valuable, but I think the approach we'll want to take is somewhat different since I think we want to work with tuples rather than vectors.

src/multivariate/composite.jl
```diff
@@ -0,0 +1,203 @@
+# Composite Multivariate distribution
+
+## Generic Composite multivariate distribution class
+
+module CompositeDistributions
```
johnmyleswhite: Putting this into a separate module really complicates the code (e.g. the many lines of imports shown below). I would not do this in production code.

src/multivariate/composite.jl
```diff
@@ -0,0 +1,203 @@
...
+
+if !isdefined(:Distributions) using Distributions end
```
johnmyleswhite: This line seems like it was relevant to a version of this code that lived outside of Distributions, but not a version that lives inside of Distributions.

src/multivariate/composite.jl
```diff
@@ -0,0 +1,203 @@
...
+
+import Base.length, Base.mean, Base.show
+import Distributions.params
+import Distributions.mean, Distributions.mode, Distributions.var, Distributions.cov
+import Distributions.entropy, Distributions.insupport
+import Distributions._logpdf, Distributions._pdf!
+import Distributions._rand!
+
+export AbstractCompositeContinuousDist, ContinuousMultivariateDistribution
```
johnmyleswhite: We don't abbreviate `Dist` in any type except the distributions of the F and t statistics. I would go for the more complete name that matches the majority of the other names.

src/multivariate/composite.jl
```diff
@@ -0,0 +1,203 @@
...
+export GenericCompositeContinuousDist, CompositeDist
+export length, params, set_params!, mean, mode, var, cov, entropy
+export insupport, _logpdf, gradlogpdf, _rand!
+
+
+abstract AbstractCompositeContinuousDist <: ContinuousMultivariateDistribution
+#abstract AbstractCompositeDiscreteDist <: DiscreteMultivariateDistribution  # An idea, but not implemented yet
+
+immutable GenericCompositeContinuousDist <: AbstractCompositeContinuousDist
+    dist::Vector{ContinuousDistribution}
```
johnmyleswhite: I think this should be a tuple. Were it to remain a vector, you'd still want this type to be a parametric type since the current representation of this type creates a quasi-black-box that type inference can't fully see into. Usually this sort of thing looks like:

```julia
immutable GenericCompositeContinuousDist{T <: ContinuousDistribution} <: AbstractCompositeContinuousDist
    dist::Vector{T}
[...]
```

eford:
Thanks for the feedback, @johnmyleswhite .
As you noticed, I've written it so that it can live outside Distributions.jl, at least until we decide whether it should go into Distributions.jl.  Of course, I'd be happy to adjust type names for consistency.  
I see the potential virtue in your suggestion of using Tuples.  I'm fluent in vectors, but would have to spend time figuring out how to implement Tuples of potentially different lenghts without it being a kludge.  So I'm not opposed to it, but I don't think that I could justify spending my time to implement that.  
If someone volunteers to help switching from Vector to Tuple, then great.  
If not, then I'd be inclined to create a separate package, e.g., CompositeDistributions.jl, so people who want to use it as is can.  (Any suggestions on package names?)

src/multivariate/composite.jl
eford: But the who point is that I want to be able to have different variables come from different distributions.   Wouldn't your suggestion for a vector of parametric types force each member of the vector to be from the same distribution?

johnmyleswhite: You're right: my suggestion won't work. Given your goal of allowing different types in the vector, the only possible way to create a new type that will satisfy `isbits` (and therefore be amenable to stack allocation) is to use a tuple instead of a vector.

johnmyleswhite:
Here's a very rough outline for how I'd do this using tuples:

```julia
import Distributions: Normal, Gamma, Distribution

# Constructor
immutable CompositeDistribution{T <: Tuple} <: Distribution
    rvs::T

    function CompositeDistribution(rvs::Tuple)
        @assert all(rv -> isa(rv, Distribution), rvs)
        new(rvs)
    end
end

# Need to create an outer constructor because of the inner constructor + parametric type
CompositeDistribution{T}(rvs::T) = CompositeDistribution{T}(rvs)

# These kinds of methods could be generated with a macro since they all look the same
Base.rand(d::CompositeDistribution) = map(rv -> rand(rv), rvs)

# Let's use this type
rvs = (Normal(0, 1), Gamma(1, 1))
d = CompositeDistribution(rvs)
tpl = rand(d)
typeof(tpl)
```

This code would be fairly inefficient on 0.4, but should be extremely fast on 0.5.

To be possibly pedantically clear: the reason I think tuples are the way to handle this is because the most tightly-typed vector that can contain a draw from a tuple of Normal, Binomial and Dirichlet is so close to being a `Vector{Any}` that it essentially will act as `Vector{Any}` when type inference looks at it.

johnmyleswhite:
I'm not sure what to call this if you put it into a package. If you were use tuples, I might call it a TupleDistribution since it's so general at that point that I don't know what else to call it. CompositeDistributions.jl works for me, but I wouldn't know offhand what that meant. If I saw something like TupleDistributions.jl I would be clear what it does, although it would take me a few minutes to realize why it would be useful.

eford:
Thanks for the suggestions.
```
# These kinds of methods could be generated with a macro since they all look the same
Base.rand(d::CompositeDistribution) = map(rv -> rand(rv), rvs)
```
That's a step in the right direction for simplicity of implementation.  But if rvs included Multivariate Distributions, this will return a Tuple of Vectors (and maybe some scalars mixed in).  
The main motivation is that some algorithms want a standard distribution and expect a flat vector of outputs.  
Perhaps we could write just two functions to flatten the outputs and unflatten the inputs, so that the code could stay fairly clean.

> If you were use tuples, I might call it a TupleDistribution since it's so general at that point that I don't know what else to call it. 

But TupleDistribution describes the implementation, rather than the purpose.  

> If I saw something like TupleDistributions.jl I would be clear what it does, although it would take me a few minutes to realize why it would be useful.

Lots of statistical algoritms take a distribution as an input.  E.g., https://github.com/eford/ABC.jl

What about MixedMultivariateDistributions?

eford:
I found https://github.com/rdeits/Flatten.jl/blob/master/src/Flatten.jl
But it doesn't support variable length arrays.

eford:
A slight adaptation of Stephan's suggestion at https://groups.google.com/forum/#!topic/julia-users/1QrIhbRA8hs seems promising:
flatten(a::Tuple) = any(map(x->isa(x,Array),a))? flatten(vcat(map(flatten,a)...)): a
flatten{T}(a::Array{T,1}) = any(map(x->isa(x,Array),a))? flatten(vcat(map(flatten,a)...)): a
flatten{T}(a::Array{T}) = reshape(a,prod(size(a)))
flatten(a)=a
using Distributions
rvs = (MvNormal([0.0,0.0], [1.0,2.0]), Gamma(1, 1))
flatten(map(rv -> mean(rv), rvs))

Do you see any problems with that solution? 
Some functions like cov would need special treatment, but hopefully the above would cover most needs.

lindahua:
It seems to me it is implicitly assumed that different components of a tuple are *independent*. In such cases, the distribution should be called a [product distribution](https://en.wikipedia.org/wiki/Product_distribution). The name *composite* here may obscure the concept.

johnmyleswhite:
> In such cases, the distribution should be called a product distribution.

Isn't a product distribution a distribution induced by multiplying two separate random variables? For example, `z` has a product distribution in the following:

```
x = rand()
y = rand()
z = x * y
```

lindahua:
Sorry. I was considering [product measure](https://en.wikipedia.org/wiki/Product_measure).

lindahua:
In some statistical learning literatures, we refer to the distribution of ``(X, Y)`` with independent priors as *product of distributions*, while those of ``X * Y`` as the *distribution of product*.

lindahua:
The issue of naming aside, +1 for @johnmyleswhite's tuple-based implementation.

johnmyleswhite:
Yeah, it's very confusing that production distribution and product measure are so dissimilar.

Also, welcome back!

matbesancon:
Is this PR covered by #722 ?
If that's the case, should this one be closed? cc @eford

matbesancon:
From what I understand, the PR has been replaced by #722 and moved to another package

PR #532, opened 2016-09-12
Fix fit(Laplace) for changes to StasBase.mad

andreasnoack:
We really need some automated way of testing downstream packages when we tag a new release.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.4
-PDMats 0.4.2 0.5
-StatsFuns 0.3.0
+PDMats 0.5
+StatsFuns 0.3.1
 Calculus
 StatsBase 0.8.3
-Compat 0.8.4
+Compat 0.9.2
--- a/src/multivariate/mvlognormal.jl
+++ b/src/multivariate/mvlognormal.jl
@@ -136,18 +136,21 @@ location(d::MvLogNormal) = mean(d.normal)
 scale(d::MvLogNormal) = cov(d.normal)
 
 #See https://en.wikipedia.org/wiki/Log-normal_distribution
-mean(d::MvLogNormal) = exp(mean(d.normal) + var(d.normal)/2)
-median(d::MvLogNormal) = exp(mean(d.normal))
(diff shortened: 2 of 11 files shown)


ararslan:
We could add something like
```bash
julia -e 'for dep in Pkg.dependents("Distributions") Pkg.test(dep) end'
```
to the .travis.yml script. That will make Travis times *skyrocket* though.

johnmyleswhite:
The R community addresses this by e-mailing all dependent package authors before tagging a new release. But their releases are far more spread out.

KristofferC:
@ararslan  Could just do it locally before tagging the release?

andreasnoack:
I'll try to do that now. I've access to some beefy machines and with `pmap` it might be possible to finish in reasonable time although it's 89 packages.

KristofferC:
Maybe run it on the last version as well to get rid of spurious failures?

ararslan:
@KristofferC Sure, I was just thinking more in terms of identifying breaking changes in PRs before they're merged.

andreasnoack:
Just tried this for this branch and there is some work to do if we want to look into all problems in dependent packages
```jl
julia> r = pmap(foo, Pkg.dependents("Distributions"), on_error=identity)

julia> mapreduce(t -> isa(t, Exception), +, r)
86
```

tkelman:
So Distributions master is massively broken for almost every package that uses it? This should be a blocker on tagging any new version of the package then. 0.10.2 was made off of a branch for this reason, and that branch should probably be used to make a not-so-breaking 0.10.3 where you backport only the deprecation and test-related fixes, without the enormous rewrite.

andreasnoack:
I'm pretty sure my result in https://github.com/JuliaStats/Distributions.jl/pull/532#issuecomment-246552504 is nonsense. Testing packages in parallel is probably not possible.

tkelman:
Yeah, in some discussions with @ararslan we dug up the PkgEval results from the day that 0.10.1 was tagged, and there's some breakage but it's a handful of package, not all of them.

PR #533, opened 2016-09-12
Various cleanup

andreasnoack:
Things that we have forgotten to remove at various refactorizations. It could be argued that we should keep the `posterior` definitions but they don't seem to be used much right now and I think they would fit better in something like `BayesianBase`.

Notice: This PR includes the bug fix for `Laplace`. I'll refactor once that PR has been merged.

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -190,12 +190,6 @@ export
     logpdf!,            # evaluate log pdf to provided storage
     logpmf,             # log probability mass
     logpmf!,            # evaluate log pmf to provided storage
-    posterior,          # get posterior distribution given prior and observed data
-    posterior_canon,    # get the canonical form of the posterior distribution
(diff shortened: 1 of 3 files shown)


ararslan:
The posterior stuff is being used by ConjugatePriors (which is currently in utter disrepair). IMO we should wait to remove it from here until it has a separate package to live in, akin to how we moved `read_rda` from DataFrames to RData.

andreasnoack:
Couldn't they just live in `ConjugatePriors` instead provided that we get that package into a state where it works?

ararslan:
That's a great idea! We should try to get ConjugatePriors at least sort of working before merging this then, IMHO, so there's no gap in function availability.

Best not to taint this beautiful frequentist package with that Bayesian nonsense anyway. :stuck_out_tongue_winking_eye:

PR #545, opened 2016-10-11
allowing Multinomial to work with 0 trials

### LINKED ISSUE #543: Multinomial distribution throws error on n=0
Why does the multinomial distribution require n > 0? It seems like it should be fine to take n = 0, in which case it would always just return the all-zeroes vector?

ajdunlap:
Attempt to fix #543. I didn't spend the time to figure out the internals of how Multinomial worked but it seemed like when the bounds check was relaxed the resulting behavior was exactly the natural degeneration.

### DIFF
--- a/test/multinomial.jl
+++ b/test/multinomial.jl
@@ -102,13 +102,13 @@ r = fit_mle(Multinomial, x, fill(2.0, size(x,2)))
 # behavior for n = 0
 d0 = Multinomial(0, p)
 @test rand(d0) == [0, 0, 0]
-@test pdf(d0,[0, 0, 0]) == 1
-@test pdf(d0,[0, 1, 0]) == 0
+@test pdf(d0, [0, 0, 0]) == 1
+@test pdf(d0, [0, 1, 0]) == 0
 @test mean(d0) == [0, 0, 0]
 @test var(d0) == [0, 0, 0]
 @test cov(d0) == zeros(3, 3)
 @test entropy(d0) == 0
-@test insupport(d0,[0, 0, 0]) == true
-@test insupport(d0,[0, 0, 4]) == false
+@test insupport(d0, [0, 0, 0]) == true
+@test insupport(d0, [0, 0, 4]) == false
 @test length(d0) == 3
 @test size(d0) == (3,)
--- a/test/multinomial.jl
+++ b/test/multinomial.jl
@@ -98,3 +98,6 @@ r = fit_mle(Multinomial, x, fill(2.0, size(x,2)))
 @test r.n == nt
 @test length(r) == length(p)
 @test_approx_eq_eps probs(r) p 0.02
+
+# behavior for n = 0
+@test rand(Multinomial(0,p)) == [0,0,0]
--- a/test/multinomial.jl
+++ b/test/multinomial.jl
@@ -100,4 +100,15 @@ r = fit_mle(Multinomial, x, fill(2.0, size(x,2)))
 @test_approx_eq_eps probs(r) p 0.02
 
 # behavior for n = 0
-@test rand(Multinomial(0,p)) == [0,0,0]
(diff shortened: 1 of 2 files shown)


johnmyleswhite:
Thank you so much for doing this!

Quick question: do methods like `pdf` work for this degenerate distribution instance?

ajdunlap:
pdf does at least:

```
julia> pdf(Multinomial(0,[0.5,0.25,0.25]),[0,0,0])
1.0

julia> pdf(Multinomial(0,[0.5,0.25,0.25]),[0,0,2])
0.0
```

johnmyleswhite:
Would you up for testing the rest of the core functions to make sure they all work?

ajdunlap:
Sure, not sure where I can find a list of all of the "core" functions but I can test the ones on the documentation page http://distributionsjl.readthedocs.io/en/latest/multivariate.html:

```
julia> using Distributions

julia> d = Multinomial(0,[0.25,0.25,0.5])
Distributions.Multinomial{Float64}(n=0, p=[0.25,0.25,0.5])

julia> length(d)
3

julia> 

julia> size(d)
(3,)

julia> mean(d)
3-element Array{Float64,1}:
 0.0
 0.0
 0.0

julia> var(d)
3-element Array{Float64,1}:
 0.0
 0.0
 0.0

julia> cov(d)
3×3 Array{Float64,2}:
 0.0  0.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0

julia> entropy(d)
0.0

julia> insupport(d,[0,0,0])
true

julia> insupport(d,[0,0,2])
false

```

simonbyrne:
Yes, those along with `pdf` are probably the main ones. Could you please add them as tests?

ajdunlap:
Sure, just pushed the additional tests.

test/multinomial.jl
```diff
@@ -98,3 +98,17 @@ r = fit_mle(Multinomial, x, fill(2.0, size(x,2)))
 @test r.n == nt
 @test length(r) == length(p)
 @test_approx_eq_eps probs(r) p 0.02
+
+# behavior for n = 0
+d0 = Multinomial(0,p)
```
johnmyleswhite: Can you insert spaces after a comma to match the style of the rest of this file?

test/multinomial.jl
```diff
@@ -98,3 +98,17 @@ r = fit_mle(Multinomial, x, fill(2.0, size(x,2)))
...
+@test rand(d0) == [0,0,0]
```
johnmyleswhite: Same here.

ajdunlap:
OK

test/multinomial.jl
```diff
@@ -98,3 +98,17 @@ r = fit_mle(Multinomial, x, fill(2.0, size(x,2)))
...
+d0 = Multinomial(0, p)
+@test rand(d0) == [0, 0, 0]
+@test pdf(d0,[0, 0, 0]) == 1
+@test pdf(d0,[0, 1, 0]) == 0
+@test mean(d0) == [0, 0, 0]
+@test var(d0) == [0, 0, 0]
+@test cov(d0) == zeros(3, 3)
+@test entropy(d0) == 0
+@test insupport(d0,[0, 0, 0]) == true
+@test insupport(d0,[0, 0, 4]) == false
```
ararslan: Space after the comma here and on lines 111, 106, and 105 as well, please. :slightly_smiling_face:

ajdunlap: ok

johnmyleswhite:
Does Travis really take this long now?

ararslan:
@johnmyleswhite Apparently Travis has had delays in boot times that led to a lengthy backlog. They've resumed builds and I guess we're only getting one worker at a time for now. See https://www.traviscistatus.com/ for reference.

Issue #547, opened 2016-10-19
ERROR: InexactError() for BetaBinomial n > 66


cjprybol:
```
               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.5.1-pre+2 (2016-09-20 03:34 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit f0d40ec* (29 days old release-0.5)
|__/                   |  x86_64-apple-darwin16.1.0

julia> using Distributions

julia> pdf(BetaBinomial(66, 2, 2));

julia> pdf(BetaBinomial(67, 2, 2));
ERROR: InexactError()
 in binomial(::Int64, ::Int64) at ./intfuncs.jl:410
 in copy!(::Array{Float64,1}, ::Base.Generator{UnitRange{Int64},Distributions.##15#16{Int64}}) at ./abstractarray.jl:477
 in pdf(::Distributions.BetaBinomial{Float64}) at /Users/Cameron/.julia/v0.5/Distributions/src/univariate/discrete/betabinomial.jl:94
```

Is there a limit to how large n can be when generating a beta-binomial distribution? I've only come across `ERROR: InexactError()` before when casting Floats to Ints.

following the error to line 410 in intfuncs.jl shows `x >>= 4` as the offending line (if I'm interpreting the traceback correctly)
```
function hex(x::Unsigned, pad::Int, neg::Bool)
    i = neg + max(pad,(sizeof(x)<<1)-(leading_zeros(x)>>2))
    a = Array{UInt8}(i)
    while i > neg
        d = x & 0xf
        a[i] = '0'+d+39*(d>9)
        x >>= 4
        i -= 1
    end
    if neg; a[1]='-'; end
    String(a)
end
```

behavior is the same for current release and master

andreasnoack:
Using `binomial` in https://github.com/JuliaStats/Distributions.jl/blob/master/src/univariate/discrete/betabinomial.jl#L85 is not going to work for large `n`. Just using the beta function might be sufficient in practice. Using `lbeta` would solve overflow problem generally but then we might introduce some precision issues. It would be great if you could try to prepare a PR with `beta` instead of `binomial`.

cjprybol:
[...]

PR #549, opened 2016-10-19
WIP: replace binomial(n,k) with beta equivalent for pdf(BetaBinomial)

cjprybol:
extends pdf() to work with n <= 170, whereas the binomial function only allowed pdf calculations up to n <= 66

https://github.com/JuliaStats/Distributions.jl/issues/547


ararslan:
It would likely be more efficient to use the beta function directly rather than computing the ratio of gammas.

cjprybol:
Thanks for the feedback @ararslan!

this method accommodates `n <= 1029`, which seems like a big improvement over the `n <= 66` of the current implementation, and the `n <= 170` of the gamma based equation

```
julia> n = 1029; k = Int(round(n/2)); 1 / ((n+1) * beta(k+1, n-k+1))
1.4298206864998805e308

julia> n = 1030; k = Int(round(n/2)); 1 / ((n+1) * beta(k+1, n-k+1))
Inf
```

Would it be appropriate to throw an error to the user for `n >= 1030` to let them know that Float64s will no longer handle the values?

src/univariate/discrete/betabinomial.jl
```diff
@@ -91,7 +91,7 @@ end
 function pdf(d::BetaBinomial)
     n, α, β = d.n, d.α, d.β
     k = 0:n
-    binoms = Float64[binomial(n, i) for i in k]
+    binoms = [1 / ((n+1) * beta(k+1, n-k+1)) for i in k]
```
ararslan: Should be `i` instead of `k`; that's what's causing the test failures

ararslan:
I'm not really sure what the precedent is in Base Julia in terms of explicitly throwing an error in cases like that, but I would think we should follow the lead there. @andreasnoack might know?

ararslan:
Also if you could add a test or two for large samples to ensure everything is working as expected, that would be great!

andreasnoack:
@cjprybol This id a great improvement. How large an `n` would be useful to handle? Is it worth bothering with larger `n`? It could easily be handled with an `lbeta` branch for large `n` but if people will never us such large `n`s it might not be necessary.

cjprybol:
Glad to help! I'm not convinced that handling larger values for `n` would be useful. If using `lbeta` will reduce precision, then I don't see any advantage that using `lbeta` with the `BetaBinomial` distribution would have over just using a continuous distribution to model the scenario.

If there's agreement on stopping here, I'm happy to close this and the corresponding issue I opened

ararslan:
I think it should be fine to merge as-is. We can always revisit adding support for larger *n* should we find that someone needs it.

ararslan:
I'm wondering if it might be worth a branch for smaller n that still uses `binomial`. Any idea whether that would be a good idea for performance or accuracy, @andreasnoack?

test/runtests.jl
```diff
@@ -28,7 +28,8 @@ tests = [
     "gradlogpdf",
     "truncate",
     "truncatednormal",
-    "generalizedextremevalue"]
+    "generalizedextremevalue",
+    "largenbetabinomialpdf"]
```
ararslan: There are existing tests, they're just part of a larger, less obvious framework. See https://github.com/JuliaStats/Distributions.jl/blob/master/test/discrete_hand_coded_test.json.

andreasnoack:
The `binomial` seems to be quite a bit faster and is, of course, also slightly more precise but it will complicate the function a bit.

ararslan:
Is complicating it necessarily that bad? Perhaps we could try it with the branch and run some benchmarks and whatnot, unless you don't think it's worth it.

cjprybol:
`beta` passes `@test isapprox(binomial(n,k), (1 / ((n+1) * beta(k+1, n-k+1))))` for everything I've thrown at it, but fails when testing with `==`, so the currently proposed change seems accurate to whatever degree of error is tolerated by `isapprox`. The first one I checked was equivalent down to 16 decimal places. I don't think I have a good intuition about whether this error is acceptable or not to the average user.

benchmarks: I tried the current version of the `pdf` function in master, the currently proposed change using `beta`, and a version that runs a `try-catch` block if `n` is too large.

Functions
```
function pdf_master(d::BetaBinomial, k::Int)
    n, α, β = d.n, d.α, d.β
    choose = Float64(binomial(n, k))
    numerator = beta(k + α, n - k + β)
    denominator = beta(α, β)
    return (numerator / denominator) * choose
end

function pdf_current(d::BetaBinomial, k::Int)
    n, α, β = d.n, d.α, d.β
    chooseinv = (n+1) * beta(k+1, n-k+1)
    numerator = beta(k + α, n - k + β)
    denominator = beta(α, β)
    return (numerator / denominator) / chooseinv
end

function pdf_both(d::BetaBinomial, k::Int)
    n, α, β = d.n, d.α, d.β
    try
        choose = Float64(binomial(n, k))
        numerator = beta(k + α, n - k + β)
        denominator = beta(α, β)
        return choose * (numerator / denominator)
    catch InexactError
        chooseinv = (n+1) * beta(k+1, n-k+1)
        numerator = beta(k + α, n - k + β)
        denominator = beta(α, β)
        return (numerator / denominator) / chooseinv
    end
end
```

benchmarks for small `n`s that should never trigger the `catch` block
```
@benchmark (n = rand(1:10); k = rand(1:n); pdf_master(BetaBinomial(n,2,2), k))
BenchmarkTools.Trial:
  samples:          10000
  evals/sample:     199
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  576.00 bytes
  allocs estimate:  6
  minimum time:     426.00 ns (0.00% GC)
  median time:      454.00 ns (0.00% GC)
  mean time:        532.99 ns (8.28% GC)
  maximum time:     13.47 μs (94.76% GC)

@benchmark (n = rand(1:10); k = rand(1:n); pdf_current(BetaBinomial(n,2,2), k))
BenchmarkTools.Trial:
  samples:          10000
  evals/sample:     181
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  864.00 bytes
  allocs estimate:  9
  minimum time:     587.00 ns (0.00% GC)
  median time:      631.00 ns (0.00% GC)
  mean time:        742.08 ns (8.10% GC)
  maximum time:     15.21 μs (93.67% GC)

@benchmark (n = rand(1:10); k = rand(1:n); pdf_both(BetaBinomial(n,2,2), k))
BenchmarkTools.Trial:
  samples:          10000
  evals/sample:     198
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  576.00 bytes
  allocs estimate:  6
  minimum time:     460.00 ns (0.00% GC)
  median time:      495.00 ns (0.00% GC)
  mean time:        581.35 ns (7.58% GC)
  maximum time:     12.59 μs (89.69% GC)
```
`beta` is definitely slower, and we can gain something back by using the `try-catch`

and benchmarks for larger `n`s that should trigger the `catch` block frequently
```
@benchmark (n = rand(1:1000); k = rand(1:n); pdf_current(BetaBinomial(n,2,2), k))
BenchmarkTools.Trial:
  samples:          10000
  evals/sample:     184
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  864.00 bytes
  allocs estimate:  9
  minimum time:     571.00 ns (0.00% GC)
  median time:      607.00 ns (0.00% GC)
  mean time:        707.83 ns (8.76% GC)
  maximum time:     13.73 μs (89.91% GC)

@benchmark (n = rand(1:1000); k = rand(1:n); pdf_both(BetaBinomial(n,2,2), k))
BenchmarkTools.Trial:
  samples:          10000
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  576.00 bytes
  allocs estimate:  6
  minimum time:     469.00 ns (0.00% GC)
  median time:      73.54 μs (0.00% GC)
  mean time:        71.22 μs (0.00% GC)
  maximum time:     256.48 μs (0.00% GC)
```
The `try-catch` helps in situations where `n` is small and using the `binomial` function would be preferable, but does substantially worse when `binomial` throws the `InexactError()` and triggers the `catch` statement to run. So it's at least clear that this particular `try-catch` is a bad idea.

Trying a naive `if` statement that just checks if `n` is getting large starts putting us back in the same runtime territory as the current `beta` calculation for calculating n choose k
```
function pdf_if(d::BetaBinomial, k::Int)
  n, α, β = d.n, d.α, d.β
  if n < 40
      choose = Float64(binomial(n, k))
      numerator = beta(k + α, n - k + β)
      denominator = beta(α, β)
      return choose * (numerator / denominator)
  else
      chooseinv = (n+1) * beta(k+1, n-k+1)
      numerator = beta(k + α, n - k + β)
      denominator = beta(α, β)
      return (numerator / denominator) / chooseinv
  end
end

@benchmark (n = rand(1:1000); k = rand(1:n); pdf_if(BetaBinomial(n,2,2), k))
BenchmarkTools.Trial:
  samples:          10000
  evals/sample:     184
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  834.00 bytes
  allocs estimate:  8
  minimum time:     565.00 ns (0.00% GC)
  median time:      598.00 ns (0.00% GC)
  mean time:        698.18 ns (8.51% GC)
  maximum time:     14.47 μs (94.34% GC)
```

It looks like adding an `if` statement wouldn't make this any slower than if we blindly used the `beta` equation for `n` of all sizes. If we went with this approach in order to retain accuracy for small `n` (won't get the speed back), I guess the question is where do we set the threshold for switching over to the `beta` version of the calculation? Have there been discussions that would set a precedent for how to balance the accuracy/simplicity tradeoff?

src/univariate/discrete/betabinomial.jl
```diff
@@ -82,18 +82,22 @@ end
 
 function pdf(d::BetaBinomial, k::Int)
     n, α, β = d.n, d.α, d.β
-    choose = 1 / ((n+1) * beta(k+1, n-k+1))
+    chooseinv = (n+1) * beta(k+1, n-k+1)
     numerator = beta(k + α, n - k + β)
     denominator = beta(α, β)
-    return choose * (numerator / denominator)
+    return (numerator / denominator) / chooseinv
```
ararslan: `numerator / (denominator * chooseinv)` saves a division operation

src/univariate/discrete/betabinomial.jl
```diff
@@ -82,18 +82,22 @@ end
...
-    k = 0:n
-    binoms = [1 / ((n+1) * beta(i+1, n-i+1)) for i in k]
-    fixed_beta = beta(α, β)
-    return binoms .* @compat(beta.(k + α, n - k + β)) / fixed_beta
+    denominator = beta(α, β)
+    values = Array{Float64, 1}()
+    for k = 0:n
+        chooseinv = (n+1) * beta(k+1, n-k+1)
+        numerator = beta(k + α, n - k + β)
+        push!(values, (numerator / denominator) / chooseinv)
```
ararslan: Rather than growing the array using `push!`, you know how large you want it to end up, so it's better to allocate the array with the desired length ahead of time. Also note that having `Array{Float64,1}` will implicitly `convert` the elements to `Float64`. In the interest of generality, I think it would be better to write this more like how it was written previously.

cjprybol: `push!` doesn't appear to have convert functionality, could you explain the conversion concern?
```
julia> a = Array(Float64, 0)
0-element Array{Float64,1}

julia> push!(a, "A")
ERROR: MethodError: Cannot `convert` an object of type String to an object of type Float64
This may have arisen from a call to the constructor Float64(...),
since type constructors fall back to convert methods.
 in push!(::Array{Float64,1}, ::String) at ./array.jl:479
```
I'll try to convince you otherwise about writing this like the current master syntax. I find the current syntax hard to follow. The first aspect is the nontrivial (but admittedly awesome) auto-expansion for `beta.(k + α, n - k + β)`. This broadcast of `beta` auto-expands the `k=0:n`, but I had no idea what the behavior was until I ran it to check. But if that works, why isn't `binoms = Float64[binomial(n, i) for i in k]` also written as an auto-expansion `Float64.(binomial.(n,k))`, and is the `Float64` conversion really necessary? Benchmarking the time it takes for `Float64 * Int64` and `Float64 * Float64` doesn't show any noticeable differences. I also didn't like macro in the middle of the equation. At that point, I felt I was asking too many questions and thought a simpler, more explicit method that used the same variable names as the `pdf(d::BetaBinomial, k::Int)` definition would be preferable.

Benchmarks also slightly prefer the explicit version (updated to pre-allocate the array)
```julia
using BenchmarkTools
using Distributions
using Compat
using Base.Test
import Distributions.pdf

function pdf_master(d::BetaBinomial)
	n, α, β = d.n, d.α, d.β
	k = 0:n
	binoms = Float64[binomial(n, i) for i in k]
	fixed_beta = beta(α, β)
	return binoms .* @compat(beta.(k + α, n - k + β)) / fixed_beta
end

function pdf_elementwise(d::BetaBinomial)
    n, α, β = d.n, d.α, d.β
	k = 0:n
	binoms = [(n+1) * beta(i+1, n-i+1) for i in k]
	fixed_beta = beta(α, β)
	return @compat(beta.(k + α, n - k + β)) ./ (fixed_beta .* binoms)
end

function pdf_preinitialized(d::BetaBinomial)
    n, α, β = d.n, d.α, d.β
    denominator = beta(α, β)
    values = Array(Float64, n+1)
    for (i,k) in enumerate(0:n)
        chooseinv = (n+1) * beta(k+1, n-k+1)
        numerator = beta(k + α, n - k + β)
        values[i] = numerator / (denominator * chooseinv)
    end
    return values
end
```
```julia
julia> # check accuracy
       d = BetaBinomial(3, 2, 2)
Distributions.BetaBinomial{Float64}(n=3, α=2.0, β=2.0)

julia> answers = [0.19999999999999998, 0.29999999999999977, 0.29999999999999977, 0.19999999999999998]
4-element Array{Float64,1}:
 0.2
 0.3
 0.3
 0.2

julia> @test isapprox(pdf_master(d), answers)
Test Passed
  Expression: isapprox(pdf_master(d),answers)

julia> @test isapprox(pdf_elementwise(d), answers)
Test Passed
  Expression: isapprox(pdf_elementwise(d),answers)

julia> @test isapprox(pdf_preinitialized(d), answers)
Test Passed
  Expression: isapprox(pdf_preinitialized(d),answers)

julia> # check efficiency
       d = BetaBinomial(10,2,2)
Distributions.BetaBinomial{Float64}(n=10, α=2.0, β=2.0)

julia> @benchmark pdf_master(d)
BenchmarkTools.Trial:
  samples:          10000
  evals/sample:     9
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  4.14 kb
  allocs estimate:  44
  minimum time:     3.00 μs (0.00% GC)
  median time:      3.06 μs (0.00% GC)
  mean time:        3.46 μs (7.92% GC)
  maximum time:     233.08 μs (96.08% GC)

julia> @benchmark pdf_elementwise(d)
BenchmarkTools.Trial:
  samples:          10000
  evals/sample:     7
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  7.23 kb
  allocs estimate:  77
  minimum time:     4.32 μs (0.00% GC)
  median time:      4.47 μs (0.00% GC)
  mean time:        5.17 μs (8.21% GC)
  maximum time:     312.74 μs (94.84% GC)

julia> @benchmark pdf_preinitialized(d)
BenchmarkTools.Trial:
  samples:          10000
  evals/sample:     7
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  6.64 kb
  allocs estimate:  70
  minimum time:     4.07 μs (0.00% GC)
  median time:      4.20 μs (0.00% GC)
  mean time:        4.70 μs (6.78% GC)
  maximum time:     308.45 μs (96.60% GC)

julia> # check larger scale efficiency
       d = BetaBinomial(1029,2,2)
Distributions.BetaBinomial{Float64}(n=1029, α=2.0, β=2.0)

julia> @benchmark pdf_elementwise(d)
BenchmarkTools.Trial:
  samples:          9779
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  612.48 kb
  allocs estimate:  6191
  minimum time:     440.05 μs (0.00% GC)
  median time:      456.44 μs (0.00% GC)
  mean time:        510.72 μs (7.88% GC)
  maximum time:     2.74 ms (71.10% GC)

julia> @benchmark pdf_preinitialized(d)
BenchmarkTools.Trial:
  samples:          10000
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  587.84 kb
  allocs estimate:  6184
  minimum time:     394.74 μs (0.00% GC)
  median time:      423.55 μs (0.00% GC)
  mean time:        471.70 μs (8.04% GC)
  maximum time:     2.84 ms (68.21% GC)
```

src/univariate/discrete/betabinomial.jl
cjprybol: Thank you! Updating now

src/univariate/discrete/betabinomial.jl
```diff
@@ -82,18 +82,22 @@ end
...
 end
 
 function pdf(d::BetaBinomial)
     n, α, β = d.n, d.α, d.β
-    k = 0:n
-    binoms = Float64[binomial(n, i) for i in k]
-    fixed_beta = beta(α, β)
-    return binoms .* @compat(beta.(k + α, n - k + β)) / fixed_beta
+    denominator = beta(α, β)
+    values = Array{Float64, 1}()
```
andreasnoack: The element type of this array should be determined from `α` and `β`, e.g. `eltype(denominator)`. It is possible to construct an instance with `Float32`s.

src/univariate/discrete/betabinomial.jl
ararslan: > `push!` doesn't appear to have convert functionality, could you explain the conversion concern?

It's not that `push!` itself converts, it's that the array is defined to have element type `Float64`, so Julia will _attempt_ to convert any value that's added to the array, be it via `push!`, `setindex!`, or what have you, to the array's defined element type. That's evident in the example you gave:
```julia
julia> push!(a, "A")
ERROR: MethodError: Cannot `convert` an object of type String to an object of type Float64
This may have arisen from a call to the constructor Float64(...),
since type constructors fall back to convert methods.
 in push!(::Array{Float64,1}, ::String) at ./array.jl:479
```
It's trying to convert the value you gave it, but in this case the conversion fails. If you were to push a `Float32` or `Int64`, for example, or any type for which a `convert` method exists for `Float64`, the conversion wouldn't fail. However, then you're locked into `Float64`; the natural type that would have resulted from `promote_op` is no longer preserved.

Do you see what I mean?

src/univariate/discrete/betabinomial.jl
```diff
@@ -85,17 +85,17 @@ function pdf(d::BetaBinomial, k::Int)
...
-    return (numerator / denominator) / chooseinv
+    return numerator / (denominator * chooseinv)
 end
 
 function pdf(d::BetaBinomial)
     n, α, β = d.n, d.α, d.β
     denominator = beta(α, β)
-    values = Array{Float64, 1}()
-    for k = 0:n
+    values = Array(Float64, n+1)
```
ararslan: This syntax is "deprecated" (according to the documentation it is, but Julia itself doesn't emit a deprecation warning for it). The preferred syntax is now `Array{T,N}(dims)`, where `T` is the type, `N` is the number of dimensions, and `dims` is the dimension specification. So you could write this as `Array{Float64,1}(n+1)`.

andreasnoack: The `N` part is not necessary though. `Array{T}(n+1)` is fine since `n+1` implicitly determines `N`.

src/univariate/discrete/betabinomial.jl
cjprybol: Yes, sorry, that's now blatantly obvious. I don't think that error message registered at all, I just checked to see if using the wrong type would fail. Thanks for the ELI5

src/univariate/discrete/betabinomial.jl
ararslan: Is `eltype(denominator)` enough? Could `numerator / (denominator * chooseinv)` have a type other than `eltype(denominator)`?

src/univariate/discrete/betabinomial.jl
cjprybol: To confirm I understand, this should be written as `values = Array{eltype(denominator)}(n+1)`?

src/univariate/discrete/betabinomial.jl
```diff
@@ -82,18 +82,22 @@ end
...
 function pdf(d::BetaBinomial)
     n, α, β = d.n, d.α, d.β
-    k = 0:n
-    binoms = Float64[binomial(n, i) for i in k]
-    fixed_beta = beta(α, β)
-    return binoms .* @compat(beta.(k + α, n - k + β)) / fixed_beta
+    denominator = beta(α, β)
+    values = Array(Float64, n+1)
+    for (i,k) in enumerate(0:n)
+        chooseinv = (n+1) * beta(k+1, n-k+1)
```
ararslan: Sorry, just a bit of style pedantry: Could you space out the additions to match how they are in the line below?

src/univariate/discrete/betabinomial.jl
```diff
@@ -82,18 +82,22 @@ end
...
 
 function pdf(d::BetaBinomial)
     n, α, β = d.n, d.α, d.β
-    k = 0:n
-    binoms = Float64[binomial(n, i) for i in k]
-    fixed_beta = beta(α, β)
-    return binoms .* @compat(beta.(k + α, n - k + β)) / fixed_beta
+    denominator = beta(α, β)
+    values = Array(Float64, n+1)
+    for (i,k) in enumerate(0:n)
```
ararslan: Wouldn't hurt to annotate this with `@inbounds`. Could go here or `@inbounds values[i] = ...`

src/univariate/discrete/betabinomial.jl
ararslan: Yep, that's right!

src/univariate/discrete/betabinomial.jl
ararslan: Okay great.

ararslan:
@andreasnoack Have all of your concerns here been sufficiently addressed? Okay to (squash and) merge?

ararslan:
Thanks for the contribution, @cjprybol!

cjprybol:
Thanks for all of the guidance! I hope it wasn't too painful being on the other end. The process was very instructive and helpful for me

ararslan:
> I hope it wasn't too painful being on the other end.

Not at all! You did a great job here.

> The process was very instructive and helpful for me

That's great, I'm very glad to hear that. Hopefully that means you'll continue to contribute to the JuliaStats ecosystem. 😉

PR #553, opened 2016-11-21
WIP: DirichletMultinomial

joshday:
Here is my first pass "everything seems to work" version.  I still need to add MLE stuff and a few more tests, which I should be able to get to tomorrow.

### DIFF
--- a/src/multivariate/dirichletmultinomial.jl
+++ b/src/multivariate/dirichletmultinomial.jl
@@ -28,17 +28,17 @@ params(d::DirichletMultinomial) = (d.n, d.α)
 # Statistics
 mean(d::DirichletMultinomial) = d.α .* (d.n / d.α0)
 function var{T <: Real}(d::DirichletMultinomial{T})
-    v = fill(d.n * (d.n + d.α0) / (one(T) + d.α0), length(d))
+    v = fill(d.n * (d.n + d.α0) / (1 + d.α0), length(d))
     p = d.α / d.α0
     for i in eachindex(v)
-        @inbounds v[i] *= p[i] * (one(T) - p[i])
+        @inbounds v[i] *= p[i] * (1 - p[i])
     end
     v
 end
 function cov{T <: Real}(d::DirichletMultinomial{T})
     v = var(d)
     c = d.α * d.α'
-    multiply!(c, -d.n * (d.n + d.α0) / (d.α0^2 * (one(T) + d.α0)))
+    multiply!(c, -d.n * (d.n + d.α0) / (d.α0^2 * (1 + d.α0)))
     for i in 1:length(d)
         @inbounds c[i, i] = v[i]
     end
--- a/src/multivariate/dirichletmultinomial.jl
+++ b/src/multivariate/dirichletmultinomial.jl
@@ -1,0 +1,39 @@
+immutable DirichletMultinomial{T <: Real} <: ContinuousMultivariateDistribution
+    n::Int
+    α::Vector{T}
+    α0::T
+
+    function DirichletMultinomial(n::Integer, α::Vector{T})
+        α0 = sumabs(α)
+        sum(α) == α0 || throw(ArgumentError("DirichletMultinomial: alpha must be a positive vector."))
+        n > 0 || throw(ArgumentError("DirichletMultinomial: n must be a positive integer."))
+        new(round(Int, n), α, α0)
+    end
+end
+DirichletMultinomial{T <: Real}(n::Integer, α::Vector{T}) = DirichletMultinomial{T}(n, α)
+DirichletMultinomial{T <: Integer}(n::Integer, α::Vector{T}) = DirichletMultinomial(n, float(α))
+DirichletMultinomial(n::Integer, k::Integer) = DirichletMultinomial(n, ones(k))
+
+Base.show(io::IO, d::DirichletMultinomial) = show(io, d, (:n, :α,))
+
+# Parameters
+ncategories(d::DirichletMultinomial) = length(d.α)
+length(d::DirichletMultinomial) = ncategories(d)
+ntrials(d::DirichletMultinomial) = d.n
+params(d::DirichletMultinomial) = (d.n, d.α)
+@inline partype{T<:Real}(d::DirichletMultinomial{T}) = T
+
+
+# Statistics
+mean(d::DirichletMultinomial) = d.α .* (d.n / d.α0)
+function var{T <: Real}(d::DirichletMultinomial{T})
+    v = fill(d.n * (d.n + d.α0) / (one(T) + d.α0), length(d))
+    p = d.α / d.α0
+    for i in eachindex(v)
+        @inbounds v[i] *= p[i] * (one(T) - p[i])
+    end
+    v
+end
+function cov(d::DirichletMultinomial)
+
+end
--- a/src/multivariate/dirichletmultinomial.jl
+++ b/src/multivariate/dirichletmultinomial.jl
@@ -1,4 +1,4 @@
-immutable DirichletMultinomial{T <: Real} <: ContinuousMultivariateDistribution
+immutable DirichletMultinomial{T <: Real} <: DiscreteMultivariateDistribution
     n::Int
     α::Vector{T}
     α0::T
@@ -7,7 +7,7 @@ immutable DirichletMultinomial{T <: Real} <: ContinuousMultivariateDistribution
         α0 = sumabs(α)
         sum(α) == α0 || throw(ArgumentError("DirichletMultinomial: alpha must be a positive vector."))
         n > 0 || throw(ArgumentError("DirichletMultinomial: n must be a positive integer."))
-        new(round(Int, n), α, α0)
+        new(Int(n), α, α0)
     end
 end
 DirichletMultinomial{T <: Real}(n::Integer, α::Vector{T}) = DirichletMultinomial{T}(n, α)
@@ -15,6 +15,7 @@ DirichletMultinomial{T <: Integer}(n::Integer, α::Vector{T}) = DirichletMultino
 DirichletMultinomial(n::Integer, k::Integer) = DirichletMultinomial(n, ones(k))
 
 Base.show(io::IO, d::DirichletMultinomial) = show(io, d, (:n, :α,))
+
 
 # Parameters
 ncategories(d::DirichletMultinomial) = length(d.α)
@@ -34,6 +35,47 @@ function var{T <: Real}(d::DirichletMultinomial{T})
     end
     v
 end
-function cov(d::DirichletMultinomial)
-
+function cov{T <: Real}(d::DirichletMultinomial{T})
+    v = var(d)
+    c = d.α * d.α'
+    multiply!(c, -(d.n / d.α0 ^ 2) * ((d.n + d.α0) / (one(T) + d.α0)))
+    for i in 1:length(d)
+        @inbounds c[i, i] = v[i]
+    end
+    c
 end
+
+
+# Evaluation
+function insupport{T<:Real}(d::DirichletMultinomial, x::AbstractVector{T})
+    k = length(d)
+    length(x) == k || return false
+    for xi in x
+        (isinteger(xi) && xi >= 0) || return false
+    end
+    return sum(x) == ntrials(d)
+end
+function _logpdf{T<:Real}(d::DirichletMultinomial, x::AbstractVector{T})
+    c = log(factorial(d.n)) + lgamma(d.α0) - lgamma(d.n + d.α0)
+    for j in eachindex(x)
+        xj, αj = x[j], d.α[j]
+        c += lgamma(xj + αj) - log(factorial(xj)) - lgamma(αj)
+    end
+    c
+end
+
+
+# Sampling
+function _rand!{T<:Real}(d::DirichletMultinomial, x::AbstractVector{T})
+    multinom_rand!(ntrials(d), rand(Dirichlet(d.α)), x)
+end
+
+
+# Fit Model
+# TODO: use https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2945396/pdf/nihms205488.pdf
+# What gets returned if there are different number of trials in observations?
+#   - Not an option for Multinomial
+# immutable DirichletMultinomialStats <: SufficientStats
+#     r::Vector{Int}
+#     s::Matrix{Int}
+# end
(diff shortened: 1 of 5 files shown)


test/dirichletmultinomial.jl
```diff
@@ -0,0 +1,46 @@
...
+@test_approx_eq(var(d), d.n * (d.n + d.α0) / (1 + d.α0) .* p .* (1.0 - p))
+
+# test Evaluation
+d = DirichletMultinomial(10, 5)
+@test !insupport(d, rand(5))
+@test insupport(d, [2, 2, 2, 2, 2])
+@test insupport(d, 2.0 * ones(5))
+@test !insupport(d, 3.0 * ones(5))
+
+for x in [2 * ones(5), [1, 2, 3, 4, 0], [3.0, 0.0, 3.0, 0.0, 4.0]]
```
ararslan: Should prefix the outer array container with `Any` to avoid the deprecation warnings this causes.

joshday: Good call.

test/dirichletmultinomial.jl
```diff
@@ -0,0 +1,46 @@
...
+d = DirichletMultinomial(10, 5)
+@test !insupport(d, rand(5))
+@test insupport(d, [2, 2, 2, 2, 2])
+@test insupport(d, 2.0 * ones(5))
+@test !insupport(d, 3.0 * ones(5))
+
+for x in Any[2 * ones(5), [1, 2, 3, 4, 0], [3.0, 0.0, 3.0, 0.0, 4.0]]
+    @test_approx_eq(
+        pdf(d, x),
+        factorial(d.n) * gamma(d.α0) / gamma(d.n + d.α0) * prod(gamma(d.α + x) ./ factorial.(x) ./ gamma(d.α))
```
ararslan: The `f.(x)` syntax means something different on Julia 0.4, which is causing the errors. I think you can just `@compat` that line.

test/dirichletmultinomial.jl
```diff
@@ -0,0 +1,46 @@
...
+    )
+    @test_approx_eq(
+        logpdf(d, x),
+        log(factorial(d.n)) + lgamma(d.α0) - lgamma(d.n + d.α0) + sum(lgamma(d.α + x) - log(factorial.(x)) - lgamma(d.α))
```
ararslan: Here too

test/dirichletmultinomial.jl
```diff
@@ -0,0 +1,46 @@
...
+@test !insupport(d, 3.0 * ones(5))
+
+for x in (2 * ones(5), [1, 2, 3, 4, 0], [3.0, 0.0, 3.0, 0.0, 4.0])
+    @test_approx_eq(
+        pdf(d, x),
+        @compat factorial(d.n) * gamma(d.α0) / gamma(d.n + d.α0) * prod(gamma(d.α + x) ./ factorial.(x) ./ gamma(d.α))
+    )
+    @test_approx_eq(
+        logpdf(d, x),
+        @compat log(factorial(d.n)) + lgamma(d.α0) - lgamma(d.n + d.α0) + sum(lgamma(d.α + x) - log(factorial.(x)) - lgamma(d.α))
```
ararslan: You'll need `log.(factorial.(x))`

src/multivariate/dirichletmultinomial.jl
```diff
@@ -0,0 +1,114 @@
+immutable DirichletMultinomial{T <: Real} <: DiscreteMultivariateDistribution
+    n::Int
+    α::Vector{T}
+    α0::T
+
+    function DirichletMultinomial(n::Integer, α::Vector{T})
+        α0 = sumabs(α)
+        sum(α) == α0 || throw(ArgumentError("DirichletMultinomial: alpha must be a positive vector."))
+        n > 0 || throw(ArgumentError("DirichletMultinomial: n must be a positive integer."))
```
ararslan: Including the type name in the error text is unnecessary since the backtrace already includes the name of the function that threw the error.

src/multivariate/dirichletmultinomial.jl
```diff
@@ -0,0 +1,114 @@
...
+function fit_mle(::Type{DirichletMultinomial}, ss::DirichletMultinomialStats;
+        tol::Float64 = 1e-8, maxiter::Int = 1000)
+    k = size(ss.s, 2)
+    α = ones(size(ss.s, 1))
+    rng = 0.0:(k - 1)
+    tolerance = Inf
+    for iter in 1:maxiter
+        α_old = copy(α)
+        αsum = sum(α)
+        denom = ss.tw * sum(1 ./ (αsum + rng))
```
ararslan: `sum(inv, αsum + rng)` may be marginally more efficient

src/multivariate/dirichletmultinomial.jl
```diff
@@ -0,0 +1,114 @@
...
+    d, m = size(x)
+    # s = [float(sum(x[i, :] .>= k)) for i in 1:size(x, 1), k in 1:n]
+    s = zeros(size(x, 1), n)
+    for k in 1:n, i in 1:m, j in 1:d
+        s[j, k] += Float64(x[j, i] >= k)
+    end
+    DirichletMultinomialStats(n, s, m)
+end
+function fit_mle(::Type{DirichletMultinomial}, ss::DirichletMultinomialStats;
+        tol::Float64 = 1e-8, maxiter::Int = 1000)
```
ararslan: Align indent with `::Type` for consistency with the rest of the repo

ararslan:
One of your test cases is triggering one of the error checks you implemented. If that's intentional, you should use `@test_throws`.

src/multivariate/dirichletmultinomial.jl
joshday: Okay, I was using Dirichlet as a template.  Want me to remove the same thing there?

joshday:
Hmm, It's not intentional.  I tested on 0.4 and 0.5 on my machine without issue.  I'll look into it.

src/multivariate/dirichletmultinomial.jl
ararslan: Oh I didn't realize it was doing that. You can leave `Dirichlet` as-is and we can revisit that another time.

src/multivariate/dirichletmultinomial.jl
```diff
@@ -85,10 +85,26 @@ function suffstats{T<:Real}(::Type{DirichletMultinomial}, x::Matrix{T})
...
     DirichletMultinomialStats(n, s, m)
 end
+function suffstats{T<:Real}(::Type{DirichletMultinomial}, x::Matrix{T}, w::Array{Float64})
+    length(w) == size(x, 2) || throw(ArgumentError("Inconsistent argument dimensions."))
+    n = sum(x[:, 1])
+    all(sum(x, 1) .== n) || error("Each sample in X should sum to the same value.")
+    d, m = size(x)
+    s = zeros(d, n)
+    tw = 0.0
+    for k in 1:n, i in 1:m, j in 1:d
```
ararslan: If you put the `@inbounds` here rather than at the assignment, you also get the benefit for the `getindex` right below this line. (Similar comments apply to other places in the diff.) That's why I had initially suggested `@inbounds for` specifically.

src/multivariate/dirichletmultinomial.jl
```diff
@@ -0,0 +1,125 @@
...
+    p = d.α / d.α0
+    for i in eachindex(v)
+        @inbounds v[i] *= p[i] * (one(T) - p[i])
+    end
+    v
+end
+function cov{T <: Real}(d::DirichletMultinomial{T})
+    v = var(d)
+    c = d.α * d.α'
+    multiply!(c, -(d.n / d.α0 ^ 2) * ((d.n + d.α0) / (one(T) + d.α0)))
```
ararslan:
```julia
multiply!(c, -d.n * (d.n + d.α0) / (d.α0^2 * (one(T) + d.α0)))
```
may be be somewhat more efficient, since it avoids a second division operation. Also for consistency we should stick to no spaces around `^`. (I had actually forgotten that was even allowed...)

src/multivariate/dirichletmultinomial.jl
```diff
@@ -0,0 +1,125 @@
...
+# Fit Model
+# Using https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2945396/pdf/nihms205488.pdf
+immutable DirichletMultinomialStats <: SufficientStats
+    n::Int
+    s::Matrix{Float64}  # s_{jk} = ∑_i x_{ij} ≥ (k - 1),  k = 1,...,(n - 1)
+    tw::Float64
+    DirichletMultinomialStats(n::Int, s::Matrix{Float64}, tw::Real) = new(n, s, Float64(tw))
+end
+function suffstats{T<:Real}(::Type{DirichletMultinomial}, x::Matrix{T})
+    n = sum(x[:, 1])
```
ararslan: You can use the dimension argument to `sum` to avoid slice.

src/multivariate/dirichletmultinomial.jl
joshday: Good catch.  All the indexing used to be on that one line, but I changed it.  I'll run through and make sure everything is covered.

ararslan:
Thanks for all of your hard work here! I'm looking forward to getting this merged. In the meantime it's always good to have more eyes on it, so I'm going to assign a couple of folks for review.

joshday:
Sounds great.  Thanks for all the quick responses!

ararslan:
Paging @andreasnoack, @nalimilan, and @jmxpearson for a second set of :eyes: on this

src/multivariate/dirichletmultinomial.jl
```diff
@@ -0,0 +1,127 @@
...
+function insupport{T<:Real}(d::DirichletMultinomial, x::AbstractVector{T})
+    k = length(d)
+    length(x) == k || return false
+    for xi in x
+        (isinteger(xi) && xi >= 0) || return false
+    end
+    return sum(x) == ntrials(d)
+end
+function _logpdf{T<:Real}(d::DirichletMultinomial, x::AbstractVector{T})
+    c = log(factorial(d.n)) + lgamma(d.α0) - lgamma(d.n + d.α0)
```
jmxpearson: Do we lose a lot of precision by not using the log Gamma function on `d.n` here? More correctly, could we allow for larger `n` if we used `lgamma(d.n + 1)`?

I suppose the pitfall is that `lgamma(d.n + 1)` would be `Float64`, so there would need to be a cast here to `T`.

I suspect that in practice, this way is fine.

src/multivariate/dirichletmultinomial.jl
```diff
@@ -0,0 +1,127 @@
...
+    DirichletMultinomialStats(n, s, m)
+end
+function suffstats{T<:Real}(::Type{DirichletMultinomial}, x::Matrix{T}, w::Array{Float64})
+    length(w) == size(x, 2) || throw(ArgumentError("Inconsistent argument dimensions."))
+    ns = sum(x, 1)
+    n = ns[1]
+    all(ns .== n) || error("Each sample in X should sum to the same value.")
+    d, m = size(x)
+    s = zeros(d, n)
+    tw = 0.0
```
jmxpearson: Is this variable used?

src/multivariate/dirichletmultinomial.jl
ararslan: We could definitely allow for larger `n` without factorial; as it stands, 20 is the largest `n` that won't overflow.

src/multivariate/dirichletmultinomial.jl
jmxpearson: Sorry, I meant line 102.

joshday: Oh, right...yes, I'll delete that.

src/multivariate/dirichletmultinomial.jl
```diff
@@ -0,0 +1,126 @@
...
+    for xi in x
+        (isinteger(xi) && xi >= 0) || return false
+    end
+    return sum(x) == ntrials(d)
+end
+function _logpdf{T<:Real, S<:Real}(d::DirichletMultinomial{S}, x::AbstractVector{T})
+    c = lgamma(S(d.n + 1)) + lgamma(d.α0) - lgamma(d.n + d.α0)
+    for j in eachindex(x)
+        @inbounds xj, αj = x[j], d.α[j]
+        c += lgamma(xj + αj) - lgamma(xj + one(xj)) - lgamma(αj)
```
ararslan: `one` shouldn't be necessary; I think we can just assume that `xj + 1` will promote as desired.

src/multivariate/dirichletmultinomial.jl
```diff
@@ -0,0 +1,126 @@
...
+function insupport{T<:Real}(d::DirichletMultinomial, x::AbstractVector{T})
+    k = length(d)
+    length(x) == k || return false
+    for xi in x
+        (isinteger(xi) && xi >= 0) || return false
+    end
+    return sum(x) == ntrials(d)
+end
+function _logpdf{T<:Real, S<:Real}(d::DirichletMultinomial{S}, x::AbstractVector{T})
+    c = lgamma(S(d.n + 1)) + lgamma(d.α0) - lgamma(d.n + d.α0)
```
ararslan: I don't think John meant convert inside `lgamma`. @jmxpearson, correct me if I'm wrong, but we should be returning `S(c)`, right?

src/multivariate/dirichletmultinomial.jl
jmxpearson: I think `typeof(xj + 1) == typeof(xj)`, so `1` is fine inside.

src/multivariate/dirichletmultinomial.jl
jmxpearson: I think this may be better. If, for some reason, you had `typeof(lgamma(d.α0)) != typeof(d.α0)`, you want to promote inside so that the return types from `lgamma` are the same in all terms.

ararslan: Ohhhh yes, I see what you mean. 👍

ararslan:
Looks like there are several other instances of `one(T)` that could similarly be replaced with 1.

ararslan:
Looks ready to me, pending green CI. What do the others think?

joshday:
The failure looks to be on Travis' side, but I don't have permissions to restart the job.

andreasnoack:
I've just restarted it.

joshday:
🎉 Thanks @ararslan and @jmxpearson for the help and comments!

ararslan:
Thank you for the contribution! 👍

PR #555, opened 2016-11-24
Update logistic.jl

dapias:
Update 2 factor needed to make it the correct logistic distribution

### DIFF
--- a/src/univariate/continuous/logistic.jl
+++ b/src/univariate/continuous/logistic.jl
@@ -4,7 +4,7 @@ doc"""
 The *Logistic distribution* with location `μ` and scale `θ` has probability density function
 
 $f(x; \mu, \theta) = \frac{1}{4 \theta} \mathrm{sech}^2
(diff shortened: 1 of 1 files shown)


dapias:
Can you correct the documentation also @ararslan ?
http://distributionsjl.readthedocs.io/en/latest/univariate-continuous.html
I did not find the way to do it
Thank you!

ararslan:
Ah right, this repo isn't using Documenter to generate documentation, so docs aren't automatically regenerated when things are merged to master... I rebuilt the docs on ReadTheDocs but the change isn't showing up. Sorry to bother you, @andreasnoack, but you seem to know this system pretty well—any chance you can help?

andreasnoack:
The documentation is in this file, https://github.com/JuliaStats/Distributions.jl/blob/master/doc/source/univariate-continuous.rst,  which wasn't updated with this PR. When that file is updated, ReadTheDocs should automatically rebuild the documentation.

PR #556, opened 2016-11-25
Update Distributions.jl

cmey:
I propose to remove:

`NormalGamma`,
`NormalInverseGamma`,
`NormalInverseWishart`,
`NormalWishart`

since they are all exported, but trying to access them (e.g. `Distributions.NormalGamma`) fails with `ERROR: UndefVarError: NormalGamma not defined`.

This is on an up-to-date Julia:
```
julia> versioninfo()
Julia Version 0.5.0
Commit 3c9d753 (2016-09-19 18:14 UTC)
Platform Info:
  System: Darwin (x86_64-apple-darwin13.4.0)
  CPU: Intel(R) Core(TM) i7-4980HQ CPU @ 2.80GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.7.1 (ORCJIT, haswell)
julia> Pkg.status("Distributions")
 - Distributions                 0.11.0
```


johnmyleswhite:
I think these all got moved to https://github.com/JuliaStats/ConjugatePriors.jl.

Removing them seems sensible to me. We might also want to consider a deprecation warning, which is what I'd assume was the intent for leaving them among the exports.

johnmyleswhite:
So I'm ok with deleting these rather than deprecating them, but I think we need to make some additional changes before we can commit this PR.

I see documentation for functionality that isn't present in `doc/source/conjugates.rst` and I also see tests for this functionality in `test/compoundvariate.jl`. We should probably delete all of these things?

Are you up for that?

Thanks for working on this.

cmey:
Great! Yes, I'm ok doing the additional changes to finish the PR.

cmey:
Anything else that needs to be modified?

doc/source/conjugates.rst
```diff
@@ -113,10 +113,6 @@ The cases for *Normal* are more involved, as they have two parameters: the mean
 	pri = NormalInverseGamma(mu, v, sh, r)
```
johnmyleswhite: Was this distribution also deleted?

johnmyleswhite:
I'm happy to merge this since it's a clear improvement over the current state, but my impression is that we need some more work after this to finish cleaning up the docs.

What do others want to do?

doc/source/conjugates.rst
cmey: Yes, fixed

johnmyleswhite:
Thanks for doing this!

PR #557, opened 2016-11-25
Document fix for logistic formula

ararslan:
Adds the fix in #555 to the docs. @andreasnoack, you said that once this is merged, ReadTheDocs will automatically update?


andreasnoack:
Looks like it works.

ararslan:
Great, thanks so much! One of these days I'll figure things out...

PR #564, opened 2016-12-08
Fix check_args macro

andreasnoack:


ararslan:
The 0.6 Linux failure is because Rmath is being built from source, whereas it's coming from Homebrew on OS X. Looks like maybe BinDeps is the culprit? If it's Rmath, we'll have to bump the version of StatsFuns, since that's how we're calling into Rmath here currently...

yuyichao:
```diff
diff --git a/src/utils.jl b/src/utils.jl
index 58c9cb1..08058df 100644
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -105,9 +105,13 @@ end
 # for checking the input range of quantile functions
 # comparison with NaN is always false, so no explicit check is required
 macro checkquantile(p,ex)
+    p = esc(p)
+    ex = esc(ex)
     :(zero($p) <= $p <= one($p) ? $ex : NaN)
 end
 macro checkinvlogcdf(lp,ex)
+    lp = esc(lp)
+    ex = esc(ex)
     :($lp <= zero($lp) ? $ex : NaN)
 end
```

andreasnoack:
@yuyichao Thanks. I've fixed it.

src/utils.jl
```diff
@@ -105,9 +105,11 @@ end
 # for checking the input range of quantile functions
 # comparison with NaN is always false, so no explicit check is required
 macro checkquantile(p,ex)
+    p = esc(p)
```
yuyichao: Ex too

andreasnoack:
Done.

src/utils.jl
```diff
@@ -2,9 +2,9 @@
 
 macro check_args(D, cond)
     quote
-        if !($cond)
+        if !($(esc(cond)))
             throw(ArgumentError(string(
-                $(string(D)), ": the condition ", $(string(cond)), " is not satisfied.")))
+                $(string(esc(D))), ": the condition ", $(string(esc(cond))), " is not satisfied.")))
```
yuyichao: Actually, I don't think you want these two `esc`.

yuyichao:
And just for the reference, the change I was testing locally before I saw this.

```patch
diff --git a/src/utils.jl b/src/utils.jl
index 58c9cb1..08058df 100644
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -2,7 +2,7 @@
 
 macro check_args(D, cond)
     quote
-        if !($cond)
+        if !($(esc(cond)))
             throw(ArgumentError(string(
                 $(string(D)), ": the condition ", $(string(cond)), " is not satisfied.")))
         end
@@ -105,9 +105,13 @@ end
 # for checking the input range of quantile functions
 # comparison with NaN is always false, so no explicit check is required
 macro checkquantile(p,ex)
+    p = esc(p)
+    ex = esc(ex)
     :(zero($p) <= $p <= one($p) ? $ex : NaN)
 end
 macro checkinvlogcdf(lp,ex)
+    lp = esc(lp)
+    ex = esc(ex)
     :($lp <= zero($lp) ? $ex : NaN)
 end
```

andreasnoack:
When https://github.com/JuliaLang/METADATA.jl/pull/7245 is merged, the tests on master can be restarted.

PR #568, opened 2016-12-18
Define support for mixture models

crbinz:
Defines support for `UnivariateGMM` using `@distr_support`, and implements `minimum(::MixtureModel)` and `maximum(::MixtureModel)`, since those depend on the individual components. This also adds relevant tests.


crbinz:
@ararslan thanks!

PR #570, opened 2017-01-21
Update to Julia 0.6

andreasnoack:
Drop Julia 0.4 support


andreasnoack:
@jmxpearson I've removed the `promote_eltype` and `convert_eltype` in favor of using conversion mechanisms similar Base's. Some of the definitions become slightly more verbose but we avoid introducing new promotion functions of which there are already too many. In particular, we avoid having a `promote_eltype` with a different meaning than the version in base.

I've also had to remove the general tests for conversions for now. They hooked into internal representations of types which have now changed and which is also generally considered a bad idea by people who knows more about this than I do. We should think of a way to test conversions of all distributions without hooking into the internals of types.

jmxpearson:
Sounds great! Makes sense that conversions should be handled with the same idioms as Base.

src/matrix/inversewishart.jl
```diff
@@ -15,12 +15,15 @@ function InverseWishart{T<:Real}(df::T, Ψ::AbstractPDMat{T})
     p = dim(Ψ)
     df > p - 1 || error("df should be greater than dim - 1.")
     c0 = _invwishart_c0(df, Ψ)
-    R = promote_type(T, typeof(c0))
-    _, prom_Ψ = promote_eltype(one(R), Ψ)
+    R = Base.promote_eltype(T, c0)
+    prom_Ψ = convert(AbstractArray{R}, Ψ)
```
ararslan: Why is this converting to an abstract type rather than a concrete type like simply `Array{R}`?

src/matrix/wishart.jl
```diff
@@ -45,11 +48,11 @@ params(d::Wishart) = (d.df, d.S, d.c0)
 
 ### Conversion
 function convert{T<:Real}(::Type{Wishart{T}}, d::Wishart)
-    P = convert_eltype(T, d.S)
+    P = AbstractMatrix{T}(d.S)
     Wishart{T, typeof(P)}(T(d.df), P, T(d.c0))
 end
 function convert{T<:Real}(::Type{Wishart{T}}, df, S::AbstractPDMat, c0)
-    P = convert_eltype(T, S)
+    P = AbstractMatrix{T}(S)
```
ararslan: Isn't `AbstractPDMat <: AbstractMatrix` already?

test/dirichletmultinomial.jl
```diff
@@ -44,14 +45,10 @@ d = DirichletMultinomial(10, 5)
...
-    @test_approx_eq(
-        pdf(d, x),
-        @compat factorial(d.n) * gamma(d.α0) / gamma(d.n + d.α0) * prod(gamma(d.α + x) ./ factorial.(x) ./ gamma(d.α))
-    )
-    @test_approx_eq(
-        logpdf(d, x),
-        @compat log(factorial(d.n)) + lgamma(d.α0) - lgamma(d.n + d.α0) + sum(lgamma(d.α + x) - log.(factorial.(x)) - lgamma(d.α))
-    )
+    @test pdf(d, x) ≈
+        @compat factorial(d.n) * gamma(d.α0) / gamma(d.n + d.α0) * prod(gamma.(d.α + x) ./ factorial.(x) ./ gamma.(d.α))
```
ararslan: If you're dropping 0.4 support, the `@compat` should no longer be needed here.

test/runtests.jl
```diff
@@ -58,7 +58,7 @@ end
 if VERSION < v"0.5.0-"
```
ararslan: Can get rid of this check since you're dropping 0.4

src/matrix/wishart.jl
ararslan: Oh never mind, I see what you're doing there, you're converting the element type of `S` without changing the parent type of `S`.

src/matrix/inversewishart.jl
andreasnoack: This is the Base idiom for converting the element type of an array regardless of the array type. In this case, it will be some kind of `PDMat`.

ararslan: Gotcha, thanks!

andreasnoack:
The I can't reproduce the stall on Mac nightly locally so I'll merge this.

Issue #550, opened 2016-10-26
NoncentralChisq is undocumented


nbecker:

ararslan:
I believe this is one of several things that are undocumented in this package. I think the best course of action would be to go through and update docstrings, then switch over the doc system to Documenter.jl. That should be easier to maintain going forward.

lbollar:
I would be interested in writing some doc strings on this and possibly in other places where it might be needed. For this one, would I just need to follow the example in the Normal Distribution docstring?

ararslan:
@lbollar That would likely be a good place to start. I'm glad someone is interested in taking this on, thanks!

lbollar:
@ararslan Does this need to be closed out?

ararslan:
Yes it does, thanks for the reminder!

PR #571, opened 2017-02-01
implemented docstring for noncentralchisq

lbollar:


lbollar:
Referencing [#550](https://github.com/JuliaStats/Distributions.jl/issues/550)

src/univariate/continuous/noncentralchisq.jl
```diff
@@ -1,3 +1,24 @@
+doc"""
```
ararslan: the `doc` part is unnecessary

src/univariate/continuous/noncentralchisq.jl
```diff
@@ -1,3 +1,24 @@
...
+NoncentralChisq(ν, λ)     # Noncentral Chi-squared distribution with ν degrees of freedom and noncentrality parameter λ
+
+params(d)    # Get the parameters, i.e. (ν, λ)
+```
+
+External links
+
+* [Noncentral Chi-squared distribution on Wikipedia](https://en.wikipedia.org/wiki/Noncentral_chi-squared_distribution)
+"""
+
```
ararslan: no line break here

src/univariate/continuous/noncentralchisq.jl
```diff
@@ -1,3 +1,24 @@
+doc"""
+    NoncentralChisq(ν, λ)
```
ararslan: should have a line break after this

src/univariate/continuous/noncentralchisq.jl
```diff
@@ -1,3 +1,24 @@
+"""
+    NoncentralChisq(ν, λ)
+
+The *Noncentral Chi-squared distribution* (typically written χ²) with `ν` degrees of freedom and non-centrality parameter `λ` has the probability density function
```
ararslan: I think both "noncentral" and "chi" should be lowercase. Also the parenthesized note about the notation is probably unnecessary. I'd also remove the hyphen in "non-centrality" for consistency with "noncentral."

lbollar:
I went ahead and changed chi to lower case in the other places it appeared as well.

src/univariate/continuous/noncentralchisq.jl
```diff
@@ -1,3 +1,24 @@
+"""
+    NoncentralChisq(ν, λ)
+
+The *noncentral chi-squared distribution* with `ν` degrees of freedom and noncentrality parameter `λ` has the probability density function
+
+$f(x; ν, λ) = \frac{1}{2}e^{-(x + λ)/2}\frac{x}{λ}^{ν/4-1/2}I_{ν/2-1}(\sqrt{λx}) \quad x>0.$
```
ararslan: This might render better as
```tex
f(x; \nu, \lambda) = \frac{1}{2} e^{-(x + \lambda)/2} \left( \frac{x}{\lambda} \right)^{\nu/4-1/2} I_{\nu/2-1}(\sqrt{\lambda x}), \quad x > 0
```
Typically TeX prefers getting its recognized sequences, e.g. `\lambda`, rather than raw Unicode, e.g. λ. Whether raw Unicode works at all often depends on the particular TeX distribution and/or whether `inputenc` is active.

Also note the parentheses around the second fraction. What I've shown above should exactly match what's on Wikipedia, with the exception of `\nu` as the noncentrality parameter instead of `k`.

src/univariate/continuous/noncentralchisq.jl
```diff
@@ -1,3 +1,24 @@
...
+
+It is the distribution of the sum of squares of `ν` independent [`Normal`](:func:`Normal`) variates with individual means $μ_i$ and
+
+$λ = \sum_{i=1}^ν{μ_i^2}$
```
ararslan: Here too with the TeX syntax: `\lambda = \sum_{i=1}^\nu \mu_i^2` should do the trick.

ararslan:
Apologies for the fragmented review; I've been doing a bunch of things at once for most of the day.

src/univariate/continuous/noncentralchisq.jl
lbollar: Thanks for this tip. TeX isn't my forte and I was wondering which was the better way to use unicode in it.

src/univariate/continuous/noncentralchisq.jl
lbollar: I had missed that parentheses. Great catch.

lbollar:
No worries, thanks for patience with a new committer trying to get involved.

ararslan:
Thanks for your willingness to help, not to mention your quality contributions and positive attitude. 🙂

PR #586, opened 2017-03-23
Generic support to truncated dist

xukai92:
Add generic type support to truncated distributions.

### DIFF
--- a/src/truncate.jl
+++ b/src/truncate.jl
@@ -41,21 +41,25 @@ pdf(d::Truncated, x::Float64) = d.lower <= x <= d.upper ? pdf(d.untruncated, x) 
 
 logpdf(d::Truncated, x::Float64) = d.lower <= x <= d.upper ? logpdf(d.untruncated, x) - d.logtp : -Inf
 
-cdf(d::Truncated, x::Float64) = x <= d.lower ? 0.0 :
-                             x >= d.upper ? 1.0 :
-                             (cdf(d.untruncated, x) - d.lcdf) / d.tp
+cdf(d::Truncated, x::Float64) =
+    x <= d.lower ? 0.0 :
+    x >= d.upper ? 1.0 :
+    (cdf(d.untruncated, x) - d.lcdf) / d.tp
 
-logcdf(d::Truncated, x::Float64) = x <= d.lower ? -Inf :
-                                x >= d.upper ? 0.0 :
-                                log(cdf(d.untruncated, x) - d.lcdf) - d.logtp
+logcdf(d::Truncated, x::Float64) =
+    x <= d.lower ? -Inf :
+    x >= d.upper ? 0.0 :
+    log(cdf(d.untruncated, x) - d.lcdf) - d.logtp
 
-ccdf(d::Truncated, x::Float64) = x <= d.lower ? 1.0 :
-                              x >= d.upper ? 0.0 :
-                              (d.ucdf - cdf(d.untruncated, x)) / d.tp
+ccdf(d::Truncated, x::Float64) =
+    x <= d.lower ? 1.0 :
+    x >= d.upper ? 0.0 :
+    (d.ucdf - cdf(d.untruncated, x)) / d.tp
 
-logccdf(d::Truncated, x::Float64) = x <= d.lower ? 0.0 :
-                                 x >= d.upper ? -Inf :
-                                 log(d.ucdf - cdf(d.untruncated, x)) - d.logtp
+logccdf(d::Truncated, x::Float64) =
+    x <= d.lower ? 0.0 :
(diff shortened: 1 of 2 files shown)


src/truncate.jl
```diff
@@ -80,6 +80,28 @@ logccdf(d::Truncated, x::Int) = x <= d.lower ? 0.0 :
                                  x >= d.upper ? -Inf :
                                  log(d.ucdf - cdf(d.untruncated, x)) - d.logtp
 
+# Callback functions for generic type
+
+pdf(d::Truncated, x::Real) = d.lower <= x <= d.upper ? pdf(d.untruncated, x) / d.tp : 0.0
```
andreasnoack: Instead of hard-coded `Float64` literals, the return type should be inferred from the inputs. This might yet be the case for all of the package but we are working towards it. You should be to find some example in the source where the input types are used to compute the correct return type.

xukai92: Get it. I've made the change.

src/truncate.jl
```diff
@@ -80,6 +79,28 @@ logccdf(d::Truncated, x::Int) = x <= d.lower ? 0.0 :
...
                                  log(d.ucdf - cdf(d.untruncated, x)) - d.logtp
 
+# Callback functions for generic type
+
+pdf{T<:Real}(d::Truncated, x::T) = d.lower <= x <= d.upper ? pdf(d.untruncated, x) / d.tp : zero(T)
+
+logpdf{T<:Real}(d::Truncated, x::T) = d.lower <= x <= d.upper ? logpdf(d.untruncated, x) - d.logtp : -T(Inf)
+
+cdf{T<:Real}(d::Truncated, x::T) = x <= d.lower ? zero(T) :
+                            x >= d.upper ? one(T) :
```
ararslan: The indentation got kind of odd here. I'd maybe align it like
```julia
cdf{T<:Real}(d::Truncated, x::T) =
    x <= d.lower ? zero(T) :
    x >= d.upper ? one(T) :
    (cdf(d.untruncated, x) - d.lcdf) / d.tp
```
Same below.

xukai92: Yeah I agree. I've changed it according to your suggestion.

ararslan:
Rather than duplicating the definitions for `Float64`, `Int`, and `T<:Real`, I would omit the float definition and let it use `Real`, and I would just define the `Int` methods to be `f(d::Truncated, x::Int) = f(d, float(x))` for each `f` here.

xukai92:
Updated!

src/truncate.jl
```diff
@@ -37,48 +37,38 @@ insupport{D<:UnivariateDistribution}(d::Truncated{D,Union{Discrete,Continuous}},
...
-
 quantile(d::Truncated, p::Float64) = quantile(d.untruncated, d.lcdf + p * d.tp)
 
-pdf(d::Truncated, x::Int) = d.lower <= x <= d.upper ? pdf(d.untruncated, x) / d.tp : 0.0
+pdf(d::Truncated, x::Int) = pdf(d, float(x))
+logpdf(d::Truncated, x::Int) = logpdf(d, float(x))
+cdf(d::Truncated, x::Int) = cdf(d, float(x))
+logcdf(d::Truncated, x::Int) = logcdf(d, float(x))
+ccdf(d::Truncated, x::Int) = ccdf(d, float(x))
+logccdf(d::Truncated, x::Int) = logccdf(d, float(x))
```
ararslan: If you wanted to get *really* fancy, you could do
```julia
for f in [:pdf, :logpdf, :cdf, :logcdf, :ccdf, :logccdf]
    @eval ($f)(d::Truncated, x::Int) = ($f)(d, float(x))
end
```
Not really necessary to change what you have, I just think that style is fun.

ararslan:
It would be great if you could add some tests for the new functionality.

src/truncate.jl
xukai92: I actually fancy your style as well :) Will change to it! Thanks!

xukai92:
I just added test using `Dual` type as a test case. I ignored some (three) distributions for the generic test as the corresponding `pdf()` functions are not generically supported in `StastFuns.jl` because I didn't implement all `pdf()` callbacks in https://github.com/JuliaStats/StatsFuns.jl/pull/21. 

For similar reason, I didn't add test for `cdf()` and other methods are they are not supported generically in `StastFuns.jl`.

xukai92:
The test error seems unrelated to this PR.

xukai92:
Hi all, any idea when this PR can be merged?

ararslan:
Thanks for the reminder! Since there's been no further comment since I submitted my review, I'll go ahead and merge. Thanks for the contribution!

Issue #554, opened 2016-11-24
Generalized inverse Gaussian


patwa67:
Someone that plans to take on the task of implementing the generalized inverse Gaussian distribution?
DOI: 10.1007/s11222-013-9387-3
https://cran.r-project.org/web/packages/GIGrvg/GIGrvg.pdf

ararslan:
I don't know that anyone here is already planning to take that on, so a PR would be welcome.

*However*, do note that the R code to which you linked it licensed under GPL. That means that you can't simply translate the code to Julia, since that would constitute a "derived work," and would require the Julia code to be licensed under GPL as well. We wouldn't be able to accept the PR, since this package is licensed under MIT.

Perhaps you already know all that, just figured I'd mention it just in case. 😉

patwa67:
I don't think I'm the right person to lead a PR for this as Julia is very new to me. I would be happy to follow the process if someone else could take the lead. There is very useful pseudo-code in the paper, but it would take weeks for me to get that into Julia.

ararslan:
If it's something you'd be interested in doing then I encourage you to try your hand at it. I find things like this can be incredibly rewarding to work on, and if you need any guidance along the way I'd be happy to lend a hand. Up to you though; if you'd rather leave it to someone else that's of course totally fine as well. I just don't want you to feel like you can't or shouldn't just because you're new. Everyone starts somewhere!

patwa67:
I can give it a try, but would need help on how to start up the PR. Could you shortly outline how to do that?

ararslan:
Of course! Here's my personally preferred workflow:

* Fork the repository to your account (click Fork in the upper right corner)
* Go somewhere on your computer, `git clone https://github.com/JuliaStats/Distributions.jl.git`
* `git remote add patwa67 https://github.com/patwa67/Distributions.jl.git` (this is your fork)
* `git checkout -b yourbranch` (where `yourbranch` is whatever you want to call it)
* Create a file `src/univariate/continuous/generalizedinversegaussian.jl`
* Create the distribution's type and define methods
  * Here we define a distribution as an _immutable_ that inherits from an abstract type.
  * See `src/univariate/continuous/inversegaussian.jl` for an example of the interface required for a distribution (things like `pdf`, `cdf`, `rand`, etc.)
  * Make sure to add doc strings (again see other distributions for examples)
* `git add src/univariate/continuous/generalizedinversegaussian.jl` (stage the file for commit)
[...]

PR #587, opened 2017-03-31
WIP: GeneralizedInverseGaussian

nsgrantham:
With respect to #554, an initial implementation of the [generalized inverse Gaussian distribution](https://en.wikipedia.org/wiki/Generalized_inverse_Gaussian_distribution). The areas that require particular attention are:

1. Sampling methods from [Hormann and Leydold, (2014)](http://link.springer.com/article/10.1007/s11222-013-9387-3) in `rand`, and
2. Estimation of the CDF [eq (5) from Lemonte and Cordeiro (2011)](https://www.ime.usp.br/~abe/lista/pdfeX5DRCVM26.pdf) in `cdf`.

The CDF algorithm requires some extra care, it is unstable for very small and large values of `x`, when the function should simply return `0.0` and `1.0` respectively, but the overflow makes these situations a bit difficult to capture.

It should also be noted that the sampling algorithms are designed for quick "one-off" sampling, as one might do in Gibbs sampling, rather than bulk sampling. Perhaps another algorithm should be included for bulk sampling.


src/univariate/continuous/generalizedinversegaussian.jl
```diff
@@ -0,0 +1,273 @@
+doc"""
```
ararslan: No `doc` here

src/univariate/continuous/generalizedinversegaussian.jl
```diff
@@ -0,0 +1,273 @@
...
+    GeneralizedInverseGaussian(a,b,p)
+
+The *generalized inverse Gaussian distribution* with parameters a > 0, b > 0, p real,
+and modified Bessel function of the second kind K_p, has probability density function
+
+$f(x; a, b, p) = \frac{(a/b)^{p/2}}{2K_p(\sqrt{ab})}x^{(p-1)}
```
ararslan: Math should be surrounded with double backticks rather than the TeX-style dollar signs

src/univariate/continuous/generalizedinversegaussian.jl
```diff
@@ -0,0 +1,273 @@
...
+e^{-(ax+b/x)/2}, \quad x > 0$
+
+``julia
```
ararslan: Three backticks for code blocks

src/univariate/continuous/generalizedinversegaussian.jl
ararslan: And all backslashes should be escaped

matbesancon:
@nsgrantham bump, what's the status on this?
There seems to be some conflicts + review points to resolve

nsgrantham:
> @nsgrantham bump, what's the status on this?
> There seems to be some conflicts + review points to resolve

The status is "indefinitely stalled." The PR was written for Julia v0.4 or v0.5 IIRC and, as I do not program in Julia these days, I am not the right person to resolve the conflicts and make everything v1.0-friendly.

If there is someone else with a strong interest in adding `GeneralizedInverseGaussian` to the already extensive `Distributions.jl` repertoire, please go ahead and move forward without me, though I am happy to answer questions should they arise.

matbesancon:
Thanks, if needed someone will use this PR for future work, we can close it

dylanfesta:
Hi, I was searching for an implementation of generalized inverse Gaussian distribution on Julia, and I stumbled on this page.  How much work does it require?  I have some experience in updating 0.6 code to 1.0 , but I am not an expert developer.  How can I help with it?

matbesancon:
hi, yes you can definitely clone the branch and work to adapt it I'll be happy to review.
Otherwise you can get inspiration from the current branch but re-code it from scratch on the current version of Distributions and Julia

dylanfesta:
Ok, I adapted the code to 1.0 (very few changes, actually) , except the testing part (I was not sure how to adapt it).  For now it's here 
https://github.com/dylanfesta/Distributions.jl 
How should I proceed? Should I make a new PR out of it?

nsgrantham:
@dylanfesta Thanks for making the changes. It may be worthwhile to review the manuscripts I linked in my first comment to ensure the code matches the theory. Moreover, I have experienced stability issues for some extreme parameter settings (you may see this if you run the tests). I suspect there is an opportunity to be more careful with the computations, perhaps with log transforms and normalizing constants in the right places.

dylanfesta:
The sampling follows quite closely the equations in the paper you linked.  And I assume you already tested that the empirical distribution converges to the PDF ?  
Regarding the CDF , that you reference as  eq (5) from Lemonte and Cordeiro , I feel a bit confused by the terms:    <a href="https://www.codecogs.com/eqnedit.php?latex=\Gamma(\lambda-j)" target="_blank"></a>   . In Julia `gamma(-1.0)` returns an error (and the value diverges to plus/minus infinity) so how can the sum over gammas be defined for any integer lambda ?   I could not find other resources online, but this is not really my field.

matbesancon:
It is a bit odd if there is no CDF, if some parameter values make the distribution "undefinable", there should be a check at its construction to always allow a valid GIG object to give you a CDF at points it is defined

dylanfesta:
The problem here is that the CDF is not in closed form. It only appears in  [equation (5) in the following paper](https://www.ime.usp.br/~abe/lista/pdfeX5DRCVM26.pdf) however it is not numerically stable even for perfectly legitimate parameters.

In other words, the following happens:
```julia-repl
julia> d = GeneralizedInverseGaussian(1., 1., 1.)
GeneralizedInverseGaussian{Float64}(a=1.0, b=1.0, p=1.0)
julia> cdf(d,1.0)
ERROR: DomainError with -1.0:
NaN result for non-NaN input.
Stacktrace:
 [1] nan_dom_err at ./math.jl:339 [inlined]
 [2] gamma at /home/dfesta/.julia/packages/SpecialFunctions/fvheQ/src/gamma.jl:558 [inlined]
 [3] cdf(::GeneralizedInverseGaussian{Float64}, ::Float64) at /home/dfesta/Repositories/Distributions/src/univariate/continuous/generalizedinversegaussian.jl:118
 [4] top-level scope at none:0
```
This is *not a bug*, just the result of a discontinuity in the Gamma functions of equation (5) in the reference linked above . Writing a CDF requires much more work than just using that formula, which is certainly mathematically correct, but unfit for numerical computation.  

And of course the (1 , 1...) parameters work correctly for the sampler, the pdf, etc...  

testexample

theogf:
Reviving this PR:
Wouldn't that it be more efficient to simply compute the integral numerically of the pdf (as descrived in [equation (4)](https://www.ime.usp.br/~abe/lista/pdfeX5DRCVM26.pdf) ) ?

matbesancon:
I think it even seems to be the only option (more or less)

theogf:
Is it already the generic fallback or should I make a new PR with it?

matbesancon:
No generic fallback for this. I think we could finish this one, with a MethodError for the cdf, and then implement it numerically in another

PR #588, opened 2017-04-07
Fix Rayleigh CDF for negative values

evanfields:
Rayleigh distribution has support `[0, inf)`, but the cdf was giving positive values for inputs `< 0`. One line fix, I think.


ararslan:
Thanks. It would be great if you could add a test.

evanfields:
Of course. I looked through the tests of the univariate distributions which are compared to R code and didn't totally understand the structure. Should the test be added along with all that, or is a line in `/test/continuous.jl` okay?

ararslan:
Oh right, I forgot we do the tests kind of weird here--I also don't understand the structure. Maybe add something to this block? https://github.com/JuliaStats/Distributions.jl/blob/83c063d6858e93a73b25e17bfa9e5787d914235f/test/ref/continuous_test.ref.json#L4005-L4013

jmxpearson:
I will take a look at this tonight.

PR #591, opened 2017-04-12
Fix kurtosis calculation for FDist

### LINKED ISSUE #590: Error in kurtosis formula of F distribution
Thanks for this library! 
I would like to report a bug: 
The [code](https://github.com/JuliaStats/Distributions.jl/blob/master/src/univariate/continuous/fdist.jl#L82) implementing the kurtosis for the Fisher F distribution seems to contain a typo. 
The current code for the denominator is `ν1 * (ν2 - 6) * (ν2 - 8) * (ν2 - 2)`, whereas the correct formula is `ν1 * (ν2 - 6) * (ν2 - 8) * (ν1 + ν2 - 2)`, see [1](https://en.wikipedia.org/wiki/F-distribution) or [2](http://www.math.uah.edu/stat/special/Fisher.html).

johnmyleswhite:
Before:

```
julia> using Distributions

julia> d = FDist(30, 40)
Distributions.FDist{Float64}(ν1=30.0, ν2=40.0)

julia> kurtosis(d)
4.016099071207431

julia> kurtosis(rand(d, 5_000_000))
2.245386630172172
```

After:

```
julia> using Distributions

julia> d = FDist(30, 40)
Distributions.FDist{Float64}(ν1=30.0, ν2=40.0)

julia> kurtosis(d)
2.2442906574394463

julia> kurtosis(rand(d, 5_000_000))
2.2924267619033687
```

Closes #590

### DIFF
--- a/src/univariate/continuous/fdist.jl
+++ b/src/univariate/continuous/fdist.jl
@@ -79,7 +79,7 @@ function kurtosis{T<:Real}(d::FDist{T})
     (ν1, ν2) = params(d)
     if ν2 > 8
         a = ν1 * (5ν2 - 22) * (ν1 + ν2 - 2) + (ν2 - 4) * (ν2 - 2)^2
-        b = ν1 * (ν2 - 6) * (ν2 - 8) * (ν2 - 2)
+        b = ν1 * (ν2 - 6) * (ν2 - 8) * (ν1 + ν2 - 2)
         return 12a / b
     else
         return T(NaN)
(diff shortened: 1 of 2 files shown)


ararslan:
Great, thanks! Could you add a test for this?

johnmyleswhite:
To be honest, I don't really have time to do that. Happy to abandon the PR and let someone else take over.

ararslan:
No problem, I can push a test to your branch. Thanks for taking the time to fix this!

PR #594, opened 2017-04-16
Revert change to multivariate log likelihood

### LINKED ISSUE #595: 0.12.4 fails
```julia
julia> dist = Distributions.MvNormal(zeros(1))
julia> Distributions.loglikelihood(dist, [5.]'')

ERROR: MethodError: no method matching _logpdf(::Distributions.MvNormal{Float64,PDMats.PDiagMat{Float64,Array{Float64,1}},Distributions.ZeroVector{Float64}}, ::Float64)
Closest candidates are:
  _logpdf(::Distributions.AbstractMvNormal, ::AbstractArray{T,1}) at /Users/monty/.julia/v0.5/Distributions/src/multivariate/mvnormal.jl:44
 in _reducedim_init(::Distributions.##29#30{Distributions.MvNormal{Float64,PDMats.PDiagMat{Float64,Array{Float64,1}},Distributions.ZeroVector{Float64}}}, ::Base.#+, ::Base.#zero, ::Base.#sum, ::Array{Float64,2}, ::Int64) at ./reducedim.jl:122
 in loglikelihood(::Distributions.MvNormal{Float64,PDMats.PDiagMat{Float64,Array{Float64,1}},Distributions.ZeroVector{Float64}}, ::Array{Float64,2}) at /Users/monty/.julia/v0.5/Distributions/src/multivariates.jl:118
```

ararslan:
This reinstates the previous definition of `loglikelihood` for multivariate distributions that was changed in #592. Unfortunately there were no tests to verify the behavior so an issue went unnoticed. Addresses https://github.com/JuliaLang/METADATA.jl/pull/8848#issuecomment-294328801.

@jrevels Let me know if you have any ideas for a better way to initialize the accumulator or to otherwise approach this.

Fixes  #595

### DIFF
--- a/src/multivariates.jl
+++ b/src/multivariates.jl
@@ -113,9 +113,17 @@ end
 
 ## log likelihood
 
+function _loglikelihood(d::MultivariateDistribution, X::AbstractMatrix)
+    ll = 0.0
+    @inbounds for i in 1:size(X, 2)
+        ll += _logpdf(d, view(X, :, i))
+    end
+    return ll
+end
+
 function loglikelihood(d::MultivariateDistribution, X::AbstractMatrix)
     size(X, 1) == length(d) || throw(DimensionMismatch("Inconsistent array dimensions."))
-    return sum(x -> _logpdf(d, x), X, 2)
+    return _loglikelihood(d, X)
 end
 
 ##### Specific distributions #####
--- a/src/multivariates.jl
+++ b/src/multivariates.jl
@@ -113,17 +113,9 @@ end
 
 ## log likelihood
 
-function _loglikelihood(d::MultivariateDistribution, X::AbstractMatrix)
-    ll = 0.0
-    @inbounds for i in 1:size(X, 2)
-        ll += _logpdf(d, view(X, :, i))
-    end
-    return ll
-end
-
 function loglikelihood(d::MultivariateDistribution, X::AbstractMatrix)
     size(X, 1) == length(d) || throw(DimensionMismatch("Inconsistent array dimensions."))
-    return _loglikelihood(d, X)
+    return sum(i -> _logpdf(d, view(X, :, i)), 1:size(X, 2))
 end
 
 ##### Specific distributions #####
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.5
-PDMats 0.5.4
+PDMats 0.6.0
 StatsFuns 0.3.1
 Calculus
 StatsBase 0.8.3
--- a/test/mvnormal.jl
+++ b/test/mvnormal.jl
@@ -55,6 +55,9 @@ function test_mvnormal(g::AbstractMvNormal, n_tsamples::Int=10^6)
         @test logpdf(g, X[:,i]) ≈ lp[i]
     end
     @test logpdf(g, X) ≈ lp
+
+    # log likelihood
+    @test loglikelihood(g, X) ≈ sum([Distributions._logpdf(g, X[:,i]) for i in 1:size(X, 2)])
 end
 
 


src/multivariates.jl
```diff
@@ -113,9 +113,17 @@ end
...
+    @inbounds for i in 1:size(X, 2)
+        ll += _logpdf(d, view(X, :, i))
+    end
+    return ll
+end
+
 function loglikelihood(d::MultivariateDistribution, X::AbstractMatrix)
     size(X, 1) == length(d) || throw(DimensionMismatch("Inconsistent array dimensions."))
-    return sum(x -> _logpdf(d, x), X, 2)
+    return _loglikelihood(d, X)
```
jrevels: Why not use:

```julia
sum(Distributions._logpdf(d, view(X, :, i)) for i in 1:size(X, 2))
```

Also probably don't want to throw back in `_loglikelihood`, since it's not used anywhere.

ararslan: Good call, thanks! I tend to avoid generators where possible since the `eltype` can't be inferred but this could just as easily be written using a lambda.

jrevels: > I tend to avoid generators where possible since the eltype can't be inferred

What do you mean? I wasn't aware generators introduced any inference issues (besides the usual closure boxing issues, but this shouldn't be affected by that). AFAICT, this statement seems well-inferred to me:

```julia
julia> f(d, X) = sum(Distributions._logpdf(d, view(X, :, i)) for i in 1:size(X, 2))

julia> d = MvNormal(rand(3), rand())

julia> X = rand(g, 10);

julia> @code_warntype f(d, X)
Variables:
  #self#::#f
  d::Distributions.MvNormal{Float64,PDMats.ScalMat{Float64},Array{Float64,1}}
  X::Array{Float64,2}
  #5::##5#6{Distributions.MvNormal{Float64,PDMats.ScalMat{Float64},Array{Float64,1}},Array{Float64,2}}

Body:
  begin
      #5::##5#6{Distributions.MvNormal{Float64,PDMats.ScalMat{Float64},Array{Float64,1}},Array{Float64,2}} = $(Expr(:new, ##5#6{Distributions.MvNormal{Float64,PDMats.ScalMat{Float64},Array{Float64,1}},Array{Float64,2}}, :(d), :(X)))
      SSAValue(3) = (Base.arraysize)(X::Array{Float64,2}, 2)::Int64
      SSAValue(1) = $(Expr(:new, UnitRange{Int64}, 1, :((Base.select_value)((Base.sle_int)(1, SSAValue(3))::Bool, SSAValue(3), (Base.sub_int)(1, 1)::Int64)::Int64)))
      SSAValue(2) = $(Expr(:new, Base.Generator{UnitRange{Int64},##5#6{Distributions.MvNormal{Float64,PDMats.ScalMat{Float64},Array{Float64,1}},Array{Float64,2}}}, :(#5), SSAValue(1)))
      return $(Expr(:invoke, MethodInstance for mapfoldl(::Base.#identity, ::Function, ::Base.Generator{UnitRange{Int64},##5#6{Distributions.MvNormal{Float64,PDMats.ScalMat{Float64},Array{Float64,1}},Array{Float64,2}}}), :(Base.mapfoldl), :(Base.identity), :(Base.+), SSAValue(2)))
  end::Float64
```

jrevels: Ah, I see this doesn't infer as well on v0.5, which Distributions still supports. I guess `sum(i -> Distributions._logpdf(d, view(X, :, i)), 1:size(X, 2))` would make more sense then.

ararslan:
Ah, the tests seem to be failing because of the `view`; `invquad` is getting called, and that expects a `StridedArray`.

Issue #436, opened 2015-11-26
use non-global rng


CarloLucibello:
I would like to have the opportunity to do something like this

``` julia
d= Binomial(10,0.1)
rng = MersenneTwister(17)
rand(rng, d)
```

Is there any way to do something similar?

simonbyrne:
Not yet: some distributions such as Binomial utilise the Rmath RNGs, which unfortunately are tied to global state. I'm hoping to fix this in the near future.

zsunberg:
This would be extremely useful to us - we're developing a [framework for decision making under uncertainty](https://github.com/JuliaPOMDP/POMDPs.jl). We use a lot of distributions, but it is really convenient to be able to specify the rng, so using Distributions.jl doesn't work well for us yet. I hope that you can implement this  soon - we would really like to enable more ease-of use and compatibility of code by using this!

richardreeve:
@matbesancon I think this can be closed now.

PR #597, opened 2017-04-27
rand(rng, d) for certain distributions (starting with MvNormal)

zsunberg:
#436 has made using Distributions.jl with POMDPs.jl difficult. Unfortunately completely addressing #436 seems to be a large task since the issue has not been addressed for a long time. To enable the use of some distributions, we have been using [this hack](https://github.com/JuliaPOMDP/POMDPToolbox.jl/blob/master/src/distributions/distributions_jl.jl), which I would like to get rid of. I am wondering if JuliaStats would be amenable to adding the `rng` argument to `rand` for only a subset of distributions. After all, people using POMDPs.jl tend to only use either `MvNormal` or the univariate distributions.

This PR contains the implementation for `AbstractMvNormal`. If you think this is a good idea, I will also add it for some of the univariate distributions.

### DIFF
--- a/doc/source/multivariate.rst
+++ b/doc/source/multivariate.rst
@@ -223,6 +223,11 @@ In addition to the methods listed in the common interface above, we also provide
 
     Write the squared Mahalanbobis distances from each column of x to the center of d to r.
 
+.. function:: rand(rng, d)
+              rand(rng, d, n)
+              rand!(rng, d, x)
+
+    Sample from distribution ``d`` using the random number generator ``rng``.
 
 Canonical form
 ~~~~~~~~~~~~~~~
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -55,9 +55,9 @@ end
 _pdf!(r::AbstractArray, d::AbstractMvNormal, x::AbstractMatrix) = exp!(_logpdf!(r, d, x))
 
 # Sampling with designated rng
-rand(rng::AbstractRNG, d::AbstractMvNormal) = _rand!(rng, d, Array{eltype(d)}(length(d)))
+rand(rng::AbstractRNG, d::AbstractMvNormal) = _rand!(rng, d, Vector{eltype(d)}(length(d)))
 rand!(rng::AbstractRNG, d::AbstractMvNormal, x::VecOrMat) = _rand!(rng, d, x)
-rand(rng::AbstractRNG, d::AbstractMvNormal, n::Int64) = _rand!(rng, d, Array{eltype(d)}(length(d), n))
+rand(rng::AbstractRNG, d::AbstractMvNormal, n::Int64) = _rand!(rng, d, Matrix{eltype(d)}(length(d), n))
 
 
 ###########################################################
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -54,6 +54,11 @@ end
 
 _pdf!(r::AbstractArray, d::AbstractMvNormal, x::AbstractMatrix) = exp!(_logpdf!(r, d, x))
 
+# Sampling with designated rng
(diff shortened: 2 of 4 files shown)


rofinn:
Is someone available to review (and maybe merge) this? I've been waiting for the api to support `rand(::AbstractRNG, ::Sampleable)` for a while. @zsunberg I'd be interested in supporting univariate distributions.

src/multivariate/mvnormal.jl
```diff
@@ -54,6 +54,11 @@ end
 
 _pdf!(r::AbstractArray, d::AbstractMvNormal, x::AbstractMatrix) = exp!(_logpdf!(r, d, x))
 
+# Sampling with designated rng
+rand(rng::AbstractRNG, d::AbstractMvNormal) = _rand!(rng, d, Array{eltype(d)}(length(d)))
```
ararslan: I'd use `Vector{eltype(d)}(length(d))` since it's clearer from a cursory glance that it's 1-dimensional, plus it's easier for the compiler, which prefers not to have to determine how many dimensions there are itself.

src/multivariate/mvnormal.jl
```diff
@@ -54,6 +54,11 @@ end
...
+rand!(rng::AbstractRNG, d::AbstractMvNormal, x::VecOrMat) = _rand!(rng, d, x)
+rand(rng::AbstractRNG, d::AbstractMvNormal, n::Int64) = _rand!(rng, d, Array{eltype(d)}(length(d), n))
```
ararslan: Likewise here except `Matrix{eltype(d)}`, since it's 2-dimensional.

ararslan:
Looks okay to me aside from the minor notes above.

rofinn:
@zsunberg It might make sense to update the documentation for `rand` in [multivariate.rst](https://github.com/JuliaStats/Distributions.jl/blob/master/doc/source/multivariate.rst) and it would be nice if the rand methods from Distributions.jl had a docstring (probably not necessary for this PR).

src/multivariate/mvnormal.jl
zsunberg: done

src/multivariate/mvnormal.jl
zsunberg: done

zsunberg:
I updated the docs with the additional rand functions (don't know how to check if the docs will build right though). I put them in the section for special functions for MvNormal since that's the only distribution that works for now, but maybe they should go in the general interface section eventually.

@rofinn, do you want to go ahead and start working on the univariate distributions?

doc/source/multivariate.rst
```diff
@@ -223,6 +223,11 @@ In addition to the methods listed in the common interface above, we also provide
 
     Write the squared Mahalanbobis distances from each column of x to the center of d to r.
 
+.. function:: rand(rng, d)
```
rofinn: I was more just thinking that you could update the existing lines in "Sampling" (line ~90) section of this file to something like:

```
.. function:: rand([rng], d)

    Sample a vector from the distribution ``d``.

.. function:: rand([rng], d, n)

    Sample n vectors from the distribution ``d``. This returns a matrix of size ``(dim(d), n)``, where each column is a sample.

.. function:: rand!([rng], d, x)

    Draw samples and output them to a pre-allocated array x. Here, x can be either a vector of length ``dim(d)`` or a matrix with ``dim(d)`` rows.
```

rofinn:
@zsunberg Running `make html` in the `doc` directory works, but that's probably cause I already have [sphinx](http://www.sphinx-doc.org/en/stable/) installed. 

Yeah, I'm happy to handle the univariate distributions.

zsunberg:
Since #603 was merged, would you mind also merging this? The check that is failing is the OSX nightly, and the failure appears unrelated to the changes.

Thanks!

ararslan:
Sure, this looks okay to me.

zsunberg:
Thanks a bunch!

Issue #598, opened 2017-04-28
Error in CDF of Cosine distribution


Planeshifter:
The CDF of the cosine distribution incorrectly returns values outside of [0,1] for values not inside of its support. For example

``` 
dist = Cosine( 2, 4 );
cdf( dist, 10 );
```

returns `1.5`. 

The current implementation looks as follows:

```
function cdf(d::Cosine, x::Real)
    z = (x - d.μ) / d.σ
    (1 + z + sinpi(z) * invπ) / 2
end
```

To correct this error, the function should be amended to return zero for `x < μ − σ` and one for `x > μ + σ` (see the [Wikipedia article][wikipedia]). The problem applies to `ccdf`, too.

[wikipedia]: https://en.wikipedia.org/wiki/Raised_cosine_distribution

andreasnoack:
Thanks for reporting this. It would be great if you could open a pull request to fix the issue.

Planeshifter:
Created a pull request.

Planeshifter:
Closing this issue as the pull request has been merged. Thank you!

PR #600, opened 2017-05-11
Return correct values for cosine CDF outside of its support

Planeshifter:
This pull request addresses issue #598.


src/univariate/continuous/cosine.jl
```diff
@@ -65,12 +65,24 @@ function logpdf{T<:Real}(d::Cosine{T}, x::Real)
...
+    end
+    if x > d.μ + d.σ 
+        return one(T)
+    end
     z = (x - d.μ) / d.σ
     (1 + z + sinpi(z) * invπ) / 2
 end
 
-function ccdf(d::Cosine, x::Real)
+function ccdf{T<:Real}(d::Cosine, x::Real)
```
ararslan: The type parameter isn't being used here. Did you mean `Cosine{T}`?

src/univariate/continuous/cosine.jl
```diff
@@ -65,12 +65,24 @@ function logpdf{T<:Real}(d::Cosine{T}, x::Real)
...
+    if x < d.μ - d.σ 
+        return one(T)
+    end
+    if x > d.μ + d.σ 
+        return zero(T)
```
ararslan: In order for this function to be type-stable, we have to determine what the return type would be if the value were in the support. Could do this with some `promote` magic.

Planeshifter: My knowledge of Julia is very basic, so please excuse that I can't follow completely. Looking at say the [CDF of the uniform distribution](https://github.com/JuliaStats/Distributions.jl/blob/master/src/univariate/continuous/uniform.jl#L71), I fail to see the difference from the code I wrote. Could you elaborate?

src/univariate/continuous/cosine.jl
Planeshifter: Thank you! Changed.

src/univariate/continuous/cosine.jl
ararslan: "Type stable" means that the type of the value returned from a function does not depend on the particular _value_ passed to it. As a contrived example, `f(x::Int) = x > 1 ? x : 1.0` is not type stable, since the result is `Float64` when the input is less than 2 and `Int` otherwise. This matters because Julia can get a lot of speedup from knowing or being able to predict the type of an object, plus it's less surprising for a user when the return type is consistent across input values of a particular type.

When values of different types interact through some operation, their types undergo "promotion," which means that the values are converted to a common type before the operation is carried out. There's a hierarchy that determines the exact mapping from combinations of types to a common type, but in general it's pretty intuitive. An example of promotion is `1 + 1.0`, which is `Int + Float64 -> Float64`.

In situations such as this, one can use the `promote` function or one of its kin to figure out what the return type should be. That's what I was referring to with my "`promote` magic" comment.

In this particular case I don't think we need to worry to much about getting the return types perfect, since as you pointed out, this style is used for other functions as well. Perhaps someone should eventually go through and try to make these things type stable, but don't worry about that now.

ararslan:
The test system in this package is pretty weird (I admit that I don't fully understand it), but would you be willing to add a test for this? It should just be a matter of adding a line to a JSON file somewhere in the test directory.

Planeshifter:
Added test cases for the `Cosine` distributions in the JSON file. Tests seem to pass when I run the `runtests.jl` file. As you, I don't really understand how unit testing of this package is set up, so hopefully I didn't do anything wrong.

src/univariate/continuous/cosine.jl
Planeshifter: Thank you for this detailed explanation!

ararslan:
Looks great. Thanks!

PR #603, opened 2017-05-13
Added support for passing in a `AbstractRNG` to some of the univariat…

rofinn:
…e `rand` calls.

* Only for univariate distributions which don't rely on RFunctions for their `rand`.
* Doesn't alter the default behaviour (ie: for distributions which don't define their own `rand`)

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -7,9 +7,10 @@ using StatsFuns
 using StatsBase
 using Compat
 
+using Base.Random
+
 import QuadGK.quadgk
 import Compat.view
-import Base.Random
 import Base: size, eltype, length, full, convert, show, getindex, scale!, rand, rand!
 import Base: sum, mean, median, maximum, minimum, quantile, std, var, cov, cor
 import Base: +, -
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -7,15 +7,15 @@ using StatsFuns
 using StatsBase
 using Compat
 
-using Base.Random
-
 import QuadGK.quadgk
 import Compat.view
+import Base.Random
 import Base: size, eltype, length, full, convert, show, getindex, scale!, rand, rand!
 import Base: sum, mean, median, maximum, minimum, quantile, std, var, cov, cor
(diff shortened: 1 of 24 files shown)


src/Distributions.jl
```diff
@@ -7,9 +7,10 @@ using StatsFuns
 using StatsBase
 using Compat
 
+using Base.Random
```
ararslan: Shouldn't this still be `import`, since we're extending `rand` from Base?

rofinn: Yeah, but we're explicitly doing a `import Base: ... rand, rand!` below. Could just add an explicit `import Base.Random: GLOBAL_RNG`.

ararslan:
Nightly macOS failure is unrelated (has to do with the Homebrew-installed GCC updating and Julia having been build on an older GCC).

rofinn:
Out of curiosity, would you prefer that I make separate PRs for the other points in the TODO list?

rofinn:
Is there anything else preventing this from being merged?

rofinn:
I'm not sure if this is the correct place to ask this (or if what I'm asking is advisable), but is there any way I could get access to push an "invenia" branch to Distributions.jl and maybe StatsBase.jl? I'd like to get my colleagues working off of these PRs before they get merged and I think `Pkg.checkout("Distributions", "invenia")` would be the most convenient way to do that.

rofinn:
Would I be able to just join the JuliaStats org, I'd be happy to help monitor repos and provide code review if necessary?
Anyways, doing stuff like 
```julia
const PR = 603  # The PR number on github
info("Checking out pull request (PR) #$PR for Distributions.jl.")
info("Once that #$PR is merged you should run `Pkg.free(\"Distributions\")`.")
cd(Pkg.dir("Distributions")) do
    run(`git fetch origin pull/$PR/head:pull-$PR`)
    run(`git checkout pull-$PR`)
end
```
in the `deps/build.jl` of our internal packages isn't really maintainable.

rofinn:
Awesome, thanks folks!

Issue #601, opened 2017-05-11
Error in CDF and PDF of Lévy distribution 


Planeshifter:
The support of the Levy distribution is the interval `[μ,∞ )`. For input values smaller than `μ`, the CDF currently throws a domain error. It should return zero.

The PDF does not return zero for `x < μ` but throws an error too.

Planeshifter:
Closing this issue as the problem is now resolved.

PR #608, opened 2017-05-16
Fix handling of input values smaller than mu for Levy distribution

Planeshifter:
This pull request addresses issue #601.


ararslan:
Thanks for all your help here. I'm not really familiar with this distribution but the changes look okay to me. I'll request a second set of eyes just in case though.

Issue #602, opened 2017-05-12
Dirichlet initialization for fit diverges when one of the pk is constant across samples


henripal:
When all the Dirichlet samples have one constant `pk`, Dirichlet fit fails because the first guess in the Newton algorithm divides by 0:
```julia
julia > dirichlet_mle_init([0.1 0.1;
                            0.5 0.3;
                            0.4 0.6])
3-element Array{Float64,1}:
Inf
Inf
Inf
```

This is because the first guess described matches the moments of the `pk` using the heurisitic described in [T. Minka's paper](https://tminka.github.io/papers/dirichlet/minka-dirichlet.pdf), which involves a division by `E(pk)^2 - E(pk^2)`.

Proposed solution: The factor `a0` can be estimated independently using any `k`, therefore instead of taking the average of the indepdent estimators, I would suggest taking the median of these independent estimators. This would allow some `pk` to be constant.

Happy to PR and add a test.

PR #610, opened 2017-05-17
Changed mean to median in initial guess for dirichlet MLE

henripal:
Addresses https://github.com/JuliaStats/Distributions.jl/issues/602.
Changed mean of independent estimators to median of independent estimators to account for potential divergence when all `pk` are identical for a given `k`. Added test.

### DIFF
--- a/src/multivariate/dirichlet.jl
+++ b/src/multivariate/dirichlet.jl
@@ -219,14 +219,15 @@ end
 function _dirichlet_mle_init2(μ::Vector{Float64}, γ::Vector{Float64})
     K = length(μ)
 
-    α0 = 0.
+    α = zeros(K) 
+
     for k = 1:K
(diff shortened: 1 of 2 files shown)


ararslan:
I'm not super familiar with the methodology here, but in general this approach seems reasonable. My only concern is that this change will likely be detrimental to performance, since it has to allocate and update an array rather than simply adding to a scalar. Though manually computing the median this way might be too annoying to deal with, so perhaps this is fine.

henripal:
agreed. if performance is an issue I would suggest simply taking the mean of the non divergent estimators rather than taking the median. let me know and I can push this to this to the pr.

matbesancon:
what's the status for this PR? Should mean/median be possibly a choice with default to median in some cases to fix #602? cc @ararslan @andreasnoack

PR #613, opened 2017-05-21
Added rng support to discreteuniform.jl with StatsBase v0.15.0.

rofinn:
Adds support for passing an rng for DiscreteUniform distributions.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,7 +2,7 @@ julia 0.5
 PDMats 0.6.0
 StatsFuns 0.3.1
 Calculus
-StatsBase 0.8.3
+StatsBase 0.15.0
 Compat 0.18.0
 QuadGK 0.1.1
 SpecialFunctions 0.1.0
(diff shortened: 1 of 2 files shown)


ararslan:
Looks good, thanks! It'd be nice if there were a solid way to test this, but I'm not sure how best to go about that. @nalimilan, any ideas?

rofinn:
Well, the new function definition is getting [covered](https://coveralls.io/builds/11617326/source?filename=src%2Funivariate%2Fdiscrete%2Fdiscreteuniform.jl), but it would be nice to have a way of systematically testing the rand methods. Maybe by using a specific rng seed and hard coding expected values?

ararslan:
That seems reasonable, but IIRC seeded random values from the same generator aren't guaranteed to be the same across Julia versions.

rofinn:
This is outside the scope of this PR (possibly even Distributions.jl), but would it make sense to implement a MockRNG type which can guarantee the sequence across julia versions when testing `rand` methods?

ararslan:
Yeah, perhaps a lightweight package that could be used as a test dependency would be good to have eventually. But in the meantime this PR looks reasonable to me.

PR #618, opened 2017-06-05
Fix a typo in matrixvariates.jl

xukai92:


ararslan:
Good catch. We must not have a test that properly exercises that code path. Would you be willing to add one?

xukai92:
Sure. 

Also seem that the `pdf` and `logpdf` of multiple matrix variables are not really vectorized, but using a loop to call the single version instead. Any idea how it can be done in the same manner as univariate and multivariate?

ararslan:
I'm not sure what you mean. Is there a problem or user-facing inconsistency that arises when using a simple loop to apply the functions element-wise that does not occur in the univariate case?

xukai92:
What I mean is that it seems that using a loop is not a real vectorization and it will not really speed-up the code. I believe this is not the problem for univariate and multivariate, where calling `pdf` or `logpdf` with multiple variables a very fast.

Also for the user-end, there IS inconsistency when calling `rand()` for multiple variables, i.e. `rand(dist, n)`:

1. Univariate returns a `Array{T,1}` (i.e. Vector{T})
2. Multivariate returns a `Array{T,2}` (i.e. Matrix{T})
3. Matrixvariate returns a `Array{T,Matrix}`

In order to be consistent, either 2. should be `Array{T,Vector}` or 3. should be `Array{T,3}`.

xukai92:
I've added the tests for calling mutiple variables for matrixvariate distribution

ararslan:
Great, thanks for the tests.

Let's see what @andreasnoack thinks about the `rand` return types.

xukai92:
Cool. Regarding the vectorisation problem, Julia built-in doesn't support vectorisation for matrices in general. I tried functions like `chol`, which also doesn't have interfaces for neither `Array{T,3}` or `Array{Matrix,1}`.

andreasnoack:
> What I mean is that it seems that using a loop is not a real vectorization and it will not really speed-up the code.

I'm still not completely understanding what you are saying here. The univariate distributions also use a loop somewhere. For some of the distributions, the loop might be inside R but it is still a loop. Are you talking about preallocating the memory and then populating it?

I agree that we are inconsistent. Either all three should a `Vector` or we should use `N=1,2,3`. I'm starting to believe that we should split out the non-scalar distributions in a separate package which should then depend on `StaticArrays` and use them to return the variates. I think it could be a nice design and most likely more efficient. At least for the cases where you draw a single variate at a time.

PR #621, opened 2017-06-09
Add LocationScale type. Needs tests/locationscale.jl

NelMuriel:
So, I created this LocationScale type which works for univariate distributions.  The most common functions are already implemented.  I didn't bother adding moment generating functions or characteristic functions, but they would be easy to add.
Arithmetic operations are also defined to make things simpler.  A function "standardize" is given which provides the standard (mean = 0, var = 1) distribution for the family.  For instance standardize(TDist(6)) gives the standard T distribution with 6 degrees of freedom defined as
 (TDist(6) - mean(TDist(6)) / std(TDist(6)).
Documentation is also added.  I don't quite understand what kind of tests are required or how to write those files (even after reading a couple of them...yes, I am rather new to Julia), which is why they are lacking, but as far as I have tested it, it works just fine.

### DIFF
--- a/doc/source/locationscale.rst
+++ b/doc/source/locationscale.rst
@@ -1,0 +1,88 @@
+.. _locationscale:
+
+Location Scale families
+========================
+
+The package provides a type named `LocationScale`, to represent distributions from a location-scale family, which is defined as:
+
+.. code:: julia
+    immutable LocationScale{D<:UnivariateDistribution, S<:ValueSupport} <: UnivariateDistribution{S}
+      base::D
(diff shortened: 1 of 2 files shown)


matbesancon:
hi @NelMuriel, have you looked at #144 and #429 ?
They are trying to achieve the same goal, it could be interesting to compare and possibly to start from the work done there :)

matbesancon:
and #635 which is closed already. I'm closing this one,
@kleinschmidt if you want inspiration for an eventual rewrite

PR #535, opened 2016-09-12
Loosen AbstractMatrix to AbstractVecOrMat in _rand for MvNormalCanon

ararslan:
Fixes #534. I'm not sure this is the right solution, I just kind of wanted to see what happens.

PR #625, opened 2017-06-28
Allow AbstractVector in _rand! for ConjugatePriors.jl

### LINKED ISSUE #534: should MvNormalCanon `_rand!` accept AbstractVecOrMat?
The current implementation of for MvNormalCanon's `_rand!` in v0.4 is:

```
_rand!(d::MvNormalCanon, x::AbstractMatrix) = add!(unwhiten_winv!(d.J, randn!(x)), d.μ)
```

This is causing problems with the test for ConjugatePriors.jl. Can we change it to  `_rand!(d::MvNormalCanon, x::AbstractVecOrMat)`? I think that still works as intended, but someone who knows their stuff should verify.

mschauer:
See https://github.com/JuliaStats/ConjugatePriors.jl/pull/16, ping @ararslan 

Edit: This fixes #534 and is basically a #535 with ambiguity resolved.


mschauer:
Test errors on nightly related to "LoadError: fft! has been moved to the package FFTW.jl" and "testing Distributions.Categorical{Float64}(K=2, p=[0.5, 0.5])"

src/multivariate/mvnormalcanon.jl
```diff
@@ -118,3 +118,5 @@ unwhiten_winv!(J::ScalMat, x::AbstractVecOrMat) = whiten!(J, x)
 
 _rand!(rng::AbstractRNG, d::MvNormalCanon, x::AbstractMatrix) = add!(unwhiten_winv!(d.J, randn!(rng,x)), d.μ)
 _rand!(d::MvNormalCanon, x::AbstractMatrix) = _rand!(Base.GLOBAL_RNG, d, x)
+_rand!(rng::AbstractRNG, d::MvNormalCanon, x::AbstractVector) = add!(unwhiten_winv!(d.J, randn!(rng,x)), d.μ)
```
ararslan: Since these are identical to the matrix methods, I'd just change the signature of the previous methods to accept `AbstractVecOrMat`.

ararslan:
Nice, thanks! Could you add a test for this? Perhaps even something as simple as
```diff
diff --git a/test/mvnormal.jl b/test/mvnormal.jl
index 8a3b33c..b0749df 100644
--- a/test/mvnormal.jl
+++ b/test/mvnormal.jl
@@ -147,6 +147,9 @@ d = MvNormalCanon(Array{Float32}(mu), Array{Float32}(h), PDMat(Array{Float32}(J)
 @test typeof(convert(MvNormalCanon{Float64}, d)) == typeof(MvNormalCanon(mu, h, PDMat(J)))
 @test typeof(convert(MvNormalCanon{Float64}, d.μ, d.h, d.J)) == typeof(MvNormalCanon(mu, h, PDMat(J)))

+d = MvNormalCanon(mu, 1)
+@test insupport(d, rand(d))
+
 ##### MLE

 # a slow but safe way to implement MLE for verification
```
Regarding the FFT error, I'm wondering whether we could just use a slow method for computing the DFT in that case rather than relying on the FFTW package. Not something that has to be handled in this PR, just thinking aloud.

src/multivariate/mvnormalcanon.jl
mschauer: Apparently not - AbstractVecOrMat is less specific and then the method is ambiguous, I made the same mistake when copying @getzdan s two lines,  https://github.com/JuliaStats/Distributions.jl/pull/625/commits/cee75c3692d924471cade1168063ea3426641582

ararslan: Whoa, weird... Alright then, carry on.

mschauer:
Uh, that was a big testing gap...

ararslan:
Made some changes in a separate PR, closed and reopened to incorporate changes in a new CI run. Hopefully this code path will actually be hit on the nightly tests now rather than just erroring out on FFTs.

ararslan:
Looks like you got it. Thanks so much!

PR #627, opened 2017-06-29
Add dependency on FFTW, bump Julia to 0.6

ararslan:
Should get the tests passing on nightly (I hope) since we use FFTs to compute the Poisson binomial PDF that the Base DFT functionality has been removed in 0.7. This requires bumping the required Julia version to 0.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ os:
   - linux
   - osx
 julia:
-  - 0.5
+  - 0.6
   - nightly
 notifications:
   email: false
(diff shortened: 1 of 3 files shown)


tkelman:
except on windows at the moment, but guess appveyor isn't on here?

ararslan:
Yes, no AppVeyor here at the moment, though I guess we should probably turn that on at some point. At any rate, this doesn't make the package any more broken on nightly than it already is on Windows.

PR #637, opened 2017-07-11
Emergency fix to get distributions to work on 0.6

### LINKED ISSUE #631: Precompilation error on Julia v0.6
Currently I am unable to compile Distributions.jl on Julia v0.6:

```julia
julia> using Distributions
INFO: Recompiling stale cache file /home/juliohm/.julia/lib/v0.6/Distributions.ji for module Distributions.
WARNING: The call to compilecache failed to create a usable precompiled cache file for module Distributions. Got:
WARNING: Module FFTW uuid did not match cache file.
ERROR: LoadError: Declaring __precompile__(true) is only allowed in module files being imported.
Stacktrace:
 [1] __precompile__(::Bool) at ./loading.jl:335
 [2] include_from_node1(::String) at ./loading.jl:569
 [3] eval(::Module, ::Any) at ./boot.jl:235
 [4] _require(::Symbol) at ./loading.jl:483
 [5] require(::Symbol) at ./loading.jl:398
while loading /home/juliohm/.julia/v0.6/Distributions/src/Distributions.jl, in expression starting on line 1
```

Similar issue has happened with `Combinatorics.jl`: https://github.com/JuliaMath/Combinatorics.jl/issues/48

What exactly is happening? How can I fix this myself?

aviks:
This reverts parts of #627, to remove the dependency on FFTW, which is currently breaking most usages of Distributions on Julia v0.6. 

I would advocate merging this, tagging, then putting the dependency on `FFTW` back, and tagging a 0.7 only release. 

I release there is no consensus on this approach right now, but I'm putting this out there at the very least for users to patch their systems locally -- use this branch, and delete `FFTW` and `AbstractFFT` from `~/.julia/v0.6/` if you are struggling with precompilation issues around Distributions and FFTW. 

Fix #631

cc: @ararslan @stefankarpinski

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -6,4 +6,3 @@ StatsBase 0.17.0
 Compat 0.18.0
 QuadGK 0.1.1
 SpecialFunctions 0.1.0
-FFTW
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -8,7 +8,6 @@ using StatsBase
 using Compat
 
 import QuadGK.quadgk
-import FFTW.fft!
 import Compat.view
 import Base.Random
 import Base: size, eltype, length, full, convert, show, getindex, scale!, rand, rand!


simonbyrne:
I think it's more important to be working on 0.6 than 0.7 at this point.

ChrisRackauckas:
Thanks for implementing this!

KristofferC:
So now put back using the FFTW package, tag a new 0.7-dev+ only version, and adopt a backport type of workflow until 0.7 is released?

ararslan:
My plan is to make a fix in FFTW then just revert this and retag.

ChrisRackauckas:
That's one way that would work. Another way which would require less effort is to just expect that users on the nightly are comfortable with git, so keep a v0.7 branch which (for now) only adds the FFTW package and tell them to check that out. Then it would be easy to keep fast-forwarding that to master, and making v0.7-specific changes as well, but it would not require backporting (and it would keep CI running on v0.6 for master).

>My plan is to make a fix in FFTW then just revert this and retag.

Or that.

PR #640, opened 2017-07-14
Documenter

rofinn:
Moved docs over to Documenter.jl which mostly involved converting the restructured text to markdown and moving most of the docs into docstrings. Unfortunately, I needed to create some extra methods for the general docstrings, but I'm open to suggestions.

Fixes #629

PR #641, opened 2017-07-15
Use a simple DFT algorithm when FFTW is not in Base

ararslan:
Avoids having to introduce a dependency on FFTW, which has binary dependencies, just for this one distribution. Distributions is already a pretty heavyweight dependency for packages; FFTW (once it actually works) will only make that worse. Thus I'm proposing this simple, standalone DFT implementation for the Poisson binomial PDF.

### DIFF
--- a/src/univariate/discrete/poissonbinomial.jl
+++ b/src/univariate/discrete/poissonbinomial.jl
@@ -149,8 +149,8 @@ end
 if VERSION >= v"0.7.0-DEV.602"
     function _dft(x::Vector{T}) where T
         n = length(x)
-        y = fill!(copy(x), zero(complex(float(T))))
-        @inbounds for j = 0:n-1, i = 0:n-1
+        y = Vector{complex(float(T))}(n)
+        @inbounds for j = 0:n-1, k = 0:n-1
             y[k+1] += x[j+1] * cis(-2π * j * k / n)
         end
         return y
--- a/src/univariate/discrete/poissonbinomial.jl
+++ b/src/univariate/discrete/poissonbinomial.jl
@@ -141,8 +141,22 @@ function poissonbinomial_pdf_fft(p::AbstractArray)
             x[n + 1 - l + 1] = conj(x[l + 1])
         end
     end
-    fft!(x)
-    [max(0, real(xi)) for xi in x]
+    [max(0, real(xi)) for xi in _dft(x)]
+end
+
+# A simple implementation of a DFT to avoid introducing a dependency
+# on an external FFT package just for this one distribution
+if VERSION >= v"0.7.0-DEV.602"
+    function _dft(x::Vector{T}) where T
+        n = length(x)
+        y = fill!(copy(x), zero(complex(float(T))))
+        @inbounds for j = 0:n-1, i = 0:n-1
+            y[k+1] += x[j+1] * cis(-2π * j * k / n)
+        end
+        return y
+    end
+else
+    _dft(x) = Base.fft(x)
 end
 
 #### Sampling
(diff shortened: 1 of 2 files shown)


ararslan:
The nightly failures are due to unescaped backslashes in docstrings, which will hopefully be fixed by #640.

src/univariate/discrete/poissonbinomial.jl
```diff
@@ -141,8 +141,22 @@ function poissonbinomial_pdf_fft(p::AbstractArray)
...
-    [max(0, real(xi)) for xi in x]
+    [max(0, real(xi)) for xi in _dft(x)]
+end
+
+# A simple implementation of a DFT to avoid introducing a dependency
+# on an external FFT package just for this one distribution
+if VERSION >= v"0.7.0-DEV.602"
+    function _dft(x::Vector{T}) where T
+        n = length(x)
+        y = fill!(copy(x), zero(complex(float(T))))
```
andreasnoack: Please look at this one again. I'm pretty sure it is not doing what you intended it to do. Furthermore, there is no reason to initialize the memory since that is done in the loop below.

src/univariate/discrete/poissonbinomial.jl
```diff
@@ -141,8 +141,22 @@ function poissonbinomial_pdf_fft(p::AbstractArray)
...
+        @inbounds for j = 0:n-1, i = 0:n-1
+            y[k+1] += x[j+1] * cis(-2π * j * k / n)
```
andreasnoack: `k` should be `i` or vice versa.

ararslan:
> Also, `_dft` -> `_dft!`?

Calling this `_dft!` would be misleading since it doesn't compute the FFT in-place on the input; it allocates a second array and modifies that. I'm honestly not familiar enough with FFT algorithms to implement anything other than the not-in-place N<sup>2</sup> algorithm. I figured that we could put something simple in place and improve it as needed.

Also the things that Andreas noted are really embarrassing. 😐 Sorry about that...

andreasnoack:
Btw. I looked at the `pdf` for the Poisson-Binomial and the `n^2` might be better suited here in any case because I think that `n` typically will be quite small.

jmxpearson:
@ararslan Ah, I see. My mistake. I got confused about it being a drop-in for `fft!`. Will double-check everything later tonight.

ararslan:
Looks like this isn't retaining sufficient accuracy for the Poisson binomial PDF tests to pass. In particular:
```
Expression: isapprox(pdf(d, k), m, atol=1.0e-15)
Evaluated: isapprox(1.0373646386341306e-15, 3.099087406079243e-76; atol=1.0e-15)
```
We could widen the tolerance for the test or find a better algorithm...

jmxpearson:
If the tolerance is widened, would it be helpful to issue a warning (at least until this workaround is unnecesary)?

ararslan:
The workaround is more or less permanent; the FFTW bindings (and the FFT functionality as a whole) were removed from Base in Julia 0.7 and this replaces the uses of it with a simple FFT implementation.

jmxpearson:
I see. I had it the wrong way around.

In the failed test, your `_dft` must be losing precision somehow, right?
Does the test pass robustly for `fft` with different values/seeds, or is
this one of those tests that fails some percentage of the time if we're not
careful?

andreasnoack:
I believe that tests here are deterministic. It loops through a fixed set of parameter values and compares the `pdf` with a slower more precise version. I printed all the values for the previous FFTW based version and it produced values around `0.0` and `eps()` while the new one is up to `8eps()` so neither of them have any relative accuracy in the tail. If we really care about the relative accuracy, we should probably change algorithm completely to the version used in the tests but I'm not sure accuracy at that level is required.

jmxpearson:
Then maybe lower test tolerance as suggested and make a note about limited
tail accuracy in the docs?

ararslan:
Good idea. Let's merge #640 first so that the note in the docs can be applied in the right place.

PR #643, opened 2017-07-17
Fix precision in _dft by reducing the magnitude of the argument in cis

andreasnoack:
Also, fix initialization of the `y` vector. (Sorry, Alex)


ararslan:
If you check out this branch then merge Rory's Documenter branch from the Invenia remote into it and `Pkg.test` locally, does it work on nightly?

ararslan:
Meh, whatever. It can't make nightly any more broken than it already is.

PR #644, opened 2017-07-18
Fixes for changes in nightly

mschauer:
Triggered by JuliaLang/julia#22800

### DIFF
--- a/src/univariate/continuous/noncentralchisq.jl
+++ b/src/univariate/continuous/noncentralchisq.jl
@@ -1,4 +1,4 @@
-"""
+doc"""
     NoncentralChisq(ν, λ)
 
 The *noncentral chi-squared distribution* with `ν` degrees of freedom and noncentrality parameter `λ` has the probability density function
--- a/test/vonmisesfisher.jl
+++ b/test/vonmisesfisher.jl
@@ -2,6 +2,7 @@
 
 using Distributions
 using Base.Test
+using SpecialFunctions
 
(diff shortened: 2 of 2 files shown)


mschauer:
I added fix related to besseli moving out of base.

src/univariate/continuous/noncentralchisq.jl
```diff
@@ -1,4 +1,4 @@
-"""
+doc"""
```
ararslan: This shouldn't be necessary. Were you seeing failures without it?

mschauer: Otherwise one can escape the backslashes, I followed "doc_str should only be used when the docstring contains $ or \ characters that should not be parsed by Julia such as LaTeX syntax or Julia source code examples containing interpolation" from https://docs.julialang.org/en/latest/manual/documentation/

ararslan: Ah, I see, thanks. #640 escapes all of the backslashes.

test/vonmisesfisher.jl
```diff
@@ -2,6 +2,7 @@
 
 using Distributions
 using Base.Test
+using SpecialFunctions
```
ararslan: I'm actually not sure how the tests work at all 0.6+ without this.

src/univariate/continuous/noncentralchisq.jl
rofinn: `doc_str` doesn't play nice with documenter as a result #640 intentionally removes most of the `doc_str`s in the code and escapes everything accordingly.

PR #654, opened 2017-08-17
Ensure that vectorized version of pdf etc work with OffsetArrays

### LINKED ISSUE #653: `pdf`, `cdf`, etc. do not work with `OffsetArrays`
For univariate distributions, the functions `pdf`, `cdf`, etc. do not work with offset arrays. Here is an example illustrating this:

~~~
using OffsetArrays
using Distributions

const N = 10 
const L = 5 

const range = linspace(-L, L, 2N+1)

const A = collect(range)
const B = OffsetArray(range, -N:N)

W = Normal(0,1)

X = cdf(W, A)
Y = cdf(W, B)
~~~

I get the following error:

~~~
ERROR: LoadError: size not supported for arrays with indices (-10:10,); see http://docs.julialang.org/en/latest/devdocs/offset-arrays/
~~~

The linked page says that the use of `size` should be replaced by `indices` for the code to work with general arrays.

adityam:
The changes are based on the suggestions provided at https://docs.julialang.org/en/latest/devdocs/offset-arrays/

Closes #653

### DIFF
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -465,7 +465,7 @@ for fun in [:pdf, :logpdf,
 
     @eval begin
         function ($_fun!)(r::AbstractArray, d::UnivariateDistribution, X::AbstractArray)
-            for i in linearindices(X)
+            for i in eachindex(X)
                 r[i] = ($fun)(d, X[i])
             end
             return r
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -465,20 +465,20 @@ for fun in [:pdf, :logpdf,
 
     @eval begin
         function ($_fun!)(r::AbstractArray, d::UnivariateDistribution, X::AbstractArray)
-            for i in 1 : length(X)
+            for i in linearindices(X)
                 r[i] = ($fun)(d, X[i])
             end
             return r
         end
 
         function ($fun!)(r::AbstractArray, d::UnivariateDistribution, X::AbstractArray)
-            length(r) == length(X) ||
+            indices(r) == indices(X) ||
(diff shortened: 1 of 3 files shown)


src/univariates.jl
```diff
@@ -465,20 +465,20 @@ for fun in [:pdf, :logpdf,
 
     @eval begin
         function ($_fun!)(r::AbstractArray, d::UnivariateDistribution, X::AbstractArray)
-            for i in 1 : length(X)
+            for i in linearindices(X)
```
andreasnoack: I think it is better to use `eachindex` here.

adityam: I was simply following the advice at https://docs.julialang.org/en/latest/devdocs/offset-arrays/ which says

> replace `1:length(A)` with `linearindices(A)`

andreasnoack: I see. However, I still think `eachindex` will be slightly better in this case. In most cases, it will return the same as `linearindices` but in a few cases it will return something better.

simonbyrne:
Shouldn't we just deprecate all these methods in favour of broadcasting?

andreasnoack:
Yes, you are right. These should be deprecated instead.

andreasnoack:
@adityam Sorry that we didn't realize this right away. We have generally deprecated automatic vectorization so the right way forward here is to call `cdf.(W, B)` and it already works now.

simonbyrne:
I'm putting together a PR, and will add these as tests.

andreasnoack:
Superseded by #655

PR #655, opened 2017-08-17
Deprecate vectorized methods for univariate distributions

simonbyrne:
This keeps the old behaviour for multivariate distributions, as I don't have any great ideas for those.

Still need to overload `Base.broadcast` for fast `UnitRange` methods (e.g. `Binomial`).

### DIFF
--- a/src/univariate/discrete/betabinomial.jl
+++ b/src/univariate/discrete/betabinomial.jl
@@ -81,8 +81,9 @@ function kurtosis(d::BetaBinomial)
     return (left * right) - 3
 end
 
-function pdf(d::BetaBinomial, k::Int)
+function pdf(d::BetaBinomial{T}, k::Int) where T
     n, α, β = d.n, d.α, d.β
+    0 <= k <= n || return zero(T)
     chooseinv = (n + 1) * beta(k + 1, n - k + 1)
     numerator = beta(k + α, n - k + β)
     denominator = beta(α, β)
--- a/src/univariate/discrete/betabinomial.jl
+++ b/src/univariate/discrete/betabinomial.jl
@@ -92,7 +92,7 @@ end
 
 entropy(d::BetaBinomial) = entropy(Categorical(pdf.(d,support(d))))
 median(d::BetaBinomial) = median(Categorical(pdf.(d,support(d)))) - 1
-mode(d::BetaBinomial{T}) = indmax(pdf.(d,support(d))) - 1
+mode(d::BetaBinomial) = indmax(pdf.(d,support(d))) - 1
 modes(d::BetaBinomial) = modes(Categorical(pdf.(d,support(d)))) .- 1
 
 quantile(d::BetaBinomial, p::Float64) = quantile(Categorical(pdf.(d, support(d))), p) - 1
--- a/src/univariate/discrete/betabinomial.jl
+++ b/src/univariate/discrete/betabinomial.jl
@@ -83,7 +83,7 @@ end
 
 function pdf(d::BetaBinomial{T}, k::Int) where T
     n, α, β = d.n, d.α, d.β
-    0 <= k <= n || return zero(T)
+    insupport(d, k) || return zero(T)
     chooseinv = (n + 1) * beta(k + 1, n - k + 1)
     numerator = beta(k + α, n - k + β)
     denominator = beta(α, β)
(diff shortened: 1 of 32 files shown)


src/univariate/discrete/binomial.jl
```diff
@@ -118,7 +118,9 @@ end
 RecursiveBinomProbEvaluator(d::Binomial) = RecursiveBinomProbEvaluator(d.n, d.p / (1 - d.p))
 nextpdf(s::RecursiveBinomProbEvaluator, pv::Float64, x::Integer) = ((s.n - x + 1) / x) * s.coef * pv
 
-function _pdf!(r::AbstractArray, d::Binomial, X::UnitRange)
+function Base.broadcast!(::typeof(pdf), r::AbstractArray, d::Binomial, X::UnitRange)
+    indices(r) == indices(X) || throw(ArgumentError("Inconsistent array dimensions."))
```
ararslan: Perhaps `DimensionMismatch` would be more appropriate than `ArgumentError`?

src/univariate/discrete/betabinomial.jl
```diff
@@ -81,8 +81,9 @@ function kurtosis(d::BetaBinomial)
     return (left * right) - 3
 end
 
-function pdf(d::BetaBinomial, k::Int)
+function pdf(d::BetaBinomial{T}, k::Int) where T
     n, α, β = d.n, d.α, d.β
+    0 <= k <= n || return zero(T)
```
ararslan: Perhaps `insupport`?

simonbyrne: I always feel that the arguments to `insupport` are the wrong way around.

src/univariates.jl
```diff
@@ -551,8 +522,8 @@ function _pdf!(r::AbstractArray, d::DiscreteUnivariateDistribution, X::UnitRange
     return r
 end
 
-
-pdf(d::DiscreteUnivariateDistribution) = isbounded(d) ? pdf(d, minimum(d):maximum(d)) :
+# TODO: should we get rid of this?
```
ararslan: Makes sense to me to get rid of it, maybe deprecate it for `pdf.(d, support(d))`?

src/univariate/discrete/betabinomial.jl
ararslan: Oh really? The current order makes sense to me. I think of it like `contains(haystack, needle)` from Base.

src/univariates.jl
simonbyrne: ok

ararslan: If the user calls it on a distribution that doesn't have finite support, that's on them. 😛

src/univariate/discrete/betabinomial.jl
simonbyrne: well, it's the opposite of `in` for example.

src/univariate/discrete/betabinomial.jl
```diff
@@ -81,29 +81,18 @@ function kurtosis(d::BetaBinomial)
...
-    return values
-end
-
-entropy(d::BetaBinomial) = entropy(Categorical(pdf(d)))
-median(d::BetaBinomial) = median(Categorical(pdf(d))) - 1
-mode(d::BetaBinomial{T}) where {T<:Real} = indmax(pdf(d)) - one(T)
-modes(d::BetaBinomial) = [x - 1 for x in modes(Categorical(pdf(d)))]
+entropy(d::BetaBinomial) = entropy(Categorical(pdf.(d,support(d))))
+median(d::BetaBinomial) = median(Categorical(pdf.(d,support(d)))) - 1
+mode(d::BetaBinomial{T}) = indmax(pdf.(d,support(d))) - 1
```
rofinn: Forgot the `where {T}`

simonbyrne: Yes, not really necessary.

ararslan: I'm confused by that response. This should either have `where {T<:Real}` or (even better) have `BetaBinomial{T}` changed to `BetaBinomial`.

simonbyrne:
Okay, this should be good to go.

src/univariate/discrete/betabinomial.jl
simonbyrne: ah, I missed that. Good point.

Issue #628, opened 2017-06-29
Make testing infrastructure less weird


ararslan:
The current JSON-based testing infrastructure in this package is rather bizarre, at least in comparison to the standard unit testing done in most other Julia packages. It almost seems like a barrier to entry for new features, since figuring out how to test any given feature is very non-obvious. It would be great if we could migrate to a standard, obvious format for testing.

PR #660, opened 2017-08-30
make quantile function of truncated distributions accept more than just Float64 arguments

jleugeri:
I have code that somewhere calls `quantile(d, x)` for a user-specified distribution `d` and a value `x` that happens to be `Float32`. At the moment this code works for general univariate distributions but it breaks when `d` is a truncated distribution, as its quantile requires `Float64`. Not sure if enforcing `Float64` is required for any specific reason, but since neither the `quantile` function for non-truncated distributions nor other functions (`pdf` etc.) for truncated distributions are as restrictive, I think the `quantile` function of truncated distributions shouldn't be, either.

Side remark: the whole definition of `Truncated` distributions is hard coded to use `Float64`:
```julia
struct Truncated{D<:UnivariateDistribution, S<:ValueSupport} <: UnivariateDistribution{S}
    untruncated::D      # the original distribution (untruncated)
    lower::Float64      # lower bound
    upper::Float64      # upper bound
    lcdf::Float64       # cdf of lower bound
    ucdf::Float64       # cdf of upper bound

    tp::Float64         # the probability of the truncated part, i.e. ucdf - lcdf
    logtp::Float64      # log(tp), i.e. log(ucdf - lcdf)
end
```
Is this desirable?

PS: have mercy with me if this is a pointless issue, it's my first PR :relaxed:

### DIFF
--- a/src/truncate.jl
+++ b/src/truncate.jl
@@ -47,7 +47,7 @@ insupport(d::Truncated{D,Union{Discrete,Continuous}}, x::Real) where {D<:Univari
 
 ### evaluation
 
-quantile(d::Truncated, p::Real) = quantile(d.untruncated, d.lcdf + p * d.tp)
+quantile(d::Truncated, p::T) where {T<:Real} = quantile(d.untruncated, d.lcdf + p * d.tp)
 
 for f in [:pdf, :logpdf, :cdf, :logcdf, :ccdf, :logccdf]
     @eval ($f)(d::Truncated, x::Int) = ($f)(d, float(x))
--- a/src/truncate.jl
+++ b/src/truncate.jl
@@ -47,7 +47,7 @@ insupport(d::Truncated{D,Union{Discrete,Continuous}}, x::Real) where {D<:Univari
 
 ### evaluation
 
-quantile(d::Truncated, p::T) where {T<:Real} = quantile(d.untruncated, d.lcdf + p * d.tp)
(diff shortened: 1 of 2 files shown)


ararslan:
Hey @jleugeri, thanks for the contribution and nice first PR! I think this is definitely a desirable change; it seems like an oversight that the signature is currently so limited. Just two notes here:

* Since `T` isn't used in the body of the function, it can be omitted and the argument can instead be annotated with `::Real`. In general, `where` is only needed when `T` is used inside the function.

* Could you add a test for this behavior? It can just be something simple, like calling `quantile` on a truncated distribution with an integer, `Float32`, or anything other than `Float64`.

That should do it I think. Let me know if you have any questions or anything like that and I'll be happy to help!

jleugeri:
Allright, I'm a bit confused about how the tests are organized, but I added some simple tests.
I didn't wrap them in a `try-catch` block like the other tests are, because a `MethodError` here is actually indicative of the problem we're testing for (`quantile` not being defined for some argument types).

Btw @ararslan you're of course right about the type.  I'm still rather new to Julia and routinely confuse when to use `::T where T<:Real`, `<:Real` and `::Real`. Thanks for the nice words ;)

ararslan:
> I'm a bit confused about how the tests are organized

Don't worry, you're _definitely_ [not alone in that](https://github.com/JuliaStats/Distributions.jl/issues/628). How you've set up your tests here look good to me.

ararslan:
Thanks again for the contribution!

PR #662, opened 2017-09-02
Fix EmpiricalUnivariateDistribution

juliohm:
This PR should fix the `quantile` and `rand` of `EmpiricalUnivariateDistribution`. I have also changed the distribution to be discrete instead of continuous because it relies on the `ecdf` from `StatsBase.jl`, which doesn't perform interpolation.

Could you please review the changes and see if I got all the corner cases right? I am bisecting the CDF values in order to find the quantile, I think this is the best we can do?

### DIFF
--- a/src/empirical.jl
+++ b/src/empirical.jl
@@ -4,33 +4,33 @@
 #
 ##############################################################################
 
-struct EmpiricalUnivariateDistribution <: ContinuousUnivariateDistribution
-	values::Vector{Float64}
-    support::Vector{Float64}
-	cdf::Function
-	entropy::Float64
-	kurtosis::Float64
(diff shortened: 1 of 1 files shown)


andreasnoack:
Remeber to check for open PRs, see #661

juliohm:
Oh, didn't see that, I told you in Discourse that I would submit a PR, that is why I didn't even check.

I like that you are using quantile(d.values, p) instead of reimplementing the bisection, I will close this one.

andreasnoack:
> Oh, didn't see that, I told you in Discourse that I would submit a PR

No worries. I started on this yesterday right after I realized how broken the existing version was.

PR #664, opened 2017-09-15
Fix most depwarns and errors on 0.7

yuyichao:


test/vonmisesfisher.jl
```diff
@@ -76,6 +76,7 @@ for (p, κ) in [(2, 1.0),
                (3, 1.0),
                (3, 5.0),
                (5, 2.0)]
+    local p
```
yuyichao: Some of these might not be necessary anymore after I wrapped each file in a module from the test runner. They shouldn't hurt though...

test/locationscale.jl
```diff
@@ -44,13 +44,13 @@ function test_location_scale_normal(μ::Float64,σ::Float64,μD::Float64,σD::Fl
...
     @test pdf(d,0.1) ≈ pdf(dref,0.1)
-    @test pdf(d,2:4) ≈ pdf(dref,2:4)
+    @test pdf.(d,2:4) ≈ pdf.(dref,2:4)
     @test logpdf(d,0.4) ≈ logpdf(dref,0.4)
     @test loglikelihood(d,[0.1,0.2,0.3]) ≈ loglikelihood(dref,[0.1,0.2,0.3])
     @test cdf(d,μ-0.4) ≈ cdf(dref,μ-0.4)
     @test logcdf(d,μ-0.4) ≈ logcdf(dref,μ-0.4)
     @test ccdf(d,μ-0.4) ≈ ccdf(dref,μ-0.4) atol=1e-100
-    @test logccdf(d,μ-0.4) ≈ logccdf(dref,μ-0.4) atol=1e-20
+    @test logccdf(d,μ-0.4) ≈ logccdf(dref,μ-0.4) atol=1e-16
```
yuyichao: AFAICT the large change of the tolerance is needed on 0.7 because this was ignored on 0.6. The value I get on both 0.6 and 0.7 are `~ -0.34` and the `eps` there is `~1e-17`. An absolute tolerance of `1e-20` here is basically the same as exactly equal.

src/Distributions.jl
```diff
@@ -21,6 +21,13 @@ if isdefined(Base, :scale)
     import Base: scale
 end
 
+if isdefined(Base, :MathConstants)
+    import Base.MathConstants: γ
+else
+    # TODO replace with Compat
+    const ℯ = e
```
yuyichao: I'm collecting things that should go in `Compat`. Meanwhile, this package uses `e` enough times that I think it's worth putting a temperary solution in place.

There's one occurance in the test that I didn't replace since it won't affect package users.

ararslan: When Base.MathConstants is defined, ASCII `e` is defined therein, so you could do the same thing I did a couple lines up with `γ`. That is,
```julia
if isdefined(Base, :MathConstants)
    import Base.MathConstants: γ, e
end
```
When MathConstants isn't defined, `e` will be defined in Main since it's exported from Base.

yuyichao: It's done this way to match the new default behavior.

src/univariate/continuous/frechet.jl
```diff
@@ -103,7 +103,7 @@ function kurtosis(d::Frechet{T}) where T<:Real
 end
 
 function entropy(d::Frechet)
-    const γ = 0.57721566490153286060  # γ is the Euler-Mascheroni constant
+    γ = 0.57721566490153286060  # γ is the Euler-Mascheroni constant
```
nalimilan: Isn't this exactly `Base.MathConstants.γ`?

test/vonmisesfisher.jl
nalimilan: Always better not have unnecessary instructions, when reading the code one will wonder why these are present.

yuyichao: They are just not necessary to suppress warnings. They shouldn't be confusing since they give the loop 1.0 semantics.

yuyichao:
Ping.

nalimilan:
Looks good to me, but why do all tests fail?

andreasnoack:
The problem is #671 and there is *a* fix in #674 but I'm not sure why @simonbyrne changed the call in #655

PR #665, opened 2017-09-19
add Wigner semicircle distribution

bicycle1885:
A simple continuous univariate distribution: https://en.wikipedia.org/wiki/Wigner_semicircle_distribution.

I think it needs more methods and tests, but I'd like to get your quick check before going further.

You will see this distribution in [Wigner's semicircle law](http://mathworld.wolfram.com/WignersSemicircleLaw.html).  Here is a simulation code:
```julia
using Distributions
using PyPlot

function symmat(d, n)
    X = Matrix{Float64}(n, n)
    for i in 1:n, j in i:n
        X[i,j] = X[j,i] = rand(d)
    end
    return X
end

N = 1000
bins = 100
fig, ax = subplots(3, sharex=true, sharey=true)
srand(1234)
for (i, d) in enumerate([Normal(), Uniform(-1, 1), DiscreteUniform(-1, 1)])
    X = symmat(d, N) ./ sqrt(N)
    Λ = eigvals(X)
    ax[i][:hist](Λ, bins=bins, label="eigen values")
    semicircle = Semicircle(2 * std(d))
    lo, hi = minimum(semicircle), maximum(semicircle)
    xs = linspace(lo, hi)
    ax[i][:plot](xs, pdf.(semicircle, xs) * (N * (hi - lo) / bins), label="pdf")
    ax[i][:set_title](repr(d))
    ax[i][:legend](loc=1)
end
fig[:tight_layout]()
fig[:savefig]("semicircle.png")
```

semicircle

### DIFF
--- a/docs/src/univariate.md
+++ b/docs/src/univariate.md
@@ -132,6 +132,7 @@ NormalCanon
 NormalInverseGaussian
 Pareto
 Rayleigh
+Semicircle
 SymTriangularDist
 TDist
 TriangularDist
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -129,6 +129,7 @@ export
     PoissonBinomial,
     QQPair,
     Rayleigh,
+    Semicircle,
     Skellam,
     SymTriangularDist,
     TDist,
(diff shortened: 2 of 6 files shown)


andreasnoack:
Needs a rebase and I also think you'd need to adjust https://github.com/JuliaStats/Distributions.jl/blob/master/docs/src/univariate.md

bicycle1885:
I think this is ready to review. The CI may fail due to #671, though.

andreasnoack:
Thanks

PR #670, opened 2017-09-26
Fixed NaN for Multinomial degenerate cases + tests.

ludkinm:
There is a bug when one of the parameters of a Multinomial is 0.
This seems to be a `0*log(0) ` issue, although under the definition of a Multinomial distribution, such a parameterization is legitimate. 
The provided tests succeed, whereas under master these give NaN in all cases.

Following [this pull request](https://github.com/JuliaStats/Distributions.jl/pull/457) if a variable is outside of the support it has pdf 0.


src/multivariate/multinomial.jl
```diff
@@ -140,21 +140,20 @@ function insupport(d::Multinomial, x::AbstractVector{T}) where T<:Real
     return s == ntrials(d)  # integer computation would not yield truncation errors
 end
 
-function _logpdf(d::Multinomial, x::AbstractVector{T}) where T<:Real
+function _logpdf{T<:Real}(d::Multinomial, x::AbstractVector{T})
```
ararslan: This change is incorrect, it should still be `where`

src/multivariate/multinomial.jl
```diff
@@ -140,21 +140,20 @@ function insupport(d::Multinomial, x::AbstractVector{T}) where T<:Real
...
     for i = 1:length(p)
         @inbounds xi = x[i]
         @inbounds p_i = p[i]
-        t += xi
-        s -= R(lgamma(xi + 1))
-        @inbounds s += xi * log(p_i)
-    end
-    return ifelse(t == n, s, -R(Inf))
+        s -= R(lgamma(R(xi) + 1))
+        @inbounds s += ifelse((xi==0) & (p_i==0), 0.0, xi * log(p_i)) # log(0^0)=0 not NaN
```
ararslan: Do you see a performance improvement with `@inbounds` here? Doesn't look like there's any indexing happening in this step, so it shouldn't be necessary. But if you want you could move the `@inbounds` annotation from in front of the variable definitions to right before `for`.

ararslan: Also for type stability should this be `zero(T)` rather than the `Float64` literal 0.0?

ararslan:
Hey @ludkinm, thanks for the contribution and for catching this bug!

src/multivariate/multinomial.jl
ludkinm: Thanks.

src/multivariate/multinomial.jl
ludkinm: 1) I just edited the maths leaving the macros as in master.
2) you are right, edited for commit.

src/multivariate/multinomial.jl
```diff
@@ -151,7 +151,7 @@ function _logpdf{T<:Real}(d::Multinomial, x::AbstractVector{T})
         @inbounds xi = x[i]
         @inbounds p_i = p[i]
         s -= R(lgamma(R(xi) + 1))
-        @inbounds s += ifelse((xi==0) & (p_i==0), 0.0, xi * log(p_i)) # log(0^0)=0 not NaN
+        s += ifelse((xi== zero(T)) & (p_i == zero(T)), 0.0, xi * log(p_i)) # log(0^0)=0 not NaN
```
ararslan: Oh sorry, I wasn't clear, I meant
```julia
s += ifelse((xi == 0) & (p_i == 0), zero(T), xi * log(p_i))
```
It's fine to use `zero` in the comparison, but adding `zero(T)` rather than 0.0, which is always `Float64`, avoids any unnecessary type promotion when summing.

src/multivariate/multinomial.jl
```diff
@@ -145,16 +145,15 @@ function _logpdf(d::Multinomial, x::AbstractVector{T}) where T<:Real
...
     for i = 1:length(p)
         @inbounds xi = x[i]
         @inbounds p_i = p[i]
-        t += xi
-        s -= R(lgamma(xi + 1))
-        @inbounds s += xi * log(p_i)
-    end
-    return ifelse(t == n, s, -R(Inf))
+        s -= R(lgamma(R(xi) + 1))
+        s += ifelse((xi== zero(T)) & (p_i == zero(T)), 0.0, xi * log(p_i)) # log(0^0)=0 not NaN
```
jmxpearson: Agree with @ararslan here.

ludkinm:
Thanks for the positive reviews. 
I think this should be merged into release v0.14.2 instead of master?

test/multinomial.jl
```diff
@@ -64,6 +64,18 @@ end
...
 @test typeof(logpdf(convert(Multinomial{Float32}, d), x1)) == Float32
 
+# test degenerate cases of logpdf
+d1 = Multinomial(1, [0.5, 0.5, 0.0])
+d2 = Multinomial(0, [0.5, 0.5, 0.0])
+x2 = [1, 0, 0]
+x3 = [0, 0, 1]
+x4 = [1, 0, 1]
+
+@test logpdf(d1, x2) == log(0.5)
```
ararslan: For float equality it's better to check `@test logpdf(d1, x2) ≈ log(0.5)`. That ensures there's a sensible tolerance. (In case you aren't familiar, `≈`, which can be completed at the REPL as `\approx<tab>`, calls the function `isapprox`.)

ararslan:
> I think this should be merged into release v0.14.2 instead of master?

Once a release is made, its content is set in stone. Since this addresses an existing bug, we can tag a new patch release (i.e. v0.14.3) once this is merged. We always tag releases from the master branch, so PRs should target master.

ludkinm:
Ok that makes sense. I've updated the test in the latest commit.

src/multivariate/multinomial.jl
```diff
@@ -145,16 +145,15 @@ function _logpdf(d::Multinomial, x::AbstractVector{T}) where T<:Real
...
     for i = 1:length(p)
         @inbounds xi = x[i]
         @inbounds p_i = p[i]
-        t += xi
-        s -= R(lgamma(xi + 1))
-        @inbounds s += xi * log(p_i)
-    end
-    return ifelse(t == n, s, -R(Inf))
+        s -= R(lgamma(R(xi) + 1))
+        s += ifelse((xi == 0) & (p_i == 0), zero(T), xi * log(p_i))
```
andreasnoack: Use `xlogy(xi, p_i)` instead. Then it should be good to go.

src/multivariate/multinomial.jl
```diff
@@ -145,16 +145,15 @@ function _logpdf(d::Multinomial, x::AbstractVector{T}) where T<:Real
...
     for i = 1:length(p)
         @inbounds xi = x[i]
         @inbounds p_i = p[i]
-        t += xi
-        s -= R(lgamma(xi + 1))
-        @inbounds s += xi * log(p_i)
-    end
-    return ifelse(t == n, s, -R(Inf))
+        s -= R(lgamma(R(xi) + 1))
+        s += ifelse((xi == 0) & (p_i == 0), zero(T), xlogy(xi, p_i))
```
andreasnoack: Sorry for not being clear here. `xlogy` is doing all the work here so you should just have `s += xlogy(xi, p_i)`. It is made exactly for this purpose.

PR #673, opened 2017-10-08
change Base.Test to Test

xiaodaigh:
Base.Test is deprecated in favour of Test

### DIFF
--- a/src/testutils.jl
+++ b/src/testutils.jl
@@ -1,7 +1,10 @@
 # Utilities to support the testing of distributions and samplers
 
-import Test: @test
-
+if VERSION >= v"0.7.0-DEV"
+    import Test: @test
+else
+    import Base.Test: @test
+end
 # auxiliary functions
 
 # to workaround issues of Base.linspace
--- a/src/testutils.jl
+++ b/src/testutils.jl
@@ -1,6 +1,6 @@
 # Utilities to support the testing of distributions and samplers
 
-import Base.Test: @test
+import Test: @test
 
 # auxiliary functions
 
--- a/test/binomial.jl
+++ b/test/binomial.jl
@@ -1,7 +1,11 @@
 using Distributions
-using Test
+if VERSION >= v"0.7.0-DEV"
+    using Test
+else
+    using Base.Test
+end
 
-# Test the consistency between the recursive and nonrecursive computation of the pdf 
+# Test the consistency between the recursive and nonrecursive computation of the pdf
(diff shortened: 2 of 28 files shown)


ararslan:
Thanks for the contribution! The Test module is only part of the standard library rather than Base in recent 0.7 and this package still supports 0.6, so we can't just replace Base.Test with Test yet. Instead we can do Compat.Test, or wrap it in a version conditional.

xiaodaigh:
What about capping the release version for 0.6 to the lasf release?

ararslan:
I don't think it's worth dropping 0.6 support at this point, especially since most of the 0.7 changes have Compat support.

xiaodaigh:
Ok I updated it to check for version

src/testutils.jl
```diff
@@ -1,7 +1,10 @@
 # Utilities to support the testing of distributions and samplers
 
-import Test: @test
-
+if VERSION >= v"0.7.0-DEV"
```
ararslan: This check isn't quite specific enough, since earlier versions of 0.7 still had Base.Test. I'd recommend just doing
```julia
using Compat
using Compat.Test
```
in each of the relevant files. This requires adding an entry for Compat in test/REQUIRE if no such entry exists, and the minimum version should be 0.34.0 (I think).

xiaodaigh: I put forward another view in that if people are using 0.7.0-DEV there is no reason to be using earlier versions of it as it's purely a nightly build version. They should just switch.

ararslan: It's not feasible for everyone to upgrade often. Making the check more specific is as easy as replacing `v"0.7.0-DEV"` in what you have here with `v"0.7.0-DEV.2004"`.

nalimilan: Sorry, but the well-established rule is that we should either find the specific commit which introduced the change, or use `isdefined` to check for the change directly.  This is useful e.g. if one needs to test an earlier development version of Julia to check for a regression.

Anyway, when there's support in Compat, it should be used instead of explicit `VERSION` checks.

oschulz:
@ararslan What's the status with this? Could we do a fix similar to JuliaStats/PDMats.jl#67 soon, and tag Distributions? This currently breaks testing of all packages that require Distributions on Julia v0.7 (due to "Distributions.jl/src/testutils.jl").

oschulz:
Here's a Compat-based version: #676

andreasnoack:
Superseded by #676

PR #676, opened 2017-10-22
Use Compat.Test instead of Base.Test for Julia v0.7 compatibility

oschulz:
See #673 and JuliaStats/PDMats.jl#67.


oschulz:
Test failure seems unrelated to changes in this PR.

andreasnoack:
Nightly still fails but now because of `DiffRules`

oschulz:
Thanks for the quick merge and the new tag!

PR #684, opened 2017-11-15
Stop exporting `pmf`.

hessammehr:
The `pmf` function is exported in `Distributions.jl` but not defined anywhere. If I understand correctly, there is a move towards using `pdf` for discrete as well as continuous distributions and this name should no longer be exported.

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -199,8 +199,6 @@ export
     loglikelihood,      # log probability of array of IID draws
     logpdf,             # log probability density
     logpdf!,            # evaluate log pdf to provided storage
-    logpmf,             # log probability mass
-    logpmf!,            # evaluate log pmf to provided storage
 
     invscale,           # Inverse scale parameter
     sqmahal,            # squared Mahalanobis distance to Gaussian center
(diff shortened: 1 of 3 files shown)


nalimilan:
There's also `logpmf`/`logpmf!`. The three functions are deprecated in deprecated.jl, so you could remove their exports too.

hessammehr:
Thanks @nalimilan!

kmsquire:
`functionals.jl` has a use of `pmf()` ([here](https://github.com/JuliaStats/Distributions.jl/blob/master/src/functionals.jl#L16)) that needs to be changed to `pdf()`.

A test for that function would also be useful--if there was one, the tests would have broken when the deprecation was removed.

kmsquire:
CI is failing on nightly because of a problem with DiffRules:

https://travis-ci.org/JuliaStats/Distributions.jl/jobs/303650559#L1610

Reported [here](https://github.com/JuliaDiff/DiffRules.jl/issues/6).

CI passes on Julia v0.6.  This can be merged.

kmsquire:
Thanks, @hessammehr!

(I explicitly did not squash these commits, as they seem logically separate to me.)

Issue #690, opened 2017-12-05
Mean and variance of truncated normal far from the mode fails


cossio:
```
julia> d = Distributions.TruncatedNormal(0,1, 100, 115)
Truncated(Distributions.Normal{Float64}(μ=0.0, σ=1.0), range=(100.0, 115.0))

julia> mean(d)
NaN
```

andreasnoack:
I don't really see how that could be different unless you special cased the distribution in that area. If the calculation is based on the untruncated version then you'll need some probability mass in the included region. If not I think `NaN` is a reasonable result. What we should do is to parameterize `Truncated` on the element type in https://github.com/JuliaStats/Distributions.jl/blob/master/src/truncate.jl such that you could do the calculation in `BigFloat`s.

cossio:
I think it can be done in ordinary Floats, see https://github.com/cossio/TruncatedNormal.jl. I am testing it a bit more first, then I'll try to PR here.

cossio:
@andreasnoack I have setup a PR, https://github.com/JuliaStats/Distributions.jl/pull/691.

PR #691, opened 2017-12-06
fix mean and var of trunc normal

cossio:
The previous implementation had trouble when the truncation ocurred away from the mode of the distribution. This commit fixes that.

Fixes https://github.com/JuliaStats/Distributions.jl/issues/690.

See https://github.com/cossio/TruncatedNormal.jl/blob/master/notes/normal.pdf for explanation.

Test values were obtained with arbitrary precision in Mathematica.

### DIFF
--- a/src/truncated/normal.jl
+++ b/src/truncated/normal.jl
@@ -26,30 +26,103 @@ end
 
 modes(d::Truncated{Normal{T},Continuous}) where {T <: Real} = [mode(d)]
 
+# do not export. Used in mean, var
+function _F1(x::Real, y::Real; thresh=1e-7)
+    @assert 0 < thresh < Inf
+    -Inf < x < Inf && -Inf < y < Inf || throw(DomainError())
+    ϵ = exp(x^2 - y^2)
+    if abs(x) > abs(y)
+        _F1(y,x)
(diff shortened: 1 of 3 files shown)


andreasnoack:
Could you please rebase this one?

cossio:
@andreasnoack How do I do that?

andreasnoack:
Basically, from your local branch you do
```
git fetch origin
git rebase origin/master
```
and there will be some conflict which you'll have to fix. Git will tell you and you can also do `git status` get see which files still have conflicts. When all conflicts in a file have been fixed you'll do a `git add FILENAME` and then when all files have been fixed, you'll do a `git rebase --continue`. If you are unlucky, there'll then be another round of fixed but most often you'll be done. See also https://git-scm.com/docs/git-rebase. Finally, you'll `git push --force` to your remote branch and the PR will be updated.

cossio:
I rebased. I guess this can be merged now. I will prepare another PR later to handle semi-infinite truncation limits. Now it relies on the original implementation if one of the truncation limits is infinite.

cossio:
I think the CI errors are not from my changes.

andreasnoack:
Thanks. The package hasn't been made ready for 0.7 yet so failures there are expected.

PR #693, opened 2017-12-11
Negative scale in location scale

bdeonovic:
Scale should be allowed to be negative in location scale family. 

e.g. Lets say I have a random variable X= -10X + 3 where X is gamma(2,2).  This should be 

```julia
LocationScale(3.0, -10.0, Gamma(2.0, 2.0))
```

Currently throws argument error


ararslan:
Looks okay to me. Could you add a test or two that would have caught this?

bdeonovic:
Actually looking over the location scale wiki it says the scale should be non-negative, which follows convention used in rest of Distributions.jl.

Issue #681, opened 2017-11-04
Support type of Discrete distributions


GordStephen:
Right now `Discrete` distributions are assumed to have supports limited to `Int` values. This makes sense if the discrete distribution's support is assumed to serve as a proxy for some other set of non-numeric values, but the reworking of `EmpiricalUnivariateDistribution` (#661) and the proposed new `Generic` distribution (#634) both provide good examples of discrete distributions with numeric but potentially non-integer supports.

As far as I can tell this assumption is primarily relevant for preallocating arrays for random draws from samplers. Could we just have sampler implementations provide their own `eltype` method instead of inferring the support type from whether the distribution is continuous or discrete?

GordStephen:
Any thoughts on this, one way or the other? If people are receptive to the idea I'm happy to submit a PR.

kmsquire:
Not very familiar with that code, but instead of having each sampler provide its own `eltype`, I would instead consider having a method, `supporttype(...)`, which defaults to `Int` for `Discrete` distributions, `Float64` (or whatever is appropriate) for `Continuous` distributions, and override that method for individual distributions whenever that is not true.  Samplers could then use this method to determine the support type for the given distribution.

GordStephen:
I agree, a default implementation with the ability to add a specialized method on concrete distributions makes the most sense. The new `supporttype` method shouldn't be required though, since it's essentially what `eltype` already does.

The existing code is actually very close to this, the only change required would be to remove the explicit coupling between continuous/discrete distributions and Float64/Int support since while those mappings are the defaults, a particular distribution is free to define their own `eltype` method. I'll submit a PR.

matbesancon:
@GordStephen I understand this one could be closed?

GordStephen:
Maybe, although there are still assumptions about discrete supports being integer and representable by a `UnitRange` floating around in the codebase. This probably won't be fully addressed until the type hierarchy is reworked.

matbesancon:
oh ok I thought #634 fixed it but I understand catching it all can be tricky. We'll close it and re-open if needed

PR #699, opened 2017-12-24
RFC: Officially allow/document overriding distribution support type

GordStephen:
Removes strict coupling between continuous/discrete distributions and support types, as discussed in #681.

No actual new code here, just an update to the the docs to flag the assumptions about support types of different distributions and the ability to override the defaults. The `eltype(::Type{<:ValueSupport})` methods are unused and wouldn't be accurate anyways in the case of an override, so they're removed.

This sets the stage for allowing discrete non-integer-support distributions (e.g. #634 and #661).

### DIFF
--- a/docs/src/extends.md
+++ b/docs/src/extends.md
@@ -4,6 +4,8 @@ Whereas this package already provides a large collection of common distributions
 
 Generally, you don't have to implement every API method listed in the documentation. This package provides a series of generic functions that turn a small number of internal methods into user-end API methods. What you need to do is to implement this small set of internal methods for your distributions.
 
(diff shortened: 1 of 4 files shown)


GordStephen:
Any concerns with this?

GordStephen:
Bumping this again. I'd like to keep moving forward with #634 but this is a blocker for it.

andreasnoack:
Some of the fallbacks assume that the outcomes are in `N_0` which was where I stalled when looking into fixing the categorical. Have you thought about how to handle that?

GordStephen:
What do you mean by `N_0`?

andreasnoack:
The natural numbers and zero

søn. 21. jan. 2018 kl. 00.15 skrev Gord Stephen <notifications@github.com>:

matbesancon:
It would fix different problems to relax this assumption and allow moving on with things like #661

matbesancon:
@GordStephen this branch is still on Julia 0.6, could you rebase on master?

GordStephen:
Ok, the branch is now rebased.

matbesancon:
Thanks, merging this in 24 hours if no objection.

matbesancon:
thanks @GordStephen

PR #710, opened 2018-04-01
cleanup some outdated syntax

bjarthur:
manually ran femtocleaner, and manually fixed some dot broadcasts.


bjarthur:
tests pass on 0.6, fail on nightly.  good enough, right?

PR #711, opened 2018-04-03
fixup 8ffb14d4

bjarthur:
only two of those six periods were needed.  i should really field test more thoroughly before submitting a PR.

what unit test could i add to make sure this doesn't happen again?  i don't see QQ being tested at all, and am not sure what it really does.


bjarthur:
tests added

tlnagy:
I was about to fix this. Thanks!

Issue #708, opened 2018-03-16
Import implementations of `norm*pdf` and `norm*cdf` from StatsFuns.


dmbates:
See https://github.com/JuliaStats/StatsFuns.jl/issues/42 for background.

Copy the implementations of the `norm*pdf` and `norm*cdf` functions from https://github.com/JuliaStats/StatsFuns.jl/blob/master/src/distrs/norm.jl to `src/univariate/continuous/normal.jl` and remove the call to `@_delegate_statsfuns` in that file.

Also relates to https://github.com/JuliaStats/StatsFuns.jl/issues/37

PR #714, opened 2018-04-10
Native pdf, cdf, quantile, and related functions for the Normal distribution

Nosferican:
Porting the native implementation from StatsFuns to Distributions. See #708

### DIFF
--- a/src/univariate/continuous/normal.jl
+++ b/src/univariate/continuous/normal.jl
@@ -76,12 +76,12 @@ zval(μ::Real, σ::Real, x::Number) = (x - μ) / σ
 # pdf
 normpdf(z::Real) = exp(-abs2(z)/2) * invsqrt2π
 normpdf(μ::Real, σ::Real, x::Real) = normpdf(zval(μ, σ, x)) / σ
-pdf(d::Normal, x::Real) = normpdf(d.μ, d.σ, x) / d.σ
+pdf(d::Normal, x::Real) = normpdf(d.μ, d.σ, x)
 
 # logpdf
 normlogpdf(z::Real) = -(abs2(z) + log2π)/2
-normlogpdf(μ::Real, σ::Real, x::Real) = -(abs2(zval(μ, σ, x)) + log2π)/2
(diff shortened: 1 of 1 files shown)


src/univariate/continuous/normal.jl
```diff
@@ -68,9 +68,157 @@ kurtosis(d::Normal{T}) where {T<:Real} = zero(T)
...
 
 
-#### Evaluation
+#### Evaluation (see JuliaStats/Distributions.jl/issues/708)
+
+xval(μ::Real, σ::Real, z::Number) = μ + σ * z
+zval(μ::Real, σ::Real, x::Number) = (x - μ) / σ
+
+# pdf
+normpdf(z::Real) = exp(-abs2(z)/2) * invsqrt2π
```
andreasnoack: I don't think we should implement the `normX` versions.

matbesancon:
@Nosferican updates on this?

matbesancon:
ping @Nosferican

src/univariate/continuous/normal.jl
andreasnoack: I think we should just define these for the `Distributions` API, i.e. this method would simply be `pdf(::Normal, z)`

simonbyrne: It could be useful to have a separate "basic" functions. i.e. `normpdf(z)`, but we don't need `normpdf(x,mu,sigma)` versions.

similarly, `normcdf`, but you don't need `normccdf` since this is trivial.

Nosferican:
I know I have been lagging, but soon I will have time to go back to this. I am defending in two weeks so probably check with me a bit after that. If anyone else wants to take a go in the meantime, feel free.

matbesancon:
Hi @Nosferican, congrats on the defense :)
Up-ing on this

matbesancon:
also, there might be some merge to do first, and addressing the comments

Nosferican:
Thanks! Seems like the branch I used for the PR was deleted sometime within the last year... I will close this PR and open a fresh one with the comments addressed.

PR #720, opened 2018-05-09
[WIP] Add drop in replacements for Rmath functions

jayvn:
Currently exporting : 
`qnorm`,  `pnorm`,  `rnorm`.


andreasnoack:
I'm pretty sure we don't want to support the R API here. People should generally prefer the existing API provided by `Distributions`, i.e. `pdf`, `quantile`, `cdf`, and `norm`.

jayvn:
OK, not the whole R API, but I thought it'd be a lot more convenient to just do things like `qnorm(.05)` rather than `cdf(Normal(), .05)` especially when writing long equations. I agree that the latter is more readable.

andreasnoack:
I think providing such an API should be a task for a different package if really needed.

jayvn:
Understood. Thanks for the response.

matbesancon:
Closing this as discussed

PR #724, opened 2018-05-31
Safer log-pmf (logpdf) for BetaBinomial distribution

freeboson:
Specialize `logpdf(::UnivariateDistribution,...)` for `BetaBinomial`. Doing `log(pdf(::BetaBinomial))` will not work if `α`, `β` are large, and Poisson approximation is to be avoided.

It makes use of the fact that `binomial(n,k) = 1/((n+1) * beta(k+1, n-k+1)`.


freeboson:
If https://github.com/JuliaMath/SpecialFunctions.jl/pull/99 gets merged, I could rewrite this to use `lbinomial` directly. I guess it would need to have the correct version required.

matbesancon:
closed for #774

PR #725, opened 2018-06-05
Fix v0.7 deprecations

gragusa:


src/multivariate/mvnormal.jl
```diff
@@ -259,12 +259,12 @@ logdetcov(d::MvNormal) = logdet(d.Σ)
 
 ### Evaluation
 
-sqmahal(d::MvNormal, x::AbstractVector) = invquad(d.Σ, x .- d.μ)
+sqmahal(d::MvNormal, x::AbstractVector) = invquad(d.Σ, broadcast(-, x, d.μ))
```
dmbates: Is there a reason to prefer the explicit call to `broadcast`?  I thought the previous version was equivalent to this version.

src/utils.jl
```diff
@@ -138,10 +138,10 @@ end
 
 # because isposdef keeps giving the wrong answer for samples
 # from Wishart and InverseWisharts
-hasCholesky(a::Matrix{Float64}) = isa(trycholfact(a), Cholesky)
+hasCholesky(a::Matrix{Float64}) = isa(trycholesky(a), Cholesky)
 
 function trycholfact(a::Matrix{Float64})
```
dmbates: Don't you need to change the name `trycholfact` to `trycholesky` here too if you are going to change it in line 141?

test/REQUIRE
```diff
@@ -3,3 +3,4 @@ ForwardDiff
 Calculus
 PDMats
 SpecialFunctions
+StatsBase
```
dmbates: Needs a newline at the end.

src/multivariate/mvnormal.jl
gragusa: Well, I thought so as well. However, 

```
g = MvNormal(3, sqrt(2.0))
typeof(g.μ)
```
gives
```
Distributions.ZeroVector{Float64}
```
And
```
[0.0, 0.0, 0.0] .- g.μ
```
gives
```
3-element Array{Array{Float64,1},1}:
 [0.0, 0.0, 0.0]
 [0.0, 0.0, 0.0]
 [0.0, 0.0, 0.0]
```
with the following warning:
```
┌ Warning: `x::Number - v::ZeroVector` is deprecated, use `broadcast(-, x, v)` instead.
│   caller = _broadcast_getindex at broadcast.jl:587 [inlined]
```

src/utils.jl
gragusa: Yeap! Note: a test for `trycholesky` should be added.

gragusa:
CI failure are due to time limit for jobs.

gragusa:
Another time limit failure.

src/multivariate/mvnormal.jl
andreasnoack: Weird. Not sure why that would be the case. However, why would this be a broadcast in the first place? Shouldn't it just be `-`?

gragusa: Yes. You right. Changing `.-` with `i` suffices

gragusa: P.S.: Not really....

```
@code_lowered [0.0, 0.0, 0.0] - g.μ
```

```
CodeInfo(:(begin
      nothing
      $(Expr(:meta, :noinline))
      #= line 54 =#
      Core.SSAValue(0) = Core.Typeof
      Core.SSAValue(1) = (Core.SSAValue(0))(Distributions.:-)
      Core.SSAValue(2) = (Base.getproperty)(Core.SSAValue(1), :name)
      Core.SSAValue(3) = (Base.getproperty)(Core.SSAValue(2), :mt)
      Core.SSAValue(4) = (Base.getproperty)(Core.SSAValue(3), :name)
      (Base.depwarn)("`x::AbstractArray - v::ZeroVector` is deprecated, use `broadcast(-, x, v)` instead.", Core.SSAValue(4))
      #= line 55 =#
      Core.SSAValue(5) = (Distributions.broadcast)(Distributions.:-, x, v)
      return Core.SSAValue(5)
  end))
```

andreasnoack: Where is this deprecation coming from? I don't see it on master
```julia
julia> [0.0, 0.0, 0.0] - Distributions.ZeroVector{Float64}(4)
3-element Array{Float64,1}:
 0.0
 0.0
 0.0
```

gragusa:
```julia> VERSION
v"0.7.0-alpha.31"
```

gragusa: I get it with this morning master:

```
               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: https://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.7.0-alpha.40 (2018-06-06 03:00 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit 3a0c6e20e3 (0 days old master)
|__/                   |  x86_64-apple-darwin16.7.0

julia> using Distributions

julia> g = MvNormal(3, sqrt(2.0))
ZeroMeanIsoNormal(
dim: 3
μ: [0.0, 0.0, 0.0]
Σ: [2.0 0.0 0.0; 0.0 2.0 0.0; 0.0 0.0 2.0]
)


julia> [0.0, 0.0, 0.0] - g.μ
┌ Warning: `x::AbstractArray - v::ZeroVector` is deprecated, use `broadcast(-, x, v)` instead.
│   caller = top-level scope
└ @ Core :0
3-element Array{Float64,1}:
 0.0
 0.0
 0.0

julia> VERSION
v"0.7.0-alpha.40"
```

andreasnoack: Hm. I don't get the deprecation warning on `0.7.0-alpha.39`. Could you print the output of `@which [0.0, 0.0, 0.0] - Distributions.ZeroVector{Float64}(4)`?

andreasnoack: Ha. It should error, though, since the lengths aren't equal but that is a different matter.

andreasnoack:
We might need to fix all the upstream deprecations first to avoid the timeout. They add a lot of overhead.

src/multivariate/mvnormal.jl
gragusa: Mystery:

```
julia> @which [0.0, 0.0, 0.0] - g.μ
-(x::AbstractArray, v::Distributions.ZeroVector) in Distributions at deprecated.jl:53
```

There is no `deprecated.jl`

andreasnoack: Ah. They were introduced by @ararslan in https://github.com/JuliaStats/Distributions.jl/commit/58cafb75fce067f94413d925a27dfd000e02f6a1. We should add something like right after the deprecations such that it is only showing the deprecation for arrays with `N!=1`.
```julia
function Base.:-(x::AbstractVector{T}, y::Distributions.ZeroVector{S}) where {T,S}
    if length(x) != length(y)
        throw(DimensionMismatch("dimensions must match"))
    end
    TS = promote_type(T, S)
    return convert(AbstractArray{TS}, x)
end
```

gragusa: Right after? I think the deprecations should be removed and the code changed to:

```
for T = (:AbstractArray, :Number), op = (:+, :-)
    @eval begin
        function Base.$op(x::AbstractVector{T}, y::Distributions.ZeroVector{S}) where {T,S}
            if length(x) != length(y)
                throw(DimensionMismatch("dimensions must match"))
            end
            TS = promote_type(T, S)
            return convert(AbstractArray{TS}, x)
        end        
    end
end
```

simonbyrne:
Included as part of #729

PR #706, opened 2018-03-09
Updates for 0.7

ararslan:
Supersedes #700, which is stalled on unmerged Compat changes. Fixes #697.

PR #729, opened 2018-07-03
More 0.7 updates

simonbyrne:
Includes #706 and #725.


src/Distributions.jl
```diff
@@ -2,33 +2,36 @@ __precompile__(true)
...
+
+using Random
+import Random: GLOBAL_RNG, RangeGenerator, rand!, SamplerRangeInt
+
+if VERSION >= v"0.7.0-beta.85"
+    import Statistics: mean, median, quantile, std, var, cov, cor
+elseif VERSION >= v"0.7.0-DEV.5238"
+    import StatsBase: mean, median, quantile, std, var, cov, cor
+else
+    import Base: mean, median, quantile, std, var, cov, cor
```
ararslan: Not a big deal since CI doesn't fail on it here but there's some trailing whitespace here

simonbyrne:
Okay the tests are now passing on nightlies, but I can't figure out what is going on with the beta.

nalimilan:
Looks like a Pkg3 bug or an issue in METADATA. Probably worth asking on Slack.

simonbyrne:
I've tried it locally which gave me a bit more info. The problem is that the stdlib Statistics package is now in deps file for StatsBase.jl: https://github.com/JuliaRegistries/Uncurated/blob/master/S/StatsBase/Deps.toml#L37-L38
(Statistics was added to stdlib after the beta).

Basically this means that any package which depends on StatsBase can't be used on the beta.

nalimilan:
So it's even incorrect since we claim to require only `julia 0.7.0-alpha`. I guess there's no way around this, since Manifest needs to mention `Statistics` to be able to use it? Or maybe adding `@static` to `if VERSION >= v"0.7.0-beta.85"` would help?

dmbates:
@simonbyrne The release tag should be "v0.16.0", not "0.16.0".  attobot is complaining.

simonbyrne:
I of all people should know that since I get all AttoBot's emails...

PR #733, opened 2018-07-09
Suppress warnings when testing on 0.7

dmbates:


dmbates:
Apparently on Win32 the wrong value of the kurtosis for `Hypergeometric(60, 80, 100)` is returned, which is causing the AppVeyor failure.  It seems that recent versions were not being tested on AppVeyor, perhaps for this reason.

andreasnoack:
It looks like https://github.com/JuliaStats/Distributions.jl/blob/117894c129420b7243a1e2d10ad82dc4bb376df1/src/univariate/discrete/hypergeometric.jl#L63-L65 needs some overflow protection. We could probably just convert `N` to a float right away. I'm curious why this wasn't caused test failure on 32 bit Linux.

ararslan:
> I'm curious why this hasn't caused test failure on 32 bit Linux.

Like most packages, we're only testing on x86-64 Linux here.

andreasnoack:
Simon had already fixed the issue in #732

PR #739, opened 2018-07-17
WIP: cholesky accessors

### LINKED ISSUE #731: Accessor for `mvnorm.Σ.chol[:U]`?
I'm not sure if this is common enough to have a method, but it'd be nice if extracting the `PDMat.chol` components from a `MvNormal` distribution was part of the public API. Maybe `LinAlg.cholesky(d) = d.Σ.chol`?

rofinn:
- Adds `cholesky` methods to distributions using an `AbstractPDMat` covariance matrix
- Depends on JuliaStats/PDMats.jl#81 getting in
- Closes #731

### DIFF
--- a/src/matrix/inversewishart.jl
+++ b/src/matrix/inversewishart.jl
@@ -47,6 +47,7 @@ dim(d::InverseWishart) = dim(d.Ψ)
 size(d::InverseWishart) = (p = dim(d); (p, p))
 params(d::InverseWishart) = (d.df, d.Ψ, d.c0)
 @inline partype(d::InverseWishart{T}) where {T<:Real} = T
+LinearAlgebra.cholesky(d::InverseWishart) = cholesky(d.Ψ)
 
 ### Conversion
 function convert(::Type{InverseWishart{T}}, d::InverseWishart) where T<:Real
--- a/src/matrix/wishart.jl
+++ b/src/matrix/wishart.jl
@@ -51,6 +51,7 @@ dim(d::Wishart) = dim(d.S)
 size(d::Wishart) = (p = dim(d); (p, p))
 params(d::Wishart) = (d.df, d.S, d.c0)
 @inline partype(d::Wishart{T}) where {T<:Real} = T
+LinearAlgebra.cholesky(d::Wishart) = cholesky(d.S)
 
 ### Conversion
 function convert(::Type{Wishart{T}}, d::Wishart) where T<:Real
--- a/src/multivariate/mvlognormal.jl
+++ b/src/multivariate/mvlognormal.jl
@@ -186,6 +186,7 @@ end
 length(d::MvLogNormal) = length(d.normal)
 params(d::MvLogNormal) = params(d.normal)
 @inline partype(d::MvLogNormal{T}) where {T<:Real} = T
+LinearAlgebra.cholesky(d::MvLogNormal) = cholesky(d.normal)
 
 """
     location(d::MvLogNormal)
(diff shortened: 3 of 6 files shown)


simonbyrne:
While I initially thought this was a good idea, I'm not such a fan now: `cholesky` of a distribution is a bit of a weird concept, especially in the case of an inverse Wishart (it could also mean the `cholesky` of a sample from that distribution).

rofinn:
Hmmm, fair point, any other recommendations then? My objective is simply to have a more public interface for extracting `d.Σ.chol` from a `MvNormal` distribution without going through multiple levels of raw fields. I suppose deprecating fields should be easier in 0.7 with `getproperty`, but I'm not sure if that's something I can rely on.

simonbyrne:
we should make `cholesky(cov(d::MvNormal))` work, if it doesn't already.

rofinn:
Yes, that works, but it's pretty wasteful. `cholesky(cov(d::MvNormal))` will rerun `cholesky` on `Matrix(d.Σ)`.

matbesancon:
Not sure either `cholesky(::Distribution)` is that intuitive

rofinn:
> Is there a reason that having choleksy(d.Σ) isn't enough?

Mostly, just that I'd prefer to have a proper interface for accessing `Σ`, rather than accessing internal fields directly. I suppose this is less of a concern now that we can deprecate field access with `getproperty` on 0.7/1.0.

kleinschmidt:
Isn't that what `params` is for though? https://github.com/JuliaStats/Distributions.jl/blob/228138ac88fbfd855bc2b26a783b134f3293a86c/src/multivariate/mvnormal.jl#L252

rofinn:
That would also work, but it also seemed like selecting the second element from a tuple returned from `params` wasn't best practice either. What if the ordering changes?

ararslan:
A bit tangential, but perhaps we could return `NamedTuple`s from `params`. (That would actually be non-breaking, since `NamedTuple`s behave like `Tuple`s in most cases.)

simonbyrne:
That might be easier with #768.

The performance penalty could be related to https://github.com/JuliaLang/julia/issues/29083.

ararslan:
Checking on master built yesterday, seems the performance penalty persists:
```
julia> struct AB
           a::Float64
           b::Float64
       end

julia> @btime AB((1.0, 2.0)...)
  0.040 ns (0 allocations: 0 bytes)
AB(1.0, 2.0)

julia> @btime AB((a=1.0, b=2.0)...)
  174.420 ns (4 allocations: 176 bytes)
AB(1.0, 2.0)
```

simonbyrne:
Interesting: it is now faster to splat keywords:
```julia
julia> using KeywordDispatch, BenchmarkTools

julia> struct AB
              a::Float64
              b::Float64
          end

julia> @kwdispatch AB()
AB

julia> @kwmethod AB(;a,b) = AB(a,b)

julia> @btime AB(1.0,2.0)
  0.026 ns (0 allocations: 0 bytes)
AB(1.0, 2.0)

julia> @btime AB((a=1.0,b=2.0)...)
  176.709 ns (4 allocations: 176 bytes)
AB(1.0, 2.0)

julia> @btime AB(;(a=1.0,b=2.0)...)
  1.844 ns (0 allocations: 0 bytes)
AB(1.0, 2.0)
```

matbesancon:
should we close this and accept that `params` is the correct way to access the data to pass on to `choleksy`?

rofinn:
I think I'd almost prefer to keep it as a field access since the deprecation cycle could just involve overloading `getproperty`. Feel free to close this though.

mschauer:
One could say that `choleksy(d.Σ)` is the `scale`

Issue #530, opened 2016-09-03
Creating new distributions from old


johnczito:
This issue has a similar motivation to #482. 

It would be nice if their existed a function that, given a distribution instance and parameters, returned a new distribution instance of the same type as the old instance, but with the new parameters. I'm thinking in the spirit of the `similar` function for Arrays.

The following works well enough:
```
julia> d1 = MvNormal([1.0, 2.0], eye(2))
FullNormal(
dim: 2
μ: [1.0,2.0]
Σ: [1.0 0.0; 0.0 1.0]
)

julia> d2 = typeof(d1)([2.0, 3.0], params(d1)[2])
FullNormal(
dim: 2
μ: [2.0,3.0]
Σ: [1.0 0.0; 0.0 1.0]
)
```
But it allows you to do things like this:
```
julia> d3 = typeof(d1)([2.0, 3.0, 4.0], params(d1)[2])
FullNormal(
dim: 3
μ: [2.0,3.0,4.0]
Σ: [1.0 0.0; 0.0 1.0]
)
```

ArnavSood:
Hi, this is a very old issue, but I wrote an implementation of this which I think does the job. Here is the code: 

```
# Similarity function, per issue #530 (and by extension, issue #482)
function similar(dist, params...)
    distType = (typeof(dist).name).wrapper
    return distType(params...)
end 
```

And here is the results: 
```
julia> similar(d1, params(d1)...)
FullNormal(
dim: 2
μ: [1.0, 2.0]
Σ: [1.0 0.0; 0.0 1.0]
)


julia> similar(d1, ([2.0, 3.0, 4.0], params(d1)[2])...)
ERROR: DimensionMismatch("The dimensions of mu and Sigma are inconsistent.")
Stacktrace:
 [1] Type at C:\Users\Arnav Sood\.julia\dev\Distributions\src\multivariate\mvnormal.jl:194 [inlined]
 [2] similar(::MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}}, ::Array{Float64,1}, ::PDMats.PDMat{Float64,Array{Float64,2}}) at .\REPL[54]:3
 [3] top-level scope at none:0
```

Issue #482, opened 2016-04-08
Immutability of distributions


scidom:
In many real applications, it becomes crucial to be able to tweak a distribution. For example, in various Monte Carlo methods (Gibbs with conditional distributions representing block updates, Metropolis-Hastings proposal), it is needed to retain the same type of distribution and only update its parameters. A very concrete example would be to update the mean `μ` of a ` d = Normal()`. It would be more reasonable to retain the same distribution and only do `d.μ = 1.2`, instead of doing `d = Normal(1.2)`. The problem with the second solution is that one needs to reallocate the distribution at each iteration - so imagine 100,000 reallocations of a normal if one has 100,000 MCMC iterations.

I thought two workarounds; we can simply turn the distributions in Distributions from `immutable` to `type` or, if you feel that this is not a good idea, I could simply define a new package with mutable distributions, say `MutableDistributions.jl`. I would favor the former, in order to avoid populating Julia's package ecosystem with multiple distribution-centric packages, but if you feel that turning distributions to mutable types is a big compromise, I am happy to set up a separate package with mutable distributions.

KristofferC:
These reallocation would most likely be optimized away by the compiler. I would guess that you would see a performance losses changing to a type because they would not be stored in line and the compiler can do fewer assumptions.

scidom:
Interesting. I tried the following test from the REPL (after enclosing the test code in functions), and it turns out that you might as well be right. The first run showed that mutating the mean of a mutable normal is faster, but then subsequent runs show that reallocating immutable normals is at least as fast, perhaps after LLVM's magic kicks in:

```
julia> using Distributions

julia> type MutableNormal{T <: Real}
         μ::T
         σ::T
       end

julia> function set_immutable_pdf(n::Int)
         for i in 1:n
           d = Normal(1.2)
         end
       end
set_immutable_pdf (generic function with 1 method)

julia> function set_mutable_pdf_mean(n::Int, d)
         for i in 1:n
           d.μ = .2
         end
       end
set_mutable_pdf_mean (generic function with 1 method)

julia> n = 1000000
1000000

julia> d = MutableNormal(0., 1.)
MutableNormal{Float64}(0.0,1.0)

julia> @time set_immutable_pdf(n)
  0.008659 seconds (12.04 k allocations: 512.825 KB)

julia> @time set_mutable_pdf_mean(n, d)
  0.002381 seconds (3.95 k allocations: 156.292 KB)

[...]

PR #740, opened 2018-07-23
`similar` function for distributions

arnavs:
In a few places (e.g., #530 and #482) it seemed that this functionality would be good. So I took a shot at it. Below is the code and tests: 

### Code
```
"""
A function to, given a distribution of type D and parameters A, B, C..., create a new D(A, B, C...).

Works so long as there is a type constructor for the distribution; i.e., Normal for Normal{Float64}. 
"""
Base.similar(dist::Distribution, params::Union{Number, AbstractArray}...) = (typeof(dist).name).wrapper(params...)
```

### Tests 
```
# Test similarity function. 
d1 = Normal()
@test similar(d1, 1.0, 3.0) == Normal(1.0, 3.0) # Vanilla case. 
@test_throws MethodError similar(d1, 1.0, 3.0, 4.0) # Error handling. 
@test similar(d1, 10.0) == Normal(10.0) # Resorts to defaults in constructor. 
d2 = MvNormal([1.0, 2.0], [1.0 0.0; 0.0 1.0]) 
@test similar(d1, params(d1)...) == d1 # Multivariate case. 
@test_throws DimensionMismatch similar(d2, [1.0, 2.0, 3.0], [1.0 0.0; 0.0 1.0]) # DimensionMismatch. 
```

cc: @ararslan 

Edit: Included what the code is now.

### DIFF
--- a/docs/src/starting.md
+++ b/docs/src/starting.md
@@ -13,7 +13,7 @@ The first step is to set up the environment:
 ```julia
 julia> using Compat, Compat.Random, Distributions
 
-julia> srand(123) # Setting the seed
+julia> Random.seed!(123) # Setting the seed
 ```
 
 Then, we create a standard-normal distribution `d` and obtain samples using `rand`:
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -5,7 +5,7 @@ module Distributions
 using StatsBase, PDMats, StatsFuns, Statistics
 
 import QuadGK: quadgk
-import Base: size, eltype, length, full, convert, show, getindex, rand
+import Base: size, eltype, length, convert, show, getindex, rand
 import Base: sum, maximum, minimum, +, -
 import Base.Math: @horner
 
--- a/src/mixtures/mixturemodel.jl
+++ b/src/mixtures/mixturemodel.jl
@@ -463,7 +463,7 @@ function _cwise_pdf!(r::AbstractMatrix, d::AbstractMixtureModel, X)
     size(r) == (n, K) || error("The size of r is incorrect.")
     for i = 1:K
         if d isa UnivariateMixture
-            view(r,:,i) .= pdf.(component(d, i), X)
+            view(r,:,i) .= pdf.(Ref(component(d, i)), X)
         else
             pdf!(view(r,:,i),component(d, i), X)
         end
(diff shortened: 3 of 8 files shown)


src/utils.jl
```diff
@@ -146,3 +146,11 @@ function trycholesky(a::Matrix{Float64})
         return e
     end
 end
+
+# Similarity function, per issue #530 (and by extension, issue #482)
+import Base.similar 
+
+function similar(dist::Distribution, params...)
```
ararslan: No need to `import`; you can extend directly by prepending `Base.` to `similar(...)`.

Also, you can simplify this a bit:
```julia
Base.similar(dist::Distribution, params...) = typeof(T).name.wrapper(params...)
```
Note though that I there seems to be an ambiguity with a deprecated method in Base, so you'll likely need a disambiguating definition as well:
```julia
Base.similar(dist::Distribution, params::Union{Integer,AbstractUnitRange}...) =
    typeof(dist).name.wrapper(params...)
```
We can remove that definition once we require Julia 1.0, since the deprecated method in Base will be removed.

Lastly, we should likely document that this will only work if a `UnionAll` constructor for the given distribution is defined, e.g. `Normal(m, v)` versus `Normal{Float32}(m, v)`. If there are any distributions in this package where that is not the case, we could add specific methods to those, either of `similar` or adding constructors.

ArnavSood: Good points re `import` and one-lining. RE docs, seems reasonable, although I think this may be more general (e.g., `Normal(m, v)` where `m::T` and `v::T` should be OK?).

ararslan: I mean that for a distribution type `D{T}` where `T` dictates the type of the parameter `x`, this method will only work if a constructor `D(x)` is defined. Note that constructors such as that are _not_ defined by default when defining a type; only `D{T}(x)` is defined by default.

ArnavSood: Done. 

```
# Similarity function, per issue #530 (and by extension, issue #482)
"""
A function to, given a distribution of type D and parameters A, B, C..., create a new D(A, B, C...).

Works so long as there is a type constructor for the distribution; i.e., Normal for Normal{Float64}. 
"""
Base.similar(dist::Distribution, params::Union{Number, AbstractArray}...) = (typeof(dist).name).wrapper(params...)
```

ArnavSood: And OK. Didn't know that re default constructors. Thanks.

arnavs:
@matbesancon It didn't seem like there was much interest in this feature, so I nuked the branch. Can resuscitate if there is.

matbesancon:
It's ok thanks, I think this is more a nice to have than a necessary for Distributions, it was to know we could close the PR

PR #751, opened 2018-08-22
Set travis mandatory to 1.0

matbesancon:


simonbyrne:
Perhaps do Appveyor as well?

simonbyrne:
can also delete the script block from .travis.yml (the default should work) and add in `using Pkg` to the `after_success` blocks.

matbesancon:
Done!
Also REQUIRE julia version set to 1.0

.travis.yml
```diff
@@ -13,9 +13,6 @@ notifications:
 branches:
   only:
   - master
-script:
-  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-  - julia --check-bounds=yes -e 'Pkg.clone(pwd()); Pkg.build("Distributions"); Pkg.test("Distributions"; coverage=true)'
 after_success:
-  - julia -e 'cd(Pkg.dir("Distributions")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder())'
+  - julia -e 'using Pkg; cd(Pkg.dir("Distributions")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder())'
   - julia -e 'Pkg.add("Documenter"); cd(Pkg.dir("Distributions")); include(joinpath("docs", "make.jl"))'
```
simonbyrne: need `using Pkg` here as well

REQUIRE
```diff
@@ -1,4 +1,4 @@
-julia 0.7.0-beta.85
+julia 1.0
```
simonbyrne: unless you're using deprecated functionality, probably better to make the `julia 0.7` to make it easier on people upgrading packages.

matbesancon: doesn't it set an upper bound too?

simonbyrne: no, you use 2 numbers to specify an upper bound.

simonbyrne:
Thanks!

PR #756, opened 2018-08-28
Base.extrema methods for Distribution and for RealInterval

yha:
e.g.
```
julia> using Distribution
julia> extrema(Uniform(1,3))
(1.0, 3.0)
julia> extrema(Normal())
(-Inf, Inf)
julia> extrema(support(Normal()))
(-Inf, Inf)
```


yha:
Can be closed in favor of #764

kmsquire:
Note that, I’m the future, you can just force push to the same branch.

PR #758, opened 2018-08-30
line break, fix undef

matbesancon:
An undef had been forgotten in `var` for Univariate mixtures


matbesancon:
actually pretty similar to #757 so cc @ararslan :)

simonbyrne:
Maybe add a test that would have caught it?

matbesancon:
check

PR #761, opened 2018-08-31
Fix a couple more missing `undef`s

ararslan:
Similar to #757 and #758. This appears to be the last remaining case.

The second commit just fixes whitespace inconsistencies.


matbesancon:
Wonderful. I feel we need a standardized juliafmt in the toolbox to avoid these

PR #763, opened 2018-08-31
complete docstrings

### LINKED ISSUE #744: the docstring for `nsampes` is unclear
Here is the current doc:
   
`nsamples(s::Sampleable)` : The number of samples contained in A. Multiple samples are often organized into an array, depending on the variate form.

`A` is not in the signature so it's hard to know what it refers to.

matbesancon:
Complete some lacking docstring, especially in the `common.jl` interface, core info should be available from the REPL. Fixes #744

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -14,6 +14,12 @@ Base.eltype(::Type{Continuous}) = Float64
 
 ## Sampleable
 
+"""
+	`Sampleable` is any type able to produce random values.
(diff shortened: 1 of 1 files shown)


simonbyrne:
@matbesancon I removed the tab indentation (this would make it appear as code blocks), and added titles.

ararslan:
And I've removed some extraneous backticks that would have shown up literally in the docstrings.

matbesancon:
Thank you both for fixing it

PR #764, opened 2018-09-04
Base.extrema methods for UnivariateDistribution and for RealInterval (take 2)

yha:
Replaces #756, which needlessly created methods for specific `UnivariateDistribution` types.
Includes tests and docs.


matbesancon:
Thanks!

PR #773, opened 2018-09-19
Update to use broadcasting per warnings

aaraun:
┌ Warning: `quantile(d::UnivariateDistribution, X::AbstractArray)` is deprecated, use `quantile.(d, X)` instead.
│   caller = top-level scope at none:0
└ @ Cor


matbesancon:
Doc PR, so merging it as-is without fear
Thanks @aaraun !

PR #778, opened 2018-09-26
Export restrict

### LINKED ISSUE #777: specify what to import with using
`using MyPkg` is convenient for small projects, but is a pain for something bigger like Distributions: it can be hard to know from reading the code where one piece is coming from, having to run all the imports for this is not optimal.

We should be using either:
* `import MyPkg`, problem is that we access all functions, including unexported
* `using MyPkg: f, g`: more verbose at import, but more explicit.

I'll submit a PR with the second option

matbesancon:
Fixes #777 
Not for StatsFuns, I could not use `StatsFuns.f` within the `@delegate_statsfuns` macro

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -1,25 +1,43 @@
 module Distributions
 
-using StatsBase, PDMats, StatsFuns, Statistics
+using StatsFuns
 
 import QuadGK: quadgk
 import Base: size, eltype, length, convert, show, getindex, rand
 import Base: sum, maximum, minimum, extrema, +, -, ==
 import Base.Math: @horner
 
-using LinearAlgebra, Printf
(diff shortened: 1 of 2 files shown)


matbesancon:
Thing is, some functions and modules come up in the code without ever being stated in the using and imports, that's a pain to search for, you always need a Julia repl open

simonbyrne:
I'm a little bit skeptical too. If this is a problem, it might be better to explicitly put the module at the call/definition site (e.g. `LinearAlgebra.cholesky(...)`).

matbesancon:
giving up this one for now, but I still find it a big cognitive overhead to have identifiers hanging around and non-determinable without remembering what's in all imported packages

PR #782, opened 2018-10-01
Update logistic.jl

janrpeters:
The prior version of this function failed the test

print("Lo(-2;6,0.01) = ",pdf4(Logistic(6,0.01),-2)," = 0, ", "Lo(-2;6,0.01) = ",pdf4(Logistic(6,0.01),6)," = 25");

PR #783, opened 2018-10-02
Update continuous.jl

janrpeters:
Test for the pdf of the logistic distribution for an extreme case which commonly happens but where the previous pdf would yield a NaN instead of zero.


simonbyrne:
Could you add this change to the other pull request (#782)? Better to have all these in the one place.

simonbyrne:
In case you were wondering, you can do it by editing the branch of the other pull request, in this case: https://github.com/janrpeters/Distributions.jl/blob/patch-1/test/continuous.jl

janrpeters:
Thanks, Simon! I have done that.

matbesancon:
@janrpeters should be close this one then?

Issue #62, opened 2013-05-22
Multinomial distribution issue


theodds:
The following code generates a BoundsError():

```
rand(Multinomial(1000, [1/3, 1/3, 1/3]), 1000)
```

The error seems to be stochastic, and related to how the individual counts are generated. The following both result in the same error:

```
for i in 1:1000
    rand(Multinomial(1000, [1/3, 1/3, 1/3]))
end
```

or

```
for i in 1:(1000^2)
    rand(Multinomial(1, [1/3, 1/3, 1/3]))
end
```

The same error occurs when using the Categorical distribution for the same purpose.

```
rand(Categorical([1/3, 1/3, 1/3], 1000^2)
```

PR #789, opened 2018-10-26
Allow Normal distribution with zero std deviation

simonbyrne:
Part of #283. I have hit this issue a few times: it can be very handy, especially for simulation.

See also https://github.com/JuliaStats/StatsFuns.jl/pull/62, which tweaks the underlying functions to be more useful.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.7
 PDMats 0.9.0
-StatsFuns 0.3.1
+StatsFuns 0.7.1
 StatsBase 0.23.0
 QuadGK 0.1.1
 SpecialFunctions 0.6.0
--- a/src/univariate/continuous/normal.jl
+++ b/src/univariate/continuous/normal.jl
@@ -27,7 +27,10 @@ struct Normal{T<:Real} <: ContinuousUnivariateDistribution
     μ::T
     σ::T
 
-    Normal{T}(μ, σ) where {T} = (@check_args(Normal, σ > zero(σ)); new{T}(μ, σ))
+    function Normal{T}(μ, σ) where {T}
+        @check_args(Normal, σ >= zero(σ))
+        new{T}(μ, σ)
+    end
 end
 
 #### Outer constructors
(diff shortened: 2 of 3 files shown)


matbesancon:
Should we add a test for the behavior of `Normal` with `std = 0.0`?

simonbyrne:
Yes, but probably need the StatsFuns change merged first.

matbesancon:
Also, the documentation should mention the behavior when continuous distributions are degenerate (mass point)

mschauer:
#62 is merged in the meantime

src/univariate/continuous/normal.jl
```diff
@@ -27,7 +27,10 @@ struct Normal{T<:Real} <: ContinuousUnivariateDistribution
     μ::T
     σ::T
 
-    Normal{T}(μ, σ) where {T} = (@check_args(Normal, σ > zero(σ)); new{T}(μ, σ))
+    function Normal{T}(μ, σ) where {T}
```
matbesancon: Mention the behavior in the documentation of `Normal`

matbesancon: also, some tests for the std=0 case should be added

matbesancon:
@mschauer #62 is an issue closed in 2013, you meant something else?

src/univariate/continuous/normal.jl
simonbyrne: Fixed!

simonbyrne:
Any final comments before I merge?

REQUIRE
```diff
@@ -1,6 +1,6 @@
 julia 0.7
 PDMats 0.9.0
-StatsFuns 0.3.1
+StatsFuns 0.7.1
```
matbesancon: why require up to patch number? Why not 0.7?

simonbyrne: 0.7.1 was the release with the necessary changes (in hindsight it should have been a minor release, I apologise).

matbesancon: ok, we can let everything so for this release, in a next one it would be good to remove patch from the other packages requirements

PR #794, opened 2018-11-05
Fix #793 (compute Kullberg-Leibler where density of `P` is zero).

### LINKED ISSUE #793: KL Divergence of Categorical when P(x) is zero
```
julia> using Distributions: Categorical, kldivergence
julia> kldivergence(Categorical([0.0, 0.1, 0.9]), Categorical([0.1, 0.1, 0.8]))
NaN
```
The contribution of the first term should be 0.0 no? as it stands I guess it's 0.0*Inf == NaN

[Wikipedia says](https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence#Definition): 
 "Whenever P(i) is zero the contribution of the i-th term is interpreted as zero because lim(x->0+) of x log(x) is 0"

mschauer:


ararslan:
I made it 6 :wink:

PR #798, opened 2018-11-29
MvNormal mean vector type constraints

willtebbutt:
This PR simply changes the type constraints on mean vector of `MvNormal`, `MvLogNormal`, and `MvNormalCanon` from `Union{Vector, ZeroVector}` to `Union{AbstractVector, ZeroVector}`. This proposal is motivated by some AD-related headaches caused by the current constraints.

### DIFF
--- a/src/multivariate/mvlognormal.jl
+++ b/src/multivariate/mvlognormal.jl
@@ -161,7 +161,7 @@ Mean vector ``\\boldsymbol{\\mu}`` and covariance matrix ``\\boldsymbol{\\Sigma}
 underlying normal distribution are known as the *location* and *scale*
 parameters of the corresponding lognormal distribution.
 """
-struct MvLogNormal{T<:Real,Cov<:AbstractPDMat,Mean<:Union{Vector, ZeroVector}} <: AbstractMvLogNormal
+struct MvLogNormal{T<:Real,Cov<:AbstractPDMat,Mean<:Union{AbstractVector, ZeroVector}} <: AbstractMvLogNormal
     normal::MvNormal{T,Cov,Mean}
 end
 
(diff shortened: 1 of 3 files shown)


ararslan:
It would be nice to add a test for the behavior when the parameter is not a regular `Vector`.

matbesancon:
@willtebbutt can you add the test? Afterwards we can make a final review and merge

willtebbutt:
This has actually turned out to be slightly more complex than I originally anticipated - there seems to be a reasonable amount of the `MvNormal` implementation which makes use of these type constraints. I suspect that changes can be made that would accommodate the looser type constraints that I'm proposing, but it will require a bit of work. I'm happy to sort this out, but it _probably_ won't happen before Christmas now.

matbesancon:
hi @willtebbutt, any news on that PR? Do you need help with parts of it?

matbesancon:
@willtebbutt ping :)

willtebbutt:
Hey, sorry for the radio silence. I'm totally rammed with other work at the minute so I've not had a chance to make progress, and I'm unlikely to be able to drive this forwards any time soon.

Separately, I noticed on slack that @simonbyrne has been doing some refactoring work and there was some discussion around refactoring all of the Normal distribution-related functionality. Perhaps there's some overlap with this PR?

matbesancon:
no problem, I can take on the remaining work from where you left it, is the latest version pushed?

willtebbutt:
Yeah, it's up to date. The issue is just that when you start to try and run things with `AbstractVector`s rather than `Vector`s stuff breaks. So I think that the way forwards is for some tests to be added that involve `AbstractVector`s and then to make stuff work with them.

matbesancon:
replaced by #837

PR #800, opened 2018-12-04
Delete EmpiricalUnivariateDistribution

### LINKED ISSUE #797: EmpiricalUnivariateDistribution assumption that cdf is a function doesn't match StatsBase
https://github.com/JuliaStats/StatsBase.jl/pull/429 changes the output of `StatsBase.ecdf` from a function to a callable type. However, `EmpiricalUnivariateDistribution` requires that the `cdf` field has type `Function`. As a result, the convenience constructor `EmpiricalUnivariateDistribution(x::Vector)` is broken with recent versions of StatsBase. (There, the type of `ecdf(x::Vector{Float64}` is `ECDF{Array{Float64,1}}`)

ararslan:
It's very, very broken to the point where a hard error on use is better than actually trying to use it. We'll keep #661 open as a starting point for a working implementation.

Closes #799 
Closes #797


matbesancon:
Not even a meaningful error (use X, Y, or Z)?

ararslan:
There isn't a particular appropriate replacement, people just shouldn't use it.

GordStephen:
`DiscreteNonParametric`'s MLE fitting should provide essentially the same functionality as `Empirical` (just without preserving the original data), no?

PR #802, opened 2018-12-10
Zerovec

### LINKED ISSUE #801: ZeroVector is not a vector
`ZeroVector` should probably be a subtype of `AbstractVector{T}` or even `StaticVector{T,S}`.
As such, it should also implement the abstract array interface, including `getindex`.

Bonus: since it is used quite a bit in Distributions, it should be documented with at least a docstring

matbesancon:
Fixes #801

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -48,7 +48,7 @@ Base.broadcast(::typeof(*), v::ZeroVector, ::Number) = v
 
 ##### Utility functions
 
-mutable struct NoArgCheck end
+struct NoArgCheck end
 
 isunitvec(v::AbstractVector{T}) where {T} = (norm(v) - 1.0) < 1.0e-12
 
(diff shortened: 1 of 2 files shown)


src/utils.jl
```diff
@@ -40,11 +44,11 @@ Base.broadcast(::typeof(-), v::ZeroVector, x::AbstractArray) = -x
 Base.broadcast(::Union{typeof(+),typeof(-)}, x::Number, v::ZeroVector) = fill(x, v.len)
 Base.broadcast(::typeof(+), v::ZeroVector, x::Number) = fill(x, v.len)
 Base.broadcast(::typeof(-), v::ZeroVector, x::Number) = fill(-x, v.len)
-
+Base.broadcast(::typeof(*), v::ZeroVector, ::Number) = v
```
andreasnoack: Eventually, we should implement proper broadcasting such that fusing will work.

Issue #781, opened 2018-09-30
pdf(d::Logistic, x::Real) is not safe!


janrpeters:
I am new to Julia but this bug affects porting my algorithms to Julia:
```julia
pdf(d::Logistic, x::Real) = (e = exp(-zval(d, x)); e / (d.θ * (1 + e)^2))
```
is not safe. For example, 
```julia
pdf(Logistic(6,0.01),-2)
```
will yield `NaN` instead of zero. The proper way to define this pdf is 
```julia
pdf(d::Logistic, x::Real) = logistic(-zval(d, x))*(1-logistic(-zval(d, x)))/d.θ;
```
which yields 0. 

May I directly patch this? Or can one of you do this? Thanks!
-Jan

andreasnoack:
Please open a PR and include a test as well. It should probably be something like
```julia
z = zval(d, x)
lz = logistic(-abs(z))
return lz*(1 - lz)/d.θ
```
to get the highest accuracy and avoid computing special functions twice.

Nosferican:
This should be closed due to #782.

andreasnoack:
Fixed by #782

PR #803, opened 2018-12-10
Fix to #781

Nosferican:
@andreasnoack for review


Nosferican:
@janrpeters It seems master already had a fix. Was that addressed already?

Issue #805, opened 2019-01-03
bug in getting mode for Binomial


ablaom:
````
julia> using Distributions

julia> mode(Binomial(2,0.25))
ERROR: type Binomial has no field prob
Stacktrace:
 [1] getproperty(::Any, ::Symbol) at ./sysimg.jl:18
 [2] mode(::Binomial{Float64}) at /Users/anthony/.julia/packages/Distributions/WHjOk/src/univariate/discrete/binomial.jl:70
 [3] top-level scope at none:0
````

The problem is here (line 68 of binomial.jl):

````
function mode(d::Binomial{T}) where T<:Real
    (n, p) = params(d)
    n > 0 ? round(Int,(n + 1) * d.prob) : zero(T)
end
````

Instead of `d.prob` it should read `d.p`

Nosferican:
Close issue?

PR #811, opened 2019-01-15
Fix bug, issue #805

ablaom:


mschauer:
Can we check some edge cases? https://en.wikipedia.org/wiki/Binomial_distribution#Mode uses floor and not round

ablaom:
Yes! Good catch. I agree it should be floor and not round (or floor and then round to get integer). Fixed just now.

src/univariate/discrete/binomial.jl
```diff
@@ -67,7 +67,7 @@ mean(d::Binomial) = ntrials(d) * succprob(d)
 var(d::Binomial) = ntrials(d) * succprob(d) * failprob(d)
 function mode(d::Binomial{T}) where T<:Real
     (n, p) = params(d)
-    n > 0 ? round(Int,(n + 1) * d.prob) : zero(T)
+    n > 0 ? round(Int,floor((n + 1) * d.p)) : zero(T)
```
ararslan: `rount(Int, floor(...))` can be written as `floor(Int, ...)`.

ararslan: Also it looks like this function is not type stable unless `T == Int`. It should either be `zero(Int)` instead of `zero(T)` or the `round`/`floor` call should be wrapped in `T()`.

ararslan: The latter is not a new problem and doesn't need to be fixed here.

matbesancon:
@ararslan good to merge?

PR #816, opened 2019-01-18
Add DiscreteNonParametric NoArgCheck method

GordStephen:
This allows calling `DiscreteNonParametric(xs, ps, NoArgCheck())` without providing explicit type parameters. My understanding is that we need to keep the type parameters on the inner constructors to support `Categorical` type aliasing.


matbesancon:
I'll merge in 24 hours if no objection

ararslan:
It would be nice to have a test for the new method.

matbesancon:
@GordStephen could you add some corresponding test(s) for the new method?

GordStephen:
Yep, will do.

PR #818, opened 2019-01-21
Support larger VonMises concentrations

jdrugo:
Previously, concentration parameters kappa > 700 led to AmosException, due to an
overflow when calling besseli(0, kappa).

The updated code uses besselix(.) instead of besseli(.), which avoids this
overflow for kappa > 700. The functions relying on the cached, and now
re-scaled, I0k are updated.

I tried to add a test case with kappa = 1000, but the circular R library does not appear to support large kappa's.


simonbyrne:
Thanks!

Could try getting test values from Wolfram Alpha, e.g. 
https://www.wolframalpha.com/input/?i=PDF%5BVonMisesDistribution%5B0.0,1000%5D,0.1%5D

jdrugo:
Indeed, that would be an option. However, I don't see how manual values could be added to [test/ref/continuous_test.ref.json](https://github.com/JuliaStats/Distributions.jl/blob/master/test/ref/continuous_test.ref.json), which would probably be the most adequate location to add further tests. All values in this file seem to be generated by R scripts, and I don't know enough R to change those significantly. Ideas?

simonbyrne:
Probably the best option is to add a few simple cases to the bottom of `test/continuous.jl`

jdrugo:
I added a few test cases. As it turns out, even Wolfram Alpha / Mathematica get's the cdf wrong:

https://www.wolframalpha.com/input/?i=CDF%5BVonMisesDistribution%5B1.1,1000%5D,0.5%5D

simonbyrne:
> I added a few test cases. As it turns out, even Wolfram Alpha / Mathematica get's the cdf wrong:
> 
> https://www.wolframalpha.com/input/?i=CDF%5BVonMisesDistribution%5B1.1,1000%5D,0.5%5D

I think the issue might be that we're calculating the CDF differently (see #320).

test/continuous.jl
```diff
@@ -52,4 +52,15 @@ using ForwardDiff
...
     @test rand(d) == 0.5
-end    
+end
+
+# Test for parameters beyond those supported in R references
+@testset "VonMises with large kappa" begin
+    d = VonMises(1.1, 1000)
+    @test var(d) ≈ 0.0005001251251957198
+    @test entropy(d) ≈ -2.034688918525470
+    @test isapprox(cf(d, 2.5), -0.921417 + 0.38047im, atol=1e-6)
```
simonbyrne: you can write this as 
```
@test cf(d, 2.5) ≈ -0.921417 + 0.38047im atol=1e-6
```
and the `@test` macro will rewrite it appropriately.

jdrugo: I don't know that. I updated those in the latest commit.

jdrugo:
Regarding the cdf: I also first thought that it might be due to a different support, but don't see how this would explain the negative cdf that Wolfram Alpha returns. As far as I know, the cdf should never be negative. Furthermore, the Mathematica documentation specifies the VonMises support over `[mu-pi, mu+pi]`, while also returning a negative value for `CDF[VonMisesDistribution[1.1, 1000], 0.5]`.

simonbyrne:
Ah, I didn't see the sign. Good point.

simonbyrne:
Thanks!

PR #822, opened 2019-01-27
Implemented interface and tests for Studentized Range distribution

BioTurboNick:
Relies on this PR for StatsFuns.jl: https://github.com/JuliaStats/StatsFuns.jl/pull/66

Uses R's functions for cdf and quantile, no pdf implementation. Tests implemented with relaxed matching tolerance (R implementation reports accuracy to only 4th decimal place).


BioTurboNick:
StatsFuns.jl PR was approved and tagged, tests should pass now.

simonbyrne:
The required StatsFuns version should be bumped to 0.8.0

REQUIRE
```diff
@@ -1,6 +1,6 @@
 julia 0.7
 PDMats 0.9.0
-StatsFuns 0.7.1
+StatsFuns 0.8.0
```
matbesancon: we shouldn't use patch numbers there, `0.8` instead of `0.8.1`

matbesancon:
one minor change and it's good to merge for me

matbesancon:
thanks @BioTurboNick

Issue #474, opened 2016-03-16
Documentation: Negative Binomial


fullerf:
Reading the documentation on the Univariate Discrete > NegativeBinomial, I was led to think that the wikipedia definition of the distribution was being used.  Upon testing it appears that the "Mathematica definition" is used instead.  You'll get good agreement between the observed mean and the formula: E[NB(n,p)] = n(1-p)/p. Also tested Var[NB(n,p)] = n(1-p)/p^2.  Both of these agreed well with the implementation in Distributions.jl.

The difference between the Wikipedia definition and the "Mathematica Definition", is just a change of variables: NB_wiki(n,1-p) = NB_mathematica(n,p).  For the wiki definition, the mean has the formula np/(1-p), variance = np/(1-p)^2.

tl;dr: Documentation doesn't quite match implementation.  It's off by a change of variables.

johnmyleswhite:
A pull request that improves the documentation would very helpful.

sivapvarma:
@fullerf The definition in [documenation](http://distributionsjl.readthedocs.io/en/latest/univariate-discrete.html#NegativeBinomial) says the samething as mathematica's documentation [here](https://reference.wolfram.com/language/ref/NegativeBinomialDistribution.html). 

Both Julia and Mathematica's implementation  count the number of failures before _r_ th success with _p_ as probability of success in an individual trial. 

I guess the confusion was due to subtle difference in pmf's between the two definitions.

For Julia/Mathematica definition `pmf(k)` will be proportional to `p^r (1-p)^k` whereas for the Wikipedia definition(number of success until _r_ failures) `pmf(k)` will be proportional to `p^k (1-p)^r`

mschauer:
This problem could be alleviated by means of keyword arguments via #823, with named arguments such as `psuccess` or `pfailure` in the constructor.

matbesancon:
I would consider this closed by #825

PR #825, opened 2019-02-03
Update NegativeBinomial external link in docstring

mohamed82008:
As highlighted in #474, the Wikipedia definition of the negative binomial distribution, linked in the docstring, is not consistent with the definition used in Julia. In this PR, I attempt to remove the source of confusion.


mschauer:
Thank you. I would retain the link to wikipedia with the clarification though.

matbesancon:
merge in 24 hours if no objection

Issue #197, opened 2014-03-11
Add RNG argument to rand calls


simonbyrne:
There seems to have been some recent additions to Base that allows for the RNG to be used as a first argument to `rand` and `rand!`. Once that has been sorted out, we should probably do it here as well: perhaps this could be done in a similar manner to the handling of `IO` objects in `show`?

sunoru:
Hi, I noticed that the dependence on neither Rmath nor dSFMT have been removed, and functions in Base have been made consistent with a first argument of RNG. So is there anyone working on this? If not I want to have a try and will make a PR.

BTW do you think we should store the RNG object in Distribution types or just add an argument when using functions like `rand`?

sunoru:
Oh.. The Rmath stuff was just moved to StatsFuns, so it's still unable to add the RNG argument to every sampling functions..

matbesancon:
should we close this in favour of #606 ? cc @simonbyrne

matbesancon:
awesome thanks

Issue #795, opened 2018-11-08
rand(::Normal, ::Dims) doesn't work


johnnychen94:
I found following generic codes https://github.com/JuliaStats/Distributions.jl/blob/117894c129420b7243a1e2d10ad82dc4bb376df1/src/genericrand.jl#L43-L47

but `rand(::Normal, ::Dims)` doesn't work though...

```julia
julia> img = ones(5,5);

julia> rand(Normal(),size(img))
ERROR: MethodError: no method matching Array{Float64,1}(::UndefInitializer, ::Tuple{Int64,Int64})
Closest candidates are:
  Array{Float64,1}(::UndefInitializer, ::Tuple{Vararg{Int64,N}}) where {T, N} at boot.jl:406
  Array{Float64,1}(::UndefInitializer, ::Tuple{Vararg{Integer,N}}) where {T, N} at sysimg.jl:155
  Array{Float64,1}(::UndefInitializer, ::Int64) where T at boot.jl:394
  ...
Stacktrace:
 [1] rand(::Normal{Float64}, ::Tuple{Int64,Int64}) at /Users/jc/.julia/packages/Distributions/WHjOk/src/univariates.jl:169
 [2] top-level scope at none:0

julia> rand(Normal(),size(img)...)
5×5 Array{Float64,2}:
  0.462929  -0.063203   2.63356   -0.996424  -0.366942
 -1.01405    0.218144   0.180686  -0.720094  -1.04636
 -0.442738  -1.26272   -0.248081   0.89015   -0.273845
 -0.127492  -0.851691  -1.99012    0.688425  -0.512067
 -2.64416   -1.18138    0.483727  -0.780079   0.0627713
```

Is this a deliberately designed behavior?

richardreeve:
@matbesancon This is now fixed by #830 and can be closed:
```julia
julia> img = ones(5,5);

julia> rand(Normal(),size(img))
5×5 Array{Float64,2}:
 -0.839027   0.431422   0.408396   -0.757633  -0.0633746
  0.311111   0.583708  -0.0504512  -1.72976    1.33694
  2.29509    0.963272  -0.693654    0.795949  -0.0731486
 -2.26709    0.458791  -1.77383     0.670062  -0.745464
  0.529966  -0.522337   0.120766    0.550852  -1.22006

julia> rand(Normal(),size(img)...)
5×5 Array{Float64,2}:
 -0.0531773   0.567695  -1.80303     1.73517    1.12941
[...]

Issue #790, opened 2018-10-31
Numerical issue with Multinomial sampling


cossio:
There seems to be a numerical issue with Multinomial sampling. I think this was not occurring in previous versions of Distributions.jl.

I found this with a long probability vector. Here is the full source to reproduce this error:
https://gist.github.com/cossio/8c5bd3557aa1313919a3d0ccccef7ca2

This is on Distributions v0.16.4, Julia 1.0.1.

Here is the stack trace:

```
# ERROR: InexactError: Int64(Int64, NaN)
# Stacktrace:
#  [1] Type at ./float.jl:700 [inlined]
#  [2] convert at ./number.jl:7 [inlined]
#  [3] rand at /opt/julia-depot/packages/Distributions/WHjOk/src/univariate/discrete/binomial.jl:111 [inlined]
#  [4] multinom_rand!(::Int64, ::Array{Float64,1}, ::Array{Int64,1}) at /opt/julia-depot/packages/Distributions/WHjOk/src/samplers/multinomial.jl:14
#  [5] _rand! at /opt/julia-depot/packages/Distributions/WHjOk/src/multivariate/multinomial.jl:161 [inlined]
#  [6] rand(::Multinomial{Float64}) at /opt/julia-depot/packages/Distributions/WHjOk/src/multivariates.jl:49
#  [7] top-level scope at none:0
```

simonbyrne:
Looks like an Rmath issue:
```
julia> using StatsFuns

julia> n = 4969027388018185
4969027388018185

julia> StatsFuns.RFunctions.binomrand(n, 0.1)
NaN
```

The solution is to switch over to our own samplers.

cossio:
@simonbyrne Can you offer some guidance on how to fix this? I could then try to prepare a PR.

simonbyrne:
There is some code here: https://github.com/JuliaStats/Distributions.jl/blob/master/src/samplers/binomial.jl
It needs testing and performance tuning.

richardreeve:
This is now fixed by #830 using the sampler code @simonbyrne suggests (which is tested, but not performance tuned I'm afraid) and can be closed:
```julia
julia> include("Distributions_error.jl")
865-element Array{Int64,1}:
   222187754457
 10469767117680
  8688751380848
 11241013423469
  9679507254818
  7883122452775
  1852447899760
  1774469014382
      339509700
  6210597487262
  3058018576781
   336395441770
  8326451185448
  2009135287017
[...]

PR #830, opened 2019-02-17
Allow alternate RNGs to be used on all distributions

richardreeve:
I need to be able to use per-thread random number generators for a large application I've written, and I discovered that Distributions.jl doesn't support it. This seems to be because of a dependency on Rfunction code, which has been sitting around waiting to be fixed in #138 since 2013. I have fixed my problem here, and generalised it as much as possible so that all distributions now work with an optional first argument of a RNG.

This required refactoring the code slightly so that - although the interface remains identical as far as I can see - the internal API for implementing RNGs for a specific distribution is simpler. Instead of multiple `rand()` and `rand!()` functions that need to be implemented in different situations, just `_rand!(::AbstractRNG, ::Distribution)` needs to be implemented for non-Rfunction-based distributions and then they will be able to do anything (though it's a bit more complicated for non-univariable distributions). For the remaining Rfunction-based distributions (I've fixed Beta, Binomial, Chisq and Gamma distributions and their dependencies, using code in the repository where possible, and otherwise taking from elsewhere),`rand(::Distribution)` needs to be implemented, to allow the same behaviour as before.

Where you need to generate random numbers for one of the remaining Rfunction distributions with a specific RNG, the code will fall back to a quantile-based approach, which is obviously slow, but at least keeps the functionality consistent. I took this approach because many (weird!) distributions use a quantile-based approach anyway.

I'm happy to fiddle with this as desired, but given the 6 years to has taken to not remove Rfunction dependencies so far, if the general idea seems sound I'd rather not wait until all of the remaining distributions have been fixed - they should now be easy to pick off one at a time... the main issue the PR is intended to fix is instead allowing multithreaded random number generation.

Issues to resolve:

- [x] What to do about partially allocated arrays of arrays being allowed as inputs to `rand!()` - currently ~a deprecated keyword argument keeps existing (allocate subarrays) behaviour by default, but can be set either way to remove deprecation.~ by default it looks to see if allocation is necessary and does it if so, but can be set to allocate or not specifically by use of an added boolean argument.
- [x] Do we need to add further tests for new RNGs? Currently they are tested anyway, since calls without specified RNGs use _rand!(GLOBAL_RNG, ...). Done for univariable distributions through extending `test_samples()`. Done
- [x] ~Can we leave the remaining RFunction distributions for now, and fix them later? I don't have code for other distributions, and it's not the purpose of this PR. Once they are all rewritten, it's trivial to remove their support structure by removing the `ContinuousUnivariateRDist <: ContinuousUnivariateDistribution` and `DiscreteUnivariateRDist <: DiscreteUnivariateDistribution` subtypes~ yes
- [x] Fix Product distribution
- [x] Add in tests for truncated normal if they are missing, or fix them if not
- [x] Decide what to do in the short term about RFunction-based distributions and do it...

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -122,6 +122,16 @@ Get the probability of failure.
 failprob(d::DiscreteUnivariateDistribution)
 
 # Temporary fix to handle RFunctions dependencies
+"""
+    @rand_rdist(::Distribution)
+
+Mark a `Distribution` subtype as requiring RFunction calls. Since these calls
+cannot accept an arbitrary random number generator as an input, this macro
+creates new `rand(::Distribution, n::Int)` and
+`rand!(::Distribution, X::AbstractArray)` functions that call the relevant
+RFunction. Calls using another random number generator still work, but rely on
+a quantile function to operate.
+"""
 macro rand_rdist(D)
     esc(quote
         function rand(d::$D, n::Int)
--- a/src/common.jl
+++ b/src/common.jl
@@ -120,3 +120,17 @@ succprob(d::DiscreteUnivariateDistribution)
 Get the probability of failure.
 """
 failprob(d::DiscreteUnivariateDistribution)
+
+# Temporary fix to identify Rfunctions
+abstract type DiscreteUnivariateRDist <: DiscreteUnivariateDistribution end
+abstract type ContinuousUnivariateRDist <: ContinuousUnivariateDistribution end
+function rand(d::Union{DiscreteUnivariateRDist, ContinuousUnivariateRDist}, n::Int)
+    [rand(d) for i in Base.OneTo(n)]
+end
+function rand!(d::Union{DiscreteUnivariateRDist, ContinuousUnivariateRDist},
+               X::AbstractArray)
+    for i in eachindex(X)
+        X[i] = rand(d)
+    end
+    return X
+end
--- a/src/common.jl
+++ b/src/common.jl
@@ -122,15 +122,16 @@ Get the probability of failure.
 failprob(d::DiscreteUnivariateDistribution)
 
 # Temporary fix to handle RFunctions dependencies
-abstract type DiscreteUnivariateRDist <: DiscreteUnivariateDistribution end
-abstract type ContinuousUnivariateRDist <: ContinuousUnivariateDistribution end
-function rand(d::Union{DiscreteUnivariateRDist, ContinuousUnivariateRDist}, n::Int)
-    [rand(d) for i in Base.OneTo(n)]
-end
-function rand!(d::Union{DiscreteUnivariateRDist, ContinuousUnivariateRDist},
-               X::AbstractArray)
-    for i in eachindex(X)
-        X[i] = rand(d)
-    end
-    return X
+macro rand_rdist(D)
+    esc(quote
+        function rand(d::$D, n::Int)
+            [rand(d) for i in Base.OneTo(n)]
+        end
+        function rand!(d::$D, X::AbstractArray)
+            for i in eachindex(X)
+                X[i] = rand(d)
+            end
+            return X
+        end
+    end)
 end
--- a/src/common.jl
+++ b/src/common.jl
@@ -121,7 +121,7 @@ Get the probability of failure.
 """
 failprob(d::DiscreteUnivariateDistribution)
 
-# Temporary fix to identify Rfunctions
+# Temporary fix to handle RFunctions dependencies
 abstract type DiscreteUnivariateRDist <: DiscreteUnivariateDistribution end
 abstract type ContinuousUnivariateRDist <: ContinuousUnivariateDistribution end
 function rand(d::Union{DiscreteUnivariateRDist, ContinuousUnivariateRDist}, n::Int)
--- a/src/matrixvariates.jl
+++ b/src/matrixvariates.jl
@@ -22,16 +22,42 @@ Return the mean matrix of `d`.
 """
 mean(d::MatrixDistribution)
 
-# sampling
+## sampling
+# multiple matrix-variates with pre-allocated array
+function rand(rng::AbstractRNG, s::Sampleable{Matrixvariate},
+              X::AbstractArray{M}) where M <: AbstractMatrix
+    smp = sampler(s)
+    for i in eachindex(X)
+        X[i] = _rand!(rng, smp, M(undef, size(s)))
+    end
+    return X
+end
 
-rand!(d::MatrixDistribution, A::AbstractArray{M}) where {M<:Matrix} = _rand!(sampler(d), A)
+# multiple matrix-variates with pre-allocated array of pre-allocated matrices
+function rand!(rng::AbstractRNG, s::Sampleable{Matrixvariate},
+               X::AbstractArray{<:AbstractMatrix})
+    smp = sampler(s)
+    for x in X
+        rand!(rng, smp, x)
+    end
+    return X
+end
 
-"""
-    rand(d::MatrixDistribution, n)
+# multiple matrix-variates, must allocate array of arrays
+rand(rng::AbstractRNG, s::Sampleable{Matrixvariate}, dims::Dims) =
+    rand(rng, s, Array{Matrix{eltype(s)}}(undef, dims))
 
-Draw a sample matrix from the distribution `d`.
-"""
-rand(d::MatrixDistribution, n::Int) = _rand!(sampler(d), Vector{Matrix{eltype(d)}}(undef, n))
+# single matrix-variate, must allocate one matrix
+rand(rng::AbstractRNG, s::Sampleable{Matrixvariate}) =
+    _rand!(rng, s, Matrix{eltype(s)}(undef, size(s)))
+
+# single matrix-variate with pre-allocated matrix
+function rand!(rng::AbstractRNG, s::Sampleable{Matrixvariate},
+               A::AbstractMatrix{<:Real})
+    size(A) == size(s) ||
+        throw(DimensionMismatch("Output size inconsistent with sample size."))
+    return _rand!(rng, s, A)
+end
 
 # pdf & logpdf
 
--- a/src/matrixvariates.jl
+++ b/src/matrixvariates.jl
@@ -36,38 +36,31 @@ function _rand!(rng::AbstractRNG, s::Sampleable{Matrixvariate},
     return m
 end
 
-# multiple matrix-variates with pre-allocated array
-function rand_!(rng::AbstractRNG, s::Sampleable{Matrixvariate},
-                X::AbstractArray{M}) where M <: AbstractMatrix
-    smp = sampler(s)
-    for i in eachindex(X)
-        X[i] = _rand!(rng, smp, M(undef, size(s)))
-    end
-    return X
-end
+# multiple matrix-variates with pre-allocated array of maybe pre-allocated matrices
+rand!(rng::AbstractRNG, s::Sampleable{Matrixvariate},
+      X::AbstractArray{<:AbstractMatrix}) =
+          @inbounds rand!(rng, s, X,
+                          !all([isassigned(X,i) for i in eachindex(X)]) ||
+                          (sz = size(s); !all(size(x) == sz for x in X)))
 
-# multiple matrix-variates with pre-allocated array of pre-allocated matrices
 function rand!(rng::AbstractRNG, s::Sampleable{Matrixvariate},
-               X::AbstractArray{<:AbstractMatrix};
-               allocate::Union{Bool, Missing} = missing)
-    if ismissing(allocate)
-        Base.depwarn("`rand!([rng::AbstractRNG, ]s::Sampleable{Matrixvariate}, X::AbstractArray{<:AbstractMatrix})` is deprecated as it allocates matrices. If you want to preserve this behaviour, the keyword argument `allocate` should be set to true or call Distributions.rand_!. If you want to not allocate memory, set `allocate=false` for now. This will become the default behaviour after the next release.", :rand!)
-        allocate = true
-    end
-    if allocate
-        return rand_!(rng, s, X)
-    end
-
+               X::AbstractArray{M}, allocate::Bool) where M <: AbstractMatrix
     smp = sampler(s)
-    for x in X
-        rand!(rng, smp, x)
+    if allocate
+        for i in eachindex(X)
+            X[i] = _rand!(rng, smp, M(undef, size(s)))
+        end
+    else
+        for x in X
+            rand!(rng, smp, x)
+        end
     end
     return X
 end
 
 # multiple matrix-variates, must allocate array of arrays
 rand(rng::AbstractRNG, s::Sampleable{Matrixvariate}, dims::Dims) =
-    rand!(rng, s, Array{Matrix{eltype(s)}}(undef, dims); allocate = true)
+    rand!(rng, s, Array{Matrix{eltype(s)}}(undef, dims), true)
 
 # single matrix-variate, must allocate one matrix
 rand(rng::AbstractRNG, s::Sampleable{Matrixvariate}) =
--- a/src/matrixvariates.jl
+++ b/src/matrixvariates.jl
@@ -23,9 +23,22 @@ Return the mean matrix of `d`.
 mean(d::MatrixDistribution)
 
 ## sampling
+
+# multivariate with pre-allocated 3D array
+function _rand!(rng::AbstractRNG, s::Sampleable{Matrixvariate},
+                m::AbstractArray{<:Real, 3})
+    @boundscheck (size(m, 1), size(m, 2)) == (size(s, 1), size(s, 2)) ||
+        throw(DimensionMismatch("Output size inconsistent with matrix size."))
+    smp = sampler(s)
+    for i in Base.OneTo(size(m,3))
+        _rand!(rng, smp, view(m,:,:,i))
+    end
+    return m
+end
+
 # multiple matrix-variates with pre-allocated array
-function rand(rng::AbstractRNG, s::Sampleable{Matrixvariate},
-              X::AbstractArray{M}) where M <: AbstractMatrix
+function rand_!(rng::AbstractRNG, s::Sampleable{Matrixvariate},
+                X::AbstractArray{M}) where M <: AbstractMatrix
     smp = sampler(s)
     for i in eachindex(X)
         X[i] = _rand!(rng, smp, M(undef, size(s)))
@@ -35,7 +48,16 @@ end
 
 # multiple matrix-variates with pre-allocated array of pre-allocated matrices
 function rand!(rng::AbstractRNG, s::Sampleable{Matrixvariate},
-               X::AbstractArray{<:AbstractMatrix})
+               X::AbstractArray{<:AbstractMatrix};
+               allocate::Union{Bool, Missing} = missing)
+    if ismissing(allocate)
+        Base.depwarn("`rand!([rng::AbstractRNG, ]s::Sampleable{Matrixvariate}, X::AbstractArray{<:AbstractMatrix})` is deprecated as it allocates matrices. If you want to preserve this behaviour, the keyword argument `allocate` should be set to true or call Distributions.rand_!. If you want to not allocate memory, set `allocate=false` for now. This will become the default behaviour after the next release.", :rand!)
+        allocate = true
+    end
+    if allocate
+        return rand_!(rng, s, X)
+    end
+
     smp = sampler(s)
     for x in X
         rand!(rng, smp, x)
@@ -45,7 +67,7 @@ end
 
 # multiple matrix-variates, must allocate array of arrays
 rand(rng::AbstractRNG, s::Sampleable{Matrixvariate}, dims::Dims) =
-    rand(rng, s, Array{Matrix{eltype(s)}}(undef, dims))
+    rand!(rng, s, Array{Matrix{eltype(s)}}(undef, dims); allocate = true)
 
 # single matrix-variate, must allocate one matrix
 rand(rng::AbstractRNG, s::Sampleable{Matrixvariate}) =
@@ -54,8 +76,8 @@ rand(rng::AbstractRNG, s::Sampleable{Matrixvariate}) =
 # single matrix-variate with pre-allocated matrix
 function rand!(rng::AbstractRNG, s::Sampleable{Matrixvariate},
                A::AbstractMatrix{<:Real})
-    size(A) == size(s) ||
-        throw(DimensionMismatch("Output size inconsistent with sample size."))
+    @boundscheck size(A) == size(s) ||
+        throw(DimensionMismatch("Output size inconsistent with matrix size."))
     return _rand!(rng, s, A)
 end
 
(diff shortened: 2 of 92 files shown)


src/matrixvariates.jl
```diff
@@ -22,16 +22,42 @@ Return the mean matrix of `d`.
 """
 mean(d::MatrixDistribution)
 
-# sampling
+## sampling
+# multiple matrix-variates with pre-allocated array
+function rand(rng::AbstractRNG, s::Sampleable{Matrixvariate},
```
mschauer: Seouls this be a bang function?

richardreeve: The problem is that I found three cases for matrix-variates - 1. Nothing is pre-allocated, just rand() 2. The matrices themselves are allocated, so almost no allocation is needed, definitely rand!() 3. The array to stick the matrices in is allocated, but not the matrices themselves - personally I wouldn’t even offer this option, but it was already there, and I couldn’t see how to dispatch differently on it, so I was left with calling it rand(). It was the only really nasty compromise I found to maintain backward compatibility... I’d prefer to drop it?

richardreeve:
Issues to resolve:

- [ ] What to do about partially allocated arrays of arrays being allowed as inputs to `rand!()` or `rand()` for matrix variates - one option is a breaking change to remove them, but I have currently implemented filling partially allocated arrays through `rand()`, so as not to conflict with `rand!()` for fully allocated arrays of arrays
- [ ] Do we need to add tests for new RNGs? Currently they are tested anyway, since calls without specified RNGs use _rand!(GLOBAL_RNG, ...)
- [ ] Can we leave the remaining Rfunction distributions for now, and fix them later? I don't have code for other distributions, and it's not the purpose of this PR. Once they are all rewritten, it's trivial to remove their support structure by removing the `ContinuousUnivariateRDist <: ContinuousUnivariateDistribution` and `DiscreteUnivariateRDist <: DiscreteUnivariateDistribution` subtypes

matbesancon:
this is awesome. About your third point:
yes we can definitely finish this in another PR, one problem of the old one from 2013 was the scope.

richardreeve:
The PoissonADSampler failed to give the right statistics on nightly, but I don’t think it’s a real bug, as I didn’t edit it and it has worked fine before... could someone restart it?

richardreeve:
This PR fixes #197, #436, #540, #573, #606, #678, #790 and #795 (if they weren't already fixed, I think #540 may be). It is also relevant to cstjean/ScikitLearn.jl#25 and JuliaStats/StatsBase.jl#240, though I think the latter may already be fixed too.

PS I see that the PoissonADSampler has failed again as outside confidence intervals... this time with MacOS, Julia 1.0. I'm guessing this is a common theme, and people see it a lot?

matbesancon:
I restarted the failing job, let's see

mschauer:
There are some stragglers such as maybe 
https://github.com/JuliaStats/Distributions.jl/blob/5a59a276a0561f150426c8817d02ff174865f670/src/multivariate/product.jl - would it be good to collect those cases as a todo-list for later?

richardreeve:
Yes, absolutely! In fact we should fix them now - I've added the Product distribution to the list above, and now you point it out, I realise that I only exhaustively looked at the univariate ones, which are tested and so have to work, and the matrix-variate ones because they were complicated, and I may have forgotten other multinomial distributions...

mschauer:
Not extremely useful because full of false positives but one can look at
```
(richardreeve-rr/rng)~/dev/Distributions$ grep -r "rand\!([^r]" src
src/common.jl:function rand!(d::Union{DiscreteUnivariateRDist, ContinuousUnivariateRDist},
src/mixtures/mixturemodel.jl:    rand!(d::Union{UnivariateMixture, MultivariateMixture}, r::AbstactArray)
src/mixtures/mixturemodel.jl:rand!(d::AbstractMixtureModel, r::AbstractArray)
src/multivariate/multinomial.jl:_rand!(d::Multinomial, x::AbstractVector{T}) where T<:Real =
src/multivariates.jl:rand!(d::MultivariateDistribution, x::AbstractArray)
src/samplers/multinomial.jl:function multinom_rand!(n::Int, p::Vector{Float64},
src/samplers/multinomial.jl:_rand!(s::MultinomialSampler, x::AbstractVector{T}) where T<:Real =

(richardreeve-rr/rng)~/dev/Distributions$ grep -r "rand([^r]" src
src/common.jl:function rand(d::Union{DiscreteUnivariateRDist, ContinuousUnivariateRDist}, n::Int)
src/common.jl:    [rand(d) for i in Base.OneTo(n)]
src/common.jl:        X[i] = rand(d)
src/genericrand.jl:rand(s::Sampleable) = rand(GLOBAL_RNG, s)
src/genericrand.jl:rand(s::Sampleable, dims::Dims) = rand(GLOBAL_RNG, s, dims)
src/genericrand.jl:rand(s::Sampleable, dim1::Int, moredims::Int...) =
src/mixtures/mixturemodel.jl:    rand(d::Union{UnivariateMixture, MultivariateDistribution})
src/mixtures/mixturemodel.jl:    rand(d::Union{UnivariateMixture, MultivariateMixture}, n)
src/mixtures/mixturemodel.jl:rand(d::AbstractMixtureModel)
src/multivariate/product.jl:_rand!(d::Product, x::AbstractVector{<:Real}) = broadcast!(dn->rand(dn), x, d.v)
src/multivariates.jl:rand(s::Sampleable{Multivariate}, n::Int) = rand(GLOBAL_RNG, s, n)
src/samplers/multinomial.jl:            xi = rand(Binomial(n, pi / rp))
src/samplers/obsoleted.jl:function rand(table::DiscreteDistributionTable)
src/samplers/obsoleted.jl:    i = rand(1:(64^9 - 1))
src/samplers/obsoleted.jl:function rand(h::HuffmanNode{T}) where T
src/samplers/obsoleted.jl:    u = rand(UInt64)
src/samplers/obsoleted.jl:            u = rand(UInt64)
src/samplers/vonmisesfisher.jl:    while κ * w + (p - 1) * log(1 - x0 * w) - c < log(rand())
src/truncated/normal.jl:function rand(d::Truncated{Normal{T},Continuous}) where T <: Real
src/truncated/normal.jl:                r = rand(Exponential(1.0 / a)) + lb
src/truncated/normal.jl:                u = rand()
src/truncated/normal.jl:                r = rand(Exponential(1.0 / a)) - ub
src/truncated/normal.jl:                u = rand()
src/truncated/normal.jl:                r = lb + rand() * (ub - lb)
src/truncated/normal.jl:                u = rand()
src/univariate/continuous/noncentralbeta.jl:function rand(d::NoncentralBeta)
src/univariate/continuous/noncentralbeta.jl:    a = rand(NoncentralChisq(2d.α, β))
src/univariate/continuous/noncentralbeta.jl:    b = rand(Chisq(2β))
src/univariate/continuous/noncentralf.jl:function rand(d::NoncentralF)
src/univariate/continuous/noncentralf.jl:    r1 = rand(NoncentralChisq(d.ν1,d.λ)) / d.ν1
src/univariate/continuous/noncentralf.jl:    r2 = rand(Chisq(d.ν2)) / d.ν2
src/univariate/discrete/discretenonparametric.jl:rand(d::DiscreteNonParametric) = rand(GLOBAL_RNG, d)
src/univariate/discrete/hypergeometric.jl:rand(d::Hypergeometric) =
src/univariate/discrete/negativebinomial.jl:rand(d::NegativeBinomial) =
src/univariate/discrete/poisson.jl:rand(d::Poisson) = convert(Int, StatsFuns.RFunctions.poisrand(d.λ))
src/univariate/discrete/skellam.jl:rand(d::Skellam) = rand(Poisson(d.μ1)) - rand(Poisson(d.μ2))
```

richardreeve:
Okay, I've added in all of the tests for the multivariate and matrix-variate distributions with the additional rng argument, and it was only the Product distribution that was missing (thanks @mschauer). Apart from checking that this is all sane (I'm sure there are other issues!), the remaining item I'm aware of is the first in https://github.com/JuliaStats/Distributions.jl/pull/830#issue-253684173 - what to do with partially allocated arrays for multivariate distributions...

mschauer:
I dislike `rand` writing into the container. It becomes especially unnatural that this is close to what`rand!` (and certainly not `rand`) would do for `SArrays`.

richardreeve:
That's fine, but then it has the same signature for dispatch as `rand!` where the matrices as well as the holding array is preallocated. I suppose a third option (as well as dropping support for this intermediate version) would be a keyword argument (`allocate`?), which throws away any existing storage for the matrices, and allocates new memory? I don't have a strong feeling either way for `rand(...)` or `rand!(...;allocate=true)`, but it needs to be resolved.

I should say that the reason this arises is that there is no way of avoiding allocating memory for matrix-variates, which I feel is much worse than either of these proposals (and inconsistent with the name `rand!`).

mschauer:
I suggest to deprecate the both modifying and allocating `rand` methods, then there is some time to adapt.

richardreeve:
That sounds like a good idea, but the problem is that I introduced this method! The current behaviour for `rand!()` for matrixvariates, _but not for univariates or multivariates_, is to always allocate memory by using this modify and allocate approach. If we want to allow (what I would argue is the correct behaviour of) not allocating on `rand!()`, then this is a breaking change as the new version would have the same signature.

The two ways around this I can think of are basically the same: (1) to have a keyword argument with a default of `allocate=missing`, and then call `Base.depwarn` directly to tell people to set allocate=true; or (2) to have a function with three arguments, force people to use it if they want allocation, deprecate the two argument version, and then reintroduce it later...

I would suggest the first is better - apart from simplicity, it also allows the multivariate distributions to generate vectors of vectors now, with the same sub-allocation or not, since at the moment multiple draws from a multivariate distribution have to go to matrices.

PS The other thing that didn't occur to me is that it would be easy to allow multiple matrixvariates to use a 3D array too. It's not clear why the default behaviour for multivariates is a matrix, but the default for matrixvariates is not 3D?

src/samplers/gamma.jl
```diff
@@ -100,8 +93,8 @@ function rand(s::GammaGDSampler)
 
     # Step 8
     @label step8
```
matbesancon: should we maybe  replace the label & goto macros with something a bit easier to read? (splitting functions, if else branches)

src/samplers/gamma.jl
```diff
@@ -189,16 +182,16 @@ function GammaMTSampler(g::Gamma)
...
         while v <= 0.0
-            x = randn()
+            x = randn(rng)
             v = 1.0 + s.c * x
         end
         v *= (v * v)
-        u = rand()
+        u = rand(rng)
         x2 = x * x
         if u < 1.0 - 0.331 * abs2(x2) || log(u) < 0.5 * x2 + s.d * (1.0 - v + log(v))
```
matbesancon: should we add a comment on the hard-coded numerical values?

src/samplers/multinomial.jl
```diff
@@ -44,25 +85,26 @@ struct MultinomialSampler <: Sampleable{Multivariate,Discrete}
     alias::AliasTable
 end
 
-MultinomialSampler(n::Int, prob::Vector{Float64}) = 
+MultinomialSampler(n::Int, prob::Vector{Float64}) =
```
matbesancon: Could we relax the type constraint from Vector of Float64 to an abstract vector of Real?

src/testutils.jl
```diff
@@ -114,8 +119,8 @@ function test_samples(s::Sampleable{Univariate, Discrete},      # the sampleable
         @assert cub[i] >= clb[i]
     end
 
-    # generate samples
-    samples = rand(s, n)
+    # generate samples using global RNG
```
matbesancon: using rng passed or global rng

src/univariate/continuous/beta.jl
```diff
@@ -116,8 +116,85 @@ gradlogpdf(d::Beta{T}, x::Real) where {T<:Real} =
 
 #### Sampling
 
-rand(d::Beta) = StatsFuns.RFunctions.betarand(d.α, d.β)
+struct BetaSampler{T<:Real, S1 <: Sampleable{Univariate,Continuous},
+                   S2 <: Sampleable{Univariate,Continuous}} <:
+    Sampleable{Univariate,Continuous}
+    γ::Bool
```
matbesancon: unicode is not practical in fields people have to type (struct fields and keyword arguments to functions, maybe switch to `gamma`, `ialpha`, `ibeta`

src/samplers/gamma.jl
mschauer: Filed this issue separately as  #835 so we can proceed here in the meantime.

mschauer:
Sorry, I was a bit slow to pick that up. How about the clumsy but principled way.
Deprecate `rand!`, but keep its quirky allocating behaviour. Introduce `rand!!` with the new behaviour of writing into the matrices themselves. Then later on act on the deprecation and rename `rand!!` to `rand!`.

matbesancon:
do we have some examples of packages using the double bang for things like this?

mschauer:
It does not have to exported, it is just a name under which the functionality can be "parked" in the meantime, any other name would do itself (though the double bang makes a bit of sense to me).

matbesancon:
I'm ok with that, but since we're introducing breaking changes, it may need more people to agree

src/common.jl
```diff
@@ -120,3 +120,17 @@ succprob(d::DiscreteUnivariateDistribution)
 Get the probability of failure.
 """
 failprob(d::DiscreteUnivariateDistribution)
+
+# Temporary fix to handle RFunctions dependencies
+abstract type DiscreteUnivariateRDist <: DiscreteUnivariateDistribution end
```
simonbyrne: Why is this part needed? I'm not a huge fan of a temporary fix that changes type hierarchies

src/univariate/continuous/beta.jl
simonbyrne: why would they have to access these?

src/univariates.jl
```diff
@@ -153,31 +153,28 @@ end
 ##### generic methods (fallback) #####
 
 ## sampling
+# single univariate, no allocation
+rand(rng::AbstractRNG, s::Sampleable{Univariate}) = _rand!(rng, s)
```
simonbyrne: I still don't quite understand why we need to define all the univariate distributions in terms of `_rand!`: I like to avoid the use of underscore prefix functions where possible. 

Also there is sort of a convention that RNGs are the one thing that don't require a `!` to mutate.

richardreeve: That's completely fair, I was just getting lazy (/ for consistency of naming) since the only thing that didn't mutate at this point was single sample univariates as all of the allocation for everything else is now done before you get to the distributions. I'll change it.

simonbyrne: Thanks. I know how much of a pain these sort of changes are...

src/common.jl
richardreeve: R-based distributions need to be able to behave differently (but in a consistent fashion) to all other distributions to be alternate-RNG aware, as they switch to using a quantile-based generator. In order to pick this up generically using dispatch I introduced the type hierarchy you see. The idea was to minimise the amount of code that needed to be changed to turn them into julia-based distributions, and maximise the visibility of the changes so people would be encouraged to fix them, rather than maintain type-cleanliness.

src/univariate/continuous/beta.jl
richardreeve: This isn't my code, but I tend to agree with @simonbyrne - these are just internal to the sampler.

src/samplers/gamma.jl
richardreeve: Thanks! I have no idea what these numbers are, so hard to comment as the code comes from the other PR...

src/univariates.jl
richardreeve: I'm going ahead and changing it, but I remember my justification now - it was because the RNG was an argument and was being changed...

simonbyrne: Yeah I know it is a little weird. I can't find the original discussion right now, but it has become the convention for better or worse (see the [the `rand` docs](https://docs.julialang.org/en/v1/stdlib/Random/#Base.rand)). I guess the simplest argument is that anything involving random number will result in mutating the RNG, so it's not a helpful distinction.

src/common.jl
simonbyrne: Could you just override the methods for those specific distributions? Could define a simple macro
```julia
macro rand_rdist(D)
    esc(quote
        function rand(d::$D, n::Int)
            [rand(d) for i in Base.OneTo(n)]
        end
        function rand!(d::$D, X::AbstractArray)
            for i in eachindex(X)
                X[i] = rand(d)
            end
            return X
        end
    end)
end
```

simonbyrne: I guess my point is no one is going to be adding _more_ `DiscreteUnivariateRDist` distributions, so there isn't much to be gained by adding it to the type hierarchy.

src/samplers/gamma.jl
simonbyrne: These labels come from the original paper (and again, probably out of scope of the PR).

richardreeve: I've replaced the gotos and added one function (to calculate q), but I can't replace the labels as I don't know what they mean...

src/samplers/multinomial.jl
richardreeve: Done

src/testutils.jl
richardreeve: Done

src/univariates.jl
richardreeve: Absolutely, yes. Anyway, done now.

richardreeve:
@mschauer I'm willing to switch to `rand!!` if that's the consensus, but I don't really understand how that is more principled than just having a keyword argument that defaults to giving a deprecation warning? The problem with what you suggest is that everyone will have to change their behaviour twice, whereas with what I'm suggesting code only has to change once (by adding the correct keyword, which will become irrelevant, but that doesn't matter). I'm probably missing something obvious though...

In the meanwhile I've fixed everything else that people have mentioned except 1. the comments on the constants and the step labels and 2. the macro for R functions - I have no objection to it at all, but I'm not a macro expert so I'd rather someone (@simonbyrne?) told me what to do if that's the plan.

I've also implemented the keyword solution to the partially allocated array issue with an added `Distributions.rand_!()` function which gives the old behaviour without deprecation or keyword arguments. I can obviously switch to the `rand!!()` instead.

I've also added a 3D multiple matrix-variate option and an array-of-vectors option for multivariates (whose default is to not allocate) along with some appropriate testing for consistency (currently only implemented for multinomials). The only inconsistency I see is that  `rand(s, n)` returns a vector for univariates, a vector of matrices for matrix-variates and but a straight matrix for multivariates (rather than a vector of vectors), but that's the historical behaviour for fairly obvious reasons.

One final issue that has cropped up. I hadn't updated the truncated normal distribution, and I can't find the associated tests, which is probably why - are there any tests for that distribution in fact? If so, where, so I can fix them...

src/common.jl
richardreeve: As I say below, I'm happy to go with this, but I'm not really a macro person so you'll have to tell me exactly what to do :)

richardreeve:
An alternative `rand!` solution would be if `allocate=missing`, which is the default, then check using `isdefined` whether the sub-arrays are allocated and if they are to use them, if not allocate them. People could set `allocate=true` or `allocate=false` if they knew what they wanted to happen. Would that satisfy people - or is it worse as it is (lightly?) breaking as allocation won't happen the second time you call `rand!`?

mschauer:
What you propose would create reasonable behaviour. I am fine with all solutions, I mostly did not like `rand` to modify.

richardreeve:
Cool. Well in that case I've switched to this newer option. A v quick benchmark suggests it doubles the speed relative to the allocating version and is less than 1% slower than the non-allocating version through doing the checks.

src/common.jl
richardreeve: Okay, I've got the hang of this. Now implemented.

richardreeve:
Okay, @mschauer I think I've addressed all of your concerns, @simonbyrne I've now fixed the RFunction calls with your macro, and @matbesancon I've cleaned up the code for the gamma rng as much as I can (so it doesn't have gotos at least!), and the rest is in #835 now. I think I've now therefore addressed everyone's comments... are there any other issues I've missed?

simonbyrne:
Do you still need `_rand`? Can't we just use `rand` for those?

src/common.jl
```diff
@@ -120,3 +120,18 @@ succprob(d::DiscreteUnivariateDistribution)
 Get the probability of failure.
 """
 failprob(d::DiscreteUnivariateDistribution)
+
+# Temporary fix to handle RFunctions dependencies
+macro rand_rdist(D)
```
matbesancon: Add a docstring for this macro

src/univariate/continuous/beta.jl
matbesancon: ok to leave it as such if not belonging to the PR, but defining new behavior on samplers for example.

src/univariate/continuous/tdist.jl
```diff
@@ -73,7 +73,11 @@ end
 
 @_delegate_statsfuns TDist tdist ν
 
-rand(d::TDist) = StatsFuns.RFunctions.tdistrand(d.ν)
+function rand(rng::AbstractRNG, d::TDist)
+    z = randn(rng)
+    ν = d.ν
```
matbesancon: `ν = d.ν` is this step required? we could use `d.ν` directly in the line below

matbesancon:
Still 3 minor comments on my side, other than that the PR will be good to merge for me

src/univariate/continuous/tdist.jl
richardreeve: I was just copying the style of all of the other functions in that file that all stored the variable locally before using it multiple times... I presumed that there was some indirection cost in using the fields? Happy to remove if not...

richardreeve:
Thanks, I’ll get this sorted when I get to a computer. Having removed _rand, I’m now wondering whether _rand! is necessary though also?

src/common.jl
richardreeve: Done.

src/univariate/continuous/tdist.jl
richardreeve: I've done this now.

matbesancon:
no all good for me, this is some terrific work

matbesancon:
Merging in 24 hours if no objection

simonbyrne:
> I’m now wondering whether `_rand!` is necessary though also, or could it be removed? Is it just that's there's often no bounds checking in that code?

From what I can remember, that is the main difference. I would be happy to change it in favour of `@boundscheck` and friends (but feel free to do that in a different PR).

matbesancon:
+1 for other PR, let's merge this

richardreeve:
Hah, yes. That sounds like a plan to me!

matbesancon:
@simonbyrne good to merge?

matbesancon:
thanks again for the huge work @richardreeve

richardreeve:
No problem - happy to have helped! I needed it anyway - that's how this stuff works really :)

Anyway, hopefully no problems crop up from things that weren't being tested...

matbesancon:
I realized this morning I needed this PR also. Don't hesitate to make others for things we said were out of scope. And yes feel free to add some tests if you can

PR #814, opened 2019-01-17
Inversegaussian fit_mle

getzze:
Fit an InverseGaussian using sufficient statistics

PR #836, opened 2019-02-27
Fork2/inversegaussian fit

matbesancon:
continuing #814


matbesancon:
The last commit should probably have been in a separate PR, but Hypergeometric was curiously broken :/

src/univariate/continuous/inversegaussian.jl
```diff
@@ -160,3 +160,41 @@ function rand(rng::AbstractRNG, d::InverseGaussian)
     u = rand(rng)
     u >= p1 ? μ^2 / x1 : x1
 end
+
+#### Fit model
+
+"""
+Sufficient statustics for `InverseGaussian`, containing the weighted
```
mschauer:
```suggestion
Sufficient statistics for `InverseGaussian`, containing the weighted
```

src/univariate/continuous/inversegaussian.jl
```diff
@@ -160,3 +160,41 @@ function rand(rng::AbstractRNG, d::InverseGaussian)
...
+sum of observations, the weighted sum of inverse points and sum of weights
```
mschauer:
```suggestion
sum of observations, the weighted sum of inverse points and sum of weights.
```

matbesancon:
This one is pretty important, fixing a fail currently on master

mschauer:
I read trough it, but did not experiment with it. While reading I found no issues.

matbesancon:
The big point I'm wondering about is how did the breaking change arrive on master (there was a method defined elsewhere for entropy?)

mschauer:
`StatsBase` has entropy method for vectors which tries to iterate, this is the one which errored on the test I think.  That method  seems to before have been shadowed by another method which did not error for `Hypergeometric`, but which now ceased to exist?

matbesancon:
The one I wrote for Hypergeometric seems to be valid for all bounded discrete, I wasn't sure whether to define it as a fallback, given that all other tests passed

matbesancon:
My guess is it changed with the new release of StatsBase?

src/univariate/continuous/inversegaussian.jl
```diff
@@ -160,3 +160,41 @@ function rand(rng::AbstractRNG, d::InverseGaussian)
...
+    InverseGaussianStats(sx, sinvx, length(x))
+end
+
+function suffstats(::Type{InverseGaussian}, x::AbstractVector{<:Real}, w::AbstractVector{<:Real})
+    n = length(x)
+    if length(w) != n
+        throw(DimensionMismatch("Inconsistent argument dimensions."))
+    end
+
+    sx = sum(w .* x)
```
nalimilan: `sum(vw * vx for (vw, vx) in zip(w, x))` or just `dot(w, x)` should avoid an allocation. Same below.

nalimilan:
That's most probably https://github.com/JuliaStats/StatsBase.jl/pull/448. But I don't understand how it can have broken something here since the old method only accepted `AbstractArray` (so it wasn't relevant) and the new method is for `::Any` (so it has the lowest precedence). What method used to be called?

matbesancon:
yes the mystery is how did that test pass in the first place, it would at least have required some fallback method for distributions

src/univariate/continuous/inversegaussian.jl
```diff
@@ -160,3 +160,41 @@ function rand(rng::AbstractRNG, d::InverseGaussian)
...
+    sx::Float64      # (weighted) sum of x
+    sinvx::Float64   # (weighted) sum of 1/x
+    sw::Float64      # sum of sample weight
+
+    InverseGaussianStats(sx::Real, sinvx::Real, sw::Real) = new(sx, sinvx, sw)
+end
+
+function suffstats(::Type{InverseGaussian}, x::AbstractVector{<:Real})
+    sx = sum(x)
+    sinvx = sum(inv.(x))
```
tlienart: I think this can be replaced by `sum(inv, x)` and avoid allocating `inv.(x)`.

```julia
julia> x = randn(10_000) .+ 100;
julia> @btime sum(inv.($x))
  6.841 μs (2 allocations: 78.20 KiB)
100.02057059002993
julia> @btime sum(inv, x)
  6.146 μs (1 allocation: 16 bytes)
100.02057059002993
```

src/univariate/continuous/inversegaussian.jl
```diff
@@ -160,3 +160,41 @@ function rand(rng::AbstractRNG, d::InverseGaussian)
...
+end
+
+function suffstats(::Type{InverseGaussian}, x::AbstractVector{<:Real}, w::AbstractVector{<:Real})
+    n = length(x)
+    if length(w) != n
+        throw(DimensionMismatch("Inconsistent argument dimensions."))
+    end
+
+    sx = dot(w, x)
+    sinvx = dot(w, inv.(x))
```
tlienart: `sum(wi/xi for (wi, xi) in zip(w, x))` doesn't allocate

nalimilan: Maybe with `@inbounds`?

tlienart: Hmm it's a bit curious

```julia
julia> @btime sum(@inbounds $x[i]/$y[i] for i in eachindex($x))
  111.423 μs (2 allocations: 64 bytes)
100011.86751975746
@btime dot($x, inv.($y))
  74.382 μs (2 allocations: 781.33 KiB)
100011.86751975697
```

(the first one is pretty much identical to the zip case). I'm trying to read the `dot` code to see if I can figure it out, in the mean time I guess this is probably just fine 🙂

tlienart: (asked [on discourse](https://discourse.julialang.org/t/dot-x-inv-y-vs-non-allocating-one-the-former-is-faster/21338) I'll come back here if someone suggests a better approach, in the mean time I guess ignore my comment!)

mschauer: I cannot reproduce that the `zip` version is slower.

tlienart: fastest as per @KristofferC is probably to write out the loop at the expense of the one liners, something like:

```julia
sw = 0
swx = 0
swinvx = 0
@inbounds @simd for i in eachitem(w)
  sw += w[i]
  swx += w[i] * x[i]
  swinvx += w[i] / x[i]
end
```

KristofferC: FWIW, the initial `= 0` are probably makes this type unstable.

matbesancon: Same result as @tlienart 
```julia
julia> function f1(x, w)
       return dot(w, inv.(x))
       end
julia> function f2(x, w)
       return sum(wi/xi for (wi, xi) in zip(w, x))
       end
julia> function f3(x, w)
       return sum(w[i]/x[i] for i in eachindex(x))
       end
julia> function f4(x, w)
       @inbounds sum(w[i]/x[i] for i in eachindex(x))
       end

julia> begin
       const x = rand(1000_000)
       const w = rand(1000_000)
       @btime f1(x, w)
       @btime f2(x, w)
       @btime f3(x, w)
       @btime f4(x, w)
       end

julia> begin
       const x = rand(1000_000)
       const w = rand(1000_000)
       @btime f1(x, w)
       @btime f2(x, w)
       @btime f3(x, w)
       @btime f4(x, w)
       end
  2.395 ms (2 allocations: 7.63 MiB)
  1.130 ms (3 allocations: 64 bytes)
  1.118 ms (2 allocations: 64 bytes)
  1.116 ms (2 allocations: 64 bytes)
```

matbesancon: broadcast is allocating more but faster

tlienart: > FWIW, the initial `= 0` are probably makes this type unstable.

yeah sorry I was fixing that probably at the same time as you were typing 🙈

matbesancon: `@inbounds` is not consistently faster:
```julia
julia> begin
       const x = rand(10_000_000)
       const w = rand(10_000_000)
       @btime f1(x, w)
       @btime f2(x, w)
       @btime f3(x, w)
       @btime f4(x, w)
       end
WARNING: redefining constant x
WARNING: redefining constant w
  45.470 ms (2 allocations: 76.29 MiB)
  12.547 ms (3 allocations: 64 bytes)
  12.560 ms (2 allocations: 64 bytes)
  12.588 ms (2 allocations: 64 bytes)
```

tlienart: I think the for loop with @simd is a factor of 2 faster than your fastest one

matbesancon: might be yes, but not sure we want to add `@simd` here, it's implementation specific (this code should already have decent performance and be generic enough)

KristofferC: > it's implementation specific

What does this mean? `dot` is already using SIMD for example. `sum` uses SIMD for arrays (but cant for iterators which is why it is not as fast here).

matbesancon: From `@simd` doc, there is a pretty explicit warning that it is an experimental feature. Whatever happens under the belt, `dot` will likely keep the same signature and do what it expected.

nalimilan: I think `@simd` is OK in this kind of situation. Even if it's marked as experimental I don't see it going away before Julia 2.0 as it would break many packages.

matbesancon: I benchmarked the same functions with three others:
```julia
julia> function f5(x, w)
       s = zero(eltype(x))
       @inbounds @simd for i in eachindex(x)
       s += sum(w[i]/x[i])
       end
       end
julia> function f6(x, w)
       s = zero(eltype(x))
       @simd for i in eachindex(x)
       s += sum(w[i]/x[i])
       end
       end
julia> function f7(x, w)
       s = zero(eltype(x))
       @inbounds for i in eachindex(x)
       s += sum(w[i]/x[i])
       end
       end
```

```julia
julia> begin
       const x = rand(10_000_000)
       const w = rand(10_000_000)
       @btime f1(x, w)
       @btime f2(x, w)
       @btime f3(x, w)
       @btime f4(x, w)
       @btime f5(x, w)
       @btime f6(x, w)
       @btime f7(x, w)
end
  38.772 ms (2 allocations: 76.29 MiB)
  13.260 ms (3 allocations: 64 bytes)
  13.716 ms (2 allocations: 64 bytes)
  13.384 ms (2 allocations: 64 bytes)
  1.539 ns (0 allocations: 0 bytes)
  5.142 ms (0 allocations: 0 bytes)
  1.586 ns (0 allocations: 0 bytes)
```

Testing bigger vectors:
```julia
julia> begin
       const x = rand(100_000_000)
       const w = rand(100_000_000)
       @btime f1(x, w)
       @btime f2(x, w)
       @btime f3(x, w)
       @btime f4(x, w)
       @btime f5(x, w)
       @btime f6(x, w)
       @btime f7(x, w)
       end
WARNING: redefining constant x
WARNING: redefining constant w
  379.512 ms (2 allocations: 762.94 MiB)
  139.878 ms (3 allocations: 64 bytes)
  137.298 ms (2 allocations: 64 bytes)
  135.333 ms (2 allocations: 64 bytes)
  1.586 ns (0 allocations: 0 bytes)
  51.724 ms (0 allocations: 0 bytes)
  1.588 ns (0 allocations: 0 bytes)
```
`@inbounds @simd` seems to be a bit faster than `@inbounds alone`

matbesancon: changed in last commit

src/univariate/continuous/inversegaussian.jl
```diff
@@ -186,10 +186,14 @@ function suffstats(::Type{InverseGaussian}, x::AbstractVector{<:Real}, w::Abstra
     if length(w) != n
         throw(DimensionMismatch("Inconsistent argument dimensions."))
     end
-
-    sx = dot(w, x)
-    sinvx = sum(w[i]/x[i] for i in eachindex(x))
-    sw = sum(w)
+    sx = zero(eltype(x))
+    sinvx = zero(eltype(x))
+    sw = zero(eltype(x))
```
tlienart: sorry to be a pain but I think this is potentially problematic in pathological cases where for instance `x` is given in floating point and `w` in Integer which is allowed by the function signature afaik. I think what I had written may be preferable i.e. (using your notations this time):

```
sw = zero(eltype(w))
sx = zero(one(eltype(w)) * one(eltype(x)))
sinvx = zero(one(eltype(w))/one(eltype(x)))
```

nalimilan: Makes sense. Though better use `zero` directly instead of `zero(one(...))`.

src/univariate/continuous/inversegaussian.jl
```diff
@@ -160,3 +160,45 @@ function rand(rng::AbstractRNG, d::InverseGaussian)
...
+"""
+Sufficient statistics for `InverseGaussian`, containing the weighted
+sum of observations, the weighted sum of inverse points and sum of weights.
+"""
+struct InverseGaussianStats <: SufficientStats
+    sx::Float64      # (weighted) sum of x
+    sinvx::Float64   # (weighted) sum of 1/x
+    sw::Float64      # sum of sample weight
+
+    InverseGaussianStats(sx::Real, sinvx::Real, sw::Real) = new(sx, sinvx, sw)
```
nalimilan: AFAICT this innner constructor is the same as the default one, so better omit it.

src/univariate/continuous/inversegaussian.jl
mschauer: Just writing a generator would have the advantage of being easier maintenance.

nalimilan: Yeah, it depends on whether performance really matters in that particular piece of code.

matbesancon: agreed on principle with @mschauer, but yeah this code might be used with lots of data points and the difference is significant (see benchmarks).
`promote_type` will help make it cleaner

matbesancon: and no problem @tlienart, in that case, pain is the path to correctness

PR #838, opened 2019-02-28
fix notation in laplace.jl docstring

JeffFessler:


matbesancon:
I'm fine to keep these conventions (better than being inconsistent), but [a great place of knowledge](https://en.wikipedia.org/wiki/Laplace_distribution) uses `b` instead of beta

JeffFessler:
I am fine with `b` too - feel free to revert.  I think I saw \beta in some places and \theta in others (and b in others) so I was just pursuing self-consistency.
I looked in several probability texts on my shelf and most of them write it in terms of \lambda = 1/b and one book used \alpha=1/b so there is no standard out there it seems. Cheers.

matbesancon:
Just waiting for another opinion and this is merged, thanks for the contribution!

matbesancon:
(Travis is failing while #836 not merged)

simonbyrne:
Should it be sigma?

matbesancon:
doesn't seem to be the convention for Laplace

simonbyrne:
we should be consistent for all location/scale families:
- Cauchy
- Logistic
- SymTriangular
- Biweight
- Triweight
- Epanechnikov
- Cosine

matbesancon:
This part will be simpler for the construction with #823, but without it, it may be dangerous to require a parameter which is not the one usually taken

simonbyrne:
I’m no sure what you mean?

JeffFessler:
I just noticed that cauchy.jl has similar inconsistencies in its docstring.
Perhaps @simonbyrne is suggesting that all docstrings should be made more self consistent.
I agree with the goal of self consistency within each docstring, but I also agree with @matbesancon that trying to unify notation across all distributions could be problematic.  For example, I see that the docstring for  cauchy.jl uses \mu and \sigma but that is nonstandard and possibly misleading notation because Cauchy does not have a mean or variance.

matbesancon:
I'd be for keeping letters different from `mu, sigma` when the location and scale parameters do not represent mean, std directly

matbesancon:
restarted the build, merging as everything goes green

matbesancon:
@simonbyrne can you restart appveyor here?

PR #842, opened 2019-03-16
add mean for truncated exponential

tpapp:


mschauer:
What about the cases e.g.
```
mean(Truncated(Exponential(1.0), -3.0, -1.0)
mean(Truncated(Exponential(1.0), -3.0, -3.0)
```
Should mean error?

tpapp:
The first leads to a nonsensical distribution that should be caught by `Truncated` (happy to make a PR, but IMO that is orthogonal to this one, I opened #843 for it), while the second one already errors.
```julia
julia> d = Truncated(Exponential(1.0), -3.0, -1.0) # THIS SHOULD ERROR
Truncated(Exponential{Float64}(θ=1.0), range=(-3.0, -1.0))

julia> extrema(d) # because it is nonsensical
(0.0, -1.0)

julia> mean(Truncated(Exponential(1.0), -3.0, -3.0))
ERROR: lower bound should be less than upper bound.
Stacktrace:
 [2] Truncated(::Exponential{Float64}, ::Float64, ::Float64) at /home/tamas/.julia/dev/Distributions/src/truncate.jl:25
 [3] top-level scope at REPL[9]:1
```

src/truncated/exponential.jl
```diff
@@ -0,0 +1,19 @@
+#####
+##### Truncated exponential distribition
+#####
+
+function mean(d::Truncated{<:Exponential,Continuous})
+    θ = d.untruncated.θ
+    l, r = extrema(d)           # l is always finite
+    if isfinite(r)
+        if l == r
```
matbesancon: shouldn't that be `\approx` instead of equality?

src/truncated/exponential.jl
```diff
@@ -0,0 +1,19 @@
...
+            r
```
matbesancon: with 2 embedded if else, maybe return statements would be better

src/truncated/exponential.jl
tpapp: I don't see why. It is just there to protect from the really degenerate case, the formula with the midpoint should work fine for `r ≈ l` (see the tests, where I check this).

src/truncated/exponential.jl
tpapp: I prefer the lispy style, and would prefer not to change this unless this is the only thing that prevents this PR from being merged.

src/truncated/exponential.jl
mschauer: I think this might be fine. The only way it could go wrong is when  l - r is small but not zero and theta is unreasonable large, e.g. `l-r = eps()`, `theta = 1e308`

tpapp: You are right, I added a case for that. If CI passes, I wonder if this could be merged now (and also please tag a minor release).

tpapp:
Friendly bump: can you please merge this? And possibly tag a minor version.

src/truncated/exponential.jl
```diff
@@ -6,8 +6,8 @@ function mean(d::Truncated{<:Exponential,Continuous})
     θ = d.untruncated.θ
     l, r = extrema(d)           # l is always finite
     if isfinite(r)
-        if l == r
-            r
+        if isapprox(l, r, rtol = √eps() * θ, atol = iszero(l) ? √eps() : 0.0)
```
mschauer: Could you add a comment regarding the interplay between ˋrtolˋ and ˋatolˋ here?

tpapp: I may have been overcomplicating it; I simplified and added a comment.

PR #847, opened 2019-03-25
Add special case for truncating Uniform.

tpapp:
This is implemented by making `Truncated(::Uniform, ...)` return a
`Uniform`.

The advantage is that no other functionality needs to be duplicated.

A minor style infraction is a `T(...)` function not constructing a `T`
object, but I think in this case this is a reasonable compromise.

### DIFF
--- a/test/truncate.jl
+++ b/test/truncate.jl
@@ -38,13 +38,14 @@ function verify_and_test_drive(jsonfile, selected, n_tsamples::Int,lower::Int,up
 
         println("    testing Truncated($(ex),$lower,$upper)")
         d = Truncated(eval(Meta.parse(ex)),lower,upper)
-        if dtype != TruncatedNormal
-            @assert isa(dtype, Type) && dtype <: UnivariateDistribution
-            @test isa(d, dtypet)
+        if dtype != Uniform # Uniform is truncated to Uniform
+            if dtype != TruncatedNormal
+                @assert isa(dtype, Type) && dtype <: UnivariateDistribution
+                @test isa(d, dtypet)
+            end
+            # verification and testing
+            verify_and_test(d, dct, n_tsamples)
         end
-
-        # verification and testing
-        verify_and_test(d, dct, n_tsamples)
     end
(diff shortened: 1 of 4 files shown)


mschauer:
There is a test against this currently: JuliaStats/Distributions.jl/test/truncate.jl:43

tpapp:
Good point, I special-cased it. If some other solution would be preferred, please let me know.

tpapp:
Anything else I can do to get this merged?

test/truncate.jl
```diff
@@ -38,13 +38,14 @@ function verify_and_test_drive(jsonfile, selected, n_tsamples::Int,lower::Int,up
 
         println("    testing Truncated($(ex),$lower,$upper)")
         d = Truncated(eval(Meta.parse(ex)),lower,upper)
-        if dtype != TruncatedNormal
-            @assert isa(dtype, Type) && dtype <: UnivariateDistribution
-            @test isa(d, dtypet)
+        if dtype != Uniform # Uniform is truncated to Uniform
```
matbesancon: just a minor comment, is this part of the tests required? The one in truncated_uniform.jl seems sufficient

matbesancon:
Added a minor comment but nothing preventing merging

test/truncate.jl
tpapp: I was trying to be conservative and only disable what was causing a problem.

matbesancon: is it still required now if the bug is gone?

tpapp: Sorry, I don't know which bug you are referring to.

But if I allow these tests for Uniform then on this branch the test suite fails. Did something change in the meantime?

matbesancon: sorry, I thought the modification of the test was to debug something. All good

matbesancon:
merging in 12 hours if no objection

matbesancon:
thanks @mschauer, I should have a bot ping me every time I write "merge in X"

Issue #850, opened 2019-03-27
support LogitNormal distribution


bgctw:
The LogitNormal distribution is important to my Model-Data Integration studies as a transformation of parameters bounded (0,1) in frameworks that work with normally distributed data.
Its also used in other applications:
https://en.wikipedia.org/wiki/Log-normal_distribution#Occurrence_and_applications

Would be nice to support it with Distributions.jl

PR #851, opened 2019-03-27
implement LogitNormal univariate distribution

bgctw:
I started an implementation of the LogitNormal distribution by adapting lognormal.jl, which provides the pdf and cumulative functionality. There is no analytical solution for the moments of the LogitNormal, but I intend to add numerical approximations in future.

#850


matbesancon:
Sorry about the delay on this. I think we should not make Distributions depend on more external packages. (the goal would be to remove more than we add at this point). I would be ok to implement `LogitNormal` and leave all non-analytical functions as error. The other solution would be to build a package with a `LogitNormal` type which implements the `Distributions` interface

bgctw:
I removed the numerical code that depends on Optim, and moved those functions to a future package [NormalTransforms](https://git.bgc-jena.mpg.de/twutz/NormalTransforms.jl).

In this way, a broader community can benefit from the partial implementation of the logitnormal distribution with default availability with Distributions.jl. For many use cases,  the mode and the quantiles are sufficient.

test/logitnormal.jl
```diff
@@ -0,0 +1,70 @@
+# Tests on Univariate LogitNormal distributions
+
+using Distributions
+using Random, Test
+using StatsFuns
+
+
+function tmpF()
```
matbesancon: is this needed?

bgctw: I removed the `tmpF()` where I stored interactive code.
The usage of `StatsFuns` is necessary for the logistic function used at line 21.

src/univariate/continuous/logitnormal.jl
```diff
@@ -0,0 +1,174 @@
...
+# stdlogitx(d::LogitNormal) = d.σ
+
+function mean(d::LogitNormal; kwargs...) 
+    # estimateMean(d,kwargs...) # moved to NormalTransforms
+    @warn "Mean of logitnomral currently not implemented. Return missing"
+    missing
+end
+
+median(d::LogitNormal) = logistic(d.μ)
+function mode(d::LogitNormal)
```
matbesancon: I don't think we should even implement them here, that way it can be implemented in a new package. Otherwise which method is used between this one and the new "real" method is not deterministic

```julia
function mean(d::LogitNormal; kwargs...) 
    # estimateMean(d,kwargs...) # moved to NormalTransforms
    @warn "Mean of logitnomral currently not implemented. Return missing"
    missing
end 
```

matbesancon:
Regarding my last comment, this is not the expected behaviour of missing (used for missing data, not missing implementations). Better having a `MethodError`, which is what will happen if it is not implemented

src/univariate/continuous/logitnormal.jl
bgctw: This means, The recommended way is to remove those methods completely? 
And maybe leave a comment in the code?

src/univariate/continuous/logitnormal.jl
```diff
@@ -0,0 +1,153 @@
...
+@inline partype(d::LogitNormal{T}) where {T<:Real} = T
+
+#### Statistics
+
+# meanlogitx(d::LogitNormal) = d.μ
+# varlogitx(d::LogitNormal) = abs2(d.σ)
+# stdlogitx(d::LogitNormal) = d.σ
+
+# mean, mode, and variance 
+# moved to NormalTrnasforms because
```
matbesancon:
```suggestion
# moved to NormalTransforms because
```

src/univariate/continuous/logitnormal.jl
```diff
@@ -0,0 +1,153 @@
...
+# they depend on Optim package.
```
matbesancon:
```suggestion
# they depend on the Optim package.
```

matbesancon:
looks good, merging in 12 hours if no objection

bgctw:
Thanks, I have no objections.

matbesancon:
@bgctw totally forgot, could you add the distribution in the corresponding bloc in `docs/src/univariate.md`?

matbesancon:
@bgctw I was thinking if some maintainers have no objection, but thanks :smile:

src/univariate/continuous/logitnormal.jl
```diff
@@ -0,0 +1,153 @@
+#using Optim # numerical estimation routines moved to NormalTransforms
+
+"""
+    LogitNormal(μ,σ)
+
+The *logit normal distribution* is the distribution of the logistic function
```
mschauer: bad wording, it is not the distribution of a function `f` but the distribution of a transformed variable `f(X)`

src/univariate/continuous/logitnormal.jl
```diff
@@ -0,0 +1,153 @@
...
+# not implemented
+
+median(d::LogitNormal) = logistic(d.μ)
+
+#### Evalution
+
+#TODO check pd and logpdf
+function pdf(d::LogitNormal{T}, x::Real) where T<:Real 
+    if zero(x) < x < one(x) 
+        return normpdf(d.μ, d.σ, logit(x)) / x / (1-x)
```
mschauer: Does it numerically make sense to use `a/(b*c)` instead of `a/b/c`?

matbesancon: a/b/c always bugs me, never sure in which order to take it

src/univariate/continuous/logitnormal.jl
```diff
@@ -131,7 +136,7 @@ invlogccdf(d::LogitNormal, lq::Real) = logistic(norminvlogccdf(d.μ, d.σ, lq))
 function gradlogpdf(d::LogitNormal{T}, x::Real) where T<:Real
     #TODO check
     (μ, σ) = params(d)
-    0 < x < 1 ? - ((log(x) - μ) / (σ^2) + 1) / x /(1-x) : zero(T)
+    0 < x < 1 ? - ((log(x) - μ) / ((σ^2) + 1) * x * (1-x)) : zero(T)
```
mschauer: Huh?

matbesancon:
thanks @bgctw!

mschauer:
@matbesancon Is there a change to runtests.jl missing?

matbesancon:
oh boy, yes absolutely

PR #615, opened 2017-05-27
Generic support for uni gmm

xukai92:
Add generic support for uni-gmm

PR #853, opened 2019-03-28
Make return type same as parameter instead of data

xukai92:
A small fix after https://github.com/JuliaStats/Distributions.jl/pull/615

Basically the current behaviour of `logpdf(mixture, data)` is to construct an intermediate vector with type `eltype(data)` and accumulate log probabilities. This might broke when data and mixture has different types, e.g. mixture is of `ForwardDiff.Dual` due to AD and data is of `Float64`.

### DIFF
--- a/src/mixtures/mixturemodel.jl
+++ b/src/mixtures/mixturemodel.jl
@@ -328,7 +328,7 @@ function _mixpdf!(r::AbstractArray, d::AbstractMixtureModel, x)
     p = probs(d)
     @assert length(p) == K
     fill!(r, 0.0)
-    t = Array{Float64}(undef, size(r))
+    t = Array{eltype(p)}(undef, size(r))
     for i = 1:K
         @inbounds pi = p[i]
         if pi > 0.0
@@ -359,7 +359,7 @@ function _mixlogpdf1(d::AbstractMixtureModel, x)
     p = probs(d)
     @assert length(p) == K
 
-    lp = Vector{eltype(x)}(undef, K)
+    lp = Vector{eltype(p)}(undef, K)
     m = -Inf   # m <- the maximum of log(p(cs[i], x)) + log(pri[i])
     for i = 1:K
         @inbounds pi = p[i]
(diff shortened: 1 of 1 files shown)


xukai92:
@matbesancon Obviously we didn't catch this before because we didn't test mixtures against ForwardDiff. Do we want this in Distributions.jl though?

src/mixtures/mixturemodel.jl
```diff
@@ -386,7 +386,7 @@ function _mixlogpdf!(r::AbstractArray, d::AbstractMixtureModel, x)
     p = probs(d)
     @assert length(p) == K
     n = length(r)
-    Lp = Matrix{eltype(x)}(undef, n, K)
+    Lp = Matrix{eltype(p)}(undef, n, K)
     m = fill(-Inf, n)
     for i = 1:K
```
matbesancon: shouldn't `@inbounds` be on the for statement?

matbesancon:
I think this is fine in the scope of Distributions

src/mixtures/mixturemodel.jl
matbesancon: also `for i in eachindex(p)` is faster

xukai92: > shouldn't @inbounds be on the for statement?

Hah. I guess that's a historical typo. Nice catch.

> also for i in eachindex(p) is faster

 Got it.

xukai92:
Comment addressed and test fixed.

matbesancon:
merging in 24 hours if no objection

src/mixtures/mixturemodel.jl
nalimilan: That's the same thing as long as `@inbounds` is applied to all `getindex` calls. Also, `eachindex` isn't faster for `Array`, but it's more generic.

matbesancon: good catch, I always thought Base.OneTo(n) was faster

Issue #841, opened 2019-03-13
Sparse matrix inversion when sampling MvNormalCanon


ElOceanografo:
After finding a [separate issue](https://github.com/JuliaStats/PDMats.jl/issues/87) in PDMats, I noticed that generating random samples from an `MvNormalCanon` with a `PDSparseMat` precision matrix currently does a dense inversion of the precision matrix, which is potentially very expensive and [shouldn't be necessary](https://math.stackexchange.com/questions/1845132/sample-points-from-a-multivariate-normal-distribution-using-only-the-precision-m/2489142)...it looks like there's no `unwhiten_winv!` method for sparse matrices, so it falls back to the one for `AbstractPDMat`: https://github.com/JuliaStats/Distributions.jl/blob/a6a9ef2799e8772feb0809b4396149d5b6e9a244/src/multivariate/mvnormalcanon.jl#L174

I was wondering if this was a conscious design decision, or an oversight?  Desired working example is  below, though note this doesn't actually work at the moment because of https://github.com/JuliaStats/PDMats.jl/issues/87

```julia
using LinearAlgebra, SparseArrays
using Distributions
using PDMats

J = sprand(10, 10, 0.05)
J = J'J
J[diagind(J)] .= 1
μ = randn(10)
h = J \ μ
d = MvNormalCanon(μ, h, PDSparseMat(J))
rand(d)
```

matbesancon:
Closed by #855

matbesancon:
@JuliaRegistrator register()

matbesancon:
@JuliaRegistrator register()

PR #855, opened 2019-04-05
Add unwhiten_winv! method for PDSparseMats

ElOceanografo:
This is in response to issue https://github.com/JuliaStats/Distributions.jl/issues/841, and adds a method to `unwhiten_winv!` for `PDSparseMat`s.  This lets you sample from an `MvNormalCanon` with a sparse precision matrix (without having to invert the whole thing).

I have not added tests yet, because a number of them require inverting the precision matrix `J`, which (a) runs into [this issue](https://github.com/JuliaStats/PDMats.jl/issues/87), and (b) it's not clear if it should be allowed, since the situations where you want a sparse precision matrix will often be the ones where a full matrix wouldn't fit in memory.  For comparison, `LinearAlgebra.inv` throws an error when you try to apply it to a sparse matrix.  Suggestions on this front welcome...


matbesancon:
Fixes #841

matbesancon:
Code looks fine, but yeah we need to test this. Why can't we set the same seed, generate samples from two matrices (one sparse PDMat, one dense) and compare results?

matbesancon:
ping @ElOceanografo

ElOceanografo:
Sorry, got busy with other stuff.  Happy to add a test like that.  Was also thinking it would be good to add a line in [here](https://github.com/JuliaStats/Distributions.jl/blob/d1053bd4bec94b9ed6ab21fa341f04a851f072d0/test/mvnormal.jl#L112) to test all the functions of an `MvNormalCanon` with a sparse precision matrix, hence my question about the desired behavior for `cov`.  That's maybe a separate issue though.

matbesancon:
let's leave that for another issue and just add the tests for now

src/multivariate/mvnormalcanon.jl
```diff
@@ -174,6 +174,7 @@ sqmahal!(r::AbstractVector, d::MvNormalCanon, x::AbstractMatrix) = quad!(r, d.J,
 unwhiten_winv!(J::AbstractPDMat, x::AbstractVecOrMat) = unwhiten!(inv(J), x)
 unwhiten_winv!(J::PDiagMat, x::AbstractVecOrMat) = whiten!(J, x)
 unwhiten_winv!(J::ScalMat, x::AbstractVecOrMat) = whiten!(J, x)
+unwhiten_winv!(J::PDSparseMat, x::AbstractVecOrMat) = x[:] = J.chol.L' \ x
```
ararslan: Isn't the lower triangular matrix _L_ from a Cholesky decomposition the transpose of the upper triangular matrix _U_? Seems like this could instead be written
```suggestion
unwhiten_winv!(J::PDSparseMat, x::AbstractVecOrMat) = x[:] = J.chol.U \ x
```
unless I'm missing something.

ElOceanografo: You're not missing anything, this is correct!

ElOceanografo:
Added a couple of tests to confirm that `MvNormal` and `MvNormalCanon` produce the same random samples when they have the same `J = inv(Σ)` matrix and the same randoms seed.  It works, but should be noted that the two samples match each other less precisely as the dimension of the precision/covariance matrix increases (I assume because of floating point error that accumulates during inversion).

test/mvnormal.jl
```diff
@@ -4,6 +4,7 @@ import PDMats: ScalMat, PDiagMat, PDMat
 
 using Distributions
 using LinearAlgebra, Random, Test
+using SparseArrays
```
matbesancon: @ElOceanografo this is the cause for CI failing, you need to add SparseArrays to the test in `Project.toml`

matbesancon:
@ElOceanografo could you merge master into your branch?

Project.toml
```diff
@@ -9,6 +9,7 @@ PDMats = "90014a1f-27ba-587c-ab20-58faa44d9150"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 QuadGK = "1fd47b50-473d-5c70-9696-f719f8f3bcdc"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
```
matbesancon: @ElOceanografo you move the line to the `compat` section, and add SparseArrays to the `test` target below

matbesancon:
Looks good, merging in few days if no objection, although @andreasnoack you opinion would be helpful here

matbesancon:
@ElOceanografo can you merge master into this branch? I think we'll be good to go then

matbesancon:
@ElOceanografo the CI is failing, re-merge master and test locally to see what's happening

ElOceanografo:
Finally got around to fixing the test failures, it was just some dependency changes on the master branch I hadn't pulled. I think this is ready to merge now... @matbesancon

matbesancon:
Great, thanks a lot for the work, I'll review tomorrow

ElOceanografo:
Another bump :) @matbesancon @johnczito @andreasnoack @simonbyrne @ararslan

Project.toml
```diff
@@ -5,17 +5,21 @@ version = "0.23.9"
...
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 
 [compat]
+
```
matbesancon, line 22:
```suggestion
```
I don't think there should be a white line before the first entry of a TOML

test/mvnormal.jl
```diff
@@ -147,6 +148,8 @@ end
     end
 end
 
+
```
matbesancon, line 151:
```suggestion
```

Project.toml
```diff
@@ -5,17 +5,21 @@ version = "0.23.9"
 
 [deps]
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
+HypothesisTests = "09f84164-cd44-5f33-b23f-e6b0d136a0d5"
```
matbesancon, line 8: It does not seem like you are using `HypothesisTests` in the PR?

ElOceanografo: No, I think I added this here by mistake.

matbesancon (approved):

matbesancon:
LGTM, will merge tomorrow when all green

PR #857, opened 2019-04-09
Promote LocationScale parameters when they are different types

wkearn:
I ran into the following problem using ForwardDiff with `LocationScale` distributions. I added a method for `LocationScale` which promotes the parameters when they are different types, like the other univariate distributions.

```julia
> using ForwardDiff, Distributions
> f(θ) = logpdf(LocationScale(1.0,θ,TDist(3.0)),10.0)
> ForwardDiff.derivative(f,1.0)
ERROR: MethodError: no method matching LocationScale(::Float64, ::ForwardDiff.Dual{ForwardDiff.Tag{typeof(f),Float64},Float64,1}, ::TDist{Float64})
Closest candidates are:
  LocationScale(::T<:Real, ::T<:Real, ::D<:Distribution{Univariate,Continuous}) where {T<:Real, D<:Distribution{Univariate,Continuous}} at /Users/wkearn/.julia/dev/Distributions/src/univariate/continuous/locationscale.jl:29
  LocationScale(::T<:Integer, ::T<:Integer, ::D<:Distribution{Univariate,Continuous}) where {T<:Integer, D<:Distribution{Univariate,Continuous}} at /Users/wkearn/.julia/dev/Distributions/src/univariate/continuous/locationscale.jl:30
Stacktrace:
 [1] f(::ForwardDiff.Dual{ForwardDiff.Tag{typeof(f),Float64},Float64,1}) at ./REPL[2]:1
 [2] derivative(::typeof(f), ::Float64) at /Users/wkearn/.julia/packages/ForwardDiff/N0wMF/src/derivative.jl:14
 [3] top-level scope at none:0
```


matbesancon:
Given that tests in Distributions are already using ForwardDiff, could you add a test for the new behaviour? (One or two will suffice)

matbesancon:
ping @wkearn

matbesancon:
I relaxed the tests that assumed `Float64`

test/locationscale.jl
```diff
@@ -14,15 +9,15 @@ function test_location_scale_normal(μ::Float64,σ::Float64,μD::Float64,σD::Fl
     @test maximum(d) == maximum(dref)
     @test extrema(d) == (minimum(d), maximum(d))
 
-    #### Conversions
+    #### Promotions and conversions
 
-    @test convert(LocationScale{Float16}, μ, σ, ρ) == LocationScale{Float16,typeof(ρ)}(Float16(μ), Float16(σ), ρ)
-    @test convert(LocationScale{Float16}, d) == LocationScale{Float16,typeof(ρ)}(Float16(μ), Float16(σ), ρ)
+    @test typeof(d.µ) === typeof(d.σ)
+    # @test convert(LocationScale{Float16}, μ, σ, ρ) == LocationScale{Float16,typeof(ρ)}(Float16(μ), Float16(σ), ρ)
```
mschauer: Outcommented tests?

matbesancon: yes sorry I should remove those. This convert is not valid in the general sense, not all Real numbers can be cast into a `Float` in the general case

Issue #852, opened 2019-03-28
Missing cdf, ccdf, logcdf and logccdf functions for the Skellam distribution


id0158:
It seems that anything else than the `pdf` function is not implemented for the `Skellam` distribution.
```julia
julia> cdf(Skellam(1, 2), 3)
ERROR: MethodError: no method matching cdf(::Skellam{Float64}, ::Int64)
```

My Julia installation :
```julia
julia> versioninfo()
Julia Version 1.0.3
Commit 099e826241 (2018-12-18 01:34 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-6600U CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.0 (ORCJIT, skylake)
Environment:
  JULIA_NUM_THREADS = 2
  JULIA_PKG_SERVER = https://pkg.juliacomputing.com/
  JULIA_PKG_TOKEN_PATH = /home/id0158/.juliapro/token.toml
```

matbesancon:
It's a good question, there doesn't seem to be a defined CDF for the Skellam dist. I found the scipy implementation, other than that nothing. If you find some reference paper on it we can add the feature

id0158:
Indeed, SciPy is where I come from.
But it seems that the [scipy implementation](https://scipy-user.scipy.narkive.com/pW7NYexA/the-skellam-distribution) itself is/was based on the [R one](https://cran.r-project.org/web/packages/skellam/index.html).

Hope that helps, didn't find anything more helpful.

simonbyrne:
SciPy seems to use the CDF of the non-central chi square (which we have available via Rmath.jl).

Sumegh-git:
We can add the CDF now based on the non-central chi square CDF maybe? 
Because I too couldn't find any proper efficient implementation without using the CDF above.

matbesancon:
Fixed by #859

PR #859, opened 2019-04-17
Added cdf for Skellam

Sumegh-git:
Implemented CDF for Skellam based on SciPy one - https://github.com/scipy/scipy/blob/v0.15.1/scipy/stats/_discrete_distns.py

Referring to #852

### DIFF
--- a/src/univariate/discrete/skellam.jl
+++ b/src/univariate/discrete/skellam.jl
@@ -1,3 +1,5 @@
+using Rmath: pnchisq
+
 """
     Skellam(μ1, μ2)
 
@@ -80,13 +82,15 @@ function cf(d::Skellam, t::Real)
     exp(μ1 * (cis(t) - 1) + μ2 * (cis(-t) - 1))
 end
 
-cdf(d::Skellam, x::Int) = throw(MethodError(cdf, (d, x)))
-cdf(d::Skellam, x::Real) = throw(MethodError(cdf, (d, x)))
+function cdf(d::Skellam, t::Real)
+    μ1, μ2 = params(d)
+    t = floor(t)
+    (t < 0) ? pnchisq(2*μ2, -2*t, 2*μ1) : 1.0 - pnchisq(2*μ1, 2*(t+1), 2*μ2)
+end
+
+cdf(d::Skellam, t::Int) = cdf(d, float(t))
 
 #### Sampling
-# TODO: remove RFunctions dependency once Poisson has its removed
-@rand_rdist(Skellam)
+
 rand(d::Skellam) = rand(Poisson(d.μ1)) - rand(Poisson(d.μ2))
 
-rand(rng::AbstractRNG, d::Skellam) =
-    rand(rng, Poisson(d.μ1)) - rand(rng, Poisson(d.μ2))
--- a/src/univariate/discrete/skellam.jl
+++ b/src/univariate/discrete/skellam.jl
@@ -86,8 +86,12 @@ end
     cdf(d::Skellam, t::Real)
 
 Implementation based on SciPy: https://github.com/scipy/scipy/blob/v0.15.1/scipy/stats/_discrete_distns.py 
-(OR)
-Refer to Eqn (5) in On an Extension of the Connexion Between Poisson and χ2 Distributions, N.L Johnson, Vol 46, No 3/4(1959), doi:10.2307/2333532 which relates the Skellam and Non-central chisquare PDFs, which is very similar to their CDFs computation as well.
+
+Refer to Eqn (5) in On an Extension of the Connexion Between Poisson and χ2 Distributions, N.L Johnson(1959)
+Vol 46, No 3/4, doi:10.2307/2333532 
+It relates the Skellam and Non-central chisquare PDFs, which is very similar to their CDFs computation as well.
+
+Computing cdf of the Skellam distribution.
 """
 function cdf(d::Skellam, t::Real)
     μ1, μ2 = params(d)
--- a/src/univariate/discrete/skellam.jl
+++ b/src/univariate/discrete/skellam.jl
@@ -81,7 +81,7 @@ function cf(d::Skellam, t::Real)
     μ1, μ2 = params(d)
     exp(μ1 * (cis(t) - 1) + μ2 * (cis(-t) - 1))
 end
-
+#based on SciPy: https://github.com/scipy/scipy/blob/v0.15.1/scipy/stats/_discrete_distns.py
 function cdf(d::Skellam, t::Real)
     μ1, μ2 = params(d)
     t = floor(t)
--- a/src/univariate/discrete/skellam.jl
+++ b/src/univariate/discrete/skellam.jl
@@ -96,7 +96,7 @@ Computing cdf of the Skellam distribution.
 function cdf(d::Skellam, t::Real)
     μ1, μ2 = params(d)
     t = floor(t)
-    (t < 0) ? nchisqcdf(2*μ2, -2*t, 2*μ1) : 1.0 - nchisqcdf(2*μ1, 2*(t+1), 2*μ2)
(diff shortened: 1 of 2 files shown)


mschauer:
Should this directly use `RMath` or should `pnchisq` be made available through one of the dependencies, e.g. `StatsFuns.jl`? A comment should refer to https://github.com/scipy/scipy/blob/v0.15.1/scipy/stats/_discrete_distns.py for later reference.

Sumegh-git:
Yes, added that @mschauer .

src/univariate/discrete/skellam.jl
```diff
@@ -79,9 +81,14 @@ function cf(d::Skellam, t::Real)
     μ1, μ2 = params(d)
     exp(μ1 * (cis(t) - 1) + μ2 * (cis(-t) - 1))
 end
+#based on SciPy: https://github.com/scipy/scipy/blob/v0.15.1/scipy/stats/_discrete_distns.py
```
simonbyrne: i'd be keen to see a better reference or derivation of this, if anyone knows of one.

Sumegh-git: You can see Eqn. (5) in https://www.jstor.org/stable/2333532?seq=1#metadata_info_tab_contents
It relates the skellam pdf(difference of two independent poisson r.v's) with noncentral chi-square pdf.

Sumegh-git: @simonbyrne Or maybe simply put x=2*lambda1 ; k=2*(x+1) ; lambda = 2*lambda2 in the PDF given here at https://en.wikipedia.org/wiki/Noncentral_chi-squared_distribution.
That will yield the Skellam PDF. 
The CDF also looks exactly similar with a change of 2*x instead of 2*(x+1) in the args which I believe might be done due to some precision issues/continuity correction

matbesancon: @Sumegh-git you should put the whole reference and computation in docstring form:
```julia
"""
    cdf(::Skellam, x)

Add the link, add the explanation
"""
```

src/univariate/discrete/skellam.jl
```diff
@@ -79,9 +81,14 @@ function cf(d::Skellam, t::Real)
...
+function cdf(d::Skellam, t::Real)
+    μ1, μ2 = params(d)
+    t = floor(t)
+    (t < 0) ? nchisqcdf(2*μ2, -2*t, 2*μ1) : 1.0 - nchisqcdf(2*μ1, 2*(t+1), 2*μ2)
+end
 
-cdf(d::Skellam, x::Int) = throw(MethodError(cdf, (d, x)))
-cdf(d::Skellam, x::Real) = throw(MethodError(cdf, (d, x)))
+cdf(d::Skellam, t::Int) = cdf(d, float(t))
```
matbesancon:
```julia
cdf(d::Skellam, t::Int) = cdf(d, float(t))
```

Is not required, the signature above will also work

Sumegh-git: It was throwing some error, that's why I added it. Will look once more then.

Sumegh-git: There are a few tests in `test/ref/discrete_test.ref.json`

Sumegh-git: Should I add more?

src/univariate/discrete/skellam.jl
Sumegh-git: Ok!

src/univariate/discrete/skellam.jl
matbesancon: hadn't seen it sorry, no it's fine :)

simonbyrne:
> You will also need a couple simple tests. I would suggest comparing the result to a sum of the PDF from -N to N, with `N` big enough

Alternatively, you can also check that the difference of two `cdf`s is equal to the sum of the `pdf`s

src/univariate/discrete/skellam.jl
Sumegh-git: Okay!

Sumegh-git:
`julia> cdf(Skellam(1,2),1)
ERROR: MethodError: Distributions.cdf(::Skellam{Float64}, ::Int64) is ambiguous. Candidates:
`

matbesancon:
could you show us the candidates?

Sumegh-git:
Updated the comment above

matbesancon:
Ok this line at `univariates.jl:340` is weird. For now you can define the signature with `Int`, sorry for the time lost

Sumegh-git:
No problem. Added the comment in docstring form.

src/univariate/discrete/skellam.jl
```diff
@@ -81,7 +81,14 @@ function cf(d::Skellam, t::Real)
...
     exp(μ1 * (cis(t) - 1) + μ2 * (cis(-t) - 1))
 end
-#based on SciPy: https://github.com/scipy/scipy/blob/v0.15.1/scipy/stats/_discrete_distns.py
+
+"""
+    cdf(d::Skellam, t::Real)
+
+Implementation based on SciPy: https://github.com/scipy/scipy/blob/v0.15.1/scipy/stats/_discrete_distns.py 
+(OR)
+Refer to Eqn (5) in On an Extension of the Connexion Between Poisson and χ2 Distributions, N.L Johnson, Vol 46, No 3/4(1959), doi:10.2307/2333532 which relates the Skellam and Non-central chisquare PDFs, which is very similar to their CDFs computation as well.
```
mschauer: Some linebreaks?

Sumegh-git: You mean the (OR)?

Sumegh-git: Ah sorry! You meant giving linebreaks in the last line I guess?

Sumegh-git: Done!

Sumegh-git:
Anything more to do @matbesancon in this?

matbesancon:
I'm good with it, merging in 12-24 hours if no objection

mschauer:
What was the conclusion about the tests?

matbesancon:
`test/univariates.jl`:

```julia
        # cdf method is not implemented for Skellam & NormalInverseGaussian
        if !isa(d, Union{Skellam, NormalInverseGaussian})
            @test isapprox(cdf(d, x), cf; atol=isa(d, NoncentralHypergeometric) ? 1e-8 : 1e-12)
        end
    end
```

This has to be modified indeed

Sumegh-git:
Oh sorry! Removing union with `Skellam` from that and replacing with only `NormalInverseGaussian`.

Sumegh-git:
Nice that you pointed. I found out the `nchisqcdf` from `StatsFuns`gives some weird values which are wrong and tests fails. But replacing that with `pnchisq` from `Rmath` tests pass.
So I guess should I replace that again by Rmath one?

matbesancon:
This is weird, StatsFuns is using RMath, in particular for `nchisqcdf`. cc @simonbyrne?

Sumegh-git:
Yes, I figured the issue. I had thought the `StatsFuns` one is same as the `Rmath` one. They have different arguments. Like `pnchisq(a,b,c)` is equal to `nchisqcdf(b,c,a)`. Sorry for the silly mistake.

Sumegh-git:
Now it's fine I guess @matbesancon . Again, sorry for the mistake earlier.

matbesancon:
Alright, merging when/if CI passes

matbesancon:
thanks @Sumegh-git!

id0158:
> Ok this line at `univariates.jl:340` is weird. For now you can define the signature with `Int`, sorry for the time lost

I am new to Julia and thus to Distributions, but from my understanding this line serves as a fallback for computing `cdf` for discrete distributions not implementing it directly by adding up `pdf` values (`univariates.jl:342-363`).

Retrospectively, I guess that `Skellam` implemented a `throw(MethodError(...))` because this fallback wasn't working for it.
Indeed, the `minimum(d)` at line `univariates.jl:345` is in contradiction with the notion of _infinite support_.
Skellam is one of rare discrete distributions not bounded on the left side.

PR #863, opened 2019-04-21
Depend on the Test package even outside of the test target.

gustafsson:
testutils.jl uses `import Test` which makes Test a dependency.

This commit resolves the following warning displayed when `using 
Distributions`:

```
[ Info: Recompiling stale cache file 
~/.julia/compiled/v1.1/Distributions/xILW0.ji for Distributions 
[31c24e10-a181-5473-b8eb-7969acd0382f]
┌ Warning: Package Distributions does not have Test in its dependencies:
│ - If you have Distributions checked out for development and have
│   added Test as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with Distributions
└ Loading Test into Distributions from project dependency, future 
warnings for Distributions are suppressed.
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,12 +13,12 @@ SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [extras]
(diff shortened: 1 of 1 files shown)


johnnychen94:
I think it's better to move [`src/testutils.jl`](https://github.com/JuliaStats/Distributions.jl/blob/fdfdbed04e2f17f9ede80d759384921405d1d056/src/testutils.jl) to a separate repo

gustafsson:
This PR merely fixes an issue introduced with 4e74c25e72b9caeda6a1c1deb993678de1acf2d2.

Moving `src/testutils.jl` to `test/testutils.jl` makes sense to me but I think it's a separate concern.

matbesancon:
closed by #866, thanks for the suggestion

PR #865, opened 2019-04-22
Add DOI and citation to README

matbesancon:


README.md
```diff
@@ -21,3 +22,47 @@ A Julia package for probability distributions and associated functions. Particul
 #### Resources
 
 * **Documentation**: <https://JuliaStats.github.io/Distributions.jl/stable/>
+
+## Citing
+
+Use the most recent DOI badge above, cite as:
+```
+@misc{dahua_lin_2019_2647520,
+  author       = {Dahua Lin and
```
ararslan: Is this list every single person who has a commit on Distributions master? If so, would we need to add another name to this every time someone contributes? I think we could shorten to a few names plus "and other contributors."

matbesancon: yes it is as generated by zenodo. I always felt weird with shortening (where do you put the cutoff), maybe replace the authors with JuliaStats?

Issue #864, opened 2019-04-21
Test.jl not in dependencies


JeffFessler:
fyi, precompiling the latest version (v0.19.1) gives this warning:

 Warning: Package Distributions does not have Test in its dependencies:
│ - If you have Distributions checked out for development and have
│   added Test as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with Distributions
└ Loading Test into Distributions from project dependency, future warnings for Distributions are suppressed.

simonbyrne:
Did you try `Pkg.resolve()`? Anyway, we should fix this by adding a Project.toml

mschauer:
See #863

JeffFessler:
oops sorry - i had searched the issues but not the pull requests...  thanks.

PR #866, opened 2019-04-23
[RFC] Update Project.toml and move testutils to test/

johnnychen94:
as an alternative to https://github.com/JuliaStats/Distributions.jl/pull/863, I think this is a cleaner way

* add `Distributed` to `Project.toml` and delete `REQUIRE` -- no longer required if I understand it right.
* add `compat` section according to the contents in `REQUIRE`
* items in `Project.toml` are listed in alphabetical order
* move `src/testutils` to `test/testutils` to reduce package dependency. This also fixes issue https://github.com/JuliaStats/Distributions.jl/issues/864

cc:@gustafsson 😄

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,11 +14,21 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 
+[compat]
+JSON = "0.20"
+PDMats = "0.9"
+QuadGK = "2.0"
+SpecialFunctions = "0.7"
+StatsBase = "0.30"
+StatsFuns = "0.8"
+julia = ">= 0.7"
+
 [extras]
+Calculus = "49dc2e85-a5d0-5ad3-a950-438e2897f1b9"
+Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
-Calculus = "49dc2e85-a5d0-5ad3-a950-438e2897f1b9"
 
 [targets]
-test = ["Test", "JSON", "ForwardDiff", "Calculus"]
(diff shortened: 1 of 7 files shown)


Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["JuliaStats"]
 version = "0.19.1"
 
 [deps]
+Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
```
matbesancon: `Distributed` is a Test dependency, we should add it in the `extras` and `targets` only

Project.toml
```diff
@@ -14,11 +15,20 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 
+[compat]
+JSON = ">= 0.4.0"
```
matbesancon: We should block on minor versions since packages can otherwise be breaking:
```suggestion
JSON = "0.4"
```
etc

matbesancon:
@johnnychen94 could you merge the modifications from @gustafsson in your PR? That way we can solve the two problems in one PR

Project.toml
johnnychen94: I'm not very familiar with the Distributions, though. Given `REQUIRE` is outdated, should I pin to the latest package version? E.g., `QuadGK = "2.0"`, regardless of patch version

johnnychen94:
#863 intends to add `Test` deps to Distributions, however, I don't think `Test` is needed in the source code. The purpose of this PR is to still keep it in `extras` section

Project.toml
matbesancon: yes, we should not commit to patch version (those should be upgraded to the latest possible)

matbesancon:
ok if testutils is moved to test
=

johnnychen94:
TODO: a patch version bump is needed to solve #864 problems for downstream packages :)

matbesancon:
yup that's planned

matbesancon:
looks good, merging if CI passes in ~6 hours

gustafsson:
@johnnychen94 thanks, much better solution than my PR. Don't forget to also remove corresponding exports in src/Distributions.jl

johnnychen94:
Oh, with these function exported, this becomes an incompatible PR, which I didn't realize.

We might need some deprecation routine here.

Issue #868, opened 2019-04-23
Register v0.19.2


ararslan:
@JuliaRegistrator `register()`

ararslan:
?? Why does it need to be in `[deps]`? It's a test dependency, so it's in `[extras]`. Halp @fredrikekre

fredrikekre:
Registrator bug, it should not error, just filter it out and not propagate it to the registry.

ararslan:
Okay thanks. What should I do in the meantime, just wait for that to be fixed?

fredrikekre:
File an issue, remove the compat entries for test-only deps and try again.

ararslan:
@JuliaRegistrator `register()`

PR #869, opened 2019-04-23
Remove test-only dependencies from [compat]

ararslan:
Due to a bug in Registrator, we can't register a new version of Distributions until either the bug is fixed or we remove `[compat]` entries for test-only dependencies. See #868 and https://github.com/JuliaComputing/Registrator.jl/issues/110.


johnnychen94:
I'm not very familiar with Distributions, though. A simple port from old `REQUIRE` is what I did in https://github.com/JuliaStats/Distributions.jl/pull/866 😂

ararslan:
Adding JSON to `[compat]` is actually correct, so what you did there was good. The issue this intends to address is to work around a bug in the package registration process.

PR #871, opened 2019-04-24
Adding Generalized Gaussian Distribution to Continuous Distributions

jsmithnh09:
Hello! This is my first contribution, I've worked within the signal processing community for some time, and the Generalized Gaussian Distribution has been essential in my work and for a number of wavelet and radar-based applications. I felt the need to add the GGD as a distribution! Let me know if I've made appropriate changes to integrate the distribution into the Julia package.


mschauer:
Welcome! @simonbyrne Do we have the incomplete gamma function somewhere (you mentioned Statsfuns) to avoid depending on GSL?

andreasnoack:
It is probably possible to use Rmath for this. The gamma cdf is basically the incomplete gamma function but maybe there is also a direct symbol for the function in Rmath.

jsmithnh09:
@andreasnoack is correct, I can utilize the pgamma.c call to utilize the incomplete gamma function. I'll make the change later this evening!

src/univariate/continuous/generalizedgaussian.jl
```diff
@@ -82,16 +82,12 @@ end
 function cdf(d::GeneralizedGaussian{T}, x::Real) where T<:Real
     """
     To calculate the CDF, the incomplete gamma function is required. To
-    access this, the GNU Scientific library is called via GSL. The CDF on wikipedia
-    indicates the normalization by Γ(1/β), with a 0.5 coefficient. The Normalized P
-    function in 8.2.4 of the National Institue of Standards and Technology  provides
-    the required function, as sf_gamma_inc_P in the GSL module.
-
-     * [Digital Library of Mathematical Functions] (https://dlmf.nist.gov/8.2#E6)
+    access this, the pgamma.c file in Rmath us used. The CDF on wikipedia
```
andreasnoack: I think this should just be a comment instead of a local string in the function (i.e. use `#` instead of `"""`).

src/Distributions.jl
```diff
@@ -18,7 +18,7 @@ import StatsBase: kurtosis, skewness, entropy, mode, modes,
                   params, params!
 
 import PDMats: dim, PDMat, invquad
-import GSL: sf_gamma_inc_P
+import Rmath: pgamma
```
andreasnoack: I think it would be better to use `cdf(Gamma(...), ...)` instead of `pgamma` directly. It effectively calls the same thing but uses the Distributions API.

src/univariate/continuous/generalizedgaussian.jl
jsmithnh09: Added!

src/Distributions.jl
jsmithnh09: I swapped the pgamma call as you suggested, the call is `cdf(Gamma( a, 1.0), b)` with a unity rate parameter.

src/univariate/continuous/generalizedgaussian.jl
```diff
@@ -0,0 +1,107 @@
...
+    return ( β / ( 2 * α * gamma(1 / β) ) ) * exp( -( abs(x - μ) / α )^β )
+end
+
+function cdf(d::GeneralizedGaussian{T}, x::Real) where T<:Real
+
+    #    To calculate the CDF, the incomplete gamma function is required. The CDF
+    #   of the Gamma distribution provides this, with the necessary 1/Γ(a) normalization.
+
+    (μ, α, β) = params(d)
+    return 1/2 + sign(x - μ) * ( cdf( Gamma( (1 / β), 1.0), (abs(x - μ) / α)^β) / 2 )
```
andreasnoack: Another thing I forgot to mention was that it would be good to avoid promotion when possible. Right now we probably don't support anything but `Float64` but once `Gamma` supports e.g. `Float32`s and `ForwardDiff.Dual`s the support would carry over here if we are careful. E.g. with something like
```suggestion
    v = cdf(Gamma(1 / β, 1), (abs(x - μ) / α)^β) / 2
    return typeof(v)(1/2) + sign(x - μ) * v
```
because `1` usually won't cause promotion in binary operations involving floating point numbers.

jsmithnh09: That's a good consideration - should this change matriculate to the other functions as well? Namely the pdf and random sample generation functions

src/univariate/continuous/generalizedgaussian.jl
```diff
@@ -0,0 +1,108 @@
...
+    μ::T
+    β::T
+
+    GeneralizedGaussian{T}(μ::T,α::T,β::T) where {T} = (@check_args(GeneralizedGaussian, α > zero(α) && β > zero(β)); new{T}(μ,α,β))
+end
+
+GeneralizedGaussian(μ::T,α::T,β::T) where {T<:Real} = GeneralizedGaussian{T}(μ,α,β)
+GeneralizedGaussian(μ::Integer,α::Integer,β::Integer) = GeneralizedGaussian(Float64(μ),Float64(α),Float64(β))
+GeneralizedGaussian(β::Float64) = GeneralizedGaussian(0.0, 1.0, β)
+GeneralizedGaussian() = GeneralizedGaussian(0.0, 1.4142, 2.0) # approximate scale with unity std deviation and shape 2
```
andreasnoack:
```suggestion
GeneralizedGaussian() = GeneralizedGaussian(0.0, √2, 2.0) # approximate scale with unity std deviation and shape 2
```

src/univariate/continuous/generalizedgaussian.jl
```diff
@@ -0,0 +1,108 @@
...
+#### Statistics
+
+mean(d::GeneralizedGaussian) = d.μ
+median(d::GeneralizedGaussian) = d.μ
+mode(d::GeneralizedGaussian) = d.μ
+
+var(d::GeneralizedGaussian{T}) where {T<:Real} = (d.α^2) * ( gamma(3 / d.β) / gamma(1 / d.β) )
+std(d::GeneralizedGaussian{T}) where {T<:Real} = (d.α) * sqrt( gamma(3 / d.β) / gamma(1 / d.β) )
+
+skewness(d::GeneralizedGaussian{T}) where {T<:Real} = 0.0
```
andreasnoack:
```suggestion
skewness(d::GeneralizedGaussian{T}) where {T<:Real} = zero(T)
```

src/univariate/continuous/generalizedgaussian.jl
andreasnoack: > should this change matriculate to the other functions as well?

Yes but I think they are mostly fine. I've made two small suggestions.

andreasnoack:
This looks good. However, it needs some tests. Once there are tests for all the methods added here, I think this should be good to go.

jsmithnh09:
@andreasnoack would this be a separate generalizedgaussian.jl file within the test folder? I should be able to follow along with the tests already outlined in that folder as well, for format and syntax.

andreasnoack:
> would this be a separate generalizedgaussian.jl file within the test folder?

Not all distributions are tested with the same setup but I'd say yes. Create that file with that name and add some tests. https://github.com/JuliaStats/Distributions.jl/blob/master/test/multinomial.jl might be okay as inspiration and you can use R to compute values to test against. @matbesancon Do you agree with this approach?

matbesancon:
Agreed, on principle, but if we can avoid the process of
`R distribution generation -> JSON -> read JSON -> test`, it would be great. I was thinking of finding another way to test things, this is the main thing slowing #823 down
Could we generate the results from R and put them in hard-coded vectors in the test file?

andreasnoack:
Yes indeed. I don't think the multinomial file I linked to is using the JSON approach which was deliberate. I also think we should just hard code the values in the test file.

jsmithnh09:
I’ve hardcoded values for the CDF and PDF tests, albeit from results pulled from MATLAB R2018b. Would the preferred approach be simulation in R instead?

matbesancon:
yes it would be better for homogeneity, R is the reference that has been used for all distributions

src/univariate/continuous/generalizedgaussian.jl
```diff
@@ -0,0 +1,108 @@
...
+    #   [1]  Gonzalez-Farias, G., Molina, J. A. D., & Rodríguez-Dagnino, R. M. (2009).
+    #   Efficiency of the approximated shape parameter estimator in the generalized
+    #   Gaussian distribution. IEEE Transactions on Vehicular Technology, 58(8),
+    #   4214-4223.
+
+    # utilizing the sampler from the Gamma distribution.
+    g = Gamma((1 / d.β), 1)
+
+    # bernoulli random variable "b" with parameter (1/2).
+    b = rand(1) .< 0.5
```
matbesancon:
```suggestion
    b = rand() .< 0.5
```
Shorter is better

src/univariate/continuous/generalizedgaussian.jl
```diff
@@ -0,0 +1,108 @@
...
+#### Sampling
+function rand(rng::AbstractRNG, d::GeneralizedGaussian)
+    #   Sampling Procecdure from [1]
+    #   [1]  Gonzalez-Farias, G., Molina, J. A. D., & Rodríguez-Dagnino, R. M. (2009).
+    #   Efficiency of the approximated shape parameter estimator in the generalized
+    #   Gaussian distribution. IEEE Transactions on Vehicular Technology, 58(8),
+    #   4214-4223.
+
+    # utilizing the sampler from the Gamma distribution.
+    g = Gamma((1 / d.β), 1)
```
matbesancon:
```suggestion
    g = Gamma(inv(d.β), 1)
```
Keeps the code generic

src/univariate/continuous/generalizedgaussian.jl
```diff
@@ -0,0 +1,108 @@
...
+
+    # bernoulli random variable "b" with parameter (1/2).
+    b = rand(1) .< 0.5
+    b = convert(Float64, b[1]) * 2 - 1
+
+    return d.μ + (1 / sqrt(d.α) ) * rand(rng, g)^(1 / d.β) * b
```
matbesancon:
```suggestion
    return d.μ + inv(sqrt(d.α)) * rand(rng, g)^inv(d.β) * b
```

jsmithnh09:
Understood, I’ll push the modification along with any changes you may have in a couple days - thank you for the constructive suggestions, everyone! This is my first pull request.

test/generalizedgaussian.jl
```diff
@@ -0,0 +1,234 @@
+using Distributions
```
matbesancon: not needed here, Distributions and Test are already imported in `runtests.jl`

test/generalizedgaussian.jl
```diff
@@ -0,0 +1,234 @@
...
+using Test
+
+
+d = GeneralizedGaussian() # mean zero, scale √2, shape 2.
```
matbesancon: to avoid putting everything in global scope, use a testset bloc:
```julia
@testset "generalized gaussian" begin
    d = GeneralizedGaussian() # mean zero, scale √2, shape 2.
    # other stuff
end
```

test/generalizedgaussian.jl
```diff
@@ -0,0 +1,234 @@
...
+
+@test mean(d) == 0
+@test median(d) == 0
+@test mode(d) == 0
+@test var(d) == 1 # unity variance with shape 2 and scale √2
+@test std(d) == 1
+@test skewness(d) == 0
+@test kurtosis(d) ≈ 0
+@test entropy(d) ≈ 1.418938533204673  #[1]
+# [1] Nadarajah, Saralees (September 2005).  "A generalized normal distribution".
```
matbesancon: is the reference used for the implementation or the tested values only?

matbesancon:
It's a great add to Distributions :)

test/generalizedgaussian.jl
jsmithnh09: The reference was for implementation, I’ll admit it is a bit haphazardly placed. I can move the reference to the src file where it is calculated instead?

matbesancon: Yes, either in the docstring of the Distribution or of the relevant function if it's just for one of the functions

matbesancon:
also, @jsmithnh09, could you add the distribution to docs/src/univariates.md in the corresponding bloc? Thanks

jsmithnh09:
I've modified the test file along with updating the docs. The PDF and CDF hard-coded vectors still need to be updated with R, I should be able to do that in a few days time after some travel. Let me know if the modifications are suitable!

test/generalizedgaussian.jl
```diff
@@ -0,0 +1,232 @@
+@testset "generalized gaussian" begin
+
+    d = GeneralizedGaussian() # mean zero, scale √2, shape 2.
+
+# PDF and CDF values from MATLAB R2018b using the same default
+# mean scale and shape parameters.
+pdftest = [
```
matbesancon: indent everything in the `@testset` block

src/univariate/continuous/generalizedgaussian.jl
```diff
@@ -0,0 +1,110 @@
...
+    #   [1]  Gonzalez-Farias, G., Molina, J. A. D., & Rodríguez-Dagnino, R. M. (2009).
+    #   Efficiency of the approximated shape parameter estimator in the generalized
+    #   Gaussian distribution. IEEE Transactions on Vehicular Technology, 58(8),
+    #   4214-4223.
+
+    # utilizing the sampler from the Gamma distribution.
+    g = Gamma(inv(d.β), 1)
+
+    # bernoulli random variable "b" with parameter (1/2).
+    b = rand() .< 0.5
```
matbesancon: Broadcast is not needed here
```suggestion
    b = rand() < 0.5
```

src/univariate/continuous/generalizedgaussian.jl
```diff
@@ -0,0 +1,110 @@
...
+    b = convert(Float64, b[1]) * 2 - 1
```
matbesancon: Goes with previous modification
```suggestion
    b = convert(Float64, b) * 2 - 1
```

matbesancon:
I should be able to simulate these points soon using R if you prefer. I added some last minor comments

jsmithnh09:
@matbesancon that'd be fantastic! I included the shape, scale, and location parameters in the comments for the default constructor in the test, the vector was from -2 to 2 in stepsizes of 0.05 -- feel free to format the test however you please! And I'll apply the other changes as well right now

jsmithnh09:
@matbesancon I added the suggestions you recommended for the test, let me know once you've had the opportunity to simulate the values for the CDF and PDF comparison! Hopefully then the distribution will be ready for addition to Distributions.

matbesancon:
@jsmithnh09 one good practice is to create a branch for each new feature you develop, merging from your master -> main master is not practical
Also, there should be a precision on which "type" of Generalized normal distribution is implemented here, see https://en.wikipedia.org/wiki/Generalized_normal_distribution

jsmithnh09:
Understood - I'm still working on the semantics of Github, any suggestions or pointers to pages that provide a better recommendation on handling would be appreciated. I had attempted to branch the package, but strangely, it would not allow me on the first go, I had to fork instead.

As far as versioning goes, the "version 2" displayed on Wikipedia is not a standard nomenclature - the literature has consistently referred to the Generalized Gaussian as "version 1", with applications and research in wavelets and image processing, see [1]-[4]. Maximum Likelihood estimation and parameter estimation are both established with this version, and the skewness model of "version 2" is not fully documented. I'd recommend the introduction of this model first as it is already well documented and utilized in the community. Wikipedia isn't necessarily always a suitable reference :) 

[1] Sharifi, K., & Leon-Garcia, A. (1995). Estimation of shape parameter for generalized Gaussian distributions in subband decompositions of video. IEEE Transactions on Circuits and Systems for Video Technology, 5(1), 52-56.
[2] Do, M. N., & Vetterli, M. (2002). Wavelet-based texture retrieval using generalized Gaussian density and Kullback-Leibler distance. IEEE transactions on image processing, 11(2), 146-158.
[3] Bouman, C., & Sauer, K. (1992). A generalized Gaussian image model for edge-preserving MAP estimation. ECE Technical Reports, 277.
[4] J. Joshi, T. R. Fischer, Comparison of generalized Gaussian and Laplacian modeling in DCTimage coding, IEEE Transactions on Signal Processing Letters, 2(1995), 81-82.

jsmithnh09:
With that being said, what would we like to do with the changes? Modifying the current branch in some way, or making some sort of modification? What next step(s) need to be made for this to be merged?

matbesancon:
> Understood - I'm still working on the semantics of Github, any suggestions or pointers to pages that provide a better recommendation on handling would be appreciated. I had attempted to branch the package, but strangely, it would not allow me on the first go, I had to fork instead.

Best way from your local working branch:
```
git branch gen-gaussian
git checkout gen-gaussian
```
At this point create a dummy commit just to register the branch and push to this branch on your fork of the repository.
Then you can create a PR with this branch instead of your version of master

matbesancon:
Thanks for the references, this is fine by me

jsmithnh09:
Excellent! Thank you

matbesancon:
@jsmithnh09 that's the same distribution as this paper right?:
https://www.researchgate.net/profile/Wolf_Dieter_Richter/publication/254282790_Simulation_of_the_p-generalized_Gaussian_distribution/links/574ec3fa08aec50945bb0b21/Simulation-of-the-p-generalized-Gaussian-distribution.pdf

That's the reference paper for the implementation in the R package. Note that they use the term `p-generalized normal distribution` to distinguish it from other generalizations. This can be in the docstring

jsmithnh09:
Yes, the very same! This is an excellent paper - I believe the sampling implementation I provided from a secondary reference is the Nardon and Pianca method.

I’ve seen the use of “p”, \beta and \theta to all describe the exact same shape parameter and distribution. I think the addition of this paper in the docstring would be well suited for context, but keeping the name “Generalized Gaussian” I believe will eliminate some of the confusion that’s associated with the name, as this is the common term across all literature, (including this article)

jsmithnh09:
That’s only my take though, if the consensus is to modify the name and attach both the Wikipedia and a reference to this article, I believe either approach is suitable.

matbesancon:
it's good on my side if it's made explicit in the docstring.
For the R version, I think there is a very slight variation from the Matlab implementation, better use R from the start then.
You will find the reproduction script here:
https://pastebin.com/ZhTGFxGB
and the data here: https://pastebin.com/8ufmBxyS

simonbyrne:
Personally I dislike simply calling it "generalized" (since there are many ways one could generalize the normal distribution, as highlighted by the Wikipedia article).

In papers I've seen it used (e.g. https://projecteuclid.org/euclid.bj/1178291835), it usually isn't named, but I've heard it referred to casually as something like "power exponential class".

jsmithnh09:
@simonbyrne what would you suggest then the change be for the name? As myself and @matbesancon have both provided references, theres a historical context associated with the name, and the Wikipedia page additionally adds the name as “Generalized Normal”.

jsmithnh09:
I think maintaining context with the R package as @matbesancon suggested and what other literature have entitled the distribution would prevent additional confusion.

simonbyrne:
`PGeneralizedGaussianDistribution` seems like it would at least be unique.

matbesancon:
We don't put distribution in other distribution names, it becomes heavy and too long. `PGeneralizedGaussian` would be explicit but concise

matbesancon:
replaced by #876

PR #874, opened 2019-04-25
Abstract array types in utils.jl (fix #812)

### LINKED ISSUE #812: Use AbstractVector for isprobvec
It would be nice if [`isprobvec`](
https://github.com/JuliaStats/Distributions.jl/blob/master/src/utils.jl) could accept an `AbstractVector` instead of a `Vector`.  
This would require to use abstract types for `isfinite`, `allfinite`, `allzeros`, etc.

Is there any drawbacks of doing this ?

maxmouchet:
Use `AbstractArray` instead of `Array` in the signatures of:
- `allfinite`
- `allzeros`
- `allnonneg`
- `isprobvec`
- `isApproxSymmmetric`
- `hasCholesky`
- `trycholesky`

This allows these functions to be used with `Array`s, `StaticArray`s, ...
See #812.


matbesancon:
@maxmouchet could you add some tests using `StaticArrays` for instance to see that the new behaviour is as expected?

matbesancon:
ping @maxmouchet

matbesancon:
FIxes #812

matbesancon:
@maxmouchet some news on this?

maxmouchet:
@matbesancon sorry for the delay, was quite busy. Sure, I'll add some tests!  
How should I add the `StaticArray` dependency ? In the `[extras]` section of `Project.toml` ?

matbesancon:
I was sure they were already. No need to add a dependency. You can use `Test.GenericArray` it will do the job
```julia
help?> Test.GenericArray
  The GenericArray can be used to test generic array APIs that program to the AbstractArray interface, in order to ensure that functions can work
  with array types besides the standard Array type.
```

maxmouchet:
Done. But it turns out that `LinearAlgebra.cholesky` does not support abstract arrays, so I reverted back to `Matrix` for `hasCholesky` and `trycholesky`.

matbesancon:
thanks a lot!
> But it turns out that LinearAlgebra.cholesky does not support abstract arrays

Yes that's the pain of depending on non-Julia software (BLAS in that case)

There seems to be a conflict between your branch and some recently merged stuff. Could you pull the latest master, and merge locally master to your version?

test/utils.jl
```diff
@@ -33,3 +33,25 @@ end
...
+@test isprobvec(GenericArray([1, 1, 1])) == false
+@test isprobvec(GenericArray([1/3, 1/3, 1/3]))
+
+# Positive definite matrix
+M = GenericArray([1.0 0.0; 0.0 1.0])
+# Non-invertible matrix
+N = GenericArray([1.0 0.0; 1.0 0.0])
+
+@test Distributions.isApproxSymmmetric(N) == false
+@test Distributions.isApproxSymmmetric(M)
```
matbesancon:
```suggestion
@test Distributions.isApproxSymmmetric(M)

```
new line at the end of file

matbesancon:
Looks good, merging when CI passes

Issue #638, opened 2017-07-12
Limited Support for ForwardDiff


chriselrod:
While the normal distribution supports ForwardDiff (has cd

```julia
julia> using Distributions, ForwardDiff

julia> n = Normal(0.2, 0.5);

julia> b = Beta(3, 4);

julia> g = Gamma(2, 6);

julia> x = rand(4); print(x')
[0.24138 0.63222 0.059794 0.153202]
julia> ForwardDiff.gradient(x -> sum(cdf(n, x)), x)'
1×4 RowVector{Float64,Array{Float64,1}}:
 0.795157  0.54913  0.767124  0.794397

julia> pdf(n, x)'
1×4 RowVector{Float64,Array{Float64,1}}:
 0.795157  0.54913  0.767124  0.794397

julia> ForwardDiff.gradient(x -> sum(cdf(b, x)), x)'
ERROR: MethodError: Cannot `convert` an object of type ForwardDiff.Dual{ForwardDiff.Tag{##3#4,0xb046287d533b082d},Float64,4} to an object of type Float64
This may have arisen from a call to the constructor Float64(...),
since type constructors fall back to convert methods.
Stacktrace:
 [1] betacdf at /home/chris/.julia/v0.6/StatsFuns/src/rmath.jl:62 [inlined]
 [2] cdf at /home/chris/.julia/v0.6/Distributions/src/univariates.jl:313 [inlined]
 [3] _cdf!(::Array{ForwardDiff.Dual{ForwardDiff.Tag{##3#4,0xb046287d533b082d},Float64,4},1}, ::Distributions.Beta{Float64}, ::Array{ForwardDiff.Dual{ForwardDiff.Tag{##3#4,0xb046287d533b082d},Float64,4},1}) at /home/chris/.julia/v0.6/Distributions/src/univariates.jl:194
 [4] cdf(::Distributions.Beta{Float64}, ::Array{ForwardDiff.Dual{ForwardDiff.Tag{##3#4,0xb046287d533b082d},Float64,4},1}) at /home/chris/.julia/v0.6/Distributions/src/univariates.jl:205
 [5] vector_mode_gradient at /home/chris/.julia/v0.6/ForwardDiff/src/gradient.jl:94 [inlined]
[...]

PR #875, opened 2019-04-25
Native Normal

Nosferican:
Fresh #714.


src/univariate/continuous/normal.jl
```diff
@@ -74,22 +73,20 @@ kurtosis(d::Normal{T}) where {T<:Real} = zero(T)
...
-
-mgf(d::Normal, t::Real) = exp(t * d.μ + d.σ^2/2 * t^2)
-cf(d::Normal, t::Real) = exp(im * t * d.μ - d.σ^2/2 * t^2)
-
+logpdf(d::Normal, x::Real) = -(x - d.μ)^2 / 2 / d.σ^2 - log(2π * d.σ^2) / 2
+pdf(d::Normal, x::Real) = exp(logpdf(d, x))
+cdf(d::Normal, x::Real) = (1 + erf((x - d.μ) / d.σ / √2)) / 2
+logcdf(d::Normal, x::Real) =
+    erf((x - d.μ) / d.σ / √2) |>
+    (value -> value < -1 ? log1p(value) - log(2) : log((1 + value) / 2))
```
mschauer: This is a bit too smart for me, can we do it without `|>`?

Nosferican:
Could you check what's going on with the truncated form?

src/univariate/continuous/normal.jl
matbesancon: +1 yup, plus @ararslan is gonna block this PR with a `|>` :stuck_out_tongue:

src/univariate/continuous/normal.jl
```diff
@@ -74,22 +73,21 @@ kurtosis(d::Normal{T}) where {T<:Real} = zero(T)
 
 entropy(d::Normal) = (log2π + 1)/2 + log(d.σ)
 
-
 #### Evaluation
 
-@_delegate_statsfuns Normal norm μ σ
-
-gradlogpdf(d::Normal, x::Real) = (d.μ - x) / d.σ^2
```
matbesancon: probably a stupid question but why remove `gradlogpdf`?

matbesancon:
@Nosferican you don't get the same test error locally?

src/univariate/continuous/normal.jl
Nosferican: I couldn't find anywhere it was being used or tested. I suspect it was an artifact, but I could be wrong...

Nosferican:
Same error, seems like a loss of precision for truncated normal. Any ideas how to debug it or improve the precision?

src/univariate/continuous/normal.jl
matbesancon: A quick search highlighted it is used and tested quite a lot. Better idea would be to let it and document it better

src/univariate/continuous/normal.jl
```diff
@@ -74,22 +73,21 @@ kurtosis(d::Normal{T}) where {T<:Real} = zero(T)
...
-
-mgf(d::Normal, t::Real) = exp(t * d.μ + d.σ^2/2 * t^2)
-cf(d::Normal, t::Real) = exp(im * t * d.μ - d.σ^2/2 * t^2)
-
+logpdf(d::Normal, x::Real) = -(x - d.μ)^2 / 2 / d.σ^2 - log(2π * d.σ^2) / 2
+pdf(d::Normal, x::Real) = exp(logpdf(d, x))
+cdf(d::Normal, x::Real) = (1 + erf((x - d.μ) / d.σ / √2)) / 2
+function logcdf(d::Normal, x::Real)
+    value = erf((x - d.μ) / d.σ / √2)
+    value < -1 ? log1p(value) - log(2) : log((1 + value) / 2)
```
matbesancon: wait two possible computations for logcdf of a normal. Do you have a reference paper/doc on this?

Nosferican: Same formula. The only difference is using the `log` or `log1p` according to the docstring of `log1p`. 
```
log1p(x)

  Accurate natural logarithm of 1+x. Throws DomainError for Real arguments
  less than -1.
```

matbesancon: ok, try to write it in a consistent way (either -log(2) or the /2 in the log function. Add a comment regarding the use of `log1p` and why not use it in all cases? You are here using it only when it will throw an error
Also, use `ifelse` instead of the ternary operation, usually a tad faster for small branches like this.

Nosferican: Actually, the codomain of the error function (`erf`) is [-1, 1], so it would be safe to always use `log1p`. Is there any preference in terms of numerical precision on whether it would be best to divide and take the log afterwards or add the log of the factorization?

matbesancon: no idea but I would think not

matbesancon: if you want to be picky, log2 can be a pre-computed constant

src/univariate/continuous/normal.jl
```diff
@@ -74,22 +73,18 @@ kurtosis(d::Normal{T}) where {T<:Real} = zero(T)
...
 
-
 #### Evaluation
-
-@_delegate_statsfuns Normal norm μ σ
-
 gradlogpdf(d::Normal, x::Real) = (d.μ - x) / d.σ^2
-
-mgf(d::Normal, t::Real) = exp(t * d.μ + d.σ^2/2 * t^2)
-cf(d::Normal, t::Real) = exp(im * t * d.μ - d.σ^2/2 * t^2)
```
mschauer: I don't see why these should go

src/univariate/continuous/normal.jl
```diff
@@ -74,22 +73,18 @@ kurtosis(d::Normal{T}) where {T<:Real} = zero(T)
 
 entropy(d::Normal) = (log2π + 1)/2 + log(d.σ)
 
-
 #### Evaluation
-
-@_delegate_statsfuns Normal norm μ σ
```
mschauer: Check https://github.com/JuliaStats/Distributions.jl/blob/2ac3a83aef47132d97bb6632e8465617cd493c9f/src/univariates.jl#L529 for a list of the functions which are created by this macro and should be implemented natively.

src/univariate/continuous/normal.jl
Nosferican: I will use `StatsFuns.logtwo`

src/univariate/continuous/normal.jl
Nosferican: Thought I had put those back with `gradlogpdf.`

src/univariate/continuous/normal.jl
Nosferican: I checked whether including some methods had any noticeable drawback from using the default   fallback and included those only if it applied. For example, I did make one change to `quantile` such that the `invlogcdf` fallback was consistent providing `NaN`.

src/univariate/continuous/normal.jl
```diff
@@ -74,22 +73,21 @@ kurtosis(d::Normal{T}) where {T<:Real} = zero(T)
...
-
 gradlogpdf(d::Normal, x::Real) = (d.μ - x) / d.σ^2
+logpdf(d::Normal, x::Real) = -(x - d.μ)^2 / 2 / d.σ^2 - log(2π * d.σ^2) / 2
+pdf(d::Normal, x::Real) = exp(logpdf(d, x))
+cdf(d::Normal, x::Real) = (1 + erf((x - d.μ) / d.σ / √2)) / 2
```
simonbyrne: This will suffer from cancellation for small values (we should have a test for this). Look at how `normcdf` is implemented:
https://github.com/JuliaStats/StatsFuns.jl/blob/316cebb69d4e28e6a4b17d72eee112e61c3b9b77/src/distrs/norm.jl#L45

src/univariate/continuous/normal.jl
```diff
@@ -74,22 +73,21 @@ kurtosis(d::Normal{T}) where {T<:Real} = zero(T)
...
+logcdf(d::Normal, x::Real) = log1p(erf((x - d.μ) / d.σ / √2)) - logtwo
```
simonbyrne: Similarly, look at
https://github.com/JuliaStats/StatsFuns.jl/blob/316cebb69d4e28e6a4b17d72eee112e61c3b9b77/src/distrs/norm.jl#L66-L68

src/univariate/continuous/normal.jl
```diff
@@ -74,22 +73,21 @@ kurtosis(d::Normal{T}) where {T<:Real} = zero(T)
...
+quantile(d::Normal, p::Real) = p > 1 ? NaN : d.μ + d.σ * √2 * erfinv(2p - 1)
```
simonbyrne: https://github.com/JuliaStats/StatsFuns.jl/blob/316cebb69d4e28e6a4b17d72eee112e61c3b9b77/src/distrs/norm.jl#L91

src/univariate/continuous/normal.jl
simonbyrne: also, you will need to define `ccdf` as well.

simonbyrne:
Please have a look at https://github.com/JuliaStats/StatsFuns.jl/blob/316cebb69d4e28e6a4b17d72eee112e61c3b9b77/src/distrs/norm.jl: though it's not perfect, a fair bit of work has gone into it to make things reasonably accurate.

We should also add some tests to make sure we don't break any of these.

src/univariate/continuous/normal.jl
matbesancon: This is not documented:
https://juliastats.github.io/Distributions.jl/stable/extends.html#Univariate-Distribution-1

matbesancon: most newly implemented distributions haven't done it I think

simonbyrne: No, but if we don't we will get the same cancellation effects, since otherwise it will fallback on `1-cdf(d,x)`

matbesancon: cc @Nosferican

matbesancon:
@Nosferican you're ok on this one?

Nosferican:
I should be. Sorry for the delay. I just moved for my job and have been assembling furniture the last couple days. Ping me if I haven't moved on it by Wednesday.

matbesancon:
ping @Nosferican

Nosferican:
I will probably try to do one more pass to make the code leaner, but any suggestions would be appreciated.

matbesancon:
@Nosferican you're not using the implementation from StatsFuns pointed out by @simonbyrne ?

Nosferican:
Aye. The implementation I am using is the one from StatsFuns. I think this should be fine for now. There were a few comments in a TODO for improving it. I kept the comments when porting the code, but that can be done at a later time.

src/univariate/continuous/normal.jl
```diff
@@ -74,22 +73,175 @@ kurtosis(d::Normal{T}) where {T<:Real} = zero(T)
...
+    end
+end
+function pdf(d::Normal, x::Real)
+    if iszero(d.σ)
+        d.μ == x ? Inf : 0.0
+    else
+        exp(-((x - d.μ) / d.σ)^2 / 2) * invsqrt2π / d.σ
+    end
+end
+function cdf(d::Normal, x::Real)
```
simonbyrne: still needs a `ccdf` implementation to deal with tails (e.g. `ccdf(Normal(),10.0)`)

src/univariate/continuous/normal.jl
```diff
@@ -74,22 +73,175 @@ kurtosis(d::Normal{T}) where {T<:Real} = zero(T)
...
+function xval(μ::Real, σ::Real, z::Number)
+    if isinf(z) && iszero(σ)
+        μ + one(σ) * z
+    else
+        μ + σ * z
+    end
+end
 
-@_delegate_statsfuns Normal norm μ σ
+function quantile(d::Normal, p::Real)
```
simonbyrne: and a `cquantile` (for values < `eps()`).

src/univariate/continuous/normal.jl
```diff
@@ -74,22 +73,175 @@ kurtosis(d::Normal{T}) where {T<:Real} = zero(T)
...
+            -Inf
+        elseif isone(p)
+            Inf
+        else
+            0.5
+        end
+    end
+    norminvcdf(d.μ, d.σ, p)
+end
+invlogcdf(d::Normal, lp::Real) = norminvlogcdf(d.μ, d.σ, lp)
```
simonbyrne: and `invlogccdf`

src/univariate/continuous/normal.jl
```diff
@@ -74,22 +73,175 @@ kurtosis(d::Normal{T}) where {T<:Real} = zero(T)
...
+    end
+end
+function cdf(d::Normal, x::Real)
+    if iszero(d.σ)
+        float(d.μ ≤ x)
+    else
+        erfc(-(x - d.μ) / d.σ * invsqrt2) / 2
+    end
+end
+function logcdf(d::Normal, x::Real)
```
simonbyrne: and `logccdf`

Nosferican:
Let me know if the thin `Normal` for `LogNormal` is fine. If not, I will add a non-distribution method. Other than that, just missing porting more tests on the precision side (e.g., `ccdf(Normal(), 10)`).

src/univariate/continuous/normal.jl
Nosferican: Added `cquantile`.

Nosferican:
@simonbyrne Let me know when you have had the time to check the revision.

matbesancon:
@Nosferican thanks for the revisions. Did you merge the last master into your branch?

Nosferican:
I did before the revision so it should be using the master as of 11 days ago.

matbesancon:
master has been moving fast :sunglasses:

matbesancon:
@Nosferican could you check that this PR is fixing:
https://github.com/JuliaStats/Distributions.jl/issues/638

matbesancon:
@Nosferican, for my comment above, on your branch, just see if:
```julia
julia> using Distributions, ForwardDiff

julia> n = Normal(0.2, 0.5);

julia> b = Beta(3, 4);

julia> g = Gamma(2, 6);

julia> x = rand(4); print(x')
[0.24138 0.63222 0.059794 0.153202]

julia> ForwardDiff.gradient(x -> sum(cdf(n, x)), x)'
```

Doesn't error anymore

Nosferican:
```
using Distributions, ForwardDiff, Random
n = Normal(0.2, 0.5)
b = Beta(3, 4)
g = Gamma(2, 6)
Random.seed!(0)
x = rand(4)
ForwardDiff.gradient(x -> sum(cdf(n, i) for i in x), x)'
```
No errors now.

matbesancon:
Still missing some tests see coverage diff, and some docstring over many functions (`logccdf`, `quantile`, etc). The default for putting docstring should be better more than not enough

matbesancon:
also @Nosferican there were some changes on master with the latest PR regarding normal, it would be wise to merge master into this branch first

Nosferican:
Sounds good. I will rebase and move to test and documentation.

Nosferican:
I have added the docstring to all the newly added functions that were lacking a docstring. After CI, I will inspect the code coverage to identify any changes that should be made to the test suite.

src/univariate/continuous/normal.jl
```diff
@@ -76,14 +76,23 @@ entropy(d::Normal) = (log2π + 1)/2 + log(d.σ)
 #### Evaluation
 
 # Helpers
+"""
+    xval(d::Normal, z::Real)
```
mschauer: inconsistent hyphenation

src/univariate/continuous/normal.jl
```diff
@@ -74,22 +73,225 @@ kurtosis(d::Normal{T}) where {T<:Real} = zero(T)
...
+    end
+end
+# logccdf
+function logccdf(d::Normal, x::Real)
+    if iszero(d.σ) && x == d.μ
+        z = zval(Normal(zero(d.μ), d.σ), one(x))
+    else
+        z = zval(d, x)
+    end
+    z > 1.0 ?
```
simonbyrne: make this an `if else` block?

src/univariate/continuous/normal.jl
simonbyrne: They're now below

Nosferican:
Will address the last issues tomorrow. I think the test coverage misses a few special handlings so will add those tests.

src/univariate/continuous/normal.jl
Nosferican: Could you elaborate? It seems the project-wide style is a four space on first line, skip line and start without indentation from inspecting various files.

mschauer: „x value“ versus „z-value“

matbesancon:
ping @Nosferican how are you doing on this?

Nosferican:
If the CI likes it, it might be ready for final review and merge.

Nosferican:
There seems to be an issue with the Linux images for the CI?

matbesancon:
indeed, I re-started the build

src/univariate/continuous/normal.jl
matbesancon: @Nosferican comment to address here

src/univariate/continuous/normal.jl
matbesancon: @Nosferican comment to address here

src/univariate/continuous/normal.jl
matbesancon: @Nosferican comment to address here

matbesancon:
@Nosferican I tagged you on a couple comments for which it was not clear from the diff if they had been addressed, can you resolve them if that's the case? :)

Nosferican:
I think this is good.

matbesancon:
Fantastic, merged it, thanks :)

Nosferican:
Thanks y'all!

PR #877, opened 2019-04-30
Add "Histogram distribution"

cjdoris:
Hello,

I have implemented a "histogram distribution", called `HistogramDist`. That is, its parameters are a vector of bin edges and a distribution over bins, and the distribution is to sample a bin, then sample uniformly from that bin.

Functionally, this is identical to a mixture of uniform distributions, but the implementation of `cdf` for example takes time `log(nbins)`. This is because the mixture components (bins) are known to be disjoint and sorted.

A few considerations I can think of:
- The name. Perhaps `ContinuousNonParametric` is better? People usually refer to fitting histograms to data as "non-parametric density estimation", so this seems suitable.
- It could be made a subtype of `AbstractMixtureModel`. In this case, maybe the name should be `UniformMixtureModel`?
- The type precomputes and stores `O(nbins)` data (fields `_pdf`, `_cdf` and `_aliastable`) for making `rand` and `cdf` fast (`O(1)` and `O(log(nbins))` respectively). Does this fit with the design of the module?

Hope this is useful to you.

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -92,6 +92,7 @@ export
     Geometric,
     Gumbel,
     Hypergeometric,
+    HistogramDist,
     InverseWishart,
     InverseGamma,
     InverseGaussian,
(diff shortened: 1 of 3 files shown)


matbesancon:
@cjdoris sorry for the delay. There is already a `StatsBase.Histogram` type, aren't these a bit conflated?

cjdoris:
@matbesancon I don't think so, `StatsBase.Histogram` is not a `Distributions.Distribution`.

The reason I implemented it is because I want to create histograms with many bins and evaluate the CDF many times. Precomputing the CDF at the bin edges means this takes time `O(log(nbins))` instead of `O(nbins)` (from using `MixtureModel(Uniform,...)`) which is a substantial time saving.

PR #878, opened 2019-05-03
loosen type restrictions to accept accept variants

adam-r-kowalski:
Flux tracked arrays do not work with the `isprobvec` and `allnonneg` functions as they are. I changed the type definitions to accept abstract vectors and abstract arrays.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -72,16 +72,10 @@ end
 
 allzeros(x::ZeroVector) = true
 
-function allnonneg(x::Array{T}) where T<:Real
-    for i = 1 : length(x)
-        if !(x[i] >= zero(T))
-            return false
-        end
-    end
-    return true
-end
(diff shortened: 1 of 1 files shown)


matbesancon:
Great thanks for the contribution, merging in 20 hours if no objection

src/utils.jl
```diff
@@ -72,16 +72,10 @@ end
...
 
-function allnonneg(x::Array{T}) where T<:Real
-    for i = 1 : length(x)
-        if !(x[i] >= zero(T))
-            return false
-        end
-    end
-    return true
-end
+allnonneg(xs::AbstractArray{T}) where T<:Real = all(x -> x >= zero(T), xs)
```
mschauer:
```suggestion
allnonneg(xs::AbstractArray{<:Real}) = all(x -> x >= 0, xs)
```
Comparisons with 0 work across types

PR #881, opened 2019-05-08
copy mean, added test

### LINKED ISSUE #120: Should mean(MultvariateNormal) return a copy, not a reference?
While working on some ideas for the MCMC.jl package, I tripped on the fact that `mean(MultivariateNormal)` returns a pointer to the distribution's current mean. I accidentally overwrote it and got very corrupt results during a long slice sampler run.

The performance arguments for avoiding a copy are compelling, but perhaps this is going to cause trouble in the future?

I don't think we need to make any decision until more users tell us about their experiences, but I wanted to note this for discussion.

matbesancon:
Fixes #120, the long-awaited  and removes annoying type prints in MvLogNormal


ararslan:
Hm, I'm not sure about this. Perhaps it's safer, but I generally don't like making defensive copies. It'd be good to hear from others as well on this.

nalimilan:
That's really the same question as https://discourse.julialang.org/t/convention-for-read-only-array-getters/19356. In general it's annoying that a copy is returned since then there's no (generic/official) way to avoid copying.

src/multivariate/mvnormal.jl
```diff
@@ -236,7 +236,7 @@ Base.show(io::IO, d::MvNormal) =
 ### Basic statistics
 
 length(d::MvNormal) = length(d.μ)
-mean(d::MvNormal) = convert(Vector, d.μ)
+mean(d::MvNormal) = copy(convert(Vector, d.μ))
```
nalimilan: This should be enough (avoiding one copy when the object isn't a `Vector` already):
```suggestion
mean(d::MvNormal) = Vector(d.μ)
```

ararslan: That still copies:
```
julia> x = [1,2,3];

julia> y = Vector(x);

julia> y === x
false
```

ararslan:
> In general it's annoying that a copy is returned since then there's no (generic/official) way to avoid copying.

Yes, this. IMO we should just tell people that they need to call `copy` manually if they're worried about mutating the result. I see `mean`, `var`, etc. as accessors when it comes to distributions, and accessors shouldn't copy by default.

matbesancon:
I see the point. In that case, the current behaviour is weird, it copies only if we have something that's not a vector, and is a no-op otherwise.

nalimilan:
Then maybe just return `d.μ` directly?

I don't have a good solution for this kind of pattern (as the Discourse thread indicates). But one advantage of *not* copying IMHO is that we can always switch to making a copy or returning a read-only wrapper in a future release without breaking anything (it will just be safer). OTC if we make a copy, we won't be able to change that as people will then rely on being able to mutate the result.

matbesancon:
@nalimilan did just that

nalimilan:
Mmm, I've just realized that `cov` and `invcov` call `Matrix(...)`. Whatever solution we choose, probably better be consistent.

test/mvnormal.jl
```diff
@@ -16,9 +16,6 @@ function test_mvnormal(g::AbstractMvNormal, n_tsamples::Int=10^6,
     d = length(g)
     μ = mean(g)
     Σ = cov(g)
-    @test partype(g) == Float64
-    @test isa(μ, Vector{Float64})
```
ararslan: Maybe keep this but adjust the type as necessary?

matbesancon: the only correct thing to test for is `typeof(mean(g)) === corresponding_type_parameter`, which is a bit odd to put

ararslan: Fair point

ararslan:
...did we get @nalimilan's opinion here?

matbesancon:
@nalimilan mentioned we should be consistent with the covariance Matrix, I wanted to merge this one first and tackle covariance according to Milan's comment in another one. Goal was to keep PRs small, bigger ones tend not to get reviewed.

nalimilan:
Sounds fine. Not sure what to do for the covariance matrix though...

matbesancon:
same thing in that case, just return the cov matrix directly?

nalimilan:
Yeah, why not (I'm not familiar with these special matrix types).

Issue #813, opened 2019-01-15
Uncompatibility rand(MvNormal()) and AutoDiff


theogf:
Hello it is unfortunately not possible to use automatic differentation with (at least) the `MvNormal` distribution.
The following code will fail at `rand(p)` due to a wrong conversion to `Float64`
```julia
using Distributions, ForwardDiff
function foo(mu)
    p = MvNormal(mu)
    sum(rand(p) for _ in 1:100)
end
ForwardDiff.gradient(foo,rand(10))
```

matbesancon:
@theogf PR welcome on that

theogf:
So I found the source of the error but it's in common.jl so not sure if the change would be breaking : 
https://github.com/JuliaStats/Distributions.jl/blob/817bd83326f9d562ec88c0e782e60d25d64862ad/src/common.jl#L51 
Says that whatever the type of the distribution is `eltype` will return `Float64` for a continuous function.  Since `eltype` is called many times for the `MvNormal` sampling, one always end up with `Float64` samples. I don't know the dependency of other distributions on `eltype` but a quick fix would be overloading it for `MvNormal` types

matbesancon:
Yes this should be overloaded for (almost?) all distributions

matbesancon:
@theogf did #882 close this? The error I get is because the function does not return a scalar, which ForwardDiff.gradient requires. It seems fixed with:
```julia
julia> function foo(mu)
           p = MvNormal(mu)
           sum(sum(rand(p) for _ in 1:100))
       end

```

andreasnoack:
Is it really reasonable to expect `rand` to be differentiable? Making the sampler structs parametric introduces a lot of complications since the samplers are usually written for a specific precision. It also touches on the issue of the variate type vs the parameter type. For most distributions, the variate type doesn't follow the parameter type.

mschauer:
Yes, I think this could be important, differentiability of `a + b*randn(rng)` conditional on the `rng` state is practical (keyword normalising flows)

devmotion:
Related: https://github.com/TuringLang/DistributionsAD.jl/issues/123

I think in many cases it is not important to implement the samplers in a differentiable way but it would be useful to add custom adjoints, probably based on ChainRulesCore.

andreasnoack:
[...]

PR #882, opened 2019-05-08
Added eltype for multivariate distributions

theogf:
Should solve #813


mschauer:
This conflates `eltype` (e.g. Vector{Float64}) with the `eltype` of the `eltype` (e.g `Float64`).

theogf:
@mschauer So I am not sure what you mean by this, but for `MvNormal`, multiple functions rely on `eltype(d::MvNormal)` and it is currently set to a default `Float64`
I removed the `eltype` for `Multonomial` as they were a problem for the tests.

matbesancon:
@mschauer in that case `eltype` could be the atomic-most type? Another issue is that what is eltype if the type of the covariance matrix and the mean vector differ?

theogf:
> @mschauer in that case `eltype` could be the atomic-most type? Another issue is that what is eltype if the type of the covariance matrix and the mean vector differ?

From the constructors it does not seem possible to have different elementary types for the mean and the covariance (also it would not make so much sense).

matbesancon:
it would if you are doing Automatic Differentiation on the mean, the covariance would be Float while the mean elements would be Dual

theogf:
I am not super familiar with the backend of AD but what would be the negative side of having both Dual?

mschauer:
I see the interface exists a while but I think if there is need for a function returning the atomic element type it should not be called `eltype`. In the case your change is "internally consistent" with the current interface we could  move this concern to a separate issue though.

theogf:
`eltype` is needed since it is called directly by `rand`

mschauer:
Then rand  calls the wrong function ...

matbesancon:
`eltype(::Dirichlet{T}) where {T<:Real} = T`
@theogf you don't need the type bound here, since by construction a Dirichlet has always `T<:Real`

matbesancon:
@mschauer are you good with fixing this behaviour here? As mentioned modifying the whole interface can go in another issue

mschauer:
Ok!

matbesancon:
ping @theogf can you merge master into this branch and adapt for the comment I let above?

theogf:
Sorry I am away till next week, I take care of it as soon as I am back.

rfourquet:
I'm not familiar with this package, but wanted to note that currently, in `Random`, the meaning of `eltype` is extended to mean: for whatever object `x` (not a type), we must have `rand(x) isa eltype(x)`. So @mschauer comments go in the direction of having `Distributions` and `Random` be consistent.

However, we are there is a discussion on whether we should have a function `gentype` instead of `eltype` (`rand(x) isa gentype(x)`), which defaults to `eltype` (e.g. `rand(collection)` returns an element of the collection, so `eltype` makes sense). The idea is that `eltype` for a distribution is not an ideal name, but more importantly, do we want the possibility of `eltype(x) != gentype(x)` ? I'm just putting this here in case you want to chime in at https://github.com/JuliaLang/julia/issues/31968 (it would be great!)

mschauer:
This needs to be rebased. Could someone chime in with a link to a good explanation?

matbesancon:
@mschauer I think this is fine with commit 
d2c36fe ?

mschauer:
I thought something went wrong because the diff of the branch versus master shows 65 files changed.

matbesancon:
ah you're right. @theogf can you from your branch:
1. merge origin/master
2. push

This should do it

theogf:
Sorry I completely messed up things with git, the force-push should have got rid of my mistakes

matbesancon:
no trouble. There seems to be some misses in codecov, can you check if there are things that should be covered?

src/multivariate/dirichlet.jl
```diff
@@ -57,6 +57,7 @@ end
 
 length(d::DirichletCanon) = length(d.alpha)
 
+eltype(::Dirichlet{T}) where T = T
```
matbesancon:
```suggestion
eltype(::Dirichlet{T}) where {T} = T
```

src/multivariate/dirichletmultinomial.jl
```diff
@@ -24,7 +24,6 @@ ntrials(d::DirichletMultinomial) = d.n
 params(d::DirichletMultinomial) = (d.n, d.α)
 @inline partype(d::DirichletMultinomial{T}) where {T<:Real} = T
```
matbesancon:
```suggestion
@inline partype(d::DirichletMultinomial{T}) where {T} = T
```

src/multivariate/mvlognormal.jl
```diff
@@ -175,6 +175,8 @@ MvLogNormal(Σ::AbstractMatrix) = MvLogNormal(MvNormal(Σ))
 MvLogNormal(σ::AbstractVector) = MvLogNormal(MvNormal(σ))
 MvLogNormal(d::Int,s::Real) = MvLogNormal(MvNormal(d,s))
 
+
+eltype(::MvLogNormal{T}) where {T<:Real} = T
```
matbesancon:
```suggestion
eltype(::MvLogNormal{T}) where {T} = T
```

src/multivariate/mvnormal.jl
```diff
@@ -219,6 +219,8 @@ MvNormal(Σ::Matrix{<:Real}) = MvNormal(PDMat(Σ))
 MvNormal(σ::Vector{<:Real}) = MvNormal(PDiagMat(abs2.(σ)))
 MvNormal(d::Int, σ::Real) = MvNormal(ScalMat(d, abs2(σ)))
 
+
+eltype(::MvNormal{T}) where {T<:Real} = T
```
matbesancon:
```suggestion
eltype(::MvNormal{T}) where {T} = T
```

matbesancon:
I made suggestions for homogeneity. 
`eltype(::D{T}) where {T}` does not  need `T<:Real` if it is already a type constraint for the distribution

mschauer:
We should at a comment which says what the `eltype` of a distribution is, that is the `eltype` of its samples.

theogf:
@matbesancon I added your suggestions. For the codecov, `eltype` should be called during sampling, I guess there are already tests for that? In any case that should probably be added to the list of tests

@mschauer It is already described this way for the function on the abstract type : https://github.com/JuliaStats/Distributions.jl/blob/432545afd3a29397e2308a99a4978b588676c671/src/common.jl#L51

matbesancon:
> I guess there are already tests for that? In any case that should probably be added to the list of tests

Yes can you check tests and add some? The fact that coverage diff is negative shows that they might not be used.

Also, can you add something in the doc: https://juliastats.github.io/Distributions.jl/stable/univariate/#Common-Interface-1
Adding the doc for `eltype` and `partype` would help people

matbesancon:
ping @theogf

theogf:
So I wrote some tests, but I quickly started to realize that there are a lot more issues. 
 - all `GenericMvTDist` constructors were restricted to Float64 (this is corrected now). 
 - `Float16` cannot be tested since there are no `PDMat{Float16}` constructor for some reason
 - The sampler for `GenerivMvTDist` make calls to the `Gamma` samplers but all `Gamma` structs (`GammIPSampler`, `GammaMTSampler`, etc)are also restricted to `Float64` components. 

Should I just continue pulling the string and parametrizing everything until this works?

src/samplers/gamma.jl
```diff
@@ -6,22 +6,22 @@
...
-    s2::Float64
-    s::Float64
-    i2s::Float64
-    d::Float64
-    q0::Float64
-    b::Float64
-    σ::Float64
-    c::Float64
-    scale::Float64
+struct GammaGDSampler{T} <: Sampleable{Univariate,Continuous}
```
matbesancon: Shouldn't `T` be constrained to `Number` or `Real`?

src/samplers/gamma.jl
```diff
@@ -6,22 +6,22 @@
...
+    b::T
+    σ::T
+    c::T
+    scale::T
 end
 
 function GammaGDSampler(g::Gamma)
     a = shape(g)
-
+    T = typeof(a)
```
matbesancon: `T` could be obtained from the parameter in the signature here

test/types.jl
```diff
@@ -21,3 +22,19 @@ using Distributions
 @assert DiscreteMatrixDistribution <: MatrixDistribution
 @assert ContinuousMatrixDistribution <: ContinuousDistribution
 @assert ContinuousMatrixDistribution <: MatrixDistribution
+
+@testset "Test Sample Type" begin
+    for T in [Float64,Float32,Dual{Nothing,Float64,0}]
+    # for T in [Float64,Float32,Float16,Tracker.TrackedReal{Float64}]
+        @testset "Type $T" begin
+            for d in [MvNormal]#,MvLogNormal,MvNormalCanon,Dirichlet]
```
matbesancon:
```
for d in (MvNormal, )
```
avoids allocating a vector

test/types.jl
```diff
@@ -21,3 +22,19 @@ using Distributions
 @assert DiscreteMatrixDistribution <: MatrixDistribution
 @assert ContinuousMatrixDistribution <: ContinuousDistribution
 @assert ContinuousMatrixDistribution <: MatrixDistribution
+
+@testset "Test Sample Type" begin
+    for T in [Float64,Float32,Dual{Nothing,Float64,0}]
```
matbesancon:
```suggestion
    for T in (Float64,Float32,Dual{Nothing,Float64,0})
```

src/samplers/gamma.jl
```diff
@@ -6,22 +6,22 @@
...
     # Step 1
     s2 = a-0.5
```
matbesancon:
```suggestion
    s2 = a - 0.5
```

src/multivariates.jl
```diff
@@ -14,6 +14,13 @@ Return the sample size of distribution `d`, *i.e* `(length(d),)`.
 """
 size(d::MultivariateDistribution)
 
+
+"""
+    eltype(d::MultivariateDistribution)
```
matbesancon:
```suggestion
    eltype(d::MultivariateDistribution)

```

src/multivariate/mvtdist.jl
```diff
@@ -32,7 +32,10 @@ end
 
 GenericMvTDist(df::Real, μ::Vector{S}, Σ::Cov) where {Cov<:AbstractPDMat, S<:Real} = GenericMvTDist(df, μ, Σ, allzeros(μ))
 
-GenericMvTDist(df::T, Σ::Cov) where {Cov<:AbstractPDMat, T<:Real} = GenericMvTDist(df, zeros(dim(Σ)), Σ, true)
+function GenericMvTDist(df::T, Σ::Cov) where {Cov<:AbstractPDMat, T<:Real}
+    R = Base.promote_eltype(T, Σ)
```
matbesancon: Are you sure `Base.promote_eltype` is what's needed here?

src/samplers/gamma.jl
theogf: Indeed! I correct it.

matbesancon:
A significant amount of code has been added, should there be some additions in the doc? In the tests?

src/multivariate/mvtdist.jl
theogf: I think so, we want a `promote_type` between `T` and `eltype` of `Σ` and `T<:Real`

theogf:
Most of the code has been about relaxing the constructors `MvTDist` to `Real` variables and to parametrize the `Gamma` samplers (previously fixed to `Float64`).
Maybe I should add a line in `extends.md` to suggest to implement `eltype` for new distributions.

matbesancon:
Yes agreed

matbesancon:
@theogf I'm ok with merging this soon-ish (this week), if you want to add tests before for lines not covered let me know. You can see lines not covered here:
https://codecov.io/gh/JuliaStats/Distributions.jl/compare/432545afd3a29397e2308a99a4978b588676c671...365802f9e4cb438a67adb32b3b1fefe376c7713e/changes

matbesancon:
@theogf sorry I let this slept longer than intended, can you merge master into this branch, I'll merge it then

theogf:
@matbesancon Done! Sorry I also did not answer you, I was first confused by the report and ended up forgetting about it!

matbesancon:
ok, this should be good in ~1hour. I think the coverage decreased because of the new `eltype` methods

matbesancon:
Thanks  @theogf

PR #883, opened 2019-05-08
Create CITATION.bib

asinghvi17:


matbesancon:
Thanks, in that case it would be better to point to this file in the README:

```
# CITING

See the `CITATION.bib` file
```

matbesancon:
Thanks @asinghvi17

PR #884, opened 2019-05-09
Fix suffstats and fit_mle with concrete types

### LINKED ISSUE #809: Sometimes fit(D, data) does not work if D is a concrete type
Works as expected:

````
julia> using Distributions
julia> fit(Normal, rand(100))
Normal{Float64}(μ=0.4713220204131837, σ=0.2850689160341615)
````

However:
````
julia> fit(Normal{Float64}, rand(100))
ERROR: suffstats is not implemented for (Normal{Float64}, Array{Float64,1}).
````
This could be resolved in `src/univariate/continuous/normal.jl` by replacing `::Type{Normal}` declarations with the more appropriate `::Type{<:Normal}` declaration (see excerpt below). I think this kind of thing is not restricted to `Normal` but is quite pervasive. 

````
function suffstats(::Type{Normal}, x::AbstractArray{T}) where T<:Real
    n = length(x)

    # compute s
    s = x[1]
    for i = 2:n
        @inbounds s += x[i]
    end
    m = s / n

    # compute s2
    s2 = abs2(x[1] - m)
    for i = 2:n
        @inbounds s2 += abs2(x[i] - m)
    end

    NormalStats(s, m, s2, n)
end
````

giordano:
Possibly fix #809.

### DIFF
--- a/docs/src/fit.md
+++ b/docs/src/fit.md
@@ -8,6 +8,17 @@ d = fit(D, x)
 
 This statement fits a distribution of type `D` to a given dataset `x`, where `x` should be an array comprised of all samples. The fit function will choose a reasonable way to fit the distribution, which, in most cases, is [maximum likelihood estimation](http://en.wikipedia.org/wiki/Maximum_likelihood).
 
+!!! note
+
+    One can use as first argument simply the distribution name, like `Binomial`,
+    or a concrete distribution with a type parameter, like `Normal{Float64}` or
+    `Exponential{Float32}`.  However, in the latter case the type parameter of
+    the distribution will be ignored:
+
+    ```julia
+    julia> fit(Cauchy{Float32}, collect(-4:4))
+    Cauchy{Float64}(μ=0.0, σ=2.0)
+    ```
 
 ## Maximum Likelihood Estimation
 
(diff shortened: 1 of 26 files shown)


matbesancon:
@giordano fantastic thanks for tackling this. Could you add a bunch of tests to verify this has been solved?

giordano:
I tried to skip that :innocent:  Ok, I'll add some tests, anyway before submitting I did try with a couple of distributions and they worked.

giordano:
Question: what should we do in case there is a mismatch between the parameter of the distribution and the eltype of the vector/matrix, e.g. `fit(Dirichlet{Float32}, ::Matrix{Float64})`?  As it's now, the output would be a `Dirichlet{Float64}`.

matbesancon:
Yes this is fine, the goal was just to avoid blowing up with exact types

matbesancon:
looks good, merging in 24 hours if no objection, thanks!

matbesancon:
edit: merging after doc change

giordano:
Done!  I've been quite honest: the parameter is simply ignored.

matbesancon:
Merged, thanks!

PR #888, opened 2019-05-10
Cleanup test

matbesancon:
1. Simplification of tests, removed unnecessary overhead
2. Disambiguation of methods, especially problematic for `DiscreteNonParametric` and `Truncate`


src/truncate.jl
```diff
@@ -49,34 +49,81 @@ insupport(d::Truncated{D,Union{Discrete,Continuous}}, x::Real) where {D<:Univari
...
     x >= d.upper ? one(T) :
     (cdf(d.untruncated, x) - d.lcdf) / d.tp
 
-logcdf(d::Truncated, x::T) where {T<:Real} =
-    x <= d.lower ? -T(Inf) :
+cdf(d::Truncated, x::Real) = _cdf(d, x)
+cdf(d::Truncated, x::Integer) = _cdf(d, x) # here for specificity
+
+_logcdf(d::Truncated, x::T) where {T<:Real} =
+    x <= d.lower ? -Inf :
```
nalimilan: `T(Inf)` is useful to ensure type stability.

matbesancon: ah point, it just broke with integers which don't have `Inf` values, these will be unstable in any case

matbesancon: solution might be to always return a `Float64`? The type of probability values does not have to be the same as type of input `x`

nalimilan: You can call `float(T)`, that will preserve `Float32`.

matbesancon: fixed it on the various functions

matbesancon:
with the last commit: TIL stale exports don't break a package and maybe they should :/

src/truncate.jl
```diff
@@ -49,33 +49,94 @@ insupport(d::Truncated{D,Union{Discrete,Continuous}}, x::Real) where {D<:Univari
...
+        zero(T)
+    end
+end
+
+function pdf(d::Truncated{D}, x::T) where {D<:ContinuousUnivariateDistribution, T<:Real}
+    _pdf(d, float(x))
+end
+
+function pdf(d::Truncated{D}, x::T) where {D<:DiscreteUnivariateDistribution, T<:Real}
+    isinteger(x) || zero(float(T))
```
nalimilan:
```suggestion
    isinteger(x) || return zero(float(T))
```

src/truncate.jl
```diff
@@ -49,33 +49,94 @@ insupport(d::Truncated{D,Union{Discrete,Continuous}}, x::Real) where {D<:Univari
...
+    _pdf(d, x)
+end
+
+function pdf(d::Truncated{D}, x::T) where {D<:DiscreteUnivariateDistribution, T<:Integer}
```
nalimilan: I don't think this method is needed, since `isinteger(::Integer)` will be determined to be `true` at compile time in the method above.

src/truncate.jl
```diff
@@ -49,33 +49,94 @@ insupport(d::Truncated{D,Union{Discrete,Continuous}}, x::Real) where {D<:Univari
...
+    _pdf_(d, float(x))
+end
+
+function _logpdf(d::Truncated, x::T) where {T<:Real}
+    if (d.lower <= x <= d.upper)
```
nalimilan:
```suggestion
    if d.lower <= x <= d.upper
```

src/truncate.jl
```diff
@@ -49,33 +49,94 @@ insupport(d::Truncated{D,Union{Discrete,Continuous}}, x::Real) where {D<:Univari
...
-    @eval ($f)(d::Truncated, x::Int) = ($f)(d, float(x))
+function _pdf(d::Truncated, x::T) where {T<:Real}
+    if d.lower <= x <= d.upper
+        pdf(d.untruncated, x) / d.tp
+    else
+        zero(T)
+    end
+end
+
+function pdf(d::Truncated{D}, x::T) where {D<:ContinuousUnivariateDistribution, T<:Real}
```
nalimilan: This is very slightly shorter. You could also switch to the short function form for one-liners.
```suggestion
function pdf(d::Truncated{<:ContinuousUnivariateDistribution}, x::T) where {T<:Real}
```

src/truncate.jl
```diff
@@ -49,33 +49,94 @@ insupport(d::Truncated{D,Union{Discrete,Continuous}}, x::Real) where {D<:Univari
...
+    return _logpdf(d, x)
 end
 
-pdf(d::Truncated, x::T) where {T<:Real} = d.lower <= x <= d.upper ? pdf(d.untruncated, x) / d.tp : zero(T)
+function logpdf(d::Truncated{D}, x::Integer) where {D<:DiscreteUnivariateDistribution}
+    _logpdf(d, x)
+end
 
-logpdf(d::Truncated, x::T) where {T<:Real} = d.lower <= x <= d.upper ? logpdf(d.untruncated, x) - d.logtp : -T(Inf)
+function logpdf(d::Truncated{D, Continuous}, x::T) where {D<:ContinuousUnivariateDistribution, T<:Real}
```
nalimilan: Why not merge this with the previous one?

src/truncate.jl
```diff
@@ -49,33 +49,94 @@ insupport(d::Truncated{D,Union{Discrete,Continuous}}, x::Real) where {D<:Univari
...
     x <= d.lower ? zero(T) :
     x >= d.upper ? one(T) :
     (cdf(d.untruncated, x) - d.lcdf) / d.tp
 
-logcdf(d::Truncated, x::T) where {T<:Real} =
-    x <= d.lower ? -T(Inf) :
-    x >= d.upper ? zero(T) :
-    log(cdf(d.untruncated, x) - d.lcdf) - d.logtp
+cdf(d::Truncated, x::Real) = _cdf(d, x)
+cdf(d::Truncated, x::Integer) = _cdf(d, float(x)) # float conversion for stability
```
nalimilan: Stability isn't the issue here, is it?

src/univariates.jl
```diff
@@ -306,17 +306,24 @@ cf(d::UnivariateDistribution, t)
 # pdf
 
 """
-    pdf(d::UnivariateDistribution, x::Real)
+    pdf(d::UnivariateDistribution, x::T) where {T<:Real}
```
nalimilan:
```suggestion
    pdf(d::DiscreteUnivariateDistribution, x::T) where {T<:Real}
```

Same below.

src/univariates.jl
```diff
@@ -306,17 +306,24 @@ cf(d::UnivariateDistribution, t)
...
+type but `x` is integer, the value is converted to `Int`.
 
 See also: [`logpdf`](@ref).
 """
-pdf(d::UnivariateDistribution, x::Real)
-pdf(d::DiscreteUnivariateDistribution, x::Int) = throw(MethodError(pdf, (d, x)))
-pdf(d::DiscreteUnivariateDistribution, x::Integer) = pdf(d, round(Int, x))
 pdf(d::DiscreteUnivariateDistribution, x::Real) = isinteger(x) ? pdf(d, round(Int, x)) : 0.0
 
+"""
```
nalimilan: This docstring would better be merged with the one above IMHO.

src/truncate.jl
```diff
@@ -49,33 +49,94 @@ insupport(d::Truncated{D,Union{Discrete,Continuous}}, x::Real) where {D<:Univari
...
+    _pdf(d, float(x))
+end
+
+function pdf(d::Truncated{D}, x::T) where {D<:DiscreteUnivariateDistribution, T<:Real}
+    isinteger(x) || zero(float(T))
+    _pdf(d, x)
+end
+
+function pdf(d::Truncated{D}, x::T) where {D<:DiscreteUnivariateDistribution, T<:Integer}
+    _pdf_(d, float(x))
```
nalimilan:
```suggestion
    _pdf(d, float(x))
```

src/truncate.jl
```diff
@@ -49,33 +49,94 @@ insupport(d::Truncated{D,Union{Discrete,Continuous}}, x::Real) where {D<:Univari
 
 quantile(d::Truncated, p::Real) = quantile(d.untruncated, d.lcdf + p * d.tp)
 
-for f in [:pdf, :logpdf, :cdf, :logcdf, :ccdf, :logccdf]
-    @eval ($f)(d::Truncated, x::Int) = ($f)(d, float(x))
+function _pdf(d::Truncated, x::T) where {T<:Real}
```
nalimilan: Do you really need to define these underscored functions? Is that to avoid ambiguities?

src/truncate.jl
matbesancon: yes in `_cdf`:
```
_cdf(d::Truncated, x::T) where {T<:Real} =
    x <= d.lower ? zero(T) :
    x >= d.upper ? one(T) :
    (cdf(d.untruncated, x) - d.lcdf) / d.tp
```
If `T` is integer, you would return its type if out of the interval, and a float otherwise

src/truncate.jl
matbesancon: (see main thread)

matbesancon:
Yes, all the underscore methods are used as "real" implementation while the exposed one is used for dispatch, otherwise we have some signatures being non-specific enough because of generic fallbacks.

src/univariates.jl
matbesancon: done thanks

src/univariates.jl
matbesancon: why here?

matbesancon: ah I see sorry, fixed in last commit

src/truncate.jl
matbesancon: its main purpose it allowing dispatch

src/truncate.jl
matbesancon: dispatch

src/truncate.jl
nalimilan: How so?

src/utils.jl
```diff
@@ -30,12 +30,12 @@ Base.convert(::Type{<:Vector}, v::ZeroVector{T}) where {T} = Vector(v)
 
 Base.convert(::Type{ZeroVector{T}}, v::ZeroVector) where {T} = ZeroVector{T}(length(v))
 
-for T = (:AbstractArray, :Number), op = (:+, :-)
-    @eval begin
-        Base.@deprecate ($op)(x::$T, v::ZeroVector) broadcast($op, x, v)
-        Base.@deprecate ($op)(v::ZeroVector, x::$T) broadcast($op, v, x)
-    end
-end
+# for T = (:AbstractArray, :Number), op = (:+, :-)
```
nalimilan: Revert or remove?

src/truncate.jl
matbesancon: there is a method generic on discrete distribution types and Integer x, so to avoid ambiguous method definition, you need to implement `pdf(::D, x::Integer)`

matbesancon: so you need this method regardless. Given that it's here, better not put the integer check here

src/utils.jl
matbesancon: ah indeed thanks, I was so happy to push once all tests passed that I forgot to remove it definitely

PR #890, opened 2019-05-10
Add a WeightedResampler

rofinn:
Works for all VariateForm and ValueSupport types

The tests should document our use case, but two questions remain.

1. We've been using `@testset`s internal, but it's unclear if that's the desired test organization for Distributions.jl
2. There is some overlap with the `AliasTable` type, so perhaps we can merge the two... or maybe the `WeightedResampler` should use an alias table internally?


ararslan:
The implementation seems good to me but I'm not that familiar with the samplers stuff here.

src/samplers/resampler.jl
```diff
@@ -0,0 +1,53 @@
...
+"""
+function WeightedResampler(obs::T, wv::AbstractWeights) where T<:AbstractArray
+    F = _variate_form(T)
+    S = _value_support(eltype(T))
+
+    _validate(obs, wv)
+    WeightedResampler{F, S}(obs, wv)
+end
+
+_variate_form(::Type{<:AbstractVector}) = Univariate
```
matbesancon: why not use `variate_form` and `value_support` instead of underscore functions?

rofinn: ... because i didn't realize those functions already existed? :+1

rofinn: Would it make sense to document those functions in this MR? https://github.com/JuliaStats/Distributions.jl/pull/893/files

matbesancon: yes indeed, let me fix that

src/samplers.jl
```diff
@@ -10,7 +10,8 @@ for fname in ["aliastable.jl",
               "vonmises.jl",
               "vonmisesfisher.jl",
               "discretenonparametric.jl",
-              "categorical.jl"]
+              "categorical.jl",
+              "resampler.jl"]
```
matbesancon:
```suggestion
              "resampler.jl",
]
```

Like this it will not modify two lines to add a new file

src/samplers/resampler.jl
```diff
@@ -0,0 +1,53 @@
+struct WeightedResampler{F<:VariateForm, S<:ValueSupport} <: Sampleable{F, S}
+    obs::AbstractArray
+    wv::AbstractWeights
+end
+
+"""
```
matbesancon: This constructor could be the default one, and the docstring moved to the struct definition

src/samplers/resampler.jl
matbesancon: @rofinn I am thinking it is not a good idea to use these functions here, variate_form and value_support are supposed to take distributions in entry, passing them Array types would be some very unexpected use for them, so keeping the underscore functions is an acceptable solution here I think

rofinn: So we shouldn't extend those functions to support passing in an array type? Seems like something other samplers and distributions might find useful?

src/samplers/resampler.jl
rofinn: I'm not sure I know what you mean by "default". Are you suggesting that it should be an inner constructor or just that the docstring should be attached to the type?

Project.toml
```diff
@@ -15,6 +15,7 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 
 [compat]
+Distances = "0.8"
```
matbesancon: Freezing on Distances can be  constraining, since it is only for tests we can leave it unconstrained for the moment. If at one point it breaks we will see it on Travis and will re-add the constraint here

src/samplers/resampler.jl
matbesancon: sorry, yes that it should be attached to the type.

src/samplers/resampler.jl
```diff
@@ -0,0 +1,53 @@
+struct WeightedResampler{F<:VariateForm, S<:ValueSupport} <: Sampleable{F, S}
```
matbesancon: `WeightedResampler` should be parameterized on `obs` real type

src/samplers/resampler.jl
matbesancon: It can lead to ambiguous behavior, we can keep this outside the scope of this PR.
Can you add some tests for the underscore functions? I know it is trivial, but we'll try to maintain the codecov high

rofinn: Sure. I believe all of those methods should already be getting hit, but more unit tests are always good 👍

nickrobinson251:
This would be useful for me :) thanks for adding, Rory!

Bump?

nickrobinson251:
Is there anything left to do, before merging this?

nickrobinson251:
Bump :)

matbesancon:
Sorry about the delay, this looks good to my untrained eyes, and been stalled too long. @rofinn we can merge this after the conflict is resolved

Project.toml
```diff
@@ -32,4 +33,6 @@ StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Calculus", "Distributed", "FiniteDifferences", "ForwardDiff", "JSON", "StaticArrays", "Test"]
+test = ["Calculus", "Distances", "Distributed", "FiniteDifferences", "ForwardDiff", "JSON", "StaticArrays", "Test"]
+
+
```
matbesancon:
```suggestion
```

src/samplers/resampler.jl
```diff
@@ -0,0 +1,53 @@
...
+
+This type supports univariate, multivariate and matrixvariate forms, so `obs` can
+be a vector of values, matrix of values or a vector of matrices.
+"""
+struct WeightedResampler{F<:VariateForm, S<:ValueSupport, T<:AbstractArray} <: Sampleable{F, S}
+    obs::T
+    wv::AbstractWeights
+end
+
+function WeightedResampler(obs::T, wv::AbstractWeights) where T<:AbstractArray
```
nalimilan: Is there any reason to restrict weights to `AbstractWeights` rather than `AbstractVector`? At https://github.com/JuliaLang/julia/pull/31395, I made functions accept any array since there's no ambiguity.

matbesancon: ping @rofinn @nickrobinson251

matbesancon:
bump on this :)

src/samplers/resampler.jl
nickrobinson251: Idk but maybe @rofinn does

But also it'd be an easy follow up to loosen it

matbesancon: ping @rofinn ?

nickrobinson251: Can we merge as is and open an issue about this? It'd be a non breaking change to loosen the constraint in future. Or just loosen it now and fix it if there's a bug report

nalimilan: I guess merging this as-is is OK. I've noticed that the weights are passed directly to `sample`, so if we loosen the signature we should also ensure we convert the argument to `AbstractWeights`, which is what `sample` expects (currently at least). But I still think loosening this is a good idea, though not a requirement.

src/samplers/resampler.jl
```diff
@@ -0,0 +1,53 @@
...
+
+A WeightedResampler is a subtype of Distributions.Sampleable which randomly selects
+observations from the raw input data (`obs`) based on the weights (`wv`) provided.
+
+This type supports univariate, multivariate and matrixvariate forms, so `obs` can
+be a vector of values, matrix of values or a vector of matrices.
+"""
+struct WeightedResampler{F<:VariateForm, S<:ValueSupport, T<:AbstractArray} <: Sampleable{F, S}
+    obs::T
+    wv::AbstractWeights
```
nalimilan: Woops, this should be a type parameter!

rofinn:
Closing due to inactivity.

Issue #156, opened 2013-10-11
One-time use categorical sampler


malmaud:
Say I have an unnormalized vector of probabilities p, and want to take a single sample from the categorical distribution parameterized by p. Would I be correct in thinking that

``` julia
rand(Categorical(p/sum(p)))
```

is relatively inefficient for this case? Is there a method in this package or julia base designed for this case, and if not, would you be interested in having one?

Issue #154, opened 2013-10-02
Exchangeable non-paramatetric distributions


malmaud:
Does this package have support for any of the common Bayesian non-parametric distributions, like the Chinese restaurant restaurant or Indian buffet process? Has thought been put into what the API would look like for representing and sampling from those kinds of stochastic objects?

YSanchezAraujo:
Hi, just wondering if there ever was a package created for this? It seems like RandomProcesses didn't end up being maintained.

PR #891, opened 2019-05-10
Ensure compatibility with SpecialFunctions.jl

Sumegh-git:
Recently following a discussion in [issue #154](https://github.com/JuliaMath/SpecialFunctions.jl/issues/154) a change was made in [PR #156](https://github.com/JuliaMath/SpecialFunctions.jl/pull/156) to rename `lgamma` function into more descriptive `loggamma` and `logabsgamma`. 
This PR ensures compatibility with the latest `SpecialFunctions.jl` functions . Also added a fallback in `deprecations.jl` file to ensure that it works even now, until SpecialFunctions is upgraded.
cc @simonbyrne

### DIFF
--- a/src/deprecates.jl
+++ b/src/deprecates.jl
@@ -1,11 +1,11 @@
 #### Deprecate on 0.6 (to be removed on 0.7)
 
 @Base.deprecate expected_logdet meanlogdet
-if isdefined(SpecialFunctions, :logabsgamma)
-    logabsgamma(x) = (lgamma(x),sign(gamma(x)))
+if !isdefined(SpecialFunctions, :logabsgamma)
+    logabsgamma(x) = SpecialFunctions.lgamma_r(x)
     loggamma(x) = lgamma(x)
 end
-if isdefined(SpecialFunctions, :logabsbeta)
+if !isdefined(SpecialFunctions, :logabsbeta)
     logabsbeta(x, w) = (lbeta(x, w), sign(beta(x, w)))
     logbeta(x, w) = lbeta(x, w)
 end
--- a/src/deprecates.jl
+++ b/src/deprecates.jl
@@ -1,6 +1,10 @@
 #### Deprecate on 0.6 (to be removed on 0.7)
 
 @Base.deprecate expected_logdet meanlogdet
+if isdefined(SpecialFunctions, :logabsgamma)
+    logabsgamma(x) = (lgamma(x),sign(gamma(x)))
+    loggamma(x) = lgamma(x)
+end
 
 function probs(d::DiscreteUnivariateDistribution)
     Base.depwarn("probs(d::$(typeof(d))) is deprecated. Please use pdf(d) instead.", :probs)
--- a/src/deprecates.jl
+++ b/src/deprecates.jl
@@ -5,6 +5,10 @@ if isdefined(SpecialFunctions, :logabsgamma)
     logabsgamma(x) = (lgamma(x),sign(gamma(x)))
     loggamma(x) = lgamma(x)
 end
+if isdefined(SpecialFunctions, :logabsbeta)
+    logabsbeta(x, w) = (lbeta(x, w), sign(beta(x, w)))
+    logbeta(x, w) = lbeta(x, w)
+end
 
 function probs(d::DiscreteUnivariateDistribution)
     Base.depwarn("probs(d::$(typeof(d))) is deprecated. Please use pdf(d) instead.", :probs)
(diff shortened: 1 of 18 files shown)


src/deprecates.jl
```diff
@@ -1,6 +1,14 @@
 #### Deprecate on 0.6 (to be removed on 0.7)
 
 @Base.deprecate expected_logdet meanlogdet
+if isdefined(SpecialFunctions, :logabsgamma)
```
simonbyrne:
```suggestion
if !isdefined(SpecialFunctions, :logabsgamma)
```
(and same below)

src/deprecates.jl
```diff
@@ -1,6 +1,14 @@
...
+    logabsgamma(x) = (lgamma(x),sign(gamma(x)))
```
simonbyrne:
```suggestion
    logabsgamma(x) = SpecialFunctions.lgamma_r(x)
```

src/deprecates.jl
Sumegh-git: Oh !!!!
Very sorry ! Yeah correcting it.

ararslan:
IMO we should just put an upper compatibility bound on SpecialFunctions in the Project.toml, then once the SpecialFunctions release with these changes lands, we can set the version as the lower compatibility bound. Then we don't have to do any of these conditionals.

simonbyrne:
Ah, it looks like one is already there:
https://github.com/JuliaStats/Distributions.jl/blob/f0fd5a80c638544a4ac9888c6fcf1a6d7d956164/Project.toml#L20

Am I correct in saying that it won't then install SpecialFunctions 0.8?

ararslan:
Yes, that's correct, since there is an implicit caret and 0.7 and 0.8 are not considered semver-compatible.

matbesancon:
ping @Sumegh-git, this is still not passing on travis

Sumegh-git:
That happened due to a time-out. The tests have passed. So is there anything I can change to fix that?

matbesancon:
indeed I hadn't seen, I just restarted the failing test. Can you merge master into your branch?

matbesancon:
```julia
  StackOverflowError:
  Stacktrace:
   [1] lgamma_r(::Dual{Nothing,Float64,0}) at /Users/travis/.julia/packages/SpecialFunctions/fvheQ/src/gamma.jl:578 (repeats 80000 times)
```

@Sumegh-git

Sumegh-git:
I didn't get the error? Can you explain once?

matbesancon:
It means the function seems to be recursively calling itself

matbesancon:
ping @Sumegh-git here

tpapp:
I think the above recursive call is coming from 
```julia
float(x) = AbstractFloat(x)
```
in `Base` and `ForwardDiff` not defining `float`, so it just falls back to itself *ad infinitum*.

Something like
```julia
function gamma(x::T) where {T <: Real}
    S = float(T)
    S === T ? throw(MethodError(gamma, x)) : gamma(S(x))
end
```
could work around this.

It would be important to fix this and merge this commit since this is showing up in many places (eg https://github.com/denizyuret/Knet.jl/issues/486)

matbesancon:
ping @Sumegh-git, @simonbyrne to finish this

matbesancon:
Also blocked by https://github.com/JuliaMath/SpecialFunctions.jl/issues/186

oschulz:
I think this is becoming more urgent now. Packages are starting to require SpecialFunctions v0.8 (e.g. Zygote v0.4.0, just released). Since both SpecialFunctions and Distributions are so fundamental, this will probably block a lot of package updates quite soon.

Anything I can do to help/contribute?

matbesancon:
@oschulz yes see the issues referenced above in SpecialFunctions, once v0.8 is patched we can upgrade it

oschulz:
Ah, I see - thanks, @matbesancon!

andreasnoack:
I think what's needed are new definitions in `DiffRules` for `loggamma` and `logabsgamma`. I'll try to fix that.

torfjelde:
I believe I have a solution over at https://github.com/JuliaMath/SpecialFunctions.jl/issues/186#issuecomment-550560982

andreasnoack:
@torfjelde Thanks for linking to your solution. I hadn't seen it. I realized that we should just use `loggamma` in this PR instead of `logabsgamma` since the argument is always positive.

andreasnoack:
Superseded by #1009

Issue #668, opened 2017-09-20
Chart of univariate probability distributions


dmbates:
It might be fun to check out how much of http://www.math.wm.edu/~leemis/chart/UDR/UDR.html we have covered.

matbesancon:
Nice one, do you think we should add a link to it in the docs?

matbesancon:
otherwise we can close this one

Issue #124, opened 2013-08-10
Missing distributions list


johnmyleswhite:
Since we're making pretty great progress at including probability distributions, I thought I'd add a list of distributions we're missing that we might want to add to rival SciPy's list:
# Continuous Distribution
- [ ] alpha - An alpha continuous random variable
- [ ] anglit - An anglit continuous random variable
- [x] arcsine - An arcsine continuous random variable
- [x] betaprime - A beta prima continuous random variable
- [ ] bradford - A Bradford continuous random variable
- [ ] burr - A Burr continuous random variable.
- [x] cosine - A cosine continuous random variable
- [ ] dgamma - A double gamma continuous random variable
- [ ] dweibull - A double Weibull continuous random variable
- [x] erlang - An Erlang continuous random variable
- [x] expon - An exponential continuous random variable
- [ ] exponweib - An exponentiated Weibull continuous random variable
- [x] exponpow - An exponential power continuous random variable
- [ ] fatiguelife - A fatigue-life (Birnbaum-Sanders) continuous random variable
- [ ] fisk - A Fisk continuous random variable
- [ ] foldcauchy - A folded Cauchy continuous random variable
- [ ] foldnorm - A folded normal continuous random variable
- [ ] frechet_r - A Frechet right (or Weibull minimum) continuous random variable
- [ ] frechet_l - A Frechet left (or Weibull maximum) continuous random variable
- [ ] genlogistic - A generalized logistic continuous random variable
- [x] genpareto - A generalized Pareto continuous random variable
- [ ] genexpon - A generalized exponential continuous random variable
- [x] genextreme - A generalized extreme value continuous random variable
- [ ] gausshyper - A Gauss hypergeometric continuous random variable
- [ ] gengamma - A generalized gamma continuous random variable
- [ ] genhalflogistic - A generalized half-logistic continuous random variable
- [ ] gilbrat - A Gilbrat continuous random variable
- [ ] gompertz - A Gompertz (or truncated Gumbel) continuous random variable
- [x] gumbel_r - A right-skewed Gumbel continuous random variable
- [ ] gumbel_l - A left-skewed Gumbel continuous random variable
- [ ] halfcauchy - A Half-Cauchy continuous random variable
- [ ] halflogistic - A half-logistic continuous random variable
- [ ] halfnorm - A half-normal continuous random variable
- [ ] hypsecant - A hyperbolic secant continuous random variable
- [ ] invweibull - An inverted Weibull continuous random variable
[...]

PR #899, opened 2019-05-16
Add univariate list, contrib page & Travis update.

taqtiqa-mark:
Update to Travis staged build.
 - The exisitng build has these drawbacks:
    - Code in the after_success: part of the build does not fail the
      build.  This makes it difficult to verify that (i) the doc build
      works and (ii) that doctest etc. still passes.
    - Doc building runs in the global environment, and is thus affected
      by the surroundings.

Add univariate listing implemented, not and won't.
 - Closes issue JuliaStats/Distributions.jl #124 and issue
   JuliaStats/Distributions.jl #668.
 - Includes (most) Wikipedia list of continuous univariate dist.

Add Contributing page to docs site.
 - Modelled on Julia and Documenter.jl page.
 - Included excerpt from README.md.


docs/src/contributing.md
```diff
@@ -0,0 +1,96 @@
...
+3. At the end of the tests, `Test.detect_ambiguities(Distributions)` is run to check method ambiguities. Verify that your modified code did not yield method ambiguities.
+4. Make according modifications to the `docs` folder, build the documentation locally with `$ julia docs/make.jl`, verify that your modifications display correctly and did not yield warnings.
+
+## Branches
+
+This package uses `release-*` branches for tagged minor versions. This follows the same approach used in the main Julia repository, albeit on a much more modest scale.
+
+Please open pull requests against the `master` branch rather than any of the `release-*` branches whenever possible.
+
+### Backports
```
matbesancon: Don't think we backport things in Distributions

docs/src/contributing.md
```diff
@@ -0,0 +1,96 @@
...
+
+Bug fixes are backported to the `release-*` branches using `git cherry-pick -x` and will become available in point releases of that particular minor version of the package.
+
+Feel free to nominate commits that should be backported by opening an issue. Requests for new point releases to be tagged in `METADATA.jl` can also be made in the same way.
+
+### `release-*` branches
```
matbesancon: We don't have release branches

docs/src/univariate-list.md
```diff
@@ -0,0 +1,269 @@
+# [Univariate Distributions List](@id univariates-list)
```
matbesancon: Is is not redundant with `univariates.md`? Plus it creates a hard to track distributions list that needs to be updated. The issue tracker might be more adapted

matbesancon:
thanks for this PR @taqtiqa-mark. There has been some changes on Travis on the last PRs, would you mind merging master into your branch? Also, one different build stage has been defined for docs in the latest commits.
I left some other comments in the code for the other parts

docs/src/contributing.md
taqtiqa-mark: Hmm, on reflection and further reading: 
What to make of the `/^v\d+\.\d+(\.\d+)?(-\S*)?$/` filter that was just added to Travis? Should this section be reworded to refer to the Travis regex naming convention? Or removed entirely?

docs/src/univariate-list.md
taqtiqa-mark: Re redundancy:  
 - Think of this as an index page, with all the functionality and convenience use cases that index pages were created for.
 - Also think of this as a cross reference tool - not all distributions are known by the same name across all SME domains.

Happy to stand corrected on the cross referencing.

Yes this distributions list would need to be updated as distributions are added - more importantly reviews would be encouraged to identify synonyms that should be cross referenced to the incoming distribution.

I think the current issues are only in the search path of devs, not general users.

docs/src/contributing.md
taqtiqa-mark: So that section would be replaced by the 1-liner:  Changes are not backported in Distributions.jl

matbesancon: Yes, or even not mention it. It's not expected for most Julia packages. Even less while Distributions is < 1.0

docs/src/contributing.md
matbesancon: It can be removed. The travis regex is just declaring to push to documentation major and minor releases

taqtiqa-mark:
@matbesancon is there a reason you picked out only the documentation job and kept the matrix instructions.  

Ack. Travis-CI is fast moving and the doc's haven't always been able to keep up.... My understanding is the the way `jobs` are used in `.travis.yml` in this PR is the updated approach.  Mixing matrix and jobs has led to some subtle pain points previously.  

Is there a reason you want to retain the superseded syntax and behavior?

matbesancon:
I copied the travis template from:
https://github.com/fredrikekre/Literate.jl/blob/master/.travis.yml
Agreed this is more of a hack, I'm fine with changing to jobs completely.

matbesancon:
Again, I'm not sold on the index page, documentation is a high burden to maintain, as shown by the huge recent PR on them. I prefer an issue for the implemented and non implemented ones, like #124, to which we can point in the README

taqtiqa-mark:
No worries.  You know where to find it if you change you mind.

matbesancon:
thanks for understanding

docs/make.jl
```diff
@@ -10,12 +10,14 @@ makedocs(
         "starting.md",
         "types.md",
         "univariate.md",
+        "univariate-list.md",
```
matbesancon:
```suggestion
```

matbesancon:
One last suggestion in the diff, and think of merging master into your branch to check that you're up to date

docs/make.jl
```diff
@@ -16,6 +16,7 @@ makedocs(
         "mixture.md",
         "fit.md",
         "extends.md",
+        "contributing.md"
```
matbesancon:
```suggestion
        "contributing.md",
```
So that adding new lines reduces the diff

docs/src/contributing.md
```diff
@@ -0,0 +1,79 @@
...
+
+1. If you need explanation on how to do X, Y using Distributions,
+feel free to ask on the Julia [Discourse](https://discourse.julialang.org/c/domain/stats)
+or [Slack](https://julialang.slack.com), get an invitation
+[here](https://slackinvite.julialang.org/).
+
+2. If you have a bug linked with Distributions, check that it has
+not been reported yet on the issues of the repository.
+If not, you can file a new issue, add your version of the package
+which you can get with this command in the Julia REPL:
```
matbesancon:
```suggestion
which you can obtain with this command in the Julia REPL:
```

docs/src/contributing.md
```diff
@@ -0,0 +1,79 @@
...
+### Requirements
+
+Distributions is a central package which many rely on,
+the following are required for contributions to be accepted:
+1. Docstrings must be added to all interface and non-trivial functions.
+2. Tests validating the modified behavior in the `test` folder. If new test files are added, do not forget to add them in `test/runtests.jl`. Cover possible edge cases. Run the tests locally before submitting the PR.
+3. At the end of the tests, `Test.detect_ambiguities(Distributions)` is run to check method ambiguities. Verify that your modified code did not yield method ambiguities.
+4. Make according modifications to the `docs` folder, build the documentation locally with `$ julia docs/make.jl`, verify that your modifications display correctly and did not yield warnings.
+
+## Branches
```
matbesancon: This section can be removed since no release branch

matbesancon:
@taqtiqa-mark not sure the build is allowing nightly failure as-is, I relaunched in doubt

.travis.yml
```diff
@@ -1,16 +1,21 @@
 language: julia
+
+if: branch=master
```
ararslan: This doesn't seem right, as it disables branch builds. Doesn't it?

ararslan:
Seems good to me overall except for the Travis changes. Ideally it should resemble that of other packages as closely as possible.

matbesancon:
@ararslan I'm good to merge whenever you are

ararslan:
I'd prefer if we didn't do the whole build stage thing. Ideally Travis setup files can be copied and pasted between repos with little to no editing required, and the vast majority of packages I've seen do not use stages this way, only for documentation.

matbesancon:
I would be fine with this one, the resulting Travis page reads well. It's
not obvious why we should remain compliant with other packages, probably
some others will switch to build stages, now that it's done for
documentation.

If we want to make this package a bit more standard, I think tests would be
a more pressing starting point (switching from JSON/text files to Julia
data structures.

mschauer:
What is the source/inspiration for the style guides?

taqtiqa-mark:
>Add Contributing page to docs site.
>
>  -  Modelled on Julia and Documenter.jl page.
>  - Included excerpt from README.md.

mschauer:
Ah, I'll have a look. As a side remark, this should have been perhaps three independent PRs.

mschauer:
@taqtiqa-mark Would you split this up into a travis update and separately a contrib page? Then we can go ahead with the travis update.

matbesancon:
ping @taqtiqa-mark for the requested changes

PR #900, opened 2019-05-16
change deploy kw

matbesancon:


matbesancon:
Documentation seems not to be displaying anything after v0.14, updating deploy doc to fix it

PR #910, opened 2019-06-05
AliasTable: fix unused rng argument

rfourquet:


rfourquet:
Oups sorry, I did `dev Distributions` in Pkg and assumed I would get the last version, but no, this is obsolete!

PR #915, opened 2019-06-16
ak/abstract_types

adam-r-kowalski:
In order to use this with Flux Tracked Arrays I needed to loosen the restrictions on the types to be more abstract. Additionally the eltype was previously hard coded to be Float64 and it should depend on the type parameter of the distribution.


matbesancon:
Thanks for the contribution @adam-r-kowalski. Have you looked at #882 which should be merged soon?

matbesancon:
closing this for the moment

PR #916, opened 2019-06-18
RFC: Allow `DiscreteNonParametric` to have non-`Real` support

DilumAluthge:
## Description

Let `vs` denote the list of support values used when creating an instance of `DiscreteNonParametric`. Currently, we require that:
- `vs::Ts` where `Ts<:AbstractVector{T}` and `T<:Real`.

This pull request relaxes the requirement so that it is now:
- `vs::Ts` where `Ts<:AbstractVector{T}`

This allows you to create instances of `DiscreteNonParametric` that have non-`Real` support.

All of the existing `DiscreteNonParametric` tests are preserved, to show that this change is backwards compatible and will have no effect on DiscreteNonParametric when `T<:Real`.

I also add some tests for `DiscreteNonParametric` when `T` is `String`.

## Motivation

Suppose you have trained a multiclass classifier, and you would now like run the classifier on a single sample. For each of the classes, the classifier will output the probability of that class. For example, the output might be `setosa=0.1, versicolor=0.7, verginica=0.2`.

Since the output of a multiclass classifier (when run on a single sample) is a discrete non-parametric distribution on the set of classes, it would be natural to store it in a `DiscreteNonParametric`. However, currently `DiscreteNonParametric` requires that the support be a list of real values. This pull request relaxes the requirement and allows you to create a `DiscreteNonParametric` where the support values can be of any type (such as `String`).

### DIFF
--- a/src/samplers/discretenonparametric.jl
+++ b/src/samplers/discretenonparametric.jl
@@ -4,19 +4,19 @@
 Data structure for efficiently sampling from an arbitrary probability mass
 function defined by support `xs` and probabilities `ps`.
 """
-struct DiscreteNonParametricSampler{T<:Real, S<:AbstractVector{T}, A<:AliasTable} <: Sampleable{Univariate,Discrete}
+struct DiscreteNonParametricSampler{T, S<:AbstractVector{T}, A<:AliasTable} <: Sampleable{Univariate,Discrete}
     support::S
     aliastable::A
 
     function DiscreteNonParametricSampler{T,S}(support::S, probs::AbstractVector{<:Real}
-    ) where {T<:Real,S<:AbstractVector{T}}
+    ) where {T,S<:AbstractVector{T}}
         aliastable = AliasTable(probs)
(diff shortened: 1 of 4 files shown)


DilumAluthge:
Bump. Any thoughts on this?

richardreeve:
Not that I have any say, but this looks like a good idea to me. However, am I right in thinking that `DiscreteNonParametric` currently rounds to integers? If so, then that seems like a slightly crazy restriction in its own right, and your code doesn't fix it... e.g.

```julia
```

DilumAluthge:
I wanted this pull request to be backwards-compatible, which is why I didn't make any changes to the behavior of `DiscreteNonParametric` when the support is `Real`.

I imagine that removing those calls to `round`, `floor`, etc. would probably be a breaking change.

richardreeve:
True, but only to people doing some seriously crazy stuff!

DilumAluthge:
Closing in favor of https://github.com/JuliaStats/Distributions.jl/pull/941

Issue #307, opened 2014-11-11
linear transformations on MvNormal


goretkin:
Are these functions within the scope of the package? These seem harmless

```
+(x::AbstractVector,n::MvNormal) = MvNormal(n.μ+x,n.Σ)
+(n::MvNormal,x::AbstractVector) = MvNormal(n.μ+x,n.Σ)
Xt_A_X
*(A::AbstractMatrix,n::MvNormal) = MvNormal(A*n.μ,  PDMats.Xt_A_X(n.Σ,A)  )
```

More contentious is allowing `+(x::MvNormal,y::MvNormal)`, assuming independence.

My main motivation was from seeing the [StateSpace](https://github.com/ElOceanografo/StateSpace.jl) package that @ElOceanografo wrote. For example: https://github.com/ElOceanografo/StateSpace.jl/blob/master/src/KalmanFilter.jl#L49 . Having functions for linear transformations of Multivariate Normal distributions might improve readability in only a minor way, but they might be worth including in Distributions anyway.

mschauer:
I am running into the same problem trying to port a Kalman filter now from 
https://github.com/mschauer/GaussianDistributions.jl/blob/master/src/GaussianDistributions.jl
to Distributions. How do you see this?

simonbyrne:
It would be good to have some way to do this, at least for constants. Mathematically, I'm a bit reluctant to overload `+`/`*` directly: the objects are intended to be distributions, not random variables. But maybe that isn't such a big deal?

Alternatively we could overload `map` (e.g. so that you would do `map(x -> x + k, d)`), but that would require much more "plumbing" to get it to work.

goretkin:
I hadn't considered making the distinction between an object that
represents a distribution and an object that represents a random variable
following that distribution, but I agree that the distinction certainly can
exist. Can we think of a case where this distinction serves a purpose?

simonbyrne:
I mean, the only other interpretation it could mean would be transforming it as a measure, but (a) that isn't very useful (since it would no longer be a probability measure), and (b) we don't treat it as a measure in other contexts, e.g. defining `(d::Distribution)(x::Interval)` to get the probability of an interval.

> independent sum deserves a new symbol like ⊕

That's a great idea!

matbesancon:
@JuliaRegistrator register()

PR #917, opened 2019-06-20
Add `MatrixNormal` distribution

johnczito:
Implements the basic features here:

https://en.wikipedia.org/wiki/Matrix_normal_distribution


src/Distributions.jl
```diff
@@ -239,8 +241,10 @@ export
 
     # reexport from StatsBase
     sample, sample!,        # sample from a source array
-    wsample, wsample!      # weighted sampling from a source array
+    wsample, wsample!,      # weighted sampling from a source array
 
+    # reexport from LinearAlgebra
+    rank
```
matbesancon: do we need to re-export stdlib functions?

src/matrix/matrixnormal.jl
```diff
@@ -0,0 +1,158 @@
+"""
+
+Matrix Normal Distribution
+
+X ~ MN(M, U, V)
```
matbesancon:
```suggestion
`X ~ MN(M, U, V)`
```

src/matrix/matrixnormal.jl
```diff
@@ -0,0 +1,158 @@
+"""
+
```
matbesancon:
```suggestion
    MatrixNormal{T <: Real, ST <: AbstractPDMat} <: ContinuousMatrixDistribution
    
```

src/Distributions.jl
johnczito: Do you consider this different from re-exporting `mean` and so on from `Statistics`?

matbesancon:
No, but this shouldn't be needed IMO, could be good to check

johnczito:
@matbesancon Unless I'm missing something, `rank` is only accessible if you load `LinearAlgebra`. So without re-exporting, if you just load `Distributions`, `rank(d::MatrixNormal)` will throw an `UndefVarError` (similarly with `mean` and so on).

matbesancon:
Yes but `Distributions` itself needs to import it, not the top-level user

matbesancon:
You don't need to re-export all the external functions you are using

matbesancon:
Yes, for `mean` you kind of expect them available when dealing with distributions. `rank` is not that obvious, (re-)exporting too many things makes the package hard to reason about

src/matrix/matrixnormal.jl
```diff
@@ -0,0 +1,159 @@
...
+V: p x p positive definite
+
+f(X) = c0 * exp( -0.5 tr[inv(V) (X - M)' inv(U) (X - M)] )
+
+c0   = (2pi) ^ {-np / 2} |V| ^ {-n / 2} |U| ^ {-p / 2}
+
+https://en.wikipedia.org/wiki/Matrix_normal_distribution
+
+"""
+struct MatrixNormal{T <: Real, ST <: AbstractPDMat} <: ContinuousMatrixDistribution
```
willtebbutt: Do the types of `U` and `V` really have to be exactly the same subtype of `AbstractPDMat`?

src/matrix/matrixnormal.jl
```diff
@@ -0,0 +1,159 @@
...
+
+    M::Matrix{T}
```
willtebbutt: Is there a reason we can't have `M` being a subtype of `AbstractMatrix`, rather than `Matrix`?

matbesancon: agreed, we are trying to make Distributions as generic as possible when there are no hard constraints

johnczito:
@matbesancon Fair enough. I took it out.

src/matrix/matrixnormal.jl
johnczito: Changed.

src/matrix/matrixnormal.jl
```diff
@@ -0,0 +1,159 @@
...
+f(X) = c0 * exp( -0.5 tr[inv(V) (X - M)' inv(U) (X - M)] )
+
+c0   = (2pi) ^ {-np / 2} |V| ^ {-n / 2} |U| ^ {-p / 2}
+
+https://en.wikipedia.org/wiki/Matrix_normal_distribution
+
+"""
+struct MatrixNormal{T <: Real, ST <: AbstractPDMat} <: ContinuousMatrixDistribution
+
+    M::AbstractMatrix{T}
```
willtebbutt: This should probably now be something like
```julia
struct MatrixNormal{T <: Real, TM <: AbstractMatrix, ST <: AbstractPDMat} <: ContinuousMatrixDistribution
    M::TM
    U::ST
    V::ST
    c0::T
end
```
Having an abstract type as a parameter here will cause type stability issues.

src/matrix/matrixnormal.jl
johnczito: No, but in every instance where they're used, they're used together, and you will get promotion to the common element type anyway:
```julia
#  logpdf
-0.5 * tr( (d.V \ At) * (d.U \ A) ) 

#  rand
A .= d.M + d.U.chol.L * X * d.V.chol.U
```

src/matrix/matrixnormal.jl
johnczito: Changed.

src/matrix/matrixnormal.jl
```diff
@@ -0,0 +1,161 @@
...
+
+    A .= d.M + d.U.chol.L * X * d.V.chol.U
+
+end
+
+#  -----------------------------------------------------------------------------
+#  Transformation
+#  -----------------------------------------------------------------------------
+
+vec(d::MatrixNormal) = MvNormal(vec(d.M), kron(d.V.mat, d.U.mat))
```
nickrobinson251: Does this use of `vec` have any precedent? 
I wonder if this should instead just be a `MvNormal` constructor?
i.e. ` MvNormal(d::MatrixNormal) = MvNormal(vec(d.M), kron(d.V.mat, d.U.mat))`

test/matrixnormal.jl
```diff
@@ -0,0 +1,128 @@
...
+#  Call every constructor to catch errors
+
+MatrixNormal(M, U,        PDV.chol)
+MatrixNormal(M, U,        PDV)
+MatrixNormal(M, PDU.chol, V)
+MatrixNormal(M, PDU.chol, PDV.chol)
+MatrixNormal(M, PDU.chol, PDV)
+MatrixNormal(M, PDU,      V)
+MatrixNormal(M, PDU,      PDV.chol)
+MatrixNormal(M, PDU,      PDV)
```
nickrobinson251: Should these be something like `@test MatrixNormal(M, U, V) isa MatrixNormal` just for the sake of these being test failures rather than errors outside of a testset?

Also should this whole file be wrapped in a `@testset "MatrixNormal begin ... end`?

johnczito: Great, thanks. I will do this shortly.

src/matrix/matrixnormal.jl
johnczito: Right. Given #307, I don't think there's agreement on this in general.

nickrobinson251: I will put a vote in for this being a `MvNormal` constructor (I also see no reason not to have both `vec` and `MvNormal` do this) -- but i will leave it to you / maintainers to decide :)

johnczito:
@matbesancon @willtebbutt Let me know what else you want to see for this to be considered for merging.

nickrobinson251:
Should add to the docs :) i think just list MatrixNormal under `Distributions` here https://github.com/JuliaStats/Distributions.jl/blob/master/docs/src/matrix.md

src/matrix/matrixnormal.jl
```diff
@@ -0,0 +1,161 @@
...
+
+U: n x n positive definite
+
+V: p x p positive definite
+
+f(X) = c0 * exp( -0.5 tr[inv(V) (X - M)' inv(U) (X - M)] )
+
+c0   = (2pi) ^ {-np / 2} |V| ^ {-n / 2} |U| ^ {-p / 2}
+
+https://en.wikipedia.org/wiki/Matrix_normal_distribution
```
nickrobinson251: This doesn't render right in the REPL - the underscores `_` are treated by Markdown as italics, so we get "Matrix _normal_ distribution" as one word and the link doesn't work. 

Better to put this as a link in the text? i.e. On the first line 
```julia
"""
    MatrixNormal{...}

[Matrix Normal Distribution](https://en.wikipedia.org/wiki/Matrix_normal_distribution)
...
"""

nickrobinson251:
This looks good to me 👍

src/matrix/matrixnormal.jl
johnczito: Right, I just noticed that. Thanks!

willtebbutt:
LGTM :)

matbesancon:
Travis failed, I just restarted

johnczito:
Thanks. It failed again, but all of the tests passed. It was a problem with building the docs. I was able to build locally. Let me know what I need to change to resolve this.

src/matrix/matrixnormal.jl
matbesancon: I think I am ok with vec, but you need to document this behaviour and add it to the doc (you can add specific methods to the doc, it's already done

matbesancon:
Looks good, merging if all green

johnczito:
Looks green.

Thanks for everyone's help!

matbesancon:
thanks, this is a cool addition!

Issue #688, opened 2017-11-21
MvNormal Matrix Types and Speed


acheck10:
I'm not sure if this qualifies as an "issue", and I apologize if there is somewhere else I should be posting this. 

I have code that runs Bayesian regression with Normal/Gamma independent priors, so I need to draw from an MvNormal distribution tens (or hundreds) of thousands of times. Back in Julia 0.4, I could form an estimate of the variance matrix and plug it right into the MvNormal function. Since Julia 0.5, when the Cholesky factorization got more rigid, I have needed to declare the variance matrix as Hermitian, or I get an error that my matrix is not positive definite (due to computer rounding error). However, MvNormal does not take Hermitian matrices, so I also need to convert to an Array{Float64,2} before using the variance matrix in the MvNormal distribution. This caused a big slowdown in Julia 0.5 (not quite as bad in 0.6). I'm wondering if there could be a more elegant (and possibly faster) way to do what I'm doing, or if the Distributions package could be updated to allow Hermitian matrices as inputs to the MvNormal function.

simonbyrne:
I personally think we should get rid of `PDMat`s altogether, and just use `Cholesky` (or `Eigen`, or `LDLt`, etc.), as there is no reason you need the full matrix at all.

nickrobinson251:
xref #940

PR #919, opened 2019-06-25
functions to convolve simple distributions

glennmoy:
These are simple functions to convolve distributions where a closed form exists. 

Edit: taken essentially from this list: https://en.wikipedia.org/wiki/List_of_convolutions_of_probability_distributions

Please advise if this is the correct place to put them or if somewhere else would be more suitable.

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -18,7 +18,7 @@ import StatsBase: kurtosis, skewness, entropy, mode, modes,
                   fit, kldivergence, loglikelihood, dof, span,
                   params, params!
 
-import PDMats: dim, PDMat, invquad
+import PDMats: dim, PDMat, invquad, ScalMat
 
 using SpecialFunctions
 
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -18,7 +18,7 @@ import StatsBase: kurtosis, skewness, entropy, mode, modes,
                   fit, kldivergence, loglikelihood, dof, span,
                   params, params!
 
-import PDMats: dim, PDMat, invquad, ScalMat
+import PDMats: dim, PDMat, invquad
 
 using SpecialFunctions
 
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -168,6 +168,7 @@ export
     componentwise_pdf,      # component-wise pdf for mixture models
     componentwise_logpdf,   # component-wise logpdf for mixture models
     concentration,      # the concentration parameter
+    convolve,           # convolve distributions of the same type
     dim,                # sample dimension of multivariate distribution
     dof,                # get the degree of freedom
     entropy,            # entropy of distribution in nats
@@ -267,6 +268,7 @@ include("samplers.jl")
 # others
 include("truncate.jl")
 include("conversion.jl")
+include("convolution.jl")
 include("qq.jl")
 include("estimators.jl")
 
--- a/src/convolution.jl
+++ b/src/convolution.jl
@@ -88,7 +88,9 @@ end
 
 
 function _check_convolution_args(p1, p2)
-    p1 == p2 || throw(ArgumentError("$(p1) != $(p2): distribution parameters must be equal"))
+    p1 ≈ p2 || throw(ArgumentError(
+    "$(p1) !≈ $(p2): distribution parameters must be approximately equal",
+    ))
 end
 
 function _check_convolution_shape(d1, d2)
--- a/src/convolution.jl
+++ b/src/convolution.jl
@@ -1,0 +1,88 @@
+"""
+    convolve(d1::T, d2::T) where T<:Distribution -> Distribution
+
+Convole two distributions of the same type where the convolution has a closed form, e.g.
+
+* Bernoulli
+* Binomial
+* NegativeBinomial
+* Geometric
+* Poisson
+* Normal
+* Cauchy
+* Chisq
+* Exponential
+* Gamma
+* Multivariate Normal
+"""
+convolve(d1, d2) = @throws(MethodError("`convolve` is not defined for $d1 and $d2"))
+
+# discrete univariate
+function convolve(d1::Bernoulli, d2::Bernoulli)
+    _check_params(d1.p, d2.p)
+    return Binomial(2, d1.p)
+end
+
+function convolve(d1::Binomial, d2::Binomial)
+    _check_params(d1.p, d2.p)
+    return Binomial(d1.n + d2.n, d1.p)
+end
+
+function convolve(d1::NegativeBinomial, d2::NegativeBinomial)
+    _check_params(d1.p, d2.p)
+    return NegativeBinomial(d1.n + d2.n, d1.p)
+end
+
+function convolve(d1::Geometric, d2::Geometric)
+    _check_params(d1.p, d2.p)
+    return NegativeBinomial(2, d1.p)
+end
+
+function convolve(d1::Poisson, d2::Poisson)
+    return Poisson(d1.λ + d2.λ)
+end
+
+
+# continuous univariate
+convolve(d1::Normal, d2::Normal) = Normal(d1.μ + d2.μ, √(d1.σ^2 + d2.σ^2))
+convolve(d1::Cauchy{T}, d2::Cauchy{T}) where T = Cauchy(d1.μ + d2.μ, d1.σ + d2.σ)
+convolve(d1::Chisq, d2::Chisq) = Chisq(d1.ν + d2.ν)
+
+function convolve(d1::Exponential, d2::Exponential)
+    _check_params(d1.θ, d2.θ)
+    return Gamma(2, d1.θ)
+end
+
+function convolve(d1::Gamma, d2::Gamma)
+    _check_params(d1.θ, d2.θ)
+    return Gamma(d1.α + d2.α, d1.θ)
+end
+
+# continuous multivariate
+convolve(d1::T, d2::T) where T<:Union{IsoNormal, ZeroMeanIsoNormal}
+    _check_params(length(d1), length(d2))
+    return MvNormal(d1.μ + d2.μ, d1.Σ.value + d2.Σ.value)
+end
+
+convolve(d1::T, d2::T) where T<:Union{DiagNormal, ZeroMeanDiagNormal}
+    _check_params(length(d1), length(d2))
+    return MvNormal(d1.μ + d2.μ, d1.Σ.diag + d2.Σ.diag)
+end
+
+convolve(d1::T, d2::T) where T<:Union{FullNormal, ZeroMeanFullNormal}
+    _check_params(length(d1), length(d2))
+    return MvNormal(d1.μ + d2.μ, d1.Σ.mat + d2.Σ.mat)
+end
+
+convolve(d1::MvNormal, d2::MvNormal)
+    _check_params(length(d1), length(d2))
+    return MvNormal(d1.μ + d2.μ, Matrix(d1.Σ) + Matrix(d2.Σ.mat))
+end
+
+
+
+function _check_params(p1, p2)
+    if p1 != p2
+        throw(ArgumentError("$(p1)!=$(p2): distribution parameters must be equal"))
+    end
+end
--- a/src/convolution.jl
+++ b/src/convolution.jl
@@ -7,19 +7,18 @@ sum of independent random variables drawn from the underlying distributions.
 The function is only defined in the cases where the convolution has a closed form as
 defined here https://en.wikipedia.org/wiki/List_of_convolutions_of_probability_distributions
 
-* Bernoulli
-* Binomial
-* Negative Binomial
-* Geometric
-* Poisson
-* Normal
-* Cauchy
-* Chisq
-* Exponential
-* Gamma
-* Multi-variate Normal
+* `Bernoulli`
+* `Binomial`
+* `NegativeBinomial`
+* `Geometric`
+* `Poisson`
+* `Normal`
+* `Cauchy`
+* `Chisq`
+* `Exponential`
+* `Gamma`
+* `MultivariateNormal`
 """
-#convolve(d1, d2) = throw(ArgumentError("`convolve` is not defined for $(typeof(d1)) and $(typeof(d2))"))
 
 # discrete univariate
 function convolve(d1::Bernoulli, d2::Bernoulli)
@@ -47,7 +46,7 @@ convolve(d1::Poisson, d2::Poisson) =  Poisson(d1.λ + d2.λ)
 
 # continuous univariate
 convolve(d1::Normal, d2::Normal) = Normal(d1.μ + d2.μ, hypot(d1.σ, d2.σ))
-convolve(d1::Cauchy{T}, d2::Cauchy{T}) where T = Cauchy(d1.μ + d2.μ, d1.σ + d2.σ)
+convolve(d1::Cauchy, d2::Cauchy) = Cauchy(d1.μ + d2.μ, d1.σ + d2.σ)
 convolve(d1::Chisq, d2::Chisq) = Chisq(d1.ν + d2.ν)
 
 function convolve(d1::Exponential, d2::Exponential)
@@ -62,19 +61,11 @@ end
 
 # continuous multivariate
 function convolve(
-    d1::Union{IsoNormal, ZeroMeanIsoNormal},
-    d2::Union{IsoNormal, ZeroMeanIsoNormal},
+    d1::Union{IsoNormal, ZeroMeanIsoNormal, DiagNormal, ZeroMeanDiagNormal},
+    d2::Union{IsoNormal, ZeroMeanIsoNormal, DiagNormal, ZeroMeanDiagNormal},
     )
     _check_convolution_shape(d1, d2)
-    return MvNormal(d1.μ .+ d2.μ, d1.Σ.value + d2.Σ.value)
-end
-
-function convolve(
-    d1::Union{DiagNormal, ZeroMeanDiagNormal},
-    d2::Union{DiagNormal, ZeroMeanDiagNormal},
-    )
-    _check_convolution_shape(d1, d2)
-    return MvNormal(d1.μ .+ d2.μ, d1.Σ.diag + d2.Σ.diag)
+    return MvNormal(d1.μ .+ d2.μ, d1.Σ + d2.Σ)
 end
 
 function convolve(
@@ -82,17 +73,17 @@ function convolve(
     d2::Union{FullNormal, ZeroMeanFullNormal},
     )
     _check_convolution_shape(d1, d2)
-    return MvNormal(d1.μ .+ d2.μ, d1.Σ.mat + d2.Σ.mat)
+    return MvNormal(d1.μ .+ d2.μ, d1.Σ.mat + d2.Σ.mat + 1e-12 * I)  # ensures positive definite
 end
 
 function convolve(d1::MvNormal, d2::MvNormal)
     _check_convolution_shape(d1, d2)
-    return MvNormal(d1.μ .+ d2.μ, Matrix(d1.Σ) + Matrix(d2.Σ))
+    return MvNormal(d1.μ .+ d2.μ, Matrix(d1.Σ) + Matrix(d2.Σ) + 1e-12 * I)
 end
 
 
 function _check_convolution_args(p1, p2)
-    p1 == p2 || throw(ArgumentError("$(p1)!=$(p2): distribution parameters must be equal"))
+    p1 == p2 || throw(ArgumentError("$(p1) != $(p2): distribution parameters must be equal"))
 end
 
 function _check_convolution_shape(d1, d2)
--- a/src/convolution.jl
+++ b/src/convolution.jl
@@ -60,6 +60,8 @@ function convolve(d1::Gamma, d2::Gamma)
 end
 
 # continuous multivariate
+# The first two methods exist for performance reasons to avoid unnecessarily converting
+# PDMats to a Matrix
 function convolve(
     d1::Union{IsoNormal, ZeroMeanIsoNormal, DiagNormal, ZeroMeanDiagNormal},
     d2::Union{IsoNormal, ZeroMeanIsoNormal, DiagNormal, ZeroMeanDiagNormal},
@@ -73,7 +75,9 @@ function convolve(
     d2::Union{FullNormal, ZeroMeanFullNormal},
     )
     _check_convolution_shape(d1, d2)
-    return MvNormal(d1.μ .+ d2.μ, d1.Σ.mat + d2.Σ.mat + 1e-12 * I)  # ensures positive definite
+    # + ϵI protects against numerical instability in the cholesky method and ensures
+    # positive definite-ness
+    return MvNormal(d1.μ .+ d2.μ, d1.Σ.mat + d2.Σ.mat + 1e-12 * I)
 end
 
 function convolve(d1::MvNormal, d2::MvNormal)
--- a/src/convolution.jl
+++ b/src/convolution.jl
@@ -19,6 +19,7 @@ defined here https://en.wikipedia.org/wiki/List_of_convolutions_of_probability_d
 * `Gamma`
 * `MultivariateNormal`
 """
+function convolve end
 
 # discrete univariate
 function convolve(d1::Bernoulli, d2::Bernoulli)
@@ -77,12 +78,12 @@ function convolve(
     _check_convolution_shape(d1, d2)
     # + ϵI protects against numerical instability in the cholesky method and ensures
     # positive definite-ness
-    return MvNormal(d1.μ .+ d2.μ, d1.Σ.mat + d2.Σ.mat + 1e-12 * I)
+    return MvNormal(d1.μ .+ d2.μ, d1.Σ.mat + d2.Σ.mat)
 end
 
 function convolve(d1::MvNormal, d2::MvNormal)
     _check_convolution_shape(d1, d2)
-    return MvNormal(d1.μ .+ d2.μ, Matrix(d1.Σ) + Matrix(d2.Σ) + 1e-12 * I)
+    return MvNormal(d1.μ .+ d2.μ, Matrix(d1.Σ) + Matrix(d2.Σ))
 end
 
 
--- a/src/convolution.jl
+++ b/src/convolution.jl
@@ -76,8 +76,6 @@ function convolve(
     d2::Union{FullNormal, ZeroMeanFullNormal},
     )
     _check_convolution_shape(d1, d2)
-    # + ϵI protects against numerical instability in the cholesky method and ensures
-    # positive definite-ness
     return MvNormal(d1.μ .+ d2.μ, d1.Σ.mat + d2.Σ.mat)
 end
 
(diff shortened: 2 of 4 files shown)


src/convolution.jl
```diff
@@ -0,0 +1,88 @@
...
+end
+
+function convolve(d1::Gamma, d2::Gamma)
+    _check_params(d1.θ, d2.θ)
+    return Gamma(d1.α + d2.α, d1.θ)
+end
+
+# continuous multivariate
+convolve(d1::T, d2::T) where T<:Union{IsoNormal, ZeroMeanIsoNormal}
+    _check_params(length(d1), length(d2))
```
mschauer: Specialize `_check_params` to `_check_convolution_shape` or make a macro, the error message suggest that `length` is a `parameter`

src/convolution.jl
```diff
@@ -0,0 +1,88 @@
...
+end
+
+convolve(d1::MvNormal, d2::MvNormal)
+    _check_params(length(d1), length(d2))
+    return MvNormal(d1.μ + d2.μ, Matrix(d1.Σ) + Matrix(d2.Σ.mat))
+end
+
+
+
+function _check_params(p1, p2)
```
mschauer: The name is a bit too generic imo, maybe `_convolution_argcheck`

src/convolution.jl
```diff
@@ -0,0 +1,88 @@
+"""
+    convolve(d1::T, d2::T) where T<:Distribution -> Distribution
+
+Convole two distributions of the same type where the convolution has a closed form, e.g.
```
mschauer: Convolve

mschauer: Mention independent sum of random variables.

src/convolution.jl
```diff
@@ -0,0 +1,88 @@
...
+* Binomial
+* NegativeBinomial
+* Geometric
+* Poisson
+* Normal
+* Cauchy
+* Chisq
+* Exponential
+* Gamma
+* Multivariate Normal
```
mschauer: Space?

src/convolution.jl
```diff
@@ -0,0 +1,88 @@
...
+"""
+convolve(d1, d2) = @throws(MethodError("`convolve` is not defined for $d1 and $d2"))
```
mschauer: Do we need to shadow the method error?

src/convolution.jl
```diff
@@ -0,0 +1,88 @@
...
+    return NegativeBinomial(2, d1.p)
+end
+
+function convolve(d1::Poisson, d2::Poisson)
+    return Poisson(d1.λ + d2.λ)
+end
+
+
+# continuous univariate
+convolve(d1::Normal, d2::Normal) = Normal(d1.μ + d2.μ, √(d1.σ^2 + d2.σ^2))
```
mschauer:
```suggestion
convolve(d1::Normal, d2::Normal) = Normal(d1.μ + d2.μ, hypot(d1.σ, d2.σ))
```

src/convolution.jl
glennmoy: How about `_check_convolution_args`?

src/convolution.jl
glennmoy: agreed. I've now made `_check_convolution_shape`.

src/convolution.jl
glennmoy: No on second thought probably not...

src/convolution.jl
glennmoy: nice!

src/convolution.jl
```diff
@@ -0,0 +1,88 @@
...
+    return Gamma(2, d1.θ)
+end
+
+function convolve(d1::Gamma, d2::Gamma)
+    _check_params(d1.θ, d2.θ)
+    return Gamma(d1.α + d2.α, d1.θ)
+end
+
+# continuous multivariate
+convolve(d1::T, d2::T) where T<:Union{IsoNormal, ZeroMeanIsoNormal}
```
ararslan: Missing `function` here? (And a few below)

glennmoy: yep, caught them! This was just a first "draft"

src/convolution.jl
```diff
@@ -0,0 +1,100 @@
...
+
+function convolve(
+    d1::Union{FullNormal, ZeroMeanFullNormal},
+    d2::Union{FullNormal, ZeroMeanFullNormal},
+    )
+    _check_convolution_shape(d1, d2)
+    return MvNormal(d1.μ .+ d2.μ, d1.Σ.mat + d2.Σ.mat)
+end
+
+function convolve(d1::MvNormal, d2::MvNormal)
```
glennmoy: I wonder if this function is sufficient for all MvNormal cases or are the above functions also necessary? Are there any expected performance benefits from dispatching on the sub-type?

glennmoy: I suspect calling `Matrix()` on some very large diagonal matrices could be an unnecessary bottleneck?

src/convolution.jl
```diff
@@ -0,0 +1,100 @@
...
+* Geometric
+* Poisson
+* Normal
+* Cauchy
+* Chisq
+* Exponential
+* Gamma
+* Multi-variate Normal
+"""
+#convolve(d1, d2) = throw(ArgumentError("`convolve` is not defined for $(typeof(d1)) and $(typeof(d2))"))
```
glennmoy: need to remove this

glennmoy:
```suggestion
```

glennmoy: forgot to remove this

test/convolution.jl
```diff
@@ -0,0 +1,226 @@
...
+
+        # erroring
+        dn3 =  MvNormal([1, 2, 3], [4.2, 5.3, 2.1])
+        @test_throws ArgumentError convolve(dn1, dn3)
+    end
+
+    @testset "full-normal" begin
+
+        L1 = [1 0; 2 1]
+        fn1 = MvNormal(ones(2), cov(L1 * L1'))
```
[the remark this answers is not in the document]
mschauer: Its fine like this

src/convolution.jl
nickrobinson251: Yes, I think the specialised methods you added have performance benefits, because they can avoid the `Matrix` call

but also I think they reveal that the `MvNormal` types are a bit wierd (https://github.com/JuliaStats/Distributions.jl/issues/921)

src/convolution.jl
```diff
@@ -0,0 +1,100 @@
...
+    _check_convolution_shape(d1, d2)
+    return MvNormal(d1.μ .+ d2.μ, d1.Σ.value + d2.Σ.value)
+end
+
+function convolve(
+    d1::Union{DiagNormal, ZeroMeanDiagNormal},
+    d2::Union{DiagNormal, ZeroMeanDiagNormal},
+    )
+    _check_convolution_shape(d1, d2)
+    return MvNormal(d1.μ .+ d2.μ, d1.Σ.diag + d2.Σ.diag)
```
nickrobinson251: this seem better without the `.diag`

```julia
julia> z1 = MvNormal(repeat([1.0], 100))
ZeroMeanDiagNormal(
dim: 100
μ: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0  …  0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
Σ: [1.0 0.0 … 0.0 0.0; 0.0 1.0 … 0.0 0.0; … ; 0.0 0.0 … 1.0 0.0; 0.0 0.0 … 0.0 1.0]
)


julia> z2 = MvNormal(ones(100), repeat([1.0], 100))
DiagNormal(
dim: 100
μ: [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0  …  1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]
Σ: [1.0 0.0 … 0.0 0.0; 0.0 1.0 … 0.0 0.0; … ; 0.0 0.0 … 1.0 0.0; 0.0 0.0 … 0.0 1.0]
)


julia> @btime MvNormal($z1.μ .+ $z2.μ, $z1.Σ + $z2.Σ);
  277.636 ns (5 allocations: 2.69 KiB)

julia> @btime MvNormal($z1.μ .+ $z2.μ, $z1.Σ.diag .+ $z2.Σ.diag);
  353.230 ns (6 allocations: 3.56 KiB)
```

src/convolution.jl
```diff
@@ -0,0 +1,100 @@
...
+    return Gamma(d1.α + d2.α, d1.θ)
+end
+
+# continuous multivariate
+function convolve(
+    d1::Union{IsoNormal, ZeroMeanIsoNormal},
+    d2::Union{IsoNormal, ZeroMeanIsoNormal},
+    )
+    _check_convolution_shape(d1, d2)
+    return MvNormal(d1.μ .+ d2.μ, d1.Σ.value + d2.Σ.value)
```
nickrobinson251: we don't need the `.value` here

```julia
julia> z1 = MvNormal(Distributions.ZeroVector(Float64, 100), I)
ZeroMeanIsoNormal(
dim: 100
μ: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0  …  0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
Σ: [1.0 0.0 … 0.0 0.0; 0.0 1.0 … 0.0 0.0; … ; 0.0 0.0 … 1.0 0.0; 0.0 0.0 … 0.0 1.0]
)


julia> z2 = MvNormal(ones(100), I)
IsoNormal(
dim: 100
μ: [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0  …  1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]
Σ: [1.0 0.0 … 0.0 0.0; 0.0 1.0 … 0.0 0.0; … ; 0.0 0.0 … 1.0 0.0; 0.0 0.0 … 0.0 1.0]
)


julia> @btime MvNormal($z1.μ .+ $z2.μ, $z1.Σ + $z2.Σ);
  87.319 ns (2 allocations: 944 bytes)

julia> @btime MvNormal($z1.μ .+ $z2.μ, $z1.Σ.value + $z2.Σ.value);
  87.682 ns (2 allocations: 944 bytes)
```

nickrobinson251: actually, the `*DiagNormal` and `*IsoNormal` can be combined
```julia
julia> z1 = MvNormal(repeat([1.0], 100))
ZeroMeanDiagNormal(
dim: 100
μ: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0  …  0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
Σ: [1.0 0.0 … 0.0 0.0; 0.0 1.0 … 0.0 0.0; … ; 0.0 0.0 … 1.0 0.0; 0.0 0.0 … 0.0 1.0]
)


julia> z2 = MvNormal(ones(100), repeat([1.0], 100))
DiagNormal(
dim: 100
μ: [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0  …  1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]
Σ: [1.0 0.0 … 0.0 0.0; 0.0 1.0 … 0.0 0.0; … ; 0.0 0.0 … 1.0 0.0; 0.0 0.0 … 0.0 1.0]
)


julia> z3 = MvNormal(Distributions.ZeroVector(Float64, 100), I)
ZeroMeanIsoNormal(
dim: 100
μ: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0  …  0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
Σ: [1.0 0.0 … 0.0 0.0; 0.0 1.0 … 0.0 0.0; … ; 0.0 0.0 … 1.0 0.0; 0.0 0.0 … 0.0 1.0]
)


julia> z4 = MvNormal(ones(100), I)
IsoNormal(
dim: 100
μ: [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0  …  1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]
Σ: [1.0 0.0 … 0.0 0.0; 0.0 1.0 … 0.0 0.0; … ; 0.0 0.0 … 1.0 0.0; 0.0 0.0 … 0.0 1.0]
)


julia> for (a, b) in Iterators.product((z1, z2), (z3, z4))
           @btime MvNormal($a.μ .+ $b.μ, $a.Σ + $b.Σ)
       end
  271.267 ns (5 allocations: 2.69 KiB)
  264.330 ns (5 allocations: 2.69 KiB)
  260.786 ns (5 allocations: 2.69 KiB)
  263.312 ns (5 allocations: 2.69 KiB)
```

nickrobinson251: (which makes since, since they both have diagonal cov matrices)

src/convolution.jl
nickrobinson251: I see the `Matrix` here is for handling the diagonal + matrix case 👍 

I'll note that the code here does not handle `Int`s: this is the only method that anything with `Int`s will hit (becuase #921)... without the `Matrix` call we'd get an `Inexact` error
```julia
julia> z = MvNormal([1, 1])
MvNormal{Int64,PDiagMat{Int64,Array{Int64,1}},Distributions.ZeroVector{Int64}}(
dim: 2
μ: [0, 0]
Σ: [1 0; 0 1]
)

julia> MvNormal(z.μ .+ z.μ, z.Σ + z.Σ)
ERROR: InexactError: Int64(0.5)
...
```

The `Matrix` call saves us from that... But we hit a problem in PDMats (I guess i'll open an issue there -- but another vote for ditching PDMats https://github.com/JuliaStats/Distributions.jl/issues/688#issuecomment-346145780)

```julia
julia> MvNormal(z.μ .+ z.μ, Matrix(z.Σ) + Matrix(z.Σ))
ERROR: MethodError: no method matching PDMat{Int64,Array{Int64,2}}(::Int64, ::Array{Int64,2}, ::Cholesky{Float64,Array{Float64,2}})
Closest candidates are:
  PDMat{Int64,Array{Int64,2}}(::Int64, ::AbstractArray{T,2}, ::Cholesky{T,S}) where {T, S} at /Users/nick/.julia/packages/PDMats/AObTs/src/pdmat.jl:9
Stacktrace:
 [1] PDMat(::Array{Int64,2}, ::Cholesky{Float64,Array{Float64,2}}) at /Users/nick/.julia/packages/PDMats/AObTs/src/pdmat.jl:16
 [2] PDMat(::Array{Int64,2}) at /Users/nick/.julia/packages/PDMats/AObTs/src/pdmat.jl:19
 [3] MvNormal(::Array{Int64,1}, ::Array{Int64,2}) at /Users/nick/.julia/packages/Distributions/Iltex/src/multivariate/mvnormal.jl:203
 [4] top-level scope at none:0
```

src/convolution.jl
```diff
@@ -0,0 +1,100 @@
...
+end
+
+function convolve(d1::MvNormal, d2::MvNormal)
+    _check_convolution_shape(d1, d2)
+    return MvNormal(d1.μ .+ d2.μ, Matrix(d1.Σ) + Matrix(d2.Σ))
+end
+
+
+function _check_convolution_args(p1, p2)
+    p1 == p2 || throw(ArgumentError("$(p1)!=$(p2): distribution parameters must be equal"))
```
nickrobinson251:
```suggestion
    p1 == p2 || throw(ArgumentError("$(p1) != $(p2): distribution parameters must be equal"))
```

nickrobinson251:
BTW `\oplus` ⊕ has been suggested as an alias for the sum of idendepent ranom Variables which is (the generalisation) of what `convolve` does. (Presuming we're okay with conflating RVs with their distributions).

Should we add that here? 

maybe @mschauer has thoughts? 
https://github.com/JuliaStats/Distributions.jl/issues/307#issuecomment-429488430

src/convolution.jl
```diff
@@ -0,0 +1,100 @@
...
+    _check_convolution_args(d1.p, d2.p)
+    return NegativeBinomial(2, d1.p)
+end
+
+convolve(d1::Poisson, d2::Poisson) =  Poisson(d1.λ + d2.λ)
+
+
+# continuous univariate
+convolve(d1::Normal, d2::Normal) = Normal(d1.μ + d2.μ, hypot(d1.σ, d2.σ))
+convolve(d1::Cauchy{T}, d2::Cauchy{T}) where T = Cauchy(d1.μ + d2.μ, d1.σ + d2.σ)
```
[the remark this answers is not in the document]
glennmoy: Initially it didn't work when I tried it that way but now it does 🤷‍♂

src/convolution.jl
glennmoy: cool thanks!

src/convolution.jl
glennmoy: nice, that simplifies things a bit

src/convolution.jl
glennmoy: That is weird to restrict to just `Float64`...

Maybe to avoid the inexact error I could do 

```julia
function convolve(
    d1::Union{IsoNormal{T}, ZeroMeanIsoNormal{T}, DiagNormal{T}, ZeroMeanDiagNormal{T}},
    d2::Union{IsoNormal{T}, ZeroMeanIsoNormal{T}, DiagNormal{T}, ZeroMeanDiagNormal{T}},
    ) where T<:Float64
```

So that anything with `Int` will get sent to the other method?

glennmoy: It's very weird that they would just be defined over `Float64`...
It would be great if https://github.com/JuliaStats/Distributions.jl/issues/921 got resolved.

src/convolution.jl
nickrobinson251: i guess we can use a const to save us a bit here
```julia
const IsoOrDiagNormal = Union{IsoNormal, ZeroMeanIsoNormal, DiagNormal, ZeroMeanDiagNormal}`

function convolve(d1::IsoOrDiagNormal, d2::IsoOrDiagNormal)
    check_convolution_shape(d1, d2)
    return MvNormal(d1.μ .+ d2.μ, d1.Σ + d2.Σ)
end
```

src/convolution.jl
```diff
@@ -0,0 +1,100 @@
...
+    d1::Union{DiagNormal, ZeroMeanDiagNormal},
+    d2::Union{DiagNormal, ZeroMeanDiagNormal},
+    )
+    _check_convolution_shape(d1, d2)
+    return MvNormal(d1.μ .+ d2.μ, d1.Σ.diag + d2.Σ.diag)
+end
+
+function convolve(
+    d1::Union{FullNormal, ZeroMeanFullNormal},
+    d2::Union{FullNormal, ZeroMeanFullNormal},
```
nickrobinson251: I'd add a comment to this method and to the one for Iso/Diag MvNormals that they exist for performance reasons

glennmoy: I just pushed so I'll add this after the next review 👍

src/Distributions.jl
```diff
@@ -18,7 +18,7 @@ import StatsBase: kurtosis, skewness, entropy, mode, modes,
                   fit, kldivergence, loglikelihood, dof, span,
                   params, params!
 
-import PDMats: dim, PDMat, invquad
+import PDMats: dim, PDMat, invquad, ScalMat
```
glennmoy:
```suggestion
import PDMats: dim, PDMat, invquad
```
don't need ScalMat anymore

src/convolution.jl
```diff
@@ -0,0 +1,91 @@
...
+    _check_convolution_shape(d1, d2)
+    return MvNormal(d1.μ .+ d2.μ, d1.Σ + d2.Σ)
+end
+
+function convolve(
+    d1::Union{FullNormal, ZeroMeanFullNormal},
+    d2::Union{FullNormal, ZeroMeanFullNormal},
+    )
+    _check_convolution_shape(d1, d2)
+    return MvNormal(d1.μ .+ d2.μ, d1.Σ.mat + d2.Σ.mat + 1e-12 * I)  # ensures positive definite
```
glennmoy: else I get this error when convolving `fn1` with itself, which I'm guessing shouldn't happen...

```julia
 Expression: Matrix(d3.Σ) == Matrix(d1.Σ + d2.Σ)
  PosDefException: matrix is not positive definite; Cholesky factorization failed.
```

glennmoy: If there's a better way to avoid it I'd be interested to hear

nickrobinson251: can you post the matrix this happened with? I wonder if it is floating point round off? Mayeb need to wrap it in `Symmetric` or `Hermitian`

glennmoy:
```julia
julia> L1 = [1 0; 2 1]
2×2 Array{Int64,2}:
 1  0
 2  1

julia> fn1 = MvNormal(ones(2), cov(L1 * L1'))
FullNormal(
dim: 2
μ: [1.0, 1.0]
Σ: [0.5 1.5; 1.5 4.5]
)


julia> MvNormal(fn1.μ .+ fn1.μ, fn1.Σ + fn1.Σ)
ERROR: PosDefException: matrix is not positive definite; Cholesky factorization failed.
```

glennmoy: Looking into it a bit more the first eigenvalue is practically 0 so this might be a bad example of a matrix to build a test on. But maybe warrants its own test.

What I can't understand is why `fn1.Σ` on its own is positive definite but `2 * fn1.Σ` isn't...
```julia
julia> eigen(fn1.Σ.mat)
2-element Array{Float64,1}:
 5.551115123125783e-17
 5.0   

julia> eigen(fn1.Σ.mat + fn1.Σ.mat)
2-element Array{Float64,1}:
  1.1102230246251565e-16
 10.0    

julia> eigen(fn1.Σ.mat + fn1.Σ.mat + 1e-12*I)
2-element Array{Float64,1}:
  1.000088900582341e-12
 10.000000000001   
```

nickrobinson251: I think this is a weird floating point thing? adding/subtracting `1e-12*I` to `Σ` makes/stops `Σ` and `rΣ` being posdef (for r>0).

nickrobinson251: I'd not do the ` + 1e-12 * I` here, and leave users to handle these weird edge cases... unless we think they're common?

glennmoy: Well, albeit contrived, I've hit it on two distributions in my tests and few others from messing around locally. I wouldn't be against adding `+ eps(Float64) * I` (or maybe wrapping it in a try-catch with a warning to users) to ensure stability because the maths should allow self-convolution. But I'll defer to the opinion of others as I'm not an expert in this at all.

nickrobinson251: I think `try ... catch ... + + eps(Float64) * I` is not too awful -- let's see what others think

matbesancon: it would be better to avoid try-catching all over, it's slower. If something happens, better error right away, the other option is to have a keyword argument for a regularizing term to add, defaulting to `eps(T)`

src/convolution.jl
```diff
@@ -0,0 +1,95 @@
...
+* `Geometric`
+* `Poisson`
+* `Normal`
+* `Cauchy`
+* `Chisq`
+* `Exponential`
+* `Gamma`
+* `MultivariateNormal`
+"""
+
```
matbesancon: The docstring should be attached to the function:
```
"""
     convolve(d1::T, d2::T) where T<:Distribution -> Distribution
[...]
"""
function convolve end
```

src/convolution.jl
```diff
@@ -0,0 +1,95 @@
+"""
+    convolve(d1::T, d2::T) where T<:Distribution -> Distribution
```
matbesancon: There might be some closed-forms for different distributions:

```suggestion
    convolve(d1::D1, d2::D2) where {D1<:Distribution, D2 <: Distribution} -> Distribution
```

src/convolution.jl
```diff
@@ -0,0 +1,95 @@
...
+end
+
+function convolve(d1::MvNormal, d2::MvNormal)
+    _check_convolution_shape(d1, d2)
+    return MvNormal(d1.μ .+ d2.μ, Matrix(d1.Σ) + Matrix(d2.Σ) + 1e-12 * I)
+end
+
+
+function _check_convolution_args(p1, p2)
+    p1 == p2 || throw(ArgumentError("$(p1) != $(p2): distribution parameters must be equal"))
```
matbesancon: Shouldn't they bey approx equal?
```suggestion
    p1 ≈ p2 || throw(ArgumentError("$(p1) != $(p2): distribution parameters must be equal"))
```

matbesancon:
> BTW \oplus ⊕ has been suggested as an alias for the sum of indedepent random Variables which is (the generalisation) of what convolve does. (Presuming we're okay with conflating RVs with their distributions).

I think convolve leads to less ambiguity, but would be ok with `\oplus` as alias

matbesancon:
In any case this could go in another PR

src/convolution.jl
mschauer: Note that
```

julia> [1;3]*[1 3]
2×2 Array{Int64,2}:
 1  3
 3  9
```
so this is not posdef and `fn1.Σ` is half of that, so also not posdef up to floating point errors. it is really not the right example. I don't really like adding a multiple of `I` though

src/convolution.jl
glennmoy: At the moment though they are only defined for the same distribution. Might it be best to leave the generalised form for when the methods with two different distributions are added?

matbesancon: Ok fine by me, this is not a blocking point for the PR

src/convolution.jl
glennmoy: Do you think it is justifiable? I can see a reason where `p` (or the distribution) is estimated non-deterministically but I'm not sure this is a strong enough case.

matbesancon: Not only in non-deterministic cases, a strict equality makes this function's behavior dependent on the order of operations and on floating point trickeries, which is not something wanted here I believe

src/convolution.jl
glennmoy: I've removed the regulariser from the function and instead used a better construction for the posdef matrices:

```julia
m = Symmetric(rand(2,2))
m2 = m^2
MvNormal(m2.data)
```

I'm not against a regulariser kwarg but I think it should be used in the wild first to determine how often it is needed.

src/convolution.jl
glennmoy: fair point 👍

matbesancon: I just applied the update

matbesancon: @glennmoy could you apply it sorry, I can't commit on this branch

glennmoy: sure that's done!

matbesancon:
alright, I think we can merge this as soon as travis passes, check if you want some last modifications

glennmoy:
> alright, I think we can merge this as soon as travis passes, check if you want some last modifications

just cleaned up the comments. merge when ready!

PR #927, opened 2019-07-10
Add `MatrixTDist`

johnczito:
Implements the basic features here:

https://en.wikipedia.org/wiki/Matrix_t-distribution


src/matrix/matrixt.jl
```diff
@@ -0,0 +1,191 @@
+"""
+    MatrixT(ν, M, Σ, Ω)
```
nickrobinson251: Should be `MatrixTDist` for consistency with `TDist`, `MvTDist`?

johnczito: I dig it.

src/matrix/matrixtdist.jl
```diff
@@ -0,0 +1,197 @@
...
+
+params(d::MatrixTDist) = (d.ν, d.M, d.Σ, d.Ω)
+
+@inline partype(d::MatrixTDist{T}) where {T <: Real} = T
+
+#  -----------------------------------------------------------------------------
+#  Evaluation
+#  -----------------------------------------------------------------------------
+
+function _matrixt_c₀(Σ::AbstractPDMat, Ω::AbstractPDMat, ν::Real)
```
matbesancon: Is the unicode necessary here? Subscripts often don't render well, if possible `_matrixt_c0` would read fine here

johnczito: No, good catch.

src/matrix/matrixtdist.jl
```diff
@@ -0,0 +1,194 @@
...
+Ω:  p x p positive definite matrix
+```
+The [Matrix *t*-Distribution](https://en.wikipedia.org/wiki/Matrix_t-distribution)
+generalizes the Multivariate *t*-Distribution from vectors to matrices. An
+``n\\times p`` matrix ``\\mathbf{X}`` with the matrix *t*-distribution has density
+
+```math
+f(\\mathbf{X} ; \\nu,\\mathbf{M},\\boldsymbol{\\Sigma}, \\boldsymbol{\\Omega}) =
+c_0 \\left|\\mathbf{I}_n + \\boldsymbol{\\Sigma}^{-1}(\\mathbf{X} - \\mathbf{M})\\boldsymbol{\\Omega}^{-1}(\\mathbf{X}-\\mathbf{M})'\\right|^{-\\frac{\\nu+n+p-1}{2}},
+```
```
nickrobinson251: This will not render in the REPL... can we just rewrite it in plain text, using unicode as needed?

johnczito: Right. I'm emulating, for instance, [this](https://github.com/JuliaStats/Distributions.jl/blob/6561a7e00132c6698a961d454219b915bfb789f6/src/univariate/continuous/fdist.jl#L1) and [this](https://github.com/JuliaStats/Distributions.jl/blob/6561a7e00132c6698a961d454219b915bfb789f6/src/multivariate/mvnormal.jl#L25).

nickrobinson251: hmm, i see the precedent, but i cannot help but think it is unfortunate, as I think docstrings should be readable from the REPL whenever possible -- that seems the most user-friendly

matbesancon:
Most importantly it should render well in the docs, if possible also be
readable from the repl

src/matrix/matrixtdist.jl
johnczito: I agree completely. But since the docstring is doing double duty, I don't know how to get around it. In hindsight, [this](https://juliastats.github.io/Distributions.jl/latest/matrix/#Distributions.MatrixNormal) is kind of horrific, but works in the REPL. I will change it very soon.

src/matrix/matrixtdist.jl
```diff
@@ -0,0 +1,194 @@
...
+
+params(d::MatrixTDist) = (d.ν, d.M, d.Σ, d.Ω)
+
+@inline partype(d::MatrixTDist{T}) where {T <: Real} = T
+
+#  -----------------------------------------------------------------------------
+#  Evaluation
+#  -----------------------------------------------------------------------------
+
+function _matrixtdist_c0(Σ::AbstractPDMat, Ω::AbstractPDMat, ν::Real)
```
nickrobinson251: this is the log of the normalisation term / partition function?

I think it warrants a comment or short docstring saying that :)

src/matrix/matrixtdist.jl
```diff
@@ -0,0 +1,194 @@
...
+    term5 = (-p / 2) * logdet(Σ)
+
+    term1 + term2 + term3 + term4 + term5
+
+end
+
+function logkernel(d::MatrixTDist, X::AbstractMatrix)
+
+    A  = (X - d.M)
+    At = Matrix(A')
```
nickrobinson251: is there any benefit to calling `Matrix` here?

src/matrix/matrixtdist.jl
```diff
@@ -0,0 +1,194 @@
...
+
+    A .= rand(rng, MatrixNormal(d.M, S, d.Ω) )
+
+end
+
+#  -----------------------------------------------------------------------------
+#  Relationship with Multivariate t
+#  -----------------------------------------------------------------------------
+
+function MvTDist(MT::MatrixTDist)
```
nickrobinson251: I think you've thought about this, but is it worth putting a comment on why this isn't a `vec` method?

src/matrix/matrixtdist.jl
johnczito: I'd love for someone to teach me how to avoid doing this. But such as it is...

`A'` will return a `LinearAlgebra.Adjoint` instance, and `PDMats.jl` doesn't have a `a \ x` method for `a::PDMat` and `x::LinearAlgebra.Adjoint`, so you have to do this to keep `d.Ω \ At` from throwing an error.

src/matrix/matrixtdist.jl
johnczito: Yes, thanks. In general `vec(d::MatrixTDist)` is not known, but if a matrix is just a row or column and it's matrix *t*-distributed, then it's equivalent to a multivariate *t*-distributed vector.

src/matrix/matrixtdist.jl
johnczito: Done. Thanks.

src/matrix/matrixtdist.jl
nickrobinson251: This works for me without the Matrix call

johnczito: Oh, cool. Sorry about that. I got an error once upon a time.

src/matrix/matrixtdist.jl
```diff
@@ -0,0 +1,196 @@
...
+
+function MatrixTDist(ν::Real, M::AbstractMatrix, Σ::AbstractPDMat, Ω::AbstractPDMat)
+
+    T = Base.promote_eltype(ν, M, Σ, Ω)
+
+    MatrixTDist(convert(T, ν), convert(AbstractArray{T}, M), convert(AbstractArray{T}, Σ), convert(AbstractArray{T}, Ω))
+
+end
+
+MatrixTDist(ν::Real, M::AbstractMatrix, Σ::AbstractMatrix,         Ω::AbstractMatrix)         = MatrixTDist(ν, M, PDMat(Σ), PDMat(Ω))
```
matbesancon: No need to align the argument on every constructor, this tends to make too big diffs every time we modify functions.

Also, some of these constructors can be merged using `Union` type bounds. Example below:
```julia
MatrixTDist(ν::Real, M::AbstractMatrix, Σ::Union{AbstractMatrix, LinearAlgebra.Cholesky}, Ω::Union{AbstractMatrix, LinearAlgebra.Cholesky}) = MatrixTDist(ν, M, PDMat(Σ), PDMat(Ω))
```

src/matrix/matrixtdist.jl
```diff
@@ -0,0 +1,196 @@
...
+
+params(d::MatrixTDist) = (d.ν, d.M, d.Σ, d.Ω)
+
+@inline partype(d::MatrixTDist{T}) where {T <: Real} = T
+
+#  -----------------------------------------------------------------------------
+#  Evaluation
+#  -----------------------------------------------------------------------------
+
+function _matrixtdist_logc0(Σ::AbstractPDMat, Ω::AbstractPDMat, ν::Real)
```
matbesancon: Underscore not needed here, it's fine to remove it

src/matrix/matrixtdist.jl
```diff
@@ -0,0 +1,196 @@
...
+function logkernel(d::MatrixTDist, X::AbstractMatrix)
+
+    n, p = size(d)
+    A = X - d.M
+
+    (-(d.ν + n + p - 1) / 2) * logdet( I + (d.Σ \ A) * (d.Ω \ A') )
+
+end
+
+_logpdf(d::MatrixTDist, X::AbstractMatrix) = logkernel(d, X) + d.logc0
```
matbesancon: Why `_logpdf` and not `logpdf`?

matbesancon:
Couple comments to address, then I think we will be good to merge

nickrobinson251:
LGTM! Thanks for adding 👍

src/matrix/matrixtdist.jl
johnczito: That would appear to be the standard? See [here](https://github.com/JuliaStats/Distributions.jl/blob/master/src/matrix/wishart.jl#L103), [here](https://juliastats.github.io/Distributions.jl/latest/matrix/#Internal-Methods-(for-creating-your-own-matrix-variate-distributions)-1), and [here](https://juliastats.github.io/Distributions.jl/latest/extends/#Create-a-Matrix-variate-Distribution-1). 

Matrix-variates have a generic `logpdf` fallback [here](https://github.com/JuliaStats/Distributions.jl/blob/d876df09f06a8d559e4733a875fbd11709f2dc60/src/matrixvariates.jl#L102) that checks dims. 

(While we're on the subject, I guess it might make more sense for the fallback to call `insupport` instead of just checking the dims, but that's a PR for another day...)

matbesancon:
Ah got it yes thanks, sorry I had forgotten

src/matrix/matrixtdist.jl
johnczito: Removed.

src/matrix/matrixtdist.jl
johnczito: Awesome. Thanks!

johnczito:
Thanks for all of the help again. Let me know if you want anything else.

A lot of the feedback for this also applies to the `MatrixNormal` type we recently merged (consolidating constructors, TeX-ing the docstring, eliminating unicode subscripts, eliminating `Matrix(A')`), so I'll prepare a PR with those changes for the `MatrixNormal` shortly.

matbesancon:
@johnczito great work, there was a recently discovered bug for the univariate case, could you check not to have the same?
https://github.com/JuliaStats/Distributions.jl/issues/929

johnczito:
> @johnczito great work, there was a recently discovered bug for the univariate case, could you check not to have the same?
> #929

Right. If you give `ν = Inf`...

- `rand()` throws an error because the `InverseWishart` can't handle infinite degrees of freedom. 
- `logpdf()` returns `NaN` because `d.logc0` is `NaN`. `d.logc0` is `NaN` because, in `matrixtdist_logc0()`, `term1` is `Inf` and `term3` is `-Inf`, so `Inf - Inf = NaN`.

Is this behavior acceptable? I don't think `ν = Inf` corresponds to anything meaningful. Unlike with the univariate Student's *t*-distribution, this is not true:
```
MT(ν, M, Σ, Ω) → MN(M, Σ, Ω) as ν → ∞
```
This is true:
```
MT(ν, M, νΣ, Ω) → MN(M, Σ, Ω) as ν → ∞
```
But you shouldn't expect to be able to just plug in `Inf * Σ` and get back a `MN(M, Σ, Ω)`.

matbesancon:
Agreed, maybe should we error at construction when a Inf is provided?

johnczito:
Done.

matbesancon:
Seems all good, that's a cool add!
If you have time to tackle the issue I linked, that would be fantastic, and we can release after

Issue #929, opened 2019-07-10
rand( TDist(Inf) ) gives NaN in v0.20.0


atteson:
I noticed that `rand( TDist(Inf) )` gives NaN in v0.20.0 but gives a value in v0.16.4 (which I assume corresponds to a standard normal variate).

`pdf( TDist(Inf), 1.0 )` works fine in both versions.

matbesancon:
Several issues here, the distribution parameter should not be taking a real type but an integer value. The type parameter could capture in some sense the infinite / finite thing and dispatch accordingly.

The documentation is also pointing to the "student" wikipedia page, which is not the Student distribution page.
PR welcome

atteson:
Actually, the t distribution is perfectly well defined for integer values of nu and I (and others) use it that way.

johnczito:
What's the preferred behavior, here? Throw an error at construction if you supply `Inf` degrees of freedom? Or effectively have `TDist(Inf)` be a `Normal` constructor?

johnczito:
I think the change happened [here](https://github.com/JuliaStats/Distributions.jl/commit/817bd83326f9d562ec88c0e782e60d25d64862ad), where
```julia
rand(d::TDist) = StatsFuns.RFunctions.tdistrand(d.ν)
```
became
```julia
rand(rng::AbstractRNG, d::TDist) = randn(rng) / sqrt(rand(rng, Chisq(d.ν))/d.ν)
```
The denominator becomes `sqrt(Inf/Inf) = NaN`. So do we want something like this?
```julia
rand(rng::AbstractRNG, d::TDist) = randn(rng) / isinf(d.ν) ? 1 : sqrt(rand(rng, Chisq(d.ν))/d.ν)
```

atteson:
My preference is to allow for the nu=Inf to mean the normal distribution since this is always the limiting behavior.  I like the way the R function works.  However, there are other limiting cases which are not handled in Distributions.  For example, a normal with sigma=0 doesn't become a delta distribution.

johnczito:
Right, so you want `TDist(Inf)` to simply return `Normal(0,1)`. That makes sense, though I'm not sure what the type stability implications of that are.

atteson:
Actually, my intent is that TDist( Inf ) remain that type but act like Normal(0,1) in rand, pdf, cdf, etc.  I think that's probably better because of type stability as you mention.  Someone might assign these to a Vector{TDist} or some such.

[...]

PR #931, opened 2019-07-12
Make TDist(Inf) behave like Normal(0,1) (addresses Issue #929)

johnczito:
#929 wants `rand( TDist(Inf) )` to return standard normal draws like it used to (currently returns `NaN`). This PR brings that back, and also introduces unit tests for confirming that `TDist(Inf)` behaves like `Normal(0,1)` in other ways: `mean`, `var`, `pdf`, etc. They currently *do not* agree on `entropy`, `gradlogpdf`, `cf`. Not sure if/how we want to handle that. Let me know.

### DIFF
--- a/test/edgecases.jl
+++ b/test/edgecases.jl
@@ -1,0 +1,25 @@
+using Test, Distributions, StatsBase
+
+@testset "TDist(Inf) is Normal(0,1)" begin
+    T = TDist(Inf)
+    N = Normal(0, 1)
+
+    x = rand(N)
+    z = rand(T, 10000000)
+
+    @test mean(T) == mean(N)
+    @test median(T) == median(N)
+    @test mode(T) == mode(N)
+    #@test var(T) == var(N)
+    @test skewness(T) == skewness(N)
+    @test kurtosis(T) == kurtosis(N)
+    #@test entropy(T) == entropy(N)
+    @test pdf(T, x) ≈ pdf(N, x)
+    @test logpdf(T, x) ≈ logpdf(N, x)
+    #@test gradlogpdf(T, x) ≈ gradlogpdf(N, x)
+    #@test cf(T, x) ≈ cf(N, x)
+
+    fnecdf = ecdf(z)
+    y = [-1.96, -1.644854, -1.281552, -0.6744898, 0, 0.6744898, 1.281552, 1.644854, 1.96]
+    @test isapprox(fnecdf(y), [0.025, 0.05, 0.1, 0.25, 0.5, 0.75, 0.9, 0.95, 0.975], atol=1e-3)
+end
--- a/test/edgecases.jl
+++ b/test/edgecases.jl
@@ -13,11 +13,11 @@ using Test, Distributions, StatsBase
     @test var(T) == var(N)
     @test skewness(T) == skewness(N)
     @test kurtosis(T) == kurtosis(N)
-    #@test entropy(T) == entropy(N)
+    @test entropy(T) == entropy(N)
     @test pdf(T, x) ≈ pdf(N, x)
     @test logpdf(T, x) ≈ logpdf(N, x)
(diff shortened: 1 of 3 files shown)


test/edgecases.jl
```diff
@@ -0,0 +1,25 @@
...
+    @test mean(T) == mean(N)
+    @test median(T) == median(N)
+    @test mode(T) == mode(N)
+    @test var(T) == var(N)
+    @test skewness(T) == skewness(N)
+    @test kurtosis(T) == kurtosis(N)
+    #@test entropy(T) == entropy(N)
+    @test pdf(T, x) ≈ pdf(N, x)
+    @test logpdf(T, x) ≈ logpdf(N, x)
+    #@test gradlogpdf(T, x) ≈ gradlogpdf(N, x)
```
matbesancon: why the commented tests?

matbesancon:
there is a minor comment on commented tests, other than that looks great

test/edgecases.jl
johnczito: `entropy`, `gradlogpdf`, and `cf` are currently not rigged to give the same output as for `Normal(0,1)`, so we have:
```julia
using Distributions
d = TDist(Inf)

julia> entropy(d)
NaN

julia> gradlogpdf(d, 1.0)
NaN

julia> cf(d, 0.25)
ERROR: AmosException with id 4: input argument magnitude too large, complete loss of accuracy by argument reduction.
Stacktrace:
 [1] _besselk(::Float64, ::Complex{Float64}, ::Int32) at /Users/johnczito/.julia/packages/SpecialFunctions/fvheQ/src/bessel.jl:239
 [2] besselk at /Users/johnczito/.julia/packages/SpecialFunctions/fvheQ/src/bessel.jl:354 [inlined]
 [3] besselk at /Users/johnczito/.julia/packages/SpecialFunctions/fvheQ/src/bessel.jl:450 [inlined]
 [4] cf(::TDist{Float64}, ::Float64) at /Users/johnczito/.julia/packages/Distributions/Iltex/src/univariate/continuous/tdist.jl:82
 [5] top-level scope at none:0
```
Those tests would currently fail if they were included.

matbesancon: ok, fine for this behaviour

matbesancon:
@johnczito do you want to cover the three missing before merging?
An other solution would be to have a type parameter on `TDist` indicating whether the parameter is finite. If it is not we can fallback everything to calling a Normal

johnczito:
Sure. I don't know how people would prefer that the edge behavior be handled, but I can commit something and you'll let me know if it's horrible.

matbesancon:
it sounds easier and more simple to use dispatch for that

johnczito:
I don't think I understand what you have in mind. Naively, isn't having `TDist(ν::Real)` return one type if `ν = Inf` and another type otherwise the definition of type instability?

matbesancon:
Ah that's a fair point, but the variation would be minor, `TDist{T, F}` where `F` is true or false for finiteness

johnczito:
Provisionally, I have made them all agree by sheer brute force; each method checks if `ν` is infinite and calls the `Normal(0,1)` method if so. This has the advantage of solving the problem without adding new type parameters to `TDist` (which could break someone's code somewhere?), but it has the disadvantage of being inelegant.

matbesancon:
I think it will be fine for now, we can improve perf later

johnczito:
Well, looking at the source code for [`Normal`](https://github.com/JuliaStats/Distributions.jl/blob/master/src/univariate/continuous/normal.jl), I don't feel so bad. It's full of `if` statements to get `Normal(μ, 0)` to behave. So this solution is in line with the handling of edge cases elsewhere in `Distributions.jl`.

matbesancon:
@johnczito good to merge?

johnczito:
If you're good with it, I'm good with it. Thanks!

PR #935, opened 2019-07-19
Add `MatrixFDist`

johnczito:
Implements the basic features of the matrix *F*-distribution. See [here](https://projecteuclid.org/euclid.ba/1515747744), and also the "generalized beta type II" distribution in chapter 5 of [Gupta and Nagar (1999)](https://www.crcpress.com/Matrix-Variate-Distributions/Gupta-Nagar/p/book/9781584880462).

(This would also make use of JuliaStats/StatsFuns.jl#72 in the long-run.)

### DIFF
--- a/src/matrix/matrixfdist.jl
+++ b/src/matrix/matrixfdist.jl
@@ -111,7 +111,6 @@ function matrixfdist_logc0(n1::Real, n2::Real, B::AbstractPDMat)
     term1 = logmvgamma(p, (n1 + n2)/2) - logmvgamma(p, n1/2) - logmvgamma(p, n2/2)
     term2 = (n2 / 2) * logdet(B)
     return term1 + term2
-    #-logmvbeta(dim(B), n1 / 2, n2 / 2) + (n2 / 2) * logdet(B)
 end
 
 function logkernel(d::MatrixFDist, Σ::AbstractMatrix)
--- a/src/matrix/matrixfdist.jl
+++ b/src/matrix/matrixfdist.jl
@@ -1,0 +1,143 @@
+"""
+```julia
+MatrixFDist(n1, n2, B)
+
+n1::Real  degrees of freedom (greater than p - 1)
+n2::Real  degrees of freedom (greater than p - 1)
+B::PDMat  p x p scale
+```
+The [matrix *F*-Distribution](https://projecteuclid.org/euclid.ba/1515747744)
+(sometimes called the matrix beta type II distribution) generalizes the
+*F*-Distribution to ``p\\times p`` real, positive definite matrices ``\\boldsymbol{\\Sigma}``.
+If ``\\boldsymbol{\\Sigma}\\sim MF_{p}(n_1/2,n_2/2,\\mathbf{B})``, then its probability density function is
+
+```math
+f(\\boldsymbol{\\Sigma} ; n_1,n_2,\\mathbf{B}) =
+\\frac{\\Gamma_p(\\frac{n_1+n_2}{2})}{\\Gamma_p(\\frac{n_1}{2})\\Gamma_p(\\frac{n_2}{2})}
+|\\mathbf{B}|^{n_2/2}|\\boldsymbol{\\Sigma}|^{(n_1-p-1)/2}|\\mathbf{B}+\\boldsymbol{\\Sigma}|^{-(n_1+n_2)/2}
+```
+
+If the joint distribution ``p(\\boldsymbol{\\Psi},\\boldsymbol{\\Sigma})=p(\\boldsymbol{\\Psi})p(\\boldsymbol{\\Sigma}|\\boldsymbol{\\Psi})``
+is given by
+
+```math
+\\begin{align*}
+\\boldsymbol{\\Psi}&\\sim W_p(n_1, \\mathbf{B})\\\\
+\\boldsymbol{\\Sigma}|\\boldsymbol{\\Psi}&\\sim IW_p(n_2, \\boldsymbol{\\Psi}),
+\\end{align*}
+```
+
+then the marginal distribution of ``\\boldsymbol{\\Sigma}`` is
+``MF_{p}(n_1/2,n_2/2,\\mathbf{B})``.
+"""
+struct MatrixFDist{T <: Real, TW <: Wishart} <: ContinuousMatrixDistribution
+    W::TW
+    n2::T
+    logc0::T
+end
+
+#  -----------------------------------------------------------------------------
+#  Constructors
+#  -----------------------------------------------------------------------------
+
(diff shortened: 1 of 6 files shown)


src/matrix/matrixfdist.jl
```diff
@@ -0,0 +1,143 @@
...
+#  -----------------------------------------------------------------------------
+#  Conversion
+#  -----------------------------------------------------------------------------
+
+function convert(::Type{MatrixFDist{T}}, d::MatrixFDist) where T <: Real
+    W = convert(Wishart{T}, d.W)
+    MatrixFDist{T, typeof(W)}(W, T(d.n2), T(d.logc0))
+end
+
+function convert(::Type{MatrixFDist{T}}, W::Wishart, n2, logc0) where T <: Real
```
matbesancon: Shouldn't it be a subtype?:
```
function convert(::Type{<:MatrixFDist{T}}, W::Wishart, n2, logc0) where T <: Real
```

src/matrix/matrixfdist.jl
```diff
@@ -0,0 +1,143 @@
...
+
+insupport(d::MatrixFDist, Σ::AbstractMatrix) = isreal(Σ) && size(Σ) == size(d) && isposdef(Σ)
+
+params(d::MatrixFDist) = (d.W.df, d.n2, d.W.S)
+
+function mean(d::MatrixFDist)
+    p = dim(d)
+    n1, n2, B = params(d)
+
+    (n2 > p + 1) ? (return (n1 / (n2 - p - 1)) * B.mat) : throw(ArgumentError("mean only defined for df2 > dim + 1"))
```
matbesancon: First check, if check fails throw, then final return is the "happy path", the normal value

src/matrix/matrixfdist.jl
```diff
@@ -0,0 +1,143 @@
...
+#  Evaluation
+#  -----------------------------------------------------------------------------
+
+function matrixfdist_logc0(n1::Real, n2::Real, B::AbstractPDMat)
+    #  returns the natural log of the normalizing constant for the pdf
+    p = dim(B)
+    term1 = logmvgamma(p, (n1 + n2)/2) - logmvgamma(p, n1/2) - logmvgamma(p, n2/2)
+    term2 = (n2 / 2) * logdet(B)
+    return term1 + term2
+    #-logmvbeta(dim(B), n1 / 2, n2 / 2) + (n2 / 2) * logdet(B)
```
matbesancon: remove unused comment

src/matrix/matrixfdist.jl
johnczito: Maybe? Frankly, for all of these I've just been naively copying how `convert` is written for `Wishart` and `InverseWishart`. 

https://github.com/JuliaStats/Distributions.jl/blob/dc2e589402fc8f814f29df40df933368badf849f/src/matrix/wishart.jl#L57-L64

test/matrixfdist.jl
```diff
@@ -0,0 +1,75 @@
...
+end
+
+@testset "MatrixFDist sample moments" begin
+    @test isapprox(mean(rand(G, 100000)), mean(G) , atol = 0.1)
+end
+
+@testset "MatrixFDist conversion" for elty in (Float32, Float64, BigFloat)
+    Gel1 = convert(MatrixFDist{elty}, G)
+    Gel2 = convert(MatrixFDist{elty}, G.W, n2, G.logc0)
+
```
matbesancon: add `isa` tests to check conversion yields types as expected

johnczito: Great. Thanks!

src/matrix/matrixfdist.jl
johnczito: Gone.

src/matrix/matrixfdist.jl
johnczito: Flipped.

src/matrix/matrixfdist.jl
```diff
@@ -0,0 +1,142 @@
...
+
+insupport(d::MatrixFDist, Σ::AbstractMatrix) = isreal(Σ) && size(Σ) == size(d) && isposdef(Σ)
+
+params(d::MatrixFDist) = (d.W.df, d.n2, d.W.S)
+
+function mean(d::MatrixFDist)
+    p = dim(d)
+    n1, n2, B = params(d)
+
+    (n2 <= p + 1) ? throw(ArgumentError("mean only defined for df2 > dim + 1")) : (return (n1 / (n2 - p - 1)) * B.mat)
```
matbesancon: could we maybe put this on several lines?
```suggestion
    if n2 <= p + 1
        throw(ArgumentError("mean only defined for df2 > dim + 1"))
    end
    return (n1 / (n2 - p - 1)) * B.mat
```

matbesancon:
@johnczito this one looks good to merge?

src/matrix/matrixfdist.jl
```diff
@@ -0,0 +1,142 @@
...
+    n2::T
+    logc0::T
+end
+
+#  -----------------------------------------------------------------------------
+#  Constructors
+#  -----------------------------------------------------------------------------
+
+function MatrixFDist(n1::Real, n2::Real, B::AbstractPDMat)
+
```
ararslan: I won't comment on this everywhere, but generally speaking we don't include blank lines between the function signature and the first statement, as well as between the last statement and the closing `end`. If you could fix those, that would be great.

src/matrix/matrixfdist.jl
```diff
@@ -0,0 +1,142 @@
+"""
+```julia
```
ararslan: Don't the docstrings usually start with the name of the distribution at the top? Might want to check others for consistency.

test/matrixfdist.jl
```diff
@@ -0,0 +1,77 @@
...
+end
+
+@testset "MatrixFDist insupport" begin
+    @test insupport(G, rand(G))
+    @test insupport(F, rand(F))
+
+    @test !insupport(G, rand(G) + rand(G) * im)
+    @test !insupport(G, randn(p, p + 1))
+    @test !insupport(G, randn(p, p))
+
```
ararslan:
```suggestion
```

src/matrix/matrixfdist.jl
johnczito: Yes. A couple don't, but most do. The first thing in the `julia` block is the name, but the effect is the same if it's above it.

src/matrix/matrixfdist.jl
johnczito: Done. Thanks!

johnczito:
@ararslan Thanks for the feedback! Your suggestions apply to some other stuff we recently merged. I'll make sure to submit those revisions in the near future.

PR #936, opened 2019-07-19
upgrade matrix-variate docs

johnczito:
Bring the documentation of `Wishart` and `InverseWishart` up to the standard of the documentation for [`FDist`](https://juliastats.github.io/Distributions.jl/stable/univariate/#Distributions.FDist), [`MvNormal`](https://juliastats.github.io/Distributions.jl/stable/multivariate/#Distributions.AbstractMvNormal), [`MatrixNormal`](https://juliastats.github.io/Distributions.jl/latest/matrix/#Distributions.MatrixNormal), [`MatrixBeta`](https://juliastats.github.io/Distributions.jl/latest/matrix/#Distributions.MatrixBeta), etc.

### DIFF
--- a/docs/src/matrix.md
+++ b/docs/src/matrix.md
@@ -2,6 +2,13 @@
 
 *Matrix-variate distributions* are the distributions whose variate forms are `Matrixvariate` (*i.e* each sample is a matrix). Abstract types for matrix-variate distributions:
 
+```julia
+const MatrixDistribution{S<:ValueSupport} = Distribution{Matrixvariate,S}
+
+const DiscreteMatrixDistribution   = Distribution{Matrixvariate, Discrete}
+const ContinuousMatrixDistribution = Distribution{Matrixvariate, Continuous}
+```
+
 ## Common Interface
 
 All distributions implement the same set of methods:
--- a/src/matrix/inversewishart.jl
+++ b/src/matrix/inversewishart.jl
@@ -1,9 +1,21 @@
 """
-    InverseWishart(nu, P)
+```julia
+InverseWishart(ν, Ψ)
 
-The [Inverse Wishart distribution](http://en.wikipedia.org/wiki/Inverse-Wishart_distribution)
-is usually used as the conjugate prior for the covariance matrix of a multivariate normal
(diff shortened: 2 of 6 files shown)


matbesancon:
@johnczito did you check that this renders well in the docs?

johnczito:
Yes. From my local build:
Screen Shot 2019-07-19 at 8 22 41 AM

matbesancon:
Wonderful, one thing is that this is getting heavy on latex syntax which gets REPL doc harder to read. Maybe one thing we can do is:
Push more Julia code and text in the docstring, put advanced math including latex in the documentation files, what do you think?

johnczito:
I agree that it's preferred to have both docs and REPL readable. Off-the-shelf, I don't know how to mix the docstring output with stuff that's hard-coded into the `docs` folder, so I'd have to play around with it first. 

It could be better to bite off as part of a larger project, since the `Univariate`s are also [very](https://juliastats.github.io/Distributions.jl/latest/univariate/#Distributions.GeneralizedExtremeValue) TeX-ed.

matbesancon:
You're right, it renders horribly in the REPL:
image

matbesancon:
For the matrix types for now, we could use Julia code blocks with unicode symbols to show equations, it renders ok in doc pages too

johnczito:
I guess for the purpose of meeting a `v0.21.0` release, I'm in favor of merging this as-is. Then for `v0.21.0` we can say that the documentation for all of the `Distribution` types conforms to the same standard -- that standard being that the web documentation is nice and fully TeX-ed, and the REPL documentation is kind of unreadable. Then after the release I can try to implement your first suggestion of finding a way to put the TeX someplace other than the docstring, and combining them in the `docs` folder. I like this approach because I don't think the REPL docs need to have any math in them anyway, however rendered.

matbesancon:
That's good by me, changing the doc is not more than a patch release, it will be good to have this release out first.

matbesancon:
if it's good on your side, we can merge this one

johnczito:
Yupp, builds fine for me.

Issue #925, opened 2019-07-08
Handling discontinuous density functions


richardreeve:
I'm thinking about problems with non-continuous distributions, and whether it really makes sense to only have `Union{Discrete, Continuous} <: ValueSupport`. There was a brief discussion about mixed discrete and continuous distributions in #332, then more recently a couple of attempts to deal with other problems with discrete distributions in #887 and #916 allowing non-integer support. However, it seems like none of them is going anywhere at the moment...

The problem I see is that as well as being the probability density function for continuous distributions, `pdf` is also the probability *mass* function for discrete distributions. I'd like to be able to define a slab-and-spike distribution (as mentioned in that first issue), but I can't see how to do it, not just because distributions in `MixtureModel`s need to have all the same `ValueSupport` subtype, but because it's not clear what the `ValueSupport` should be, nor what the `pdf` function should return - if we treat it as a density function, it's infinite at point masses, whereas if we treat it as a mass function, then it's zero everywhere else.

My feeling is that there's a problem with reusing `pdf` in both cases... `pmf` seems like a better bet. This can obviously be done in a non-breaking way by aliasing `pmf` to `pdf`, but then the mixed distributions can define both. What do people think? What do these distributions actually have to provide?

The end goal of this from my perspective is to be able to construct zero-inflated, hurdle and slab-and-spike distributions easily from their constituent parts (a point mass and another discrete or continuous distribution) as they all get used a reasonable amount in the real world... [in fact there isn’t even a point mass distribution at the moment I don’t think?] But it's also to clean up the handling of discontinuous-but-not-discrete and discrete distributions more generally.

[...]

Issue #887, opened 2019-05-10
DiscreteNonParametric should be <: Continuous


abraunst:
The documentation at https://github.com/JuliaStats/Distributions.jl/blob/master/docs/src/types.md declares that distributions with domain in the reals should be subtypes of `Continuous`. However, `DiscreteNonParameteric` in https://github.com/JuliaStats/Distributions.jl/blob/master/src/univariate/discrete/discretenonparametric.jl is declared as subtype of `Discrete`, violating this rule. This seems a bug (see also my comments at the end of #332).

matbesancon:
Agreed on that.
Sort-of related to https://github.com/JuliaStats/Distributions.jl/issues/771

richardreeve:
This also seems to be related to #916 though that PR actually only adds non-real support. It would be good to fix this `ValueSupport` bug, and I don't see why the `Discrete` subtype has to have `Int` elements at all, it just needs to have a finite (countably infinite?) number of whatever its element type is.

mschauer:
At this point where we need to define `Discrete <: Continuous` , we become aware that we all want to abolish the type params :-D

richardreeve:
Well, I have to admit I'm not thinking of going quite that far! I was imagining `DiscreteSupport <: DiscontinuousSupport <: ValueSupport` and `ContinuousSupport <: ValueSupport`, and allowing DiscreteSupport to allow support over any countable set, whereas ContinuousSupport is uncountable, and Discontinuous is a statement about have Dirac deltas in the density function.

DilumAluthge:
Yeah, I would argue that "discrete" doesn't mean "the support is a subset of integers", but rather "the support is a countable set".

mschauer:
And continuous does not mean "support is a subset of the reals" but should mean perhaps that the distribution has a density (we are not going to model cantor distributions...)

richardreeve:
Does continuous require that support is a `Number` though? That would cover a lot of bases (including complex numbers, `Unitful` numbers, etc.), though we would also need the underlying number type (? I'm not sure there's anything that extracts this at the moment) to be `<: AbstractFloat` I imagine?

mschauer:
Just to state it clearly: The type system needs fixing, but  `DiscreteNonParametric` should not be `<: ContinuousDistribution`

richardreeve:
[...]

Issue #332, opened 2015-01-12
MixtureModel with Continuous and Discrete


dfagnan:
Is there a reason not to allow for mixtures with both continuous and discrete underlying distributions?

It seems like right now it would break the model, since it would be neither discrete nor continuous in terms of ValueSupport. I guess this would be a major change.

```
MixtureModel([Normal(); Geometric(0.5)])

ERROR: `variate_form` has no method matching variate_form(::Type{Sampleable{F<:V
ariateForm,S<:ValueSupport}})
 in MixtureModel at C:\Users\David\.julia\v0.3\Distributions\src\mixturemodel.jl
:20
 in MixtureModel at C:\Users\David\.julia\v0.3\Distributions\src\mixturemodel.jl
:29
```

lindahua:
Yes, all components should be on the same space. We haven't explicitly enforced this constraint yet, but need to work on this.

abraunst:
What about atomic distributions, are they continuous or discrete? More in particular, how does a distribution which is a mixture of continuous and point-mass fits here? Is it continuous or discrete? (math answer IMO: it is neither -- continuous and discrete are not complementary)

Real example:  I'm interested in spike-and-slab or Gauss-Bernoulli (mixture of an atomic point-mass distribution and a normal). Right now it seems the only way of producing this is by using a degenerate normal with sigma=0 instead of a point-mass. Is this intended?

matbesancon:
@abraunst I think in that case you should construct your own mixture type in that case, not use the common one, assuming same space for the two distributions seems expectable

abraunst:
While I agree with your statement, I think that "continuous" is a bit ambiguous: it may mean that the distribution is defined on the real line (i.e. a distinction on the space on which is defined, as you mention) or that it is absolutely continuous. I think that the former is more useful.

So, in my humble opinion, discrete should mean that the distribution is defined on a subset of Z or Z^n (and continuous all other cases). In this interpretation the point mass (and mixtures of point mass) should be continuous.

mschauer:
The absolute continuity is best seen as a property of the `pdf`. Continuous then just means that the pdf is given with respect to the Lebesgue measure.

PR #941, opened 2019-07-26
Fix ValueSupport to allow non-integer discrete support

richardreeve:
I've made this PR to provide a better way of handling non-integer discrete support, since it doesn't currently exist in Distributions. There are cleverer things that could be done, but this seems to work and so far provides all of the functionality people are talking about that I've seen.

- [x] ~New ValueSupport type hierarchy, allowing non-integer discrete support through CountableSupport <: ValueSupport~ [now in #945]
- [x] Default to `pmf()` and `logpmf()` for distributions with countable support as we have mass not density, but allow fallback to `[log]pdf()` for compatibility.
- [x] Incorporates #916 allowing DiscreteNonParametric to have non-integer discrete support, addressing #887
- [x] Preparatory work for #925 to allow discontinuous densities
- [x] Incorporate Dirac distribution from #861 and fix to new ValueSupport.
- [x] Add Slab-and-Spike (SpikeSlab) distribution for TuringLang/Turing.jl#847
- [x] Add general mixture of continuous and discrete (CompoundDistribution) addressing #332
- [x] Implement general zero-inflated (ZeroInflated) and hurdle (Hurdle) distributions as special cases of CompoundDistribution
- [ ] Add tests

This PR is now split, with the `ValueSupport` changes in #945 and the new distributions using the new type hierarchy here. This PR contains both, but the important stuff is the latter, as the former will go once #945 is merged.


DilumAluthge:
I like it

matbesancon:
@richardreeve could you merge master into your branch? Very sorry for the mess, it's a big PR that just got merged (but boilerplaty so the diff should be trivial)

richardreeve:
Done. Any comments welcome by the way @matbesancon and others! I need to put in some tests for the new distributions, but apart from that I’m pretty happy with this and I think it does what I need, and I think @DilumAluthge is also okay with it...

src/common.jl
```diff
@@ -85,17 +97,34 @@ abstract type Distribution{F<:VariateForm,S<:ValueSupport} <: Sampleable{F,S} en
 const UnivariateDistribution{S<:ValueSupport}   = Distribution{Univariate,S}
 const MultivariateDistribution{S<:ValueSupport} = Distribution{Multivariate,S}
 const MatrixDistribution{S<:ValueSupport}       = Distribution{Matrixvariate,S}
-const NonMatrixDistribution = Union{UnivariateDistribution, MultivariateDistribution}
+const NonMatrixDistribution = Union{UnivariateDistribution,
```
matbesancon: I would advise against this.
`const NonX = Union{NonX1, NonX2}` is only valid as long as you only have `X1, X2` as other subtypes.

src/univariate/continuous/normal.jl
```diff
@@ -169,11 +169,11 @@ invlogccdf(d::Normal, lp::Real) = xval(d, -norminvlogcdf(lp))
 function quantile(d::Normal, p::Real)
```
matbesancon: Wait what? Some the code wrong before? If so, could you add a test fixed by this

src/univariate/discrete/bernoulli.jl
```diff
@@ -80,8 +80,8 @@ end
 
 #### Evaluation
 
-pdf(d::Bernoulli, x::Bool) = x ? succprob(d) : failprob(d)
-pdf(d::Bernoulli, x::Int) = x == 0 ? failprob(d) :
+pmf(d::Bernoulli, x::Bool) = x ? succprob(d) : failprob(d)
+pmf(d::Bernoulli, x::Int) = x == 0 ? failprob(d) :
```
matbesancon: no need to add two methods nor to hard-code `Int`:
`Bool <: Integer` and `true == T(1)` for any `T <: Integer`

matbesancon:
ok before going any further, this PR is huge, should probably be chunked in multiple ones, there should be one for the type hierarchy, then we'll see for the rest of it

src/common.jl
richardreeve: This is existing code - actually this particular thing was just an automatic whitespace fix because it was going over the line length. I'm happy to undo my accidental edit, but I didn't do this...

src/univariate/discrete/bernoulli.jl
richardreeve: Again, this isn't my code - all I've done is searched and replaced pdf with pmf for discrete distributions. I'm sure there are lots of problems with the existing code, but I felt that I was already being ambitious enough!

src/univariate/continuous/normal.jl
richardreeve: The code previously was giving the right answer by accident as the final line of code (the xval() call) worked on its own anyway even in these edge cases so there's nothing to test. I just corrected the earlier code that was not previously returning anything. So two mistakes cancelled each other out and there's nothing to test.

richardreeve:
I did wonder about that, but it's worth pointing out that what you're spotting up there are not my introductions. Although the changes touch a lot of files, it's mostly just search and replace of pdf for pmf...

richardreeve:
It's also true that the Dirac work has been sitting around as a #861 for a while and needed updating to this new hierarchy, and the same is true in #916 for the DiscreteNonParametric and Categorical distributions - I've just merged and updated these, so they have been largely reviewed already. Also, the type hierarchy is all in common.jl (except for exports in Distributions.jl and the doc updates), so it can be reviewed almost entirely in isolation. However, I can split it out if you really think it will be better as I obviously want this to be merged.

src/common.jl
richardreeve: Reverted.

richardreeve:
Type hierarchy is now in #945.

matbesancon:
@richardreeve maybe wait for #945 before continuing this? Otherwise you risk running into merge conflicts

richardreeve:
Absolutely. I'm just keeping the two synced by merging in rr/countable and checking it still works...

richardreeve:
Keeping up to date with #941, so we now have `ContiguousSupport` as a subtype of `CountableSupport`.

richardreeve:
This PR is now split, with the `ValueSupport` changes in #945 and the new distributions using the new type hierarchy here. This PR contains both, but the important stuff is the latter, as the former will go once #945 is merged.

cscherrer:
Can you give some more detail of the benefit of splitting pdf/pmf? Formally, a pmf is still a pdf over counting measure. Breaking them apart forces everything to consider two separate cases without any obvious benefit. This can be a pain point for PPL implementation in particular.

I can see a benefit if you're describing something that's singular over the base measure like a sample from a Dirichlet process, but I don't see that addressed here.

richardreeve:
Once you start having arbitrary mixtures of distributions, that includes mixtures of continuous and discrete distributions. If you do that, you get infinite spikes in the (pdf) density that represent discrete probability masses (in the pmf). However, in the pdf there is no way of quantifying the size of the spikes (as far as I'm aware).

For cumulative densities this isn't a problem, but I don't see a way around having two functions for the density and mass functions if you want to generally be able to describe the density and the mass of distributions with discontinuous cdfs but continuous support (like spike-and-slab and hurdle distributions).

cscherrer:
> Once you start having arbitrary mixtures of distributions, that includes mixtures of continuous and discrete distributions. If you do that, you get infinite spikes in the (pdf) density that represent discrete probability masses (in the pmf). However, in the pdf there is no way of quantifying the size of the spikes (as far as I'm aware).

Ok, so this addresses the problem of singularities. That makes sense. Say you have a mixture like this, and `d` is in the support of the discrete component, while `c` is outside it. Does your implementation return 

```julia
pdf(mix, d) == Inf
pdf(mix, c) == continuousWeight * pdf(continuousComponent, c)
pmf(mix, d) == discreteWeight * pmf(discreteComponent, d)
pmf(mix, c) == 0
```

? Is that what it "should" return, or am I thinking about it wrong?

> For cumulative densities this isn't a problem, but I don't see a way around having two functions for the density and mass functions if you want to generally be able to describe the density and the mass of distributions with discontinuous cdfs but continuous support (like spike-and-slab and hurdle distributions).

Me neither. I hadn't considered carefully representing mixed continuous/discrete distributions as even being a possibility, but it would be pretty great. It seems like it would also allow things like `MvNormal`s with positive semidefinite covariance, which could help for working with low-rank models.

Have you seen such general representation of distributions in other systems? It would be helpful to know what trouble people have had to be sure we're learning from their mistakes.

richardreeve:
> Ok, so this addresses the problem of singularities. That makes sense. Say you have a mixture like this, and `d` is in the support of the discrete component, while `c` is outside it. Does your implementation return
> 
> ```julia
> pdf(mix, d) == Inf
> pdf(mix, c) == continuousWeight * pdf(continuousComponent, c)
> pmf(mix, d) == discreteWeight * pmf(discreteComponent, d)
> pmf(mix, c) == 0
> ```
> 
> ? Is that what it "should" return, or am I thinking about it wrong?

Yes, that's exactly what I was thinking of! `SpikeSlab` and `CompoundDistribution` in this PR implement this in fact, though I haven't put any testing in yet to check as I got distracted by creating #945.

> > For cumulative densities this isn't a problem, but I don't see a way around having two functions for the density and mass functions if you want to generally be able to describe the density and the mass of distributions with discontinuous cdfs but continuous support (like spike-and-slab and hurdle distributions).
> 
> Me neither. I hadn't considered carefully representing mixed continuous/discrete distributions as even being a possibility, but it would be pretty great. It seems like it would also allow things like `MvNormal`s with positive semidefinite covariance, which could help for working with low-rank models.

Great, I hadn't thought of that. Obviously it's important that this extension to the type hierarchy is as extensible as possible - though it doesn't necessarily have to be able to handle everything on day one. At the moment, I'm not at all that confident that the `UnionSupport` type that I suggest is the right way to go, so I'm open to other suggestions?

> Have you seen such general representation of distributions in other systems? It would be helpful to know what trouble people have had to be sure we're learning from their mistakes.

Unfortunately not. It's only having the ability to work on `Distributions.jl` in Julia directly that gave me the idea that I might be able to do this at all...

cscherrer:
I guess to be really rigorous about it we should be working (eventually) in terms of the [refinement to Lebesgue's decomposition](https://en.wikipedia.org/wiki/Lebesgue%27s_decomposition_theorem).

richardreeve:
I may have to take your word for that!

richardreeve:
In the meanwhile, it would be good to know what you think of this first step...

cscherrer:
I think it's definitely a step in the right direction. And if it's really non-breaking, I say go for it. But I'm having trouble seeing how this can be non-breaking. If we have mixed discrete/continuous distributions, then 
```julia
pdf(mix, d) == Inf
pdf(mix, c) == continuousWeight * pdf(continuousComponent, c)
pmf(mix, d) == discreteWeight * pmf(discreteComponent, d)
pmf(mix, c) == 0
```

seems sensible. But say we have some simple example of this, like `mix` is a mixture of a standard `Normal()` and a `Bernoulli(0.5)`. Then we'd have 
```julia
pdf(mix, 0.0) == Inf
pdf(Bernoulli(0.5)) == 0.5
```

This would seem to imply that `pdf(Normal(),0.0) == Inf`, which of course isn't right. The problem is that we're implicitly switching base measures.

richardreeve:
You're right, but it's non-breaking in the sense that everything behaves exactly as it did before except for new distributions, which obviously can't be breaking however you code them. And up till now it hasn't been possible to make these mixed distributions at all, so I'm safe...

My feeling is that we ought to move to a system where pdf and pmf are distinct though, and so the `const pdf(d::CountableDistribution) = pmf(d)` compatibility bit should be dropped, for exactly the reason you bring up.

The only other option I can think of would be to have a third (mostly optional) argument to `pdf`, that allowed you to choose the type of density to return. By default it would return the appropriate density:

```julia
pdf(d::CountableDistribution, x) = pdf(d, x, Mass())
pdf(d::ContinuousDistribution, x) = pdf(d, x, Density())
pdf(::ContinuousDistribution, _, ::Mass) = 0.0
pdf(d::CountableDistribution, x, ::Density) = insupport(d, x) ? Inf : 0.0
```

But this would not be defined for mixed distributions, so you'd have to select the appropriate density type explicitly. I don't have any feeling for whether that's a good idea, but it would allow us to maintain backward compatibility indefinitely...

PR #942, opened 2019-07-29
Optional checks of the constructor parameters

matbesancon:


matbesancon:
@ararslan the warnings I get are odd:

```julia
struct Beta{T<:Real}
    a::T
end

function Beta(a::T) where {T <: Real}
    # do stuff
    return Beta{T}(a)
end
```

Should be ok and in fact is in the REPL, but yields warning in the package

matbesancon:
It looks like the `Beta{T}(...)` constructor is over-written, when it should not be

src/univariate/continuous/arcsine.jl
```diff
@@ -23,18 +23,19 @@ External links
...
 """
 struct Arcsine{T<:Real} <: ContinuousUnivariateDistribution
     a::T
     b::T
-
-    Arcsine{T}(a::T, b::T) where {T} = (@check_args(Arcsine, a < b); new{T}(a, b))
 end
 
-Arcsine(a::T, b::T) where {T<:Real} = Arcsine{T}(a, b)
+Arcsine(a::T, b::T) where {T} = (@check_args(Arcsine, a < b); new{T}(a, b))
```
ararslan: `new` is not valid in constructors defined outside of the type definition block.

ararslan:
Hm, does it work if you define an explicit inner constructor in addition to these methods? For example,
```julia
struct Arcsine{T<:Real} <: ContinuousUnivariateDistribution
    a::T
    b::T

    Arcsine{T}(a::T, b::T) where {T<:Real} = new{T}(a, b)
end

Arcsine(...) = ...
```

matbesancon:
Indeed it supresses the warning, I'm re-adding the inner constructors, a fair amount of boilerplate but it doesn't seem to be possible otherwise

matbesancon:
Should be good now

src/univariate/continuous/arcsine.jl
```diff
@@ -23,18 +23,20 @@ External links
...
 end
 
-Arcsine(a::T, b::T) where {T<:Real} = Arcsine{T}(a, b)
+Arcsine(a::T, b::T) where {T <: Real} = (@check_args(Arcsine, a < b); Arcsine{T}(a, b))
+Arcsine(a::T, b::T, ::NoArgCheck) where {T} = Arcsine{T}(a, b)
+
 Arcsine(a::Real, b::Real) = Arcsine(promote(a, b)...)
 Arcsine(a::Integer, b::Integer) = Arcsine(Float64(a), Float64(b))
-Arcsine(b::Real) = Arcsine(0.0, b)
+Arcsine(b::T) where {T <: Real} = Arcsine(zero(T), b)
```
bkamins: I am not sure this is an OK definition, as later we e.g. define: `kurtosis(d::Arcsine{T}) where {T<:Real} = -T(3/2)`, which will fail if we write `Arcsine(1)`.
A general question is if for continuous distributions, in cases like this one, we should not add restriction for parameters to be `AbstractFloat`?

bkamins: or add a conversion if parameters are `<:Integer` to float?

src/univariate/continuous/beta.jl
```diff
@@ -29,18 +29,22 @@ External links
...
+    return Beta{T}(α, β)
+end
+
+function Beta(α::T, β::T, ::NoArgCheck) where {T<:Real}
+    return Beta{T}(α, β)
 end
 
-Beta(α::T, β::T) where {T<:Real} = Beta{T}(α, β)
 Beta(α::Real, β::Real) = Beta(promote(α, β)...)
 Beta(α::Integer, β::Integer) = Beta(Float64(α), Float64(β))
```
bkamins: I am not sure if using `float` instead of `Float64` is not better (the difference is minimal - only probably for `BigInt` argument in practical cases).

matbesancon: good point indeed, these hardcoded Float64 never felt right

src/univariate/continuous/biweight.jl
```diff
@@ -46,7 +51,7 @@ function ccdf(d::Biweight{T}, x::Real) where T<:Real
     u = (d.μ - x) / d.σ
     u <= -1 ? one(T) :
     u >= 1 ? zero(T) :
-    (u + 1)^3/16 * @horner(u,8,-9,3)
+    (u + 1)^3/16 * @horner(u, 8, -9,3)
```
bkamins:
```suggestion
    (u + 1)^3/16 * @horner(u, 8, -9, 3)
```

src/univariate/continuous/arcsine.jl
matbesancon: `Arcsine(1)` will dispatch to `Arcsine(a::Integer, b::Integer)`:
```julia
julia> kurtosis(Arcsine(1))
-1.5
```

Seems fine?

matbesancon: > or add a conversion if parameters are <:Integer to float?

there is one on line 38

src/univariate/continuous/exponential.jl
```diff
@@ -67,7 +71,8 @@ xval(d::Exponential, z::Real) = z * d.θ
 
 pdf(d::Exponential, x::Real) = (λ = rate(d); x < 0 ? zero(λ) : λ * exp(-λ * x))
 function logpdf(d::Exponential{T}, x::Real) where T<:Real
-    (λ = rate(d); x < 0 ? -T(Inf) : log(λ) - λ * x)
+    λ = rate(d)
+    x < 0 ? -T(Inf) : log(λ) - λ * x
```
bkamins: This will fail if `T<:Integer`

src/univariate/continuous/arcsine.jl
bkamins: OK, I have not noticed it, sorry. It should probably use `float` instead of `Float64` as noted below.

src/univariate/continuous/exponential.jl
matbesancon: Which cannot happen by construction:
```julia
Exponential(θ::Integer) = Exponential(Float64(θ))
```

bkamins: OK - I see that we have a conversion to float in the constructor

src/univariate/continuous/beta.jl
matbesancon: I did this in the last commit

matbesancon:
For the record, @bkamins approved

nalimilan:
Couldn't we use something like `@noargcheck` (similar to `@inbounds`) or `argcheck=false` (leaving inter-procedural optimization statically eliminate the checks)? That `NoArgCheck()` system does sound quite unusual to me.

matbesancon:
> That NoArgCheck() system does sound quite unusual to me.

Agreed on not seeing the pattern quite often, but the two advantages I saw:
Using vanilla dispatch, thus simplicity and familiarity. Using vanilla dispatch, thus static at construction. The default remains checking arguments, but there is an opt-out method.

> @noargcheck (similar to @inbounds)

`inbounds` is tricky, a dispatch solution is easier to scope: when I call `@inbounds do_stuff(inner_func())`, where does the inbound propagate, etc.

One other possibility is having as mentioned a `argcheck=false`, but introduces a branching, if not elimitated, it can be a bother when creating lots of distributions

nalimilan:
> One other possibility is having as mentioned a `argcheck=false`, but introduces a branching, if not elimitated, it can be a bother when creating lots of distributions

I think it will always be eliminated when `false` is written as a literal.

PR #943, opened 2019-07-29
Allow AbstractVector for mean of MvTDist

kleinschmidt:
In my continuing quest to use static arrays in distributions, this widens the
mean of GenericMvTDist to be any `<:AbstractVector`, and adds a type parameter
to indicate the actual type.  I didn't widen the "convenience constructor"
methods, so anyone not calling the (un-exported) `GenericMvTDist` constructor
directly shouldn't be affected by this.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -27,7 +27,8 @@ Calculus = "49dc2e85-a5d0-5ad3-a950-438e2897f1b9"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Calculus", "Distributed", "ForwardDiff", "JSON", "Test"]
+test = ["Calculus", "Distributed", "ForwardDiff", "JSON", "StaticArrays", "Test"]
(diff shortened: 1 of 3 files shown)


matbesancon:
@kleinschmidt cool idea! The tests still seem to be failing

kleinschmidt:
Well would you look at that....

kleinschmidt:
Okay the problem was that I hadn't redefined the aliases `MvTDist` etc. to include the mean vector type parameter, so the roundtrip test `dd = typeof(d)(params(d)...)` was failing because the only other constructor that takes type parameters is the inner constructor that takes all kinds of stuff besides the params *phew*.

So tests should pass but the whole thing feels a little...icky to me, since if anyone actually tries to USE an "exotic" vector type for the mean then they won't be able to round-trip in the same way.  Anyway I've added a test to demonstrate this behavior so you can see what I mean...

matbesancon:
Is it linked to the comment:
```julia
 # depends on PDMats#101 (merged but not released) 
```

kleinschmidt:
No that's just using a static matrix for the covariance.  The problem is that:

```julia
julia> d = GenericMvTDist(1., @SVector([1., 2]), PDMat([3. 2; 2 2]));

julia> typeof(d)
GenericMvTDist{Float64,PDMat{Float64,Array{Float64,2}},SArray{Tuple{2},Float64,1,2}}

julia> typeof(d) <: MvTDist
false
```

That is, `typeof(d)` returns a parametric `GenericMvTDist`, and all the external constructors for `GenericMvTDist` don't have type parameters; only the `const` aliases like `MvTDist` are parametric. I'm not sure why/when that happened; it could maybe be fixed by adding a method like

```julia
GenericMvTDist{T,Cov,Mean}(df, Sigma, mu) where {T,Cov,Mean} = GenericMvTDist(covert(T, df), convert(Cov, Sigma), convert(Mean, mu))
```

But I'll have to see if that works...

kleinschmidt:
okay with that method tests pass locally (including the round-trip test that I'd marked as broken) and I don't really see a downside.

matbesancon:
Looks good, thanks!

PR #944, opened 2019-07-30
Add rank

johnczito:
Make `rank` a standard method for matrix variate distributions.

Similar in spirit to `size` and `length`, `rank(d::MatrixDistribution)` returns the rank of each sample from `d`. Right now all of `Distributions.jl`'s matrix variate distributions are non-singular, so this functionality is a bit boring. But if/when we implement some singular matrix variate distributions (whose samples are reduced-rank matrices), then this will be more obviously useful.

This already exists for `MatrixNormal`, `MatrixTDist`, `MatrixBeta`, and `MatrixFDist`. This PR adds it for `Wishart` and `InverseWishart` and lists it in the docs. I am also in favor of re-exporting `rank` in the same way that we re-export `mean`, `var`, etc from `Statistics.jl`, but I know that this is more controversial, so for now I haven't done that.

### DIFF
--- a/src/matrix/inversewishart.jl
+++ b/src/matrix/inversewishart.jl
@@ -56,6 +56,7 @@ insupport(d::InverseWishart, X::Matrix) = size(X) == size(d) && isposdef(X)
 dim(d::InverseWishart) = dim(d.Ψ)
 size(d::InverseWishart) = (p = dim(d); (p, p))
 size(d::InverseWishart, i) = size(d)[i]
+rank(d::InverseWishart) = dim(d)
 params(d::InverseWishart) = (d.df, d.Ψ, d.c0)
 @inline partype(d::InverseWishart{T}) where {T<:Real} = T
 
--- a/docs/src/matrix.md
+++ b/docs/src/matrix.md
@@ -19,6 +19,7 @@ All distributions implement the same set of methods:
 ```@docs
 size(::MatrixDistribution)
 length(::MatrixDistribution)
+rank(::MatrixDistribution)
 mean(::MatrixDistribution)
 pdf{T<:Real}(d::MatrixDistribution, x::AbstractMatrix{T})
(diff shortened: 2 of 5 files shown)


matbesancon:
looks good to me

matbesancon:
@johnczito also I'm in favor of not re-exporting, so that people don't get in scope things they don't use

src/matrix/inversewishart.jl
```diff
@@ -56,6 +56,7 @@ insupport(d::InverseWishart, X::Matrix) = size(X) == size(d) && isposdef(X)
 dim(d::InverseWishart) = dim(d.Ψ)
 size(d::InverseWishart) = (p = dim(d); (p, p))
 size(d::InverseWishart, i) = size(d)[i]
+rank(d::InverseWishart) = dim(d)
```
ararslan: Unless `rank` has been `import`ed, this will need a module qualification in order to extend the binding from LinearAlgebra rather than defining a new local binding.

johnczito: It's imported:

https://github.com/JuliaStats/Distributions.jl/blob/91b9df8ab88fb08aabc4294b7dfcbb2ee1c7ae07/src/Distributions.jl#L12

ararslan: Okay perfect.

ararslan:
I'm generally in favor of reexporting things that are extended, but I don't care that much in this case.

matbesancon:
thanks @johnczito

johnczito:
Sorry, hold on. The docstring is not currently displaying. Without re-exporting, you need to list `Distributions.rank(::MatrixDistribution)` in `docs/.../matrix.md`. Can I still push that here?

matbesancon:
Woops, I merged a bit hastily. Sure you can push here and re-open a PR

PR #945, opened 2019-07-30
New type hierarchy for ValueSupport

richardreeve:
Allows non-Int, non-Float64 eltypes for countable and continuous support respectively, and discontinuous distributions in general.


richardreeve:
I've just stripped type hierarchy out of #941 without any improvements to the distributions as requested by @matbesancon. Do feel free to stick with the original PR if it's manageable on its own! May be of interest to @mschauer also.

docs/src/types.md
```diff
@@ -41,8 +41,10 @@ The `ValueSupport` sub-types defined in `Distributions.jl` are:
 
 **Type** | **Element type** | **Descriptions**
 --- | --- | ---
-`Discrete` | `Int` | Samples take discrete values
-`Continuous` | `Float64` | Samples take continuous real values
+`DiscreteSupport{T}` | `T` | Samples take any discrete values
+`Discrete = DiscreteSupport{Int}` | `Int` | Samples take `Int` values
+`ContinuousSupport{T <: Number}` | `T` | Samples take continuous values
+`Continuous = ContinuousSupport{Float64}` | `Float64` | Samples take continuous `Float64` values
```
matbesancon: Should we use a `T <: Real` for Continuous?

src/common.jl
```diff
@@ -50,13 +61,14 @@ Base.size(s::Sampleable{Multivariate}) = (length(s),)
...
+    eltype(::ValueSupport)
 
 The default element type of a sample. This is the type of elements of the samples generated
 by the `rand` method. However, one can provide an array of different element types to
 store the samples using `rand!`.
 """
-Base.eltype(s::Sampleable{F,Discrete}) where {F} = Int
-Base.eltype(s::Sampleable{F,Continuous}) where {F} = Float64
+Base.eltype(::Sampleable{F, <: ValueSupport{N}}) where {F, N} = N
+Base.eltype(::ValueSupport{N}) where N = N
```
matbesancon:
```suggestion
Base.eltype(::ValueSupport{N}) where {N} = N
```

src/common.jl
```diff
@@ -87,15 +99,27 @@ const MultivariateDistribution{S<:ValueSupport} = Distribution{Multivariate,S}
 const MatrixDistribution{S<:ValueSupport}       = Distribution{Matrixvariate,S}
 const NonMatrixDistribution = Union{UnivariateDistribution, MultivariateDistribution}
```
matbesancon: as mentioned on #941, this seems not required nor robust

src/common.jl
```diff
@@ -87,15 +99,27 @@ const MultivariateDistribution{S<:ValueSupport} = Distribution{Multivariate,S}
...
+const DiscreteMultivariateDistribution =
+    CountableMultivariateDistribution{Discrete}
+const ContinuousMultivariateDistribution = MultivariateDistribution{Continuous}
+
+const CountableMatrixDistribution{C<:CountableSupport} = MatrixDistribution{C}
+const DiscreteMatrixDistribution = CountableMatrixDistribution{Discrete}
+const ContinuousMatrixDistribution = MatrixDistribution{Continuous}
+
 
 variate_form(::Type{Distribution{VF,VS}}) where {VF<:VariateForm,VS<:ValueSupport} = VF
```
matbesancon:
```suggestion
variate_form(::Type{Distribution{VF, <:ValueSupport}}) where {VF<:VariateForm} = VF
```

matbesancon:
minor comments, otherwise this will be good with some tests

matbesancon:
alright, we'll freeze this one and wait for #945 to be merged, then the diff will be much smaller

src/common.jl
richardreeve: As I mentioned, this is not my code. There's lots of things I might change if I had freedom of action, but I didn't want to break anything as this is exported.

matbesancon: this PR is going to be breaking anyway, I think we can remove this, it is defined and exported but not used anywhere else in the code base.

richardreeve: I have gone to a lot of trouble to make it non-breaking - even with respect to `pdf()` still working with discrete distributions - but I'm willing to make it breaking if you think that's really okay?

matbesancon: Breaking `NonMatrixDistribution` seems ok to me. That's why Distributions is still not 1.0, we are still allowing ourselves to fix the type system. If you want, you can add a deprecation for it

docs/src/types.md
richardreeve: No, because we can have continuous distributions over things like complex numbers and Unitful numbers. It would be nice to have a continuous trait and have a constraint based on that, but this would be quite complex to implement, and can anyway be added in after this basic framework is up and running.

matbesancon: "`T` is the `eltype` of the distribution", without specifying a subtype? This would allow other floating point, bignum, etc

richardreeve:
Was that last message intended for #941 @matbesancon? In the meanwhile I've added some very minor fixes to distributions (and tests) that are currently overriding the default behaviour because they use non-Float64's in continuous distributions.

matbesancon:
> Was that last message intended for #941 @matbesancon?

Yes absolutely sorry!

src/common.jl
richardreeve: No, it's fine. I'll delete it!

matbesancon:
Thanks for appveyor good catch

matbesancon:
@richardreeve just did, you can merge master into your branch :)

richardreeve:
Yeah, sorry - I was just trying to keep everything synced between these different PRs and then I realised that I couldn't see the code coverage because appveyor was failing... anyway, I'm now pretty sure that this is complete. The only other thing I could add to this branch is switching pdf() -> pmf() for all of the discrete distributions.

matbesancon:
Cool, let's see what the coverage is saying

docs/src/types.md
richardreeve: I'm not quite sure what you're suggesting here by the way? Is this good as is?

richardreeve:
Aside from the comment you made above (that I still don't understand!), is this good to go?

matbesancon:
Sorry for the misunderstanding, I was thinking `pdf -> pmf` can go in the next PR, like #941 

I'm going to have a last review and let others voice their opinion, but looks good yup :)

richardreeve:
Ah, no - I understood the comment about pmf() - it was the comment about the docs above that I didn't understand.

matbesancon:
Ah ok sorry, it's about not forcing it to `Float64`, but leaving other "continuous" types (other floats and other continuous-like real types)

richardreeve:
Okay, right. The new "default" is to use `ContinuousSupport{T <: Number}`, but just for backward compatibility I have `Continuous = ContinuousSupport{Float64}` since most distributions are hard coded with `Float64`s.

At some point it would be nice to generalise this, but it's non-trivial. In particular being about to extend to allow Unitful numbers for many continuous distributions, which is the first thing I'd like to do, requires us to do detailed type checking since `typeof(μ) ≠ typeof(σ²)`, for instance.

matbesancon:
having this alias may create a barrier to these generic re-writes, (which we very much want to achieve). I would use `ContinuousSupport{Float64}` where it needs hard-coding only, otherwise people will tend to use `Continuous`, shorter names tend to be preferred.

richardreeve:
It would be a major breaking change to get rid of `Discrete` and `Continuous` (and `DiscreteUnivariateDistribution`, `ContinuousUnivariateDistribution`, etc.) completely. I'm not opposed to it - I think we ought to be much more flexible - but I imagine everyone who has ever written a distribution of their own will stop working.

It feels like a separate PR to do this to me, just because I feel like it's more of a political issue, where as this PR is a practical one to enable functionality in a non-breaking way...

matbesancon:
Agreed

docs/src/types.md
richardreeve: We've agreed to leave this for now.

matbesancon:
@richardreeve can you merge master into this branch?

richardreeve:
Okay, that's done. It would be nice to get this merged. The only question I have come up with is whether to separate out the CountableSupport with integer type from the specific ones whose support is `low:high`... as a subtype? There's a strong presumption in the code that anything with integer support has no gaps...

Apart from that, I've now got a third(!) PR ready which eliminates all reference to `Discrete` and `Continuous` and all of the `Discrete[*]Distribution` stuff, and parameterises `Continuous[*]Distribution` as `Continuous[*]Distribution{T}` for the floating point type. There were lots of distributions that misreported their eltype it turns out. It is obviously hugely breaking, but that can be mitigated by reintroducing those terms (though they will not be inside the package), but `ContinuousDistribution{T}` will still be breaking and obviously it also won't force people to correct their code...

richardreeve:
Since there was a bit of time(!), I've now added in `ContiguousSupport` as a subtype of `CountableSupport`, as it stops `DiscreteNonParametric` and `Categorical` acquiring inappropriate contiguous traits if they have integer support.

src/functionals.jl
```diff
@@ -18,6 +18,12 @@ function expectation(distr::DiscreteUnivariateDistribution, g::Function, epsilon
     sum(x -> f(x)*g(x), leftEnd:rightEnd)
 end
 
+function expectation(distr::CountableUnivariateDistribution,
+                     g::Function, epsilon::Real)
+    f = x->pdf(distr,x)
+    sum(x -> f(x)*g(x), support(distr))
```
oxinabox: How about just
```
sum(support(distr)) do x
    pdf(distr,x) * g(x)
end
```

src/mixtures/mixturemodel.jl
```diff
@@ -125,7 +125,7 @@ the components given by ``params``, and a prior probability vector.
 If no `prior` is provided then all components will have the same prior probabilities.
 """
 function MixtureModel(::Type{C}, params::AbstractArray) where C<:Distribution
-    components = C[_construct_component(C, a) for a in params]
+    components = [_construct_component(C, a) for a in params]
```
oxinabox: Why was this changed?

richardreeve: Because `C` is often not used in a parametrised way (e.g. `Normal`, since the eltype can be inferred from `params`), using `C[]` resulted in the code generating a vector of an abstract type. Just creating the vector creates a concrete type though, as all elements are the same.

src/functionals.jl
richardreeve: Sure - I just copied from the existing code, but I'm happy to change it. Is there some reason to choose this?

oxinabox: Not imensemsly it should all inline

oxinabox:
This to me feels like the minimal nonbreaking way to do this.
We should merge this, while we think about more comprehensive (and breaking) solutions.

richardreeve:
Thanks for the suggestion @cscherrer - I also suggested something more complex like this on slack - it wasn’t DomainSet but I can’t remember what it was now - and there was opposition because it would add a dependency. It looks great for updating the continuous support side of things, which I haven’t actually touched (as far as possible except getting some eltypes right), but it seems less relevant to the countable, finite, and discontinuous support I was worried about. 

I agree with @oxinabox though - DomainSet could be added later as a broad solution to the former problem, but in the meantime it will be nice to merge this as I’ve got a few distributions that I’d like to add that depend on it.

cscherrer:
Yes @richardreeve I agree too. This is definitely a step in the right direction, and has the benefit of being non-breaking. I think the pmf distinction causes more problems than it does, but looks like that's another discussion.

Thanks for your work on this!

richardreeve:
The `pmf` changes that you're referring to (that I propose in #941) are also non-breaking, at least at the moment, because `[log]pdf` is aliased to `[log]pmf` for discrete distributions. Something like that is essential for continuous distributions with Dirac delta functions in them though, otherwise `pdf` would just be `Inf` at the spikes, with no actual mass assigned to them... anyway, any comments you have on that would be very welcome across on #941!

richardreeve:
Bump @mschauer...

mschauer:
Sorry for coming late to the party. First, what is the "support" in terms of this PR? If I have a density which is `1/2` on `[0,1]` and `[2,3]`, and `0` everywhere else, so also `0` on , `(1,2)`, where am I in this system?

richardreeve:
Hi @mschauer. At the moment nearly all distributions in `Distributions.jl` have a continuous domain with a `Float64` eltype, and have a continuous cdf within that domain, or they have a discrete contiguous domain of `Int64` eltype. However most of the distributions don't actually behave like this - in particular most of them are quite flexible about their eltype. What this PR does is to try to fix problems with the eltype, by parameterising `ValueSupport` on the eltype, and to try to extend the discrete domains so they don't have to be numeric, never mind contiguous integers. It also allows the distributions to have discontinuous cdfs by allowed mixed distributions, such as "Spike and Slab" distributions, examples of which you can see ready to merge in #941.

It does not try to address the issue of discontinuous *domains* - you might be able to do that using the `UnionSupport` type it introduces, but there is no other support for it other than being able to define the support. The reason it doesn't try to do this is that (as far as I'm aware) there are no PRs or issues to do with the inability of `Distributions.jl` to handle discontinuous domains, whereas there are several about failures to handle different kinds of discrete and mixture distributions. @cscherrer has suggested ways of generalising this still further [above](https://github.com/JuliaStats/Distributions.jl/pull/945#pullrequestreview-270478894) and in #941, but this would add dependencies or make the changes breaking, and I was trying to avoid either.

Does that make sense?

mschauer:
I think there is really a difference between:

1. the type of the random variable outcomes and 
2. the range of values a random variable will take with certain probability and 
3. the reference measure for the density. 

For example a Normal(0,1) and a spike and slab mixture of a Dirac 
mass at zero and a standard normal take exactly the same values, but one has a density and the other not and both of them might produce a variety of different floating point values.

cscherrer:
@mschauer yes this is my concern as well. I really like the idea of being able to represent such general distributions, and it looks like @richardreeve has done a great job with it. What leads me to hesitate is that this library is central for anyone doing probabilistic work in Julia.

Once we have spike and slab, why not low-rank normals? Or a Dirichlet process? I'd **love** to be able to represent these kinds of things, and I expect others will want them as well. That's great, but we have to do it right, and I don't yet see what "doing it right" would look like.

My sense is this will involve the [Lebesgue decomposition theorem](https://en.wikipedia.org/wiki/Lebesgue%27s_decomposition_theorem) and [Radon-Nikodym derivatives](https://en.wikipedia.org/wiki/Radon%E2%80%93Nikodym_theorem). Maybe it's a list of pairs of a base measure and a density? And `pmf` and `pdf` would be special cases of this?

Anyway... I don't expect to solve this in one PR. And I'm certainly not opposed to this one. I just hope we can put a little thought into where we'd like this library to be in a few years, to make sure every step forward takes us closer to this end-goal.

richardreeve:
Sorry. My previous message doesn’t actually make a huge amount of sense now I’m looking back on it - I’m blaming it on being laid up ill in bed. If you want to understand what’s going in here, I recommend #941. This was requested to be split off from it to simplify the PR, but maybe the rationale has been lost in the meanwhile. I just want to create some distributions that Distributions does not allow me to define.

mschauer:
>  I just want to create some distributions that Distributions does not allow me to define.

100% this, this is in my mind too. I am just not sold on any specific way to get there. I had the hope to abolish an hierarchy all-together and just have `Distribution{T}` with `T` the type of the samples.

mschauer:
> Maybe it's a list of pairs of a base measure and a density? 

Sorry to also get all technical here, but I think about making the Radon-Nikodym reference measure a trait to keep things slim. There are even situations in high dimensions where you would like to use a different reference measure for a multivariate Gaussian than the multivariate Lebesgue measure even if the distribution has of course a Lebesgue density. It should just be a property of the pdf, maybe a signature `pdf(d::Distribution, g::ReferenceMeasure=referencemeasure(d))` could do the trick.

richardreeve:
I’d be delighted if you guys came up with a general solution, but in the meanwhile these PRs fix bugs and extend functionality in a non-breaking way...

cscherrer:
Thanks @richardreeve for your patience with this. From looking back over this, I think I may have been confounding these two PRs. 

Is it correct to say that this PR only makes an extension of the `ValueSupport`, with no effect on pmf or pdf? And that idea is in #941?

richardreeve:
Correct.

richardreeve:
From @mschauer's comment above, it sounds to me that something more like my [suggestion](https://github.com/JuliaStats/Distributions.jl/pull/941#issuecomment-518018428) in #941 on handling multiple `pdf`s rather than introducing `pmf` may be the way forward for that, given the similarity between the two?

mschauer:
Yes, this is the same! If the reference measure is `Lebesgue` then the pdf is a `Density` in your case and if the reference measure is the counting measure the pdf is a `Mass` (probability mass function). 

I think with such an device it is sufficient to define one additional "support" besides the one we have currently named say `SampleType{T}` which just states what the type of the samples is, but which states nothing about the reference measure (allowing e.g. a random variable returning (float) 1/2 and -1/2 with probability 1/2 each. )

richardreeve:
Okay, I'll reimplement #941 with just `pdf()` then. No problem. However, I don't think that your single `SampleType{T}` is sufficient for many reasons if I understand it correctly:

First, the current `Continuous` type assumes `Float64`, which is already wrong for many distributions, and also assumes continuity in the support and the cdf, which may not be true for new distributions (including the ones I'm trying to introduce and your example above). Parameterising `Continuous` makes more sense to me that using `SampleType{T}` for discrete and continuous support. And discontinuous cdfs with continuous support should be handled separately, because you end up with singularities at specific values of the support, which need to be handled specially, and there's currently no way of dispatching on this.

Second, `Discrete` assumes contiguous `Int64`s for the support - the former is again wrong for many existing distributions, and both are wrong for `DiscreteNonParametric` and `Categorical` already. If `Discrete` is parameterised, then it can already handle the discrete version of the `SampleType{T}` case if I understand you correctly.

Finally, hurdle-type (mixture) distributions, which have a point mass at 0 (normally), and then any other distribution that doesn't include 0 in its support (as opposed to zero-inflated, where 0 is in the support of the other distribution) have neither continuous nor discrete support (necessarily) in the sense currently used, but would fit within the `UnionSupport` type that I'm proposing here.

Anyway, that's my feeling - it's very likely that I'm misunderstanding you, but I think `SampleType{T}` would just be a fudge that *to avoid immediate breaking changes to `Distributions.jl`* would have to encompass non-contiguous countable support, non-continuous uncountable support, non-continuous cdfs, and a bunch of other special cases, which doesn't seem like a good mix...

mschauer:
The thing is, when do we actually need to now something about discrete, continuous, contiguous, support etc?

If you check https://juliastats.github.io/Distributions.jl/stable/extends/#Univariate-Distribution-1 
then except `pdf` and `logpdf` the  first type `support` type parameter of a univariate` Distribution` does not contain necessary information beyond the type of the samples, making it a `SampleType{T}` as I propose covers that.

richardreeve:
Ah, now I understand where all of this is coming from! I'd never seen those docs before - I was just basing the API on what was actually in the code. I'm afraid that documentation is only correct insofar as other functions are provided automatically ***and incorrectly in most cases*** by reference to the `ValueSupport` type you profess to have. For instance `eltype`, `support`, and I'm sure there are others, but those are the most egregiously wrong at the moment. This is one of the things I'm fixing with this PR...

richardreeve:
PS and obviously I believe implementation of default `support` (and indeed `insupport`) and `eltype` functions requires the kind of information I'm proposing, as will a default implementation of the `referencemeasure` function you propose.

richardreeve:
PPS Also `expectation`, `kldivergence`...

mschauer:
Really, you can just implement `support` and friends without using the predefined magic. For example for a Dirac measure
```
struct SampleType{T} <: ValueSupport end
....
Base.eltype(s::Sampleable{F,SampleType{T}}) where {F,T} = T
...

"""
    Dirac(x)

Distribution where the value `x` has probability `1`.

P(X = z) = \\begin{cases}
1 & \\quad \\text{for } z = x, \\\\
0 & \\quad \\text{for } z ≠ x.
\\end{cases}
"""
struct Dirac{T<:Real} <: Distribution{Univariate,SampleType{T}}
    x::T

    Dirac{T}(x::T) where {T <: Real} = new{T}(x)
end

function Dirac(x::T) where {T <: Real}
    return Dirac{T}(x)
end

support(d::Dirac) = [d.x] # could make that an interval too
minimum(d::Dirac) = d.x
maximum(d::Dirac) = d.x
insupport(d::Dirac, x) = d.x == x
...

rand(rng::AbstractRNG, d::Dirac) = d.x

expectation(d::Dirac, g::Function) = g(d.x)
```

It even says in the file `univariates.jl` about the support:
```
Generic fallback methods are provided, but it is often the case that `insupport` can be
done more efficiently, and a specialized `insupport` is thus desirable.
You should also override this function if the support is composed of multiple disjoint intervals.
```

richardreeve:
Absolutely. You could require everyone to implement all of these functions for every distribution. Obviously so far even inside this package they have failed to do this, and doing so (getting people to repeat the same code over and over again with tiny differences) is anyway a famously bad design strategy that leads to subtle bugs. But apart from anything else it isn’t how this package was written.  It was written with a bunch of fallback methods, and all I’m trying to do is to get them to work correctly. I think you believe this is a philosophical difference and I suspect we’re therefore not going to agree on it.

I see it as a purely pragmatic issue - I believe my solution is most in keeping with how the package is currently written, best demonstrated by the fact that I believe it is completely non-breaking, it fixes a lot of bugs and extends the functionality of the package. I think you believe your solution is where the package ought to be headed, and I don’t disagree with that but it’s irrelevant because doing this doesn’t hinder your long term solution. The only reason I’m not implementing a trait based approach here is because there isn’t a standard trait implementation yet in Julia and I think it would be confusing until there is. 

I implemented all of the thread-safe random number generation because I needed it and it worked, and I don’t understand why this can’t be accepted on the same basis?

mschauer:
I get that it is frustrating to have this discussion now, but please bear with me. I am worried about the following: we are having type parameters to the `Distribution` type which are too specific. Not all distributions do fit into the binary scheme `Discrete` vs `Continuous`. 

Now we could try to make the type parameter more complicated to cover all edge cases,
but that appears to me as a mistake of the "not working yet? try more of the same"-category. The problem stems from having these type parameters in the first place. Adding an opt-out type parameter, `NonOfTheAbove()` or `SampleType{T}` if you want, at least give a developer the possibility escape that. 

You replace 
```
	struct Discrete   <: ValueSupport end
	struct Continuous <: ValueSupport end
	abstract type ValueSupport{N} end
```
by a considerably more complicated hierarchy
```
	struct ContinuousSupport{N <: Number} <: ValueSupport{N} end
	abstract type CountableSupport{C} <: ValueSupport{C} end
	struct ContiguousSupport{C <: Integer} <: CountableSupport{C} end
	struct UnionSupport{N1, N2,
	                    S1 <: ValueSupport{N1},
	                    S2 <: ValueSupport{N2}} <:
	                        ValueSupport{Union{N1, N2}} end
	
	const Discrete = ContiguousSupport{Int}
	const Continuous = ContinuousSupport{Float64}
	const DiscontinuousSupport{I, F} =
	    UnionSupport{I, F, <: CountableSupport{I},
	                 ContinuousSupport{F}} where {I <: Number, F <: Number}
	const Discontinuous = DiscontinuousSupport{Int, Float64}
```

I just would like to make sure this is what we want.

> You could require everyone to implement all of these functions for every distribution. 

I am not suggesting that. The system in place covers most of the cases, but not all.  I suggest that handling the *remaining* cases separately might be a good idea.

src/multivariate/mvtdist.jl
```diff
@@ -103,7 +103,6 @@ logdet_cov(d::GenericMvTDist) = d.df>2 ? logdet((d.df/(d.df-2))*d.Σ) : NaN
 
 params(d::GenericMvTDist) = (d.df, d.μ, d.Σ)
 @inline partype(d::GenericMvTDist{T}) where {T} = T
-eltype(::GenericMvTDist{T}) where {T} = T
```
matbesancon: Why remove the eltype here?

matbesancon:
The way I see it is that this PR is extending the type parameters not for a "nicer" conceptual structure, but from the empirical experience we have from trying to implement some distributions that were fundamentally broken (Dirac) with the previous structure. In that sense, it's a much safer approach, it's like we had these broken tests before that we were just hiding under the carpet (Dirac and friends) and that can now be modeled without cringing of making them `Continuous`.

mschauer:
I am not in favour of the status quo.

src/multivariate/mvtdist.jl
richardreeve: Because it's handled by the default case now. In `src/common.jl`:
```julia
Base.eltype(::Sampleable{F, <: ValueSupport{N}}) where {F, N} = N
```

richardreeve:
>> You could require everyone to implement all of these functions for every distribution.
>
> I am not suggesting that. The system in place covers most of the cases, but not all. I suggest that handling the remaining cases separately might be a good idea.

Unfortunately you are and you don't realise it. The system in place covers almost no distributions (with the honourable exception of `GenericMvtDist` above that @matbesancon noticed!), because most are parameterised by their eltype but don't report that they are (and there are other problems that tend to be restricted to individual distributions). Short of fixing this redundantly on a case by case basis, we need a fallback that works instead. My proposal of including the eltype as a parameter of the `ValueSupport` fixes this problem immediately for all cases, and avoids any future confusion.

Furthermore you are not comparing like with like when you say I am massively expanding the complexity of the system. I have added in some const aliases for backward compatibility, but in #951 I remove them all from use in the package, and potentially from the package as a whole if we are not concerned about distributions outside the packages breaking.

What I actually do here is replace `Continuous` and `Discrete` with `ContinuousSupport{T<:Number}` and `ContiguousSupport{T<:Integer}`, which serve the same purpose *but do not falsely assert that the eltype is `Float64` and `Int`*, respectively. In the latter case, the name is also clearer about the assumption in `Discrete` that the numbers are contiguous integers. I then introduce `CountableSupport{T}` as a supertype of `ContiguousSupport{T}` for the situation where the element are not integers or not contiguous, *which is currently not covered by the type system*. Finally I add in something for discontinuous distributions, which is the bit I came in to fix before I realised how many other problems there were! This is four types replacing two, fixing the first two that were already broken, and adding two rather than your proposed one because the two non-covered cases are not related and so it would make no sense to merge them:

```julia
struct ContinuousSupport{N <: Number} <: ValueSupport{N} end
abstract type CountableSupport{C} <: ValueSupport{C} end
struct ContiguousSupport{C <: Integer} <: CountableSupport{C} end
struct UnionSupport{N1, N2,
	            S1 <: ValueSupport{N1},
	            S2 <: ValueSupport{N2}} <:
    ValueSupport{Union{N1, N2}} end
```

I actually believe there is an interesting discussion to be had here about what the best way is to handle discontinuous distributions, and some of that conversation has gone on in #941 with @cscherrer. I can only restate that this is a temporary solution, but I feel it's *temporary until traits are included into base Julia*, which may be some way off, and so we should try to make the temporary solution as good as possible.

richardreeve:
For those interested, the conversation I've been having in my head while this has been playing out is whether the more elegant temporary solution would be this:

```julia
abstract type UncountableSupport{N <: Number} <: ValueSupport{N} end
struct ContinuousSupport{N <: Number} <: UncountableSupport{N} end
abstract type CountableSupport{C} <: ValueSupport{C} end
struct ContiguousSupport{C <: Integer} <: CountableSupport{C} end
```
`UncountableSupport` would then be for non-continuous cdfs (though it would obviously include continuous ones as a subtype). The advantage of this is the reduction in complexity introduced by the `UnionSupport` type, the disadvantage is that compound distributions would have to say exactly what they were - two contiguous distributions would be contiguous, two continuous ones would be continuous, and a continuous and a contiguous would be uncountable, for instance. That may be a reasonable price to pay though... ?

src/multivariate/mvtdist.jl
matbesancon: Got it thank you

mschauer:
You write
> which serve the same purpose but do not falsely *assert that the eltype is Float64 and Int*,

To be sure, why do you think that it is not sufficient to define `eltype` depending on the type parameter of the actual distribution (and therefore not to not interpret `Continuous` as necessarily `Float64`?)

richardreeve:
Because if you read `src/common.jl`, that’s what it says:

```julia
Base.eltype(s::Sampleable{F,Discrete}) where {F} = Int
Base.eltype(s::Sampleable{F,Continuous}) where {F} = Float64
```

and only one or two distributions like `GenericMvtDist` ever bother to override this!

mschauer:
Well, lets override this then!

richardreeve:
In the meanwhile, if @matbesancon or @cscherrer have thoughts about whether the suggestion [above](https://github.com/JuliaStats/Distributions.jl/pull/945#issuecomment-520482570) is a better alternative that would be great.

matbesancon:
I tend to prefer `UnionSupport`, types describing items by negative (Non-something) tend to be more confusing IMO

richardreeve:
Okay, I see your point, but as I’m sure you know uncountable is a technical term about the cardinality of the set, so although it is a negation, it is also fully a word in its own right.

mschauer:
Please comment why it is insufficient to add the correct `eltype` declarations at the individual distribution level (and extend `ValueSupport` only as far as needed to cover the missing cases such as `Dirac`-`Normal` mixtures.)

richardreeve:
[Already done](https://github.com/JuliaStats/Distributions.jl/pull/945#issuecomment-520069881) in response to the last time you suggested this.

matbesancon:
I think we should move forward with this issue, as it is non-breaking and fixing some pain we have experienced.

mschauer:
I would like to react to this first, can we agree on a time-line when this is merged?

matbesancon:
> can we agree on a time-line when this is merged?

Let us say in one week from now?

mschauer:
That would be great, thank you a lot.

richardreeve:
Why? We’ve already had a week and we’ve achieved nothing except that I now have very little time or energy left to finish this work. I introduced this PR to improve the package, as the several thousand other lines of code that I’ve contributed to this and Distributions.jl, which have added functionality, fixed bugs and added improved testing. Why are we waiting any longer?

matbesancon:
There will likely be no release in a week from now, I still need to fix the "no-argument-check" thing, the PR I created introduces too much complexity, fixing it will not happen before then

richardreeve:
That’s fine, but there are two more PRs before this is all done - #941 and maybe #951...

matbesancon:
#951 being breaking, it would be preferable in a separate release later. #941 introduces new features, so same, it should maybe come with a new release

mschauer:
Talking about it, I think PR this will break 

https://github.com/baggepinnen/LowLevelParticleFilters.jl/blob/5617653967a35677d218f56c1c93cf28f208d88f/src/utils.jl#L116

https://github.com/alan-turing-institute/MLJBase.jl/blob/f4047a84f005e18d53d9440593bedfa7b339d5c0/src/distributions.jl#L30

https://github.com/richardreeve/Phylo.jl/blob/dcd4f9b332a4e3c88fc91a0a33021d961cafb71c/src/rand.jl#L9

I am not sure about

https://github.com/brian-j-smith/Mamba.jl/blob/b91816da7f368d24a932f08cf73d7dfd759dbf38/src/distributions/extensions.jl#L38

matbesancon:
This PR or #941?

richardreeve:
I can do #951 in a non-breaking way if you think it’s a desirable... it would basically involve removing things like `ContinuousUnivariateDistribution` as well as `Continuous` from the package code and using `UnivariateDistribution{ContinuousSupport{T}}` instead everywhere. Then keeping `Continuous` and `ContinuousUnivariateDistribution` but relegating them to `src/deprecates.jl` even though they can’t actually be deprecated and removing them from the documentation, and then it could be included immediately.

I actually think that’s desirable as there are too many const aliases knocking around, but it’s not currently implemented...

mschauer:
> This PR or #941?

This one, because subtyping `<: ValueSupport` instead of now `ValueSupport{T}` will give an `invalid subtyping in definition of ...` error:

```
julia> using Phylo
[ Info: Precompiling Phylo [aea672f4-3940-5932-aa44-993d1c3ff149]
ERROR: LoadError: LoadError: invalid subtyping in definition of Phylogenetics
```

richardreeve:
Fair enough. I hadn’t spotted this. Phylo is my package so easily fixed, but in any event I can easily fix this by changing `ValueSupport` to a different name - `SupportType` probably - and adding another const alias.

matbesancon:
>  I can easily fix this by changing ValueSupport to a different name - SupportType probably - and adding another const alias.

It sounds reasonable yes. Maybe we could deprecate `ValueSupport` for people to use `SupportType{T}` instead in that case? (maybe `Support` is better than `SupportType` here)

mschauer:
Well, it is a PR which substantially changes the internals, so some breakage is rather expected. This is just what I found when looking. So much from myself now.

richardreeve:
Yes, I was thinking about `Support` - I was only hesitating because short exported names always make me a bit nervous in case there are name clashes with other packages...

matbesancon:
@richardreeve did you have time to do the changes?

richardreeve:
Hi @matbesancon - I was actually walking in the alps last week (must stop checking github when I get a signal!), so not yet, but I'm back now and I'll get it done tonight!

src/common.jl
```diff
@@ -67,10 +80,11 @@ into an array, depending on the variate form.
 nsamples(t::Type{Sampleable}, x::Any)
 nsamples(::Type{D}, x::Number) where {D<:Sampleable{Univariate}} = 1
 nsamples(::Type{D}, x::AbstractArray) where {D<:Sampleable{Univariate}} = length(x)
-nsamples(::Type{D}, x::AbstractVector) where {D<:Sampleable{Multivariate}} = 1
+nsamples(::Type{D}, x::AbstractArray{<:AbstractVector}) where {D<:Sampleable{Multivariate}} = length(x)
```
mschauer: Is this an unrelated change?

src/functionals.jl
```diff
@@ -13,9 +13,17 @@ end
...
     (leftEnd, rightEnd) = getEndpoints(distr, epsilon)
-    sum(x -> f(x)*g(x), leftEnd:rightEnd)
+    sum(leftEnd:rightEnd) do x
+        pdf(distr,x) * g(x)
+    end
+end
+
+function expectation(distr::CountableUnivariateDistribution,
+                     g::Function, epsilon::Real)
+    sum(support(distr)) do x
```
mschauer: Does this make sense, this is a loop over typically 2ˆ64 elements?

src/univariate/discrete/discretenonparametric.jl
```diff
@@ -156,8 +156,10 @@ end
 
 minimum(d::DiscreteNonParametric) = first(support(d))
 maximum(d::DiscreteNonParametric) = last(support(d))
-insupport(d::DiscreteNonParametric, x::Real) =
+insupport(d::DiscreteNonParametric{T}, x::T) where {T} =
     length(searchsorted(support(d), x)) > 0
+insupport(d::DiscreteNonParametric{T}, x::Number) where {T<:Number} =
+    length(searchsorted(support(d), convert(T, x))) > 0
```
mschauer: I would prefer `support` to do the right thing here and not `insupport` to second guess the needed transformation

src/functionals.jl
richardreeve: This is just a slight editing of the original code:
```julia
    f = x->pdf(distr,x)
    (leftEnd, rightEnd) = getEndpoints(distr, epsilon)
    sum(x -> f(x)*g(x), leftEnd:rightEnd)
```
I don't have a strong opinion about whether it's sensible.

mschauer: You are dropping the epsilon

src/common.jl
richardreeve: I was fixing the handling of numbers throughout the code and I spotted that multivariate was missing this function - this isn't a deletion... the edited code is in the next line:
```julia
nsamples(::Type{D}, x::AbstractVector{<:Number}) where {D<:Sampleable{Multivariate}} = 1
```

mschauer:
FYI: I still have the "postponed by a week" in mind and will come back later today.

mschauer:
One thing, to be clear, what would you consider the set of features I would have to reproduce when I would like to replace this PR by my own attempt? Besides making the additional tests pass?

src/univariate/discrete/discretenonparametric.jl
richardreeve: I'm not sure what you expect `support` to do here? It's just returning the support, and then we're checking for presence of the value in that set.

richardreeve:
No, that's not it unfortunately - the problem with having split this off from #941 is that the new tests, which will emerge from the new distributions, are not included. This PR provides a mechanism for supporting non-Int, non-Float64 eltypes for countable and continuous support respectively, and discontinuous distributions in general, in a consistent and comprehensible way. It does this whilst keeping consistency with the old code base so there are no breaking changes. The tests changed here mostly just change the test type hierarchy to match the new type hierarchy.

richardreeve:
**Note: This PR is now largely superseded by #951**, which includes all of these changes, but also totally removes all references to the old const aliases from the codebase, and in so doing removes several new bugs that weren't identified here. This means that anyone referring to the code to understand how to write a new distribution will see the new API in action. I would strongly recommend merging that rather than this. Although it adds a few hundred lines of code, they are nearly all just replacing old aliases and replacing with more explicit types.

src/functionals.jl
richardreeve: Ah, I see. That's because this is for non-`ContiguousSupport` types - they are not necessarily ordered, so we can't use `getEndpoints()`. At the moment, the only type that falls into this category is `DiscreteNonParametric`, but it will also include `Dirac`, neither of which are expected to have anything like 2^64 elements..

PR #948, opened 2019-08-01
fix rank doc

johnczito:


matbesancon:
@johnczito this should be `LinearAlgebra.rank` I think

johnczito:
That only works if you add `using LinearAlgebra` to `docs/make.jl`. But it's the same as this (in my local build, anyway).

matbesancon:
`import LinearAlgebra` in the build doc would do it yes, I find it clearer to know where the functions are defined

johnczito:
Agreed. In the built docs it still displays `LinearAlgebra.rank`, not `Distributions.rank`. I figured loading `LinearAlgebra` into `make.jl` would be overkill. But if you also want it to read `LinearAlgebra.rank` in the doc source, then I'm happy to do it.

matbesancon:
If the doc displays correctly then it's fine

matbesancon:
thanks!

Issue #921, opened 2019-07-03
MvNormal aliases should have looser type constraints


nickrobinson251:
Is there a reason for these to be ` Float64` rather than `Real`?

https://github.com/JuliaStats/Distributions.jl/blob/432545afd3a29397e2308a99a4978b588676c671/src/multivariate/mvnormal.jl#L180-L186

It leads to stuff like this:
```julia
julia> z1 = MvNormal([1.0, 1.0])
ZeroMeanDiagNormal(
dim: 2
μ: [0.0, 0.0]
Σ: [1.0 0.0; 0.0 1.0]
)

julia> z2 = MvNormal([1, 1])
MvNormal{Int64,PDiagMat{Int64,Array{Int64,1}},Distributions.ZeroVector{Int64}}(
dim: 2
μ: [0, 0]
Σ: [1 0; 0 1]
)
```

PR #951, opened 2019-08-04
Change to ValueSupport interface

richardreeve:
This change to the interface removes `Discrete` and `Continuous` from the code as aliases for `ContiguousSupport{Int}` and `ContinuousSupport{Float64}` - introduced in #945 - from a suggestion by @matbesancon, and to be done after or instead of #945, which it currently includes. All of the `Discrete[*]Distribution` and `Continuous[*]Distribution` are removed from the code too, as there was concern expressed in #945 that I was introducing too many type aliases. This is done because most of the distributions that use these aren't actually constrained to have `Int` or `Float64` as their `eltype, and there were problems with breaking some external packages.

This could be made breaking by removing `Discrete` and `Continuous` and the `*Distribution` aliases from `src/deprecates.jl` where they have been moved to, but this wouldn't affect the `Distributions` package as it no longer uses them at all (except in `test/types.jl`). Keeping them in might discourage people from fixing their code. Unfortunately I can't see how to deprecate them as they are used as const aliases for types, and `typeof(DeprecatedThing)` isn't a type...

~~The PR would still be mildly breaking even then, though, as `ContinuousSupport` becomes `ContinuousSupport{T<:Number}`, and this can't be avoided unless someone has a synonym for continuous. I'm not sure how many people use `ContinuousSupport` outside the package though - I can't find it on github in any julia code. (`UncountableSupport{T}` is an option, but at the moment `ContinuousSupport` requires support to be over a single continuous domain...)~~

Edit: I'd forgotten that I introduced `ContinuousSupport` in #945!


richardreeve:
This PR is no longer breaking as it stands, because code has been added in to `src/deprecates.jl` that provides all of the missing types again. However the package itself no longer uses these "deprecated" const aliases.

mschauer:
Thank you, this is already nicer. This is also much closer in spirit to what I have in mind -- but again we do not need to be breaking: we can just avoid making the `eltype` of the distributions part of the `ValueSupport` parameter, c.f. #958. This is important. Needlessly enforcing type restrictions breaks automatic differentiation, static arrays, and others. Making them part of the abstract super type is the same kind of mistake. We have learned this in the past, see #957, #921, #943, #896 and the entire bunch of related issues/prs.

As I keep arguing, `eltype` is just sufficient to capture the sample type. No need to redo everything with `Support{T}` if `Support` and `eltype` are reach enough. I have not seen convincing arguments to the contrary. I am not against extending the system by `Discrete`, `ContiguousIntegerSupport`, `Mixture` or whatever is needed.

richardreeve:
Can I stress that this is not breaking, and also that it has been here for weeks - #945 was just the type hierarchy split out. I'll look at the other issues later when I have time.

richardreeve:
As far as I can see, your example problems above are cases where people have previously made exactly the class of errors I am proposing to eliminate (#896) - I have fixed 10s of these bugs in this PR - or are fixing times when people have got their types confused. I certainly don't see anything I'm doing as needlessly enforcing type restrictions - I don't believe I add any more restrictions, though it is possible I have inadvertently - I am just correctly parameterising on the types already used to avoid this class of error arising.

There's a lot more work to be done here incidentally. What I'd like to be able to do down the line is to seamlessly handle any non-`Real` type in Continuous distributions - at the moment the code will not handle correctly when `x` and `x^2` do not have the same units - e.g. `Unitful` units, so that I can just say that a distribution has a mean of `1.0m` and a variance of `2.0m^2`. The problem is I believe this either needs something like [computed field types](https://github.com/vtjnash/ComputedFieldTypes.jl), or traits, neither of which are currently in base Julia.

mschauer:
I do not want `T` to be part of the abstract type here. 

One example. It would be the same as if the current implementation would be

```
deriv(::typeof(sin), x) = cos(x)
```
or equivalently, before we had function types
```
abstract type Functor
end
struct Sin <: Functor
end
deriv(::Sin, x) = cos(x)
```
and you would propose to change it by
by
```
abstract type NewFunctor{T}
end
const Functor = NewFunctor{Float64} # this is a breaking change in the interface
struct Sin{T} <: Functor{T} # and this too
end
deriv(::Sin{T}, x) = cos(convert(T, x))
```

richardreeve:
I think you're ignoring the fact that most of these distributions are already parameterised on their eltypes - e.g.:

```julia
struct LogNormal{T<:Real} <: ContinuousUnivariateDistribution
```
All I'm doing is connecting that parameterisation up to their supertypes. Where are the abstract supertypes being used in a way that causes this kind of problem?

richardreeve:
And this is not what I've done:
```julia
abstract type NewFunctor{T}
end
const Functor = NewFunctor{Float64} # this is a breaking change in the interface
struct Sin{T} <: Functor{T} # and this too
end
deriv(::Sin{T}, x) = cos(convert(T, x))
```
Where the type is already parameterised - i.e. `Sin{T}` already present - I've just done this:
```julia
abstract type NewFunctor{T}
end
struct Sin{T} <: NewFunctor{T} # not breaking
end
# deriv(::Sin{T}, x) = cos(convert(T, x)) # I don't know why you need to do this
deriv(::Sin{T}, x::T) = cos # I would propose this, but I don't know what you're trying to achieve here
```
Where the type is **not** already parameterised, I've done this:
```julia
abstract type NewFunctor{T}
end
const Functor = NewFunctor{Float64} # **not** a breaking change in the interface
struct Sin <: Functor # also **not** breaking
end
deriv(::Sin, x) = cos(x) # unchanged
```

mschauer:
I am ignoring nothing. Currently we can choose to make `T` part of the type, after your change we are forced.  By the way I have done exactly this type of mistake in 
```
https://github.com/mschauer/Bridge.jl/blob/master/src/types.jl#L23
```
and I am slowly getting rid of all code using the `T` in functions instead the super type instead of the type of the arguments. The analogue of
```
const Continuous = NewContinuous{Float64}
const Functor = NewFunctor{Float64} # respective
```
*is* a breaking change, because before the meaning of `Continuous` was fuzzy (does it mean `Float64` or does it mean the existence of a pdf) and now you are cementing a particular interpretation of it (it means continuous with and only with sample type `Float64`) and that will get in our way, as it got in my way before.)

richardreeve:
I completely agree that Continuous was fuzzy - or as I like to say it *usually misrepresenting what is going on* because of the fallback assumption of Float64 - and *I am fixing that bug*. You have still failed to give any examples where there is a problem since your previous examples misstated how my code works.

mschauer:
This is a breaking change. You are breaking
```
https://github.com/JuliaMath/RandomMatrices.jl/blob/603b9e9acd95133839e804d6ef068cdfee4cbbfb/src/densities/Semicircle.jl#L9
```
and other things.

richardreeve:
How am I breaking this? It was already breaking itself because it was incorrectly reporting its eltype. The fact that so many people have got this wrong is the absolute proof that we ought not to do things in your "make a comment in the source code" way, but instead should do it properly.

richardreeve:
This is pointless by the way - as in #945 you are determined that it's okay to let people silently fail in their own code, and I am equally certain that this is shockingly bad software design. Can @matbesancon or someone just make a decision because I'm bored of repeating the same arguments over and over again.

mschauer:
Can you stop being personal?

richardreeve:
I'm sorry that you feel that way, but I'm afraid don't believe that I am being personal. If your position is not as I characterise is then I'd be happy to be corrected, but I'm simply pointing out that this discussion is not going anywhere because we hold two fundamentally different philosophical positions that I am trying to articulate clearly so someone can make a decision about how to go forward.

I understand your philosophical position, but I think it is trumped by my pragmatic concerns that it has so far resulted in *lots* of bugs in this package. I am still fixing more of them as we speak incidentally, but it's slow and boring work, all of which arises from the existing "fallback `Float64`" assumption.

mschauer:
I am also not so very happy with a "I'm sorry that you feel that way"-response, but let's just try to get this discussion on the right track again.

mschauer:
I took some time thinking and I can live with this, conditional on acknowledgment that this is a substantial change of Distributions and should be introduced carefully and not within a minor version update or rushed. We also have to look at the dependences.

I do think we are reducing the space of things which people can express as `<: Distribution" by this, but this is ok if we do it consciously.

I do not think we need to define `ContiguousSupport{<:Int}`, just returning a (possibly truncated) range for `support(x, eps = 0.0)` with  `eps` a truncation parameter will achieve effectively the same and keep the type system tidier. Otherwise downstream will need some `::Union{Distribution{Univariate, DiscreteSupport{T}},Distribution{Univariate,ContiguousSupport{S})} where {T,S}` where it is now `::DiscreteUnivariateDistribution`.

richardreeve:
I'm happy to be guided on how breaking it is - you even spotted that an earlier version of the PR was breaking on one of my own packages! - that's why I contacted you in the first place to ask for your feedback on #945. I'm still not sure whether the latest version of this PR is breaking anything that correctly implements the Distributions interface, but I completely accept that no-one actually implements that interface, so it may be a bigger problem than I'm able to tell. It will certainly be catastrophically breaking if we remove all of these type aliases - I believe there are hundreds of packages that use some of them - but it'll probably be much less breaking if we don't... ?

I want to stress that I've found hundreds of bugs like #960 in how Distributions handle eltypes inside the package now I'm looking for it in more detail, which is honestly quite shocking. I've been pretty cynical about this, but this is much worse than I appreciated! I've added a few preliminary tests, and it's going to be slow going. I've pushed the first few tests (without their fixes for now) into this PR so you can see what's happening. It may be this should be another PR tbh.

richardreeve:
I'm not sure I understand your point about `ContiguousSupport` by the way. Do you realise that `ContiguousSupport <: CountableSupport`? You mention `DiscreteSupport`, but I assume you mean `CountableSupport`, so a union would be unnecessary.

Incidentally, despite how it might appear, I'm not remotely committed to this solution. I didn't want to make such a big change - I just pulled on a thread and it has all unravelled! I'd be delighted if we could cause fewer problems, but I just don't see it at the moment - there are real problems in here.

richardreeve:
I'd also be interested to know how we are "reducing the space of things which people can express as `<: Distribution`" - I absolutely don't want to do that if at all possible.

richardreeve:
Okay, there are definitely 100s of bugs in eltype handling, and they spread into other packages. I've fixed `Normal` (I think!) as an example - it can now handle even `Unitful` numeric parameters - but fixing them all needs to be put into one (or more!) other PRs. If this works as expected, I think it's about ready to merge though...

matbesancon:
@richardreeve what is now the expected answer to the test removed in 0d66c20?

Could we edit the test instead of removing it?

richardreeve:
Sorry, I meant to comment on that, but it was a bit late when I submitted this stuff and I forgot. I don't know what the right answer is here. On master, as the test suggests, `partype(Normal(1//1, 2//1)) == Rational{Int}`, but `partype(Normal(1, 2)) == Float64`.

The latter is correct, since `partype(Normal(m, s)) == eltype(Normal(m,s))`, it would be a lot of work to break that relationship, and we can't have a rational eltype for a continuous random number. As a result of countability checking I've introduced, `Normal(1//1, 2//1)` currently fails. This is perhaps not ideal, especially as it does introduce a breaking change, but it's a breaking change that definitely doesn't do anything useful at the moment, as the Normal distribution over the rationals doesn't work right anyway...

One option is to leave it like this (for now), one is to introduce a specific case for rational numbers and normal distributions (which seems desperately inelegant), and the final one is to say that all countable parameters are promoted to floating point. My feeling is that the latter is the right approach, but it's best achieved using traits, and I was really hoping some trait-based system would be added to base Julia before I did anything like this...

What do you think?

matbesancon:
I think we could leave the `partype` as the types provided (`Rational{Int}` here), while `eltype` itself needs generalizing (to Float64 in this case)

richardreeve:
I’m not sure that’s wise. First, this would be a major breaking change to nearly every distribution - nearly all of them assume that their partype and their eltype are the same. Second, it would become impossible to describe what your eltype was (without adding a type argument, as it wouldn’t be the same as the partype input), and would require type conversion all through the code whenever there was a mismatch.

matbesancon:
I understand the point. So `Rational{Int}` would be converted to Float64 in the params?

richardreeve:
I could add that as a temporary fix. At the moment it fails.

matbesancon:
Yes it would be better not to get errors

mschauer:
> I'm happy to be guided on how breaking it is

Yes, of course, I try to keep keeping an eye on this topic.

richardreeve:
Thanks. @matbesancon - that’s done now, and I think this is about as good as this PR (replacing #945) is going to get, otherwise it'll get enormous. The net coverage increases, and I’m not aware of any breaking effects. There are an unbelievable number of bugs in the eltype handling still outstanding, which you can get an idea about in `test/eltypes.jl` - starting on line 41 the distributions present are *at best* Float64 only despite most permitting any eltype) but it’s going to be a lot of work to fix them, so I think it’s best to merge this now, then get #941 merged, and then fix all of the remaining issues (like #960) later...

richardreeve:
Bump.

matbesancon:
@richardreeve can you get the last bit of coverage up? I'm not always a fan of it but it helps to keep healthy habits for such large code base

src/utils.jl
```diff
@@ -9,6 +9,38 @@ macro check_args(D, cond)
     end
 end
 
+isuncountable(x) = false
```
matbesancon: I would be for adding only one of `isuncountable`, `iscountable`, given that one is the negation of the other.
`iscountablereal` should also not be necessary, and can rely on `iscountable` and type dispatch

richardreeve:
The overall coverage does go up. The problem with the patch coverage is that there is code in the PR that isn’t used yet, because the actual distributions that use it are in the next PR! The easiest way of increasing coverage for that is to add in the new distributions :)

matbesancon:
@richardreeve can you fix the two conflicts? The PR that created them was tiny boilerplate, so it shouldn't be a big deal

matbesancon:
Nevermind, I fixed the two conflicts

src/deprecates.jl
```diff
@@ -40,4 +40,19 @@ for fun in [:pdf, :logpdf,
...
+
+const DiscreteDistribution{F<:VariateForm}   = Distribution{F,Discrete}
+const ContinuousDistribution{F<:VariateForm} = Distribution{F,Continuous}
+
+const DiscreteUnivariateDistribution     = Distribution{Univariate,    Discrete}
+const ContinuousUnivariateDistribution   = Distribution{Univariate,    Continuous}
+const DiscreteMultivariateDistribution   = Distribution{Multivariate,  Discrete}
+const ContinuousMultivariateDistribution = Distribution{Multivariate,  Continuous}
+const DiscreteMatrixDistribution         = Distribution{Matrixvariate, Discrete}
+const ContinuousMatrixDistribution       = Distribution{Matrixvariate, Continuous}
```
nickrobinson251: Can we use `Base.@deprecate_binding` for these?

matbesancon: good idea thanks, it preserves compatibility but signals what is changing

richardreeve: Cool. I had no idea that existed - I see it isn't exported in fact, so I don't feel too bad. Anyway, I'll definitely do that.

richardreeve: I notice that it doesn't handle parametric types, so I can't do it for all of the deprecated types:

```julia
const DiscreteDistribution{F<:VariateForm}   = Distribution{F,Discrete}
const ContinuousDistribution{F<:VariateForm} = Distribution{F,Continuous}
```

but I've covered most of them now. However, testing them is another matter:

```julia
julia> @test_deprecated DiscreteMatrixDistribution
WARNING: Distributions.DiscreteMatrixDistribution is deprecated, use Distribution{Matrixvariate, Discrete} instead.
  likely near REPL[6]:1
Log Test Failed at REPL[6]:1
  Expression: $(Expr(:escape, :DiscreteMatrixDistribution))
  Log Pattern: (:warn, r"deprecated"i, Ignored(), :depwarn) match_mode = :any
  Captured Logs:

```

Is the message returned by `@test_deprecated` not a deprecation warning?

src/utils.jl
richardreeve: Actually, there's a problem there.

Nearly all distributions cannot currently handle not-`Float64` eltypes, but will be able to handle `<: Real` eltypes with only a modest amount of work. However, it's much harder to get them to handle (un)countable non-real `<: Number` eltypes, such as Unitful types. Nonetheless, the aspiration should be to handle all of these types, so I want to be able to move easily from the `iscountablereal(T)` to `iscountable(T)` checks which have been created (`@check_countable()`, etc.).

Similarly, `iscountable(T)` is not just `!isuncountable(T)`. Non-`Number` types are rejected by both for now, as they are well beyond the scope of the current series of PRs. I could call them `iscountablenumbner()` and `isuncountablenumber()` to make it clear that they are not inverses?

richardreeve: PS The other option I thought of (as you suggest) was to have type dispatch on the macro for the time being (which is the only user of `isuncountablereal()`, and this function is also not exported incidentally), but I assume that's still not possible?

richardreeve: Okay, I've got rid of this conflict - I've added three level logic - `missing` - for the cases where we haven't yet resolved whether it's countable or uncountable, then the two are inverses of each other.

richardreeve:
OK @matbesancon. We’re now ahead on all metrics and These issues are dealt with...

src/univariate/continuous/normal.jl
```diff
@@ -83,19 +85,19 @@ entropy(d::Normal) = (log2π + 1)/2 + log(d.σ)
 
 Computes the x-value based on a Normal distribution and a z-value.
 """
-function xval(d::Normal, z::Real)
+function xval(d::Normal{T}, z::Real) where {T}
     if isinf(z) && iszero(d.σ)
         d.μ + one(d.σ) * z
```
mschauer:
```suggestion
        d.μ + oneunit(d.σ) * typeof(one(T))(z)
```

src/univariate/continuous/normal.jl
```diff
@@ -70,10 +72,10 @@ mode(d::Normal) = d.μ
 
 var(d::Normal) = abs2(d.σ)
 std(d::Normal) = d.σ
-skewness(d::Normal{T}) where {T<:Real} = zero(T)
-kurtosis(d::Normal{T}) where {T<:Real} = zero(T)
+skewness(d::Normal{T}) where {T} = zero(one(T))
```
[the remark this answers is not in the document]
mschauer: Ah, we should define `zeronounit(T) = zero(one(T))` internally

src/univariate/continuous/normal.jl
```diff
@@ -83,19 +85,19 @@ entropy(d::Normal) = (log2π + 1)/2 + log(d.σ)
...
 end
 """
     zval(d::Normal, x::Real)
 
 Computes the z-value based on a Normal distribution and a x-value.
 """
-zval(d::Normal, x::Real) = (x - d.μ) / d.σ
+zval(d::Normal, x::Number) = (x - d.μ) / d.σ
 
 gradlogpdf(d::Normal, x::Real) = -zval(d, x) / d.σ
```
mschauer: should be unitless? how to strip the unit?

src/univariate/continuous/normal.jl
```diff
@@ -165,30 +167,30 @@ invlogcdf(d::Normal, lp::Real) = xval(d, norminvlogcdf(lp))
...
         elseif isone(p)
-            Inf
+            T(Inf)
         else
-            0.5
+            T(0.5)
         end
     end
-    xval(d, -erfcinv(2p) * sqrt2)
+    T(xval(d, -erfcinv(2p) * sqrt2))
```
mschauer:
```suggestion
    xval(d, -erfcinv(2p) * sqrt2)
```
`xval` should do the right thing

src/univariate/continuous/normal.jl
```diff
@@ -296,7 +298,8 @@ cf(d::Normal, t::Real) = exp(im * t * d.μ - d.σ^2 / 2 * t^2)
 
 #### Sampling
 
-rand(rng::AbstractRNG, d::Normal{T}) where {T} = d.μ + d.σ * randn(rng, T)
+rand(rng::AbstractRNG, d::Normal{T}) where {T} =
+    d.μ + d.σ * randn(rng, typeof(one(T)))
```
mschauer: `typeof(one(T))`... we need internal functions making the intention (stripping the unit) clear

src/univariates.jl
```diff
@@ -376,36 +395,41 @@ function cdf(d::DiscreteUnivariateDistribution, x::Integer, ::Type{FiniteSupport
     return c
 end
 
-cdf(d::DiscreteUnivariateDistribution, x::Real) = cdf(d, floor(Int,x))
-cdf(d::ContinuousUnivariateDistribution, x::Real) = throw(MethodError(cdf, (d, x)))
-
+cdf(d::UnivariateDistribution{<:ContiguousSupport}, x::Real) =
+    cdf(d, floor(Int,x))
```
mschauer: Should use `T` instead of `Int`? Otherwise a problem if using a `Int64` distribution on a `Int == Int32` system.
```suggestion
    cdf(d, floor(T, x))
```

src/univariates.jl
```diff
@@ -376,36 +395,41 @@ function cdf(d::DiscreteUnivariateDistribution, x::Integer, ::Type{FiniteSupport
...
 """
     ccdf(d::UnivariateDistribution, x::Real)
 
 The complementary cumulative function evaluated at `x`, i.e. `1 - cdf(d, x)`.
 """
-ccdf(d::UnivariateDistribution, x::Real) = 1.0 - cdf(d, x)
-ccdf(d::DiscreteUnivariateDistribution, x::Integer) = 1.0 - cdf(d, x)
-ccdf(d::DiscreteUnivariateDistribution, x::Real) = ccdf(d, floor(Int,x))
+ccdf(d::UnivariateDistribution, x) = 1.0 - cdf(d, x)
+ccdf(d::UnivariateDistribution{<:ContiguousSupport}, x::Integer) =
```
mschauer: Method not needed?

src/univariates.jl
```diff
@@ -376,36 +395,41 @@ function cdf(d::DiscreteUnivariateDistribution, x::Integer, ::Type{FiniteSupport
...
     logcdf(d::UnivariateDistribution, x::Real)
 
 The logarithm of the cumulative function value(s) evaluated at `x`, i.e. `log(cdf(x))`.
 """
-logcdf(d::UnivariateDistribution, x::Real) = log(cdf(d, x))
-logcdf(d::DiscreteUnivariateDistribution, x::Integer) = log(cdf(d, x))
-logcdf(d::DiscreteUnivariateDistribution, x::Real) = logcdf(d, floor(Int,x))
+logcdf(d::UnivariateDistribution, x) = log(cdf(d, x))
+logcdf(d::UnivariateDistribution{<:ContiguousSupport}, x::Integer) =
+    log(cdf(d, x))
```
mschauer: method not needed?

src/univariates.jl
```diff
@@ -439,10 +463,13 @@ invlogccdf(d::UnivariateDistribution, lp::Real) = quantile(d, -expm1(lp))
 
 # gradlogpdf
 
-gradlogpdf(d::ContinuousUnivariateDistribution, x::Real) = throw(MethodError(gradlogpdf, (d, x)))
+gradlogpdf(d::UnivariateDistribution{<:ContinuousSupport}, x::Real) =
+    throw(MethodError(gradlogpdf, (d, x)))
```
mschauer: Is this method needed?

src/univariates.jl
```diff
@@ -522,7 +551,7 @@ loglikelihood(d::UnivariateDistribution, X::AbstractArray) = sum(x -> logpdf(d,
 
 macro _delegate_statsfuns(D, fpre, psyms...)
     dt = eval(D)
-    T = dt <: DiscreteUnivariateDistribution ? :Int : :Real
+    T = eltype(dt)
```
mschauer: Before `T` was a symbol, does this work as well?

src/utils.jl
```diff
@@ -9,6 +9,34 @@ macro check_args(D, cond)
     end
 end
 
+isuncountable(x) = missing
```
mschauer: Doc string?

src/univariate/continuous/normal.jl
```diff
@@ -27,21 +27,25 @@ External links
...
+        @check_uncountable(Normal, μ)
+        @check_args(Normal, σ >= zero(σ))
+    end
     return Normal{T}(μ, σ)
 end
 
 #### Outer constructors
-Normal(μ::Real, σ::Real) = Normal(promote(μ, σ)...)
+Normal(μ::Number, σ::Number) = Normal(promote(μ, σ)...)
+Normal(μ::Rational, σ::Rational) = Normal(float(μ), float(σ))
```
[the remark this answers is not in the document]
mschauer: Ah, we have a constraint: the mean and the eltype of the distribution need to agree.  Makes sense for a `Normal`, it can neither be sampled nor the pdf evaluated in `Rational`.

test/utils.jl
```diff
@@ -10,7 +10,7 @@ r = RealInterval(1.5, 4.0)
 
 @test partype(Gamma(1, 2)) == Float64
 @test partype(Gamma(1.1, 2)) == Float64
-@test partype(Normal(1//1, 2//1)) == Rational{Int}
+@test partype(Normal(1//1, 2//1)) == Float64
```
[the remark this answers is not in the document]
mschauer: Ah, we have a constraint: the mean and the eltype of the distribution need to agree. Makes sense for a Normal, it can neither be sampled nor the pdf evaluated in Rational.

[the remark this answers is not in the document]
nickrobinson251: Worth adding that as a comment?

matbesancon:
@richardreeve I updated the branch to fix conflicts, can you check @mschauer's comments above and make the corresponding modifications?

richardreeve:
I'll try to get it done as soon as possible, but this is my busiest time of the year for teaching - the summer was when I had time to work on this unfortunately.

src/utils.jl
```diff
@@ -9,6 +9,63 @@ macro check_args(D, cond)
...
+                                       ": the type of variable ",
+                                       $(string(var)), " (",
+                                       string(typeof($(esc(var)))),
+                                       ") is not countable.")))
+        end
+    end
+end
+
+
+## a type to indicate zero vector
```
matbesancon: @richardreeve I messed up my conflict merge, could you remove things related to ZeroVector in this file, it was deleted in:
https://github.com/JuliaStats/Distributions.jl/pull/1020

Issue #939, opened 2019-07-23
use Kronecker.jl for `kron`?


nickrobinson251:
https://github.com/MichielStock/Kronecker.jl

johnczito:
I have a PR in mind for `PDMats.jl` that would create `kron(A::PDMat, B::PDMat) -> PDMat` so that we can get rid of things like this:

https://github.com/JuliaStats/Distributions.jl/blob/d00c780d680993519cf3dd0ce2b71dadd477afcf/src/matrix/matrixnormal.jl#L125

What other use cases do you have in mind?

nickrobinson251:
I think all uses of `kron` could use this package to avoid actually computing the big "kron matrix" except when completely necessary. I think `kron` is only used in MatrixNormal and MatrixTDist. If we got rid of the `kron` call in MatrixNormal, then I guess it's only a question if this would be better for the MatrxiTDist use

johnczito:
https://github.com/JuliaStats/Distributions.jl/blob/d00c780d680993519cf3dd0ce2b71dadd477afcf/src/matrix/matrixnormal.jl#L125

https://github.com/JuliaStats/Distributions.jl/blob/d00c780d680993519cf3dd0ce2b71dadd477afcf/src/matrix/matrixtdist.jl#L165

Completely agree. We are currently ignoring the fact that 
```
cholesky(A ⊗ B) = cholesky(A) ⊗ cholesky(B)
```
Since in both cases the two factors (`V`, `U` and `Σ`, `Ω`) are already `PDMat`, you should be able to go straight to the kronecker product as a `PDMat` instead of naively computing the full product and then computing its cholesky factor. But I think we will still need the full, dense kronecker product of the cholesky factors.

nickrobinson251:
yeah, `cholesky` would be a good addition to Kronecker.jl (fyi https://github.com/MichielStock/Kronecker.jl/issues/10)

I think needing a `PDMat` is a reason to not use Kronecker.jl, but a better reason to just remove PDMats (I'm not sure they're worth the bother)

johnczito:
Yeah, I will confess I don't understand what `PDMat` accomplishes that `Cholesky` doesn't (other than using twice the memory...).

MichielStock:
Will try to add `cholesky` and others to `Kronecker.jl` today!

johnczito:
[...]

PR #954, opened 2019-08-09
Use `kron` for PDMats

johnczito:
The latest release of `PDMats` includes JuliaStats/PDMats.jl#100, which adds a `kron` method that takes `PDMat` instances. This PR uses this to improve two spots [here](https://github.com/johnczito/Distributions.jl/blob/a15c8ba2ced774910d0bd6e47087946a7ac8824a/src/matrix/matrixnormal.jl#L125) and [here](https://github.com/johnczito/Distributions.jl/blob/a15c8ba2ced774910d0bd6e47087946a7ac8824a/src/matrix/matrixtdist.jl#L165).

### DIFF
--- a/src/matrix/matrixnormal.jl
+++ b/src/matrix/matrixnormal.jl
@@ -122,4 +122,4 @@ end
 #  Transformation
 #  -----------------------------------------------------------------------------
 
-vec(d::MatrixNormal) = MvNormal(vec(d.M), kron(d.V.mat, d.U.mat))
+vec(d::MatrixNormal) = MvNormal(vec(d.M), kron(d.V, d.U))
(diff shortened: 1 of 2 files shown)


johnczito:
Ah, right. Is it above my paygrade to bump it here?

ararslan:
Of course not! You just add an entry in the `[compat]` section of Project.toml that's something like `PDMats = ">=x.y.z"`, where `x.y.z` is the version of PDMats that added `kron`.

johnczito:
(Sorry, still a novice) Right, but is bumping the minimum version req something people would want to discuss in a separate PR, where they can judge the effects of the other changes in the latest `PDMat` release?

ararslan:
I guess I don't know what all changed in the most recent PDMats release, but note that the Project.toml does not currently place any version restrictions on the PDMats dependency, so users could have ol' version of PDMats, including the newest. If there was an explicit version specification in place already then I'd say yeah maybe we should think about it, but in this case the only effect will be to ensure if someone installs Distributions, their version of PDMats has the `kron` thing.

johnczito:
Actually, the comments [here](https://github.com/JuliaStats/Distributions.jl/pull/917#discussion_r297628281) and [here](https://github.com/JuliaStats/Distributions.jl/issues/939#issuecomment-515930335) are making me realize that there is a lot that I need to improve about how the matrix-variate distributions utilize `PDMats.jl`, so this PR is premature. I'll come back with a better-considered one.

PR #955, opened 2019-08-14
add var and cov methods for matrix-variates

johnczito:
Let `X` be a `n`x`p` random matrix with distribution `d`. `var(d)` returns the `n`x`p` matrix of element-wise variances. `cov(d)` returns the `np`x`np` covariance matrix of the random vector `vec(X)`.

Note that in general `cov(d)` will return a positive semi-definite (reduced rank) matrix because `X` is often symmetric, so `vec(X)` contains two copies of each free element in `X`. But the covariance matrix is still valid.


johnczito:
@codecov-io What a lovely sea of blood.

mschauer:
For `var` this is the most natural convention. For `cov` I am unsure if the returned object should be a `np x np` matrix or some 4 dimensional array. But in any case, returning the `cov` of the `vec` is a sane option.

johnczito:
Setting `cov(X):=cov(vec(X))` is a common convention. See [here](https://stats.stackexchange.com/questions/203299/covariance-between-two-random-matrices), [here](https://math.stackexchange.com/questions/1900448/how-to-compute-the-covariance-of-a-random-matrix), and a textbook like Gupta/Nagar. Though I agree that it's not necessarily the most useful form for the output.

src/matrix/matrixbeta.jl
```diff
@@ -84,6 +84,23 @@ mean(d::MatrixBeta) = ((n1, n2) = params(d); Matrix((n1 / (n1 + n2)) * I, dim(d)
 
 @inline partype(d::MatrixBeta{T}) where {T <: Real} = T
 
+#  Konno (1988 JJSS) Corollary 3.3.i
+function cov(d::MatrixBeta, i::Int, j::Int, k::Int, l::Int)
```
matbesancon: Could we relax the parameters to `Integer`?

src/matrix/matrixbeta.jl
```diff
@@ -84,6 +84,23 @@ mean(d::MatrixBeta) = ((n1, n2) = params(d); Matrix((n1 / (n1 + n2)) * I, dim(d)
...
+    n1, n2 = params(d)
+    n = n1 + n2
+    p = dim(d)
+    Ω = Matrix{partype(d)}(I, p, p)
+    n1 * n2 * inv(n * (n-1) * (n+2)) * (-(2/n) * Ω[i,j] * Ω[k,l] + Ω[j,l] * Ω[i,k] + Ω[i,l] * Ω[k,j])
+end
+
+function var(d::MatrixBeta, i::Int, j::Int)
```
matbesancon: Same here

src/matrixvariates.jl
```diff
@@ -45,6 +45,39 @@ Return the mean matrix of `d`.
...
+        i, j = Tuple(m)
+        V[m] = var(d, i, j)
+    end
+    return V
+end
+
+"""
+    cov(d::MatrixDistribution)
+
+Compute the covariance matrix for vec(X), where X is a random matrix with distribution `d`.
```
matbesancon:
```suggestion
Compute the covariance matrix for `vec(X)`, where `X` is a random matrix with distribution `d`.
```

matbesancon:
Thanks for this PR, I only had minor comments. My suspicion is that `cov(X) := cov(vec(X))` is not implemented because of the lack of N-D arrays in other ecosystems and because tensors are not commonly in some domains (pure speculation here).
In the future, that would be nice to have an additional method, with an argument `flattened::Val{false}` to return the non-flattened cov

src/matrix/matrixfdist.jl
```diff
@@ -95,6 +95,25 @@ end
...
 @inline partype(d::MatrixFDist{T}) where {T <: Real} = T
 
+#  Konno (1988 JJSS) Corollary 2.4.i
+function cov(d::MatrixFDist, i::Int, j::Int, k::Int, l::Int)
+    p = dim(d)
+    n1, n2, PDB = params(d)
+    n2 > p + 3 || throw(ArgumentError("cov only defined for df2 > dim + 3"))
+    n = n1 + n2
+    B = Matrix(PDB)
+    n1 * (n - p - 1) * inv((n2-p) * (n2-p-1) * (n2-p-3)) * (2inv(n2-p-1) * B[i,j] * B[k,l] + B[j,l] * B[i,k] + B[i,l] * B[k,j])
```
mschauer:
```suggestion
    n1*(n - p - 1)*inv((n2 - p)*(n2 - p - 1)*(n2 - p - 3))*(2inv(n2 - p - 1)*B[i,j]*B[k,l] + B[j,l]*B[i,k] + B[i,l]*B[k,j])
```
Nicer to align precedence of `*` vs `+` with the typography/spaces here

src/matrix/matrixbeta.jl
johnczito: Yupp, done.

src/matrix/matrixbeta.jl
johnczito: Yupp, done.

src/matrixvariates.jl
```diff
@@ -45,6 +45,42 @@ Return the mean matrix of `d`.
...
+    end
+    return V + tril(V, -1)'
+end
+
+"""
+    cov(d::MatrixDistribution, flattened = Val(false))
+
+Compute the 4-dimensional array whose `(i, j, k, l)` element is `cov(X[i,j], X[k, l])`.
+"""
+function cov(d::MatrixDistribution, ::Val{false})
```
matbesancon: Should we add a `Val{true}` that fallsback to the normal cov?

johnczito: Did I not already do that [here](https://github.com/johnczito/Distributions.jl/blob/03b54b0c171173faef20d3142944bb2fd8e5e526/src/matrixvariates.jl#L60)?

matbesancon: You did indeed, I mis-parsed it

johnczito:
@matbesancon @mschauer Thanks for the feedback. Let me know what else you want to see for this to be considered for merging.

matbesancon:
@mschauer good to me

Issue #956, opened 2019-08-22
Sampling from Dirichlet Dist breaks with Float32


dinarior:
`rand(Dirichlet([Float32(10),Float32(10)]))`

Results in the following error:

> ERROR: MethodError: no method matching Distributions.GammaGDSampler(::Float32, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float32, ::Float32, ::Float32, ::Float32)
> Closest candidates are:
>   Distributions.GammaGDSampler(::T<:Real, ::T<:Real, ::T<:Real, ::T<:Real, ::T<:Real, ::T<:Real, ::T<:Real, ::T<:Real, ::T<:Real, ::T<:Real) where T<:Real at /home/dinari/.julia/packages/Distributions/tfkz4/src/samplers/gamma.jl:10
> Stacktrace:
>  [1] Distributions.GammaGDSampler(::Gamma{Float32}) at /home/dinari/.julia/packages/Distributions/tfkz4/src/samplers/gamma.jl:57
>  [2] rand(::Random.MersenneTwister, ::Gamma{Float32}) at /home/dinari/.julia/packages/Distributions/tfkz4/src/univariate/continuous/gamma.jl:98
>  [3] _rand!(::Random.MersenneTwister, ::Dirichlet{Float32}, ::Array{Float32,1}) at /home/dinari/.julia/packages/Distributions/tfkz4/src/multivariate/dirichlet.jl:190
>  [4] rand(::Dirichlet{Float32}) at /home/dinari/.julia/packages/Distributions/tfkz4/src/multivariates.jl:83
>  [5] top-level scope at REPL[17]:1

Probaly related to [this](https://github.com/JuliaStats/Distributions.jl/pull/882)

PR #957, opened 2019-08-22
Relax Gamma Sampler types

dinarior:
* Relax T into Real in gamma sampler

### DIFF
--- a/src/samplers/gamma.jl
+++ b/src/samplers/gamma.jl
@@ -6,17 +6,17 @@
 
 # suitable for shape >= 1.0
 
-struct GammaGDSampler <: Sampleable{Univariate,Continuous}
-    a::Real
-    s2::Real
-    s::Real
-    i2s::Real
-    d::Real
-    q0::Real
-    b::Real
-    σ::Real
-    c::Real
-    scale::Real
+struct GammaGDSampler{T<:Real} <: Sampleable{Univariate,Continuous}
+    a::T
+    s2::T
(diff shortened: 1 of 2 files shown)


dinarior:
Fix for this:
https://github.com/JuliaStats/Distributions.jl/issues/956

mschauer:
I think this might come at performance cost because functions with `GammaGDSampler` will not dispatch on the type parameter `T` anymore.

dinarior:
Changed the ctor instead to have explicit casting to the correct type for all params (and not just some of them, as previously).

mschauer:
Yes, this is better (Ctor = constructor I guess)

src/samplers/gamma.jl
```diff
@@ -54,7 +54,7 @@ function GammaGDSampler(g::Gamma{T}) where {T}
         c = 0.1515/s
     end
 
-    GammaGDSampler(a,s2,s,i2s,d,q0,T(b),T(σ),T(c),scale(g))
+    GammaGDSampler(T(a),T(s2),T(s),T(i2s),T(d),T(q0),T(b),T(σ),T(c),scale(g))
```
matbesancon:
```suggestion
    GammaGDSampler(T(a), T(s2), T(s), T(i2s), T(d), T(q0), T(b), T(σ), T(c), scale(g))
```

matbesancon:
Thanks for this

PR #958, opened 2019-08-22
Possible replacement for #945

mschauer:
This is how I would approach #945, just as a start. It basically shows how to get the support business working without making the `eltype` a type parameter of the `support`. It helps to define `support(d, epsilon)` to find the numerical relevant support of unbounded distributions.

I illustrate how the type signature of `DiscreteNonParametric` can be extended to allow uni- and multivariate nonparametric distributions. 

The crux is, that `Discrete` does not need to imply that the support is only `Int`. All discrete distributions in the original meaning of the word (outcomes can be enumerated, e.g https://en.wikipedia.org/wiki/Probability_mass_function) can also by of `ValueSupport` type `Discrete`. I introduced some errors, so this is really just a sketch.

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -1,5 +1,7 @@
 module Distributions
 
+Base.ceil(::Type{Any}, x) = error("nonsensical")
+
 using StatsBase, PDMats, StatsFuns, Statistics
 using StatsFuns: logtwo, invsqrt2, invsqrt2π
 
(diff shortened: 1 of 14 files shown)


richardreeve:
As I've said before, I think we just fundamentally disagree about what constitutes a good design pattern for software. My code falls back to an error if you haven't implemented the correct methods. Yours falls back to the wrong answer (adding a comment in some places that this "is just a fallback" does not help). When (admittedly a long time ago!) I did my PhD in CS, this was considered a bad idea. I'll look at the rest later.

PR #965, opened 2019-09-02
Bounded distribution outside of support checking

m-wells:
I noticed an issue in the cdf of the Epanechnikov distribution.  The cdf should be

    0 for u <= -1
    1 for u >= 1
    1/2 + (3/4)u - (1/4)u^3 between -1 and 1

Currently it is of the form

    1 for u <= -1
    0 for u >= 1
    1/2 + (3/4)u - (1/4)u^3 between -1 and 1

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -50,7 +50,7 @@ const tests = [
     "discretenonparametric",
     "functionals",
     "chernoff",
-    "univariate_bounds"
+    "univariate_bounds",
 ]
 
 printstyled("Running tests:\n", color=:blue)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -50,6 +50,7 @@ const tests = [
     "discretenonparametric",
     "functionals",
     "chernoff",
+    "univariate_bounds"
 ]
 
 printstyled("Running tests:\n", color=:blue)
--- a/test/univariate_bounds.jl
+++ b/test/univariate_bounds.jl
@@ -1,7 +1,8 @@
 using Distributions
+using InteractiveUtils
 using Test
 
-dists = InteractiveUtils.subtypes(UnivariateDistribution)
+dists = subtypes(UnivariateDistribution)
 filter!(x -> hasmethod(x, ()), dists)
 filter!(x -> isbounded(x()), dists)
 
--- a/test/univariate_bounds.jl
+++ b/test/univariate_bounds.jl
@@ -1,0 +1,34 @@
+using Distributions
+using Test
+
+dists = subtypes(UnivariateDistribution)
+filter!(x -> hasmethod(x, ()), dists)
+filter!(x -> isbounded(x()), dists)
+
+@testset "bound checking $dist" for dist in dists
+    d = dist()
+    lb,ub = float.(extrema(support(d)))
+    lb = prevfloat(lb)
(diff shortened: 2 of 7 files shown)


matbesancon:
wow that's an ugly bug, thanks for spotting it. Could you add a bunch of tests to the CDF to validate the change?

matbesancon:
ping @m-wells

m-wells:
Will do. Been busy recently sorry for the delay.

matbesancon:
no problem, thanks for doing it

mschauer:
We can perhaps test this automatically for all distributions with interval support, so we could go with the fix for now and tackle this on a larger scale.

m-wells:
I added tests to check for the expected values for `cdf`, `logcdf`, `ccdf`, `logccdf`, `pdf`, and `logpdf` below and above the support bounds.  I found (and corrected) issues in `biweight.jl`, `logitnormal.jl`, `triweight.jl`, and `vonmises.jl`.

m-wells:
I implemented the `subtypes` function from `InteractiveUtils` as `get_subtypes` in the `univariate_bounds.jl` test file to get all the `univariate` distributions to test.  I would have done `using InteractiveUtils` which passes on `julia 1.0` but fails on `nightly`.  To avoid possible issues (and as it is only needed for testing) I just threw it into the `univariate_bounds.jl` test file.

test/univariate_bounds.jl
```diff
@@ -0,0 +1,88 @@
+using Distributions
+using Test
+
+# to make sure that subtypes provides the required behavoir without having to add
+# a dependency to InteractiveUtils
+function _subtypes(m::Module, x::Type, sts=Base.IdSet{Any}(), visited=Base.IdSet{Module}())
```
matbesancon: the function here is non-trivial and does some heavy reflection, could we document it a bit more (takes input this and that, returns that)

test/runtests.jl
```diff
@@ -50,6 +50,7 @@ const tests = [
     "discretenonparametric",
     "functionals",
     "chernoff",
+    "univariate_bounds"
```
matbesancon:
```suggestion
    "univariate_bounds",
```

test/univariate_bounds.jl
m-wells: This function is being used during testing only and I don't know how much time I want to spend documenting functions that won't be used outside of a single test file.  [Here is the implementation from InteractiveUtils.jl](https://github.com/JuliaLang/julia/blob/c6da87ff4bc7a855e217856757ad3413cf6d1f79/stdlib/InteractiveUtils/src/InteractiveUtils.jl#L175).  In addition, any documentation of this function should really be an upstream issue and doing that work here seems redundant.

mschauer: Ok with me.

m-wells:
No idea why the above test says pending.  If you go to details it says that both of them have passed.

test/univariate_bounds.jl
matbesancon: yeah it was more for maintainers to know what to do with this and what to expect from it, even some few months from now

matbesancon:
Since the `_subtypes` thing is on the test side and @mschauer is ok with, I'm ok with merging the PR as-is

PR #976, opened 2019-09-20
arg check with keyword

matbesancon:


matbesancon:
@nalimilan I'm finally checking `NoArgCheck` for a keyword argument in constructor, referring to our discussion from a month (?) ago

src/multivariate/multinomial.jl
```diff
@@ -23,18 +23,21 @@ struct Multinomial{T<:Real} <: DiscreteMultivariateDistribution
     n::Int
     p::Vector{T}
 
-    function Multinomial{T}(n::Integer, p::Vector{T}) where T
+    Multinomial{T}(n::Integer, p::Vector{T}) where {T} = new{T}(round(Int, n), p)
```
nalimilan:
```suggestion
    Multinomial{T}(n::Integer, p::Vector{T}) where {T} = new{T}(Int(n), p)
```
But is that constructor even needed?

src/multivariate/multinomial.jl
```diff
@@ -23,18 +23,21 @@ struct Multinomial{T<:Real} <: DiscreteMultivariateDistribution
...
+end
+
+function Multinomial(n::Integer, p::Vector{T}; arg_check = true) where {T<:Real}
```
nalimilan: Pure bikeshedding, but I'd have called it `checkargs` or `check_args`. Also I don't think we use spaces around `=` for keyword arguments.

matbesancon: I was thinking of `check_args` as the operation / action, while `arg_check` stands for "argument check", as in "is there an argument check"

matbesancon:
regarding the additional constructor: yes otherwise there is an ambiguity, because it creates a constructor with inferred type

src/multivariate/multinomial.jl
nalimilan: Yeah, but I have the impression that we generallly use verbs rather that nouns.

matbesancon:
For functions yes, but for a keyword, being a noun is ok IMO, arg_check =
true <=> is there argument checks

matbesancon:
@nalimilan just changed the keyword formatting

halleysfifthinc:
I will add as an occasional user of Distributions that I agree with @nalimilan that `check_args` reads better. I also think it matches better with the `@check_args` macro. It should have strictly less mental cost in connecting the `check_args` keyword with the `@check_args` macro (which is not to say that `arg_check` => `@check_args` is a very difficult connection to make).

matbesancon:
alright, not a nit-picking worth sweating on, I reversed the order of words in the last commit

nalimilan:
Why not `check_args` (or even `checkargs`)? :-p

src/univariate/continuous/beta.jl
```diff
@@ -32,19 +32,15 @@ struct Beta{T<:Real} <: ContinuousUnivariateDistribution
...
+function Beta(α::T, β::T; check_arg=true) where {T<:Real}
+    check_arg && @check_args(Beta, α > zero(α) && β > zero(β))
     return Beta{T}(α, β)
 end
 
 Beta(α::Real, β::Real) = Beta(promote(α, β)...)
 Beta(α::Integer, β::Integer) = Beta(float(α), float(β))
 Beta(α::Real) = Beta(α, α)
-Beta() = Beta(1.0, 1.0, NoArgCheck())
+Beta() = Beta(1.0, 1.0, check_arg = false)
```
nalimilan: Not sure what's the most common style in this repo, but having spaces around `=` for keyword arguments here is inconsistent with above.

matbesancon:
I thought it was no space for definition and spaces for use.
f(;kw=true)
f(;kw = false)

nalimilan:
AFAIK the most common convention is to have no spaces. But it may vary depending on the repo I guess.

PR #990, opened 2019-10-05
Make Multinomial support Tracker AD

mohamed82008:
This PR makes the `Multinomial` distribution support `Tracker` AD. Here is an example that works with this PR but not on `master`:
```julia
using Tracker, Distributions
Tracker.gradient((p) -> logpdf(Multinomial(2, p / sum(p)), [2, 0]), fill(0.5, 2))
```


matbesancon:
There is already `StaticArrays` in the test dependencies, maybe the tests here could include a static vector to test the effectiveness of the generic container type

mohamed82008:
Will get to it this weekend.

matbesancon:
Thanks!

PR #997, opened 2019-10-17
fix type inference issue for cdf(::Normal, ...)

### LINKED ISSUE #996: cdf(::Normal, x) does not infer well when `x` is not a `Float64`
MWE:
```julia
julia> using Distributions, ForwardDiff, Test

julia> @inferred cdf(Normal(0, 0.5), ForwardDiff.Dual(1.0, 2.0))
ERROR: return type ForwardDiff.Dual{Nothing,Float64,1} does not match inferred return type Union{Float64, ForwardDiff.Dual{Nothing,Float64,1}}
Stacktrace:
 [2] top-level scope at REPL[9]:1
```

I am wondering if [branch](https://github.com/JuliaStats/Distributions.jl/blob/fdf808cb57f67e3dfb66d1b9c2b192b670b35aaf/src/univariate/continuous/normal.jl#L142) introduced by #875 is really necessary, `erfc` should take care of this automatically.

tpapp:
Closes #996.

### DIFF
--- a/src/univariate/continuous/normal.jl
+++ b/src/univariate/continuous/normal.jl
@@ -139,9 +139,9 @@ function logccdf(d::Normal, x::Real)
     log1p(-erfc(-z * invsqrt2) / 2)
 end
 # cdf
-function cdf(d::Normal, x::Real)
+function cdf(d::Normal{T}, x::S) where {T, S <: Real}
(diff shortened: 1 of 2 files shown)


matbesancon:
Thanks!

Issue #979, opened 2019-09-25
fit Rayleigh


smeirlap:
Fitting for the Rayleigh distribution is not yet supported it seems. Would be useful.

matbesancon:
PR welcome for that

matbesancon:
@JuliaRegistrator register()

Nosferican:
Should this be closed?

matbesancon:
By all means

PR #1005, opened 2019-10-31
Add fit_mle for Rayleigh distribution

alainchau:
Addresses #979

### DIFF
--- a/src/univariate/continuous/rayleigh.jl
+++ b/src/univariate/continuous/rayleigh.jl
@@ -89,3 +89,17 @@ quantile(d::Rayleigh, p::Real) = sqrt(-2d.σ^2 * log1p(-p))
 #### Sampling
 
 rand(rng::AbstractRNG, d::Rayleigh) = d.σ * sqrt(2 * randexp(rng))
+
+
+#### Fitting
+
+function fit_mle(::Type{<:Rayleigh}, x::AbstractArray{T}) where T<:Real
+    # Compute MLE (and unbiasd estimator) of σ^2
+    n = length(x)
+    s2 = x[1]*x[1]
+    for i = 2:n
+        s2 += x[i]*x[i]
+    end
+    s2 /= (2*n)
+    Rayleigh(sqrt(s2))
+end
--- a/src/univariate/continuous/rayleigh.jl
+++ b/src/univariate/continuous/rayleigh.jl
@@ -95,11 +95,11 @@ rand(rng::AbstractRNG, d::Rayleigh) = d.σ * sqrt(2 * randexp(rng))
 
 function fit_mle(::Type{<:Rayleigh}, x::AbstractArray{T}) where {T<:Real}
     # Compute MLE (and unbiasd estimator) of σ^2
-    n = length(x)
-    s2 = x[1]*x[1]
-    for i = 2:n
-        s2 += x[i]*x[i]
+    s2 = zero(T)
+    for xi in x
+        s2 += xi^2
     end
-    s2 /= (2*n)
-    Rayleigh(sqrt(s2))
+
+    s2 /= (2*length(x))
+    return Rayleigh(sqrt(s2))
 end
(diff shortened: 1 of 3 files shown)


src/univariate/continuous/rayleigh.jl
```diff
@@ -89,3 +89,17 @@ quantile(d::Rayleigh, p::Real) = sqrt(-2d.σ^2 * log1p(-p))
 #### Sampling
 
 rand(rng::AbstractRNG, d::Rayleigh) = d.σ * sqrt(2 * randexp(rng))
+
+
+#### Fitting
+
+function fit_mle(::Type{<:Rayleigh}, x::AbstractArray{T}) where T<:Real
```
matbesancon:
```suggestion
function fit_mle(::Type{<:Rayleigh}, x::AbstractArray{T}) where {T<:Real}
```
for consistent style

src/univariate/continuous/rayleigh.jl
```diff
@@ -89,3 +89,17 @@ quantile(d::Rayleigh, p::Real) = sqrt(-2d.σ^2 * log1p(-p))
...
+    # Compute MLE (and unbiasd estimator) of σ^2
```
matbesancon: More simply, you can write it:
```julia
x2 = zero(T)
for xi in x
    s2 += xi^2
end
// rest of the function
```

src/univariate/continuous/rayleigh.jl
```diff
@@ -89,3 +89,17 @@ quantile(d::Rayleigh, p::Real) = sqrt(-2d.σ^2 * log1p(-p))
...
+    n = length(x)
+    s2 = x[1]*x[1]
+    for i = 2:n
+        s2 += x[i]*x[i]
+    end
+    s2 /= (2*n)
+    Rayleigh(sqrt(s2))
```
matbesancon:
```suggestion
    return Rayleigh(sqrt(s2))
```

matbesancon:
Thanks for the PR, see the few comments above before we can move forward

matbesancon:
ping @alainchau

alainchau:
Fixed it!

matbesancon:
Ok one last step to make sure this is fine is to check it all works fine with differentiation, something like:

the derivative of the mean of the estimated distribution is positive with the variation of one data point

dmean(R(x)) / dxi > 0 forall i

alainchau:
I'm not sure where to put it but I think we can do something like this?

```julia
@testset "Testing fit for Rayleigh" begin
    for func in funcs, dist in (Rayleigh, Rayleigh{Float64})
        x = func[2](dist(3.6), N)
        d = fit(dist, x)

        @test isa(d, dist)
        @test isapprox(mode(d), 3.6, atol=0.1)

        # ==== NEW PART: Check dμ/dxi > 0
        # We can just check differences since f'(x) ≈ (f(x+h)-f(x-h))/2h
        M = 100  # number of points to check per iteration
        h = 1e-4  # step size

        idx = rand(1:N, M)
        differences = zeros(M)

        x0 = copy(x)
        x1 = copy(x)
        for (t,i) in enumerate(idx)
            x0[i] -= h
            x1[i] += h
            μ0 = fit(Rayleigh, x0) |> mean
            μ1 = fit(Rayleigh, x1) |> mean
            differences[t] = (μ1 - μ0)
            x0[i] = x1[i] = x[i]
        end

        @test all(differences .> 0)
    end
end
```

matbesancon:
The point is to test automatic differentiation on this.

```
f(x) = mean(fit(Rayleigh, x))
@test all(ForwardDiff.gradient(f, x) .>= 0)
```

matbesancon:
Looks good, merging as soon as CI passes

alainchau:
I see. What does this check signify if you dont mind me asking? property of mle or rayleigh dist?

matbesancon:
So sorry I didn't get back earlier on this. ForwardDiff uses a Real type (`Dual`) which is not float, making sure ForwardDiff works helps us ensure the functions we build are fully generic and do not propagate hardcoded number types

matbesancon:
Thanks for the PR!

PR #1009, opened 2019-11-07
Upgrade to SpecialFunctions 0.8

andreasnoack:
Continuation of #891

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ os:
   - osx
 julia:
   - 1.0
+  - 1.2
   - nightly
 matrix:
   allow_failures:
--- a/Project.toml
+++ b/Project.toml
@@ -19,7 +19,7 @@ PDMats = "0.9"
 QuadGK = "2"
 SpecialFunctions = "0.8"
 StatsBase = "0.32"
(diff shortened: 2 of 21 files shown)


andreasnoack:
I think it might be timing out because of the depwarns from StatsFuns. We can try testing again once https://github.com/JuliaStats/StatsFuns.jl/pull/77 has been included in a release.

matbesancon:
I don't get how CI just froze (at a consistent point), infinite loop? There are still some warnings around

andreasnoack:
I believe the remaining warnings are all from StatsFuns. My suspicion was that it might be calling a deprecated method many many times and deprecations are really slow. I'll look more carefully once the new StatsFuns release is out.

andreasnoack:
I think it was the deprecations. It looks like things are working now.

andreasnoack:
@matbesancon I'm not sure how much stuff has accumulated on master so I'll let you handle the release but it would probably be good to get one out fairly soon since Distributions has hold back package updates for a while because of the SpecialFunctions restriction.

oschulz:
Cool! Will you guys tag a Distributions v0.21.6 with this?

PR #1010, opened 2019-11-07
Implement `eltype` for types instead of instances

devmotion:
Currently `eltype` is implemented for instances instead of types, although [according to the documentation](https://docs.julialang.org/en/v1/base/collections/#Base.eltype)

>  the form that accepts a type argument should be defined for new types.

I noticed this issue when I tried to implement a sampling algorithm as an iterator by following the example in the [documentation](https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-iteration-1) which advises to implement `eltype(IterType)`. Due to the missing implementation of `eltype` for types in Distributions.jl an implementation such as `eltype(::Type{<:CustomIterator{S}}) where S = eltype(S)` just leads to an element type of `Any`.

(BTW, even if the type-based implementation is added, Distributions.jl still requires some special casing due to the fact that `rand(dist::Distribution)` does not necessarily return a sample of type `eltype(dist)`, as opposed to when, e.g., [sampling from a collection](https://docs.julialang.org/en/v1/stdlib/Random/#Generating-values-from-a-collection-1)).

### DIFF
--- a/docs/src/multivariate.md
+++ b/docs/src/multivariate.md
@@ -18,7 +18,7 @@ The methods listed as below are implemented for each multivariate distribution, 
 ```@docs
 length(::MultivariateDistribution)
 size(::MultivariateDistribution)
-eltype(d::MultivariateDistribution)
+eltype(::Type{MultivariateDistribution})
 mean(::MultivariateDistribution)
 var(::MultivariateDistribution)
 cov(::MultivariateDistribution)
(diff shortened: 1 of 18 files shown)


devmotion:
Any comments or suggestions? :smiley: 

I'm not sure why the code coverage tests fail, I didn't add any new methods - maybe it's because I removed the superfluous `eltype` implementation for `ZeroVector`?

matbesancon:
it might be because the new eltype method is considered not tested

devmotion:
Since `eltype(x)` falls back to `eltype(typeof(x))` code coverage should not change because of the modified eltype methods, it seems?

matbesancon:
Could we adapt the tests / add some tests for `eltype(T)`?

devmotion:
Sure, I updated the existing tests.

Issue #995, opened 2019-10-16
MvNormal should accept Symmetric Matrices.


MrinankSharma:
Hi, 

I've been trying to use the `Distributions.MvNormal` constructor (apologies if this is the incorrect terminology - I'm still quite new to Julia). I tagged a matrix as symmetric, but the `Distributions.MvNormal` constructor does not accept a symmetric matrix. For example

```
A = [1 0; 0 1]
dist = Distributions.MvNormal(Symmetric(A))
```

Throws a `MethodError`. This seems to be as definitely undesirable behaviour. 

Please do let me know if I've misunderstood something!

Best,
Mrinank

PR #1013, opened 2019-11-14
Make MvNormal accept Symmetric and Diagonal matrices

treigerm:
Fixes the issue #995 and also makes MvNormal accept Diagonal matrices. I additionally updated the tests. 

I'm new to Julia and never made a pull request before so please let me know if I can do anything to improve this pull request.


matbesancon:
Fantastic, this looks good, merging soon thanks

PR #1016, opened 2019-11-15
fix type stability

### LINKED ISSUE #1014: logpdf(Normal, x) is not type stable
```julia
julia> @code_warntype logpdf(Normal(0.0f0, 0.0f0), 0.0f0)
Variables
  #self#::Core.Compiler.Const(Distributions.logpdf, false)
  d::Normal{Float32}
  x::Float32

Body::Union{Float32, Float64}
```
The Infs literals in
https://github.com/JuliaStats/Distributions.jl/blob/dbf0741e654d7f4d686273bc26b40dfd2a126403/src/univariate/continuous/normal.jl#L104 needs promotion.

cc: @Nosferican

matbesancon:
Fixes #1014


test/normal.jl
```diff
@@ -35,3 +36,31 @@ using Test, Distributions
         @inferred cdf(d, ForwardDiff.Dual(x, ∂x)) ≈ ForwardDiff.Dual(cdf(d, x), ∂x * pdf(d, x))
     end
 end
+
+@testset "Normal logpdf & pdf type inference" begin
```
andreasnoack: Probably a good idea to test the values here. In particular the `NaN` cases.

matbesancon:
> It'll be slower for the Inf/NaN case but I don't think it matters and then you don't need the extra methods for Union{Integer,Rational}.

Promotion of NaN is slower or it's slower because we don't branch directly on these values?

Nosferican:
I think a function barrier might be right approach... there are three cases,
- Value is finite (maybe `promote_type`)
- Is `NaN` (either `oftype` or `T(x)`)
- Type does not support the `NaN` and should be promoted to `Float`

Consider, `Dual(NaN)` what would work. I was thinking of using a `Val` for method to nonfinite, `convert`, `T`, `float`...

matbesancon:
We can consider it's the job of the caller to ensure the `Real` they use has a NaN and Inf defined

Nosferican:
Seems good.

andreasnoack:
I still don't think the `NaN` case is handled and I think it's easiest to use the actual transformations for promotion instead of relying on `promote_type` here. I believe this version should work.

```julia
function logpdf(d::Normal, x::TX) where {TX <: Real}
    ret = -(zval(d, x)^2 + log2π) / 2
    isfinite(ret)
        return ret - log(std(d))
    elseif mean(d) == x && iszero(std(d))
        return oftype(ret, Inf)
    else
        return ret
    end
end
```

andreasnoack:
> Promotion of NaN is slower or it's slower because we don't branch directly on these values?

My suggestion above is to unconditionally compute `ret = -(zval(d, x)^2 + log2π) / 2` such that we get the type right. That is more expensive than testing `mean(d) == x && iszero(std(d))` but I don't think it matters that the exceptional case is fast.

matbesancon:
> I still don't think the NaN case is handled

In what configuration would it fail?

andreasnoack:
> In what configuration would it fail?

See below. There is also an type stability problem for rationals because of the use of `promote_type`.

```julia
julia> function newlogpdf(d::Normal, x::TX) where {TX <: Real}
           T = promote_type(TX, eltype(d))
           if iszero(d.σ)
               mean(d) == x ? T(Inf) : T(-Inf)
           else
               -(Distributions.zval(d, x)^2 + Distributions.log2π) / 2 - log(std(d))
           end
       end
newlogpdf (generic function with 1 method)

julia> newlogpdf(Normal(1.0,0.0), NaN)
-Inf

julia> @inferred newlogpdf(Normal(1//1, 1//1), 1//1)
ERROR: return type Float64 does not match inferred return type Union{Rational{Int64}, Float64}
Stacktrace:
 [2] top-level scope at REPL[1303]:1
```

Nosferican:
In this the type stability is sort of pushed through since the type technically does depend on the values not the type itself. I think it is a good compromise to rely on the type having `Inf`/`NaN` defined for it as well as a constructor rather than a `convert` method. The `Integer`/`Rational` seem like a sensible rule.

andreasnoack:
@matbesancon I'm fine with finishing this one if you okay with me pushing to your branch.

matbesancon:
Completely yes thank you

matbesancon:
@andreasnoack did you manage to work on the last edge cases?

andreasnoack:
I realized that quite a few of the functions had to be modified and had to postpone it a bit. However, I think we can actually just copy the exact versions from StatsFuns since, I believe, they are actually handling the issues discussed correctly.

andreasnoack:
I've now updated this one. I've copied the definitions from `StatsFuns` with minimal adjustments. I've also removed redundant methods for quantile calculations and extended the tests for inferability to cover all of `pdf`, `logpdf`, `cdf`, `logcdf`, `ccdf`, `logccdf`, `quantile`, and `cquantile`. Finally, I've also fixed a bug in `quantile` and `cquantile` when `σ=0` and `0<p<1`.

andreasnoack:
Any idea what's up with the `NaN` testing on master?
```julia
Normal logpdf & pdf type inference: Test Failed at /home/travis/build/JuliaStats/Distributions.jl/test/normal.jl:59

  Expression: #= /home/travis/build/JuliaStats/Distributions.jl/test/normal.jl:59 =# @inferred(logpdf(Normal(0.0, 0.0), NaN)) === NaN

   Evaluated: NaN === NaN
```
looks weird.

matbesancon:
Agreed, not sure how it arrived here, it should be:
```
@inferred isnan(logpdf(Normal(0.0, 0.0), NaN))
```

matbesancon:
Oh no `===` should work for NaN

matbesancon:
> Any idea what's up with the NaN testing on master?

I thought you meant on Distributions master sorry for the two commits

matbesancon:
@andreasnoack I allowed travis failure on nightly for now, is this fine on your side?

matbesancon:
I just tested on 1.3, the tests pass

andreasnoack:
I tried locally on Julia master and was unable to reproduce. We might have some interesting debugging to do in 4-6 months when we are ready to release Julia 1.4. For now, I think this is ready to go.

PR #1018, opened 2019-11-18
Fixes out of bounds error for loop in rand(DiscreteNonParametric)

### LINKED ISSUE #1017: rand(DiscreteNonParametric) fails for long probability vectors with many small elements
This is similar to #872 in spirit but for the other end of the boundary condition. I get a out of bounds error at the [following line](https://github.com/JuliaStats/Distributions.jl/blob/f20dd9240c5c498e97fd7ae5c40fa584008d2e23/src/univariate/discrete/discretenonparametric.jl#L79) due to `p[i += 1]`. On the previous iteration of the loop, `cp += p[i += 1]` should result in `cp == 1` which in turn means `cp < draw` must be false. But if the probabilities being added are sufficiently small (e.g. ~`10e-8`), then due to numerical errors, the sum is less than 1. As a result, another iteration of the loop runs resulting in an out of bounds error. Note that the probability vector correctly passes the `isprobvec` test upon creating the distribution.

The conditions to make this error occur are `p` is long enough, `p` has non-zero probability for almost all elements and the probabilities are very small, and `draw == 1`.

In these cases, we should be returning `x[end]`. So, I propose adding the check `i <= length(p)` to the loop condition which should produce the correct return. I have time for PR which I will do immediately if it makes sense to people.

darsnack:
Fixes #1017 

Changes:
- Add condition `i < length(p)` to while loop in `rand(DiscreteNonParametric)`

### DIFF
--- a/test/discretenonparametric.jl
+++ b/test/discretenonparametric.jl
@@ -112,4 +112,14 @@ p = [1 - eps(Float32), eps(Float32)]
 d = Categorical(p)
 @test ([rand(d) for _ = 1:100_000]; true)
 
+# Numerical stability w/ large prob vectors;
+# see issue #1017
+n = 20000 # large vector length
+p = Float32[0.5; fill(0.5/(n ÷ 2) - 3e-8, n ÷ 2); fill(eps(Float64), n ÷ 2)]
+d = Categorical(p)
+struct DumbRNG <: AbstractRNG end
+Base.rand(::DumbRNG, ::Type{T}) where {T<:Number} = one(T)
+rng = DumbRNG()
+@test (rand(rng, d); true)
+
 end
--- a/test/discretenonparametric.jl
+++ b/test/discretenonparametric.jl
@@ -2,6 +2,10 @@ import StatsBase: ProbabilityWeights
 using Random, Distributions
 using Test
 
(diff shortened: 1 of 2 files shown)


matbesancon:
Thanks for the PR, could you add a test that was failing with https://github.com/JuliaStats/Distributions.jl/issues/1017?

darsnack:
Wow that was a hard one to get a consistent test case for! The most recent commit should pass with the PR but fail (throw a BoundsError) without it.

test/discretenonparametric.jl
```diff
@@ -112,4 +112,14 @@ p = [1 - eps(Float32), eps(Float32)]
 d = Categorical(p)
 @test ([rand(d) for _ = 1:100_000]; true)
 
+# Numerical stability w/ large prob vectors;
+# see issue #1017
+n = 20000 # large vector length
+p = Float32[0.5; fill(0.5/(n ÷ 2) - 3e-8, n ÷ 2); fill(eps(Float64), n ÷ 2)]
+d = Categorical(p)
+struct DumbRNG <: AbstractRNG end
```
matbesancon: The DumbRNG definition and method implementation should be at the top-level scope, outside the testset macro (see Travis error on 1.0)

Issue #826, opened 2019-02-04
Standard MvNormal, Inconsistency with Normal, and I


jessebett:
`MvNormal` is inconsistent with the syntax for `Normal`

Constructing a standard univariate Normal is achieved by `Normal()`
However, `MvNormal()` is not defined.
I understand this is ambiguous in dimensionality. However it isn't ambiguous to do something like evaluate the pdf of a d-dimensional vector under an `MvNormal`, as it will necessarily be `d` dimensional. I wonder if there is a way to have the standard normal defined without dimension stated, for evaluating densities, but sampling from this will obviously not be defined.

When `Normal` dispatches on a single argument it corresponds to the mean, but when `MvNormal` dispatches on a single argument it corresponds to the diagonal vector of its sigma... this is pretty strange.
An extreme example is `Normal(1.)` is a gaussian with `mu=1` and `sig=1`, but `MvNormal([1.])` is a 1D gaussian with `mu=0` and `sig=0`... 

`MvNormal` has not been updated to take `I` for the identity. 
The documentation needs to change as well, as it refers to `eye(d)`.

The current way to define a standard d-dimensional multivariate normal isn't great `MvNormal(zeros(d),Matrix(I,d,d))`

I would like `MvNormal(d=3)` to produce a 3-dimensional standard multivariate normal.
`MvNormal(zeros(3),I)` should work as well.

matbesancon:
@jessebett PR very welcome on this

andreasnoack:
Let's continue this conversation in https://github.com/JuliaStats/Distributions.jl/issues/826

Issue #584, opened 2017-03-13
Inconsistency in MvNormal constructors, covariance or deviation


dehann:
Hi,

I'm using Distributions.jl in [IncrementalInference.jl](http://www.github.com/dehann/IncrementalInference.jl) and found what looks to be inconsistent behavior when constructing MvNormals. One dimensional takes deviation while higher dimensions seem to take covariance.

```julia
using Distributions
julia> Base.std(rand(Normal(0.0,100.0), 1000))
96.7377564234093

julia> Base.std(rand(MvNormal([0.0],[100.0]), 1000), 2)
1×1 Array{Float64,2}:
 103.241

julia> Base.std(rand(MvNormal([0.0;0.0],[100.0;10.0]), 1000), 2)
2×1 Array{Float64,2}:
 101.026 
  10.0854

julia> Base.std(rand(MvNormal([0.0;0.0],[[100.0;0.0]';[0.0;10.0]']), 1000), 2)
2×1 Array{Float64,2}:
 10.1583 
  3.18971
```

Notice the last call expects covariance where all others take standard deviation. Do you expect to persist this behavior? I don't really mind, but it's a bit confusing and cost me some time debugging.

Thanks for putting this package out there!

felixrehren:
This inconsistency also got me.

Mixing variance and volatility/deviation is in some sense type-instable for dimensional quantities (e.g. for market prices measured in USD, the deviation is in USD, the variance in USD^2). 

It feels surprising that 
* you can either provide a diagonal matrix, or the diagonal of the same matrix, but these two will construct different distributions
* univariate `MvNormal` is different from multivariate `MvNormal` (if the correlation is specified -- which is presumably the main reason to use multivariate normal distributions), which makes d=1 a very special case indeed

This is the number 2 result for google > "julia distributions", so it seems to strike a chord. But I have no idea how this could be changed without breaking people's code. Perhaps the documentation of the arguments should be put into all-caps?

dehann:
I would vote for not having special cases, and if this is going to be fixed it should be before Julia 1.0...  Agreed that this would have deep affects in the package code base -- the only way would be clear documentation, announcements, and maximum lead-time warning upon `using Distributions`?
```julia
Warning:  Distributions.Normal and Distributions.MvNormal constructors are being standardized to deviation/covariance/?.  Julia 1.0 versions will use the new standard, i.e. Distributions v???+.
```
[...]

PR #1019, opened 2019-11-24
[Breaking] Treat `UniformScaling` in constructor of `MvNormal` as matrix

devmotion:
There exist different issues about inconsistencies in the constructor of `MvNormal` (see, e.g., https://github.com/JuliaStats/Distributions.jl/issues/584 and https://github.com/JuliaStats/Distributions.jl/issues/826). It caught me by surprise (https://github.com/TuringLang/Turing.jl/issues/991) that actually
```julia
MvNormal(ones(3), 3 * I) != MvNormal(ones(3), Matrix(3 * I, 3, 3))
```
I would have assumed that `3 * I` is treated liked other matrices as a proper covariance matrix. IMO the current behaviour of
```julia
MvNormal(ones(3), 3 * I) == MvNormal(ones(3), 3)
```
is unintuitive (which, of course, is also due to the fact that vectors and scalars are treated as standard deviations - I assume to be more consistent with the constructor of `Normal`). This PR changes this behaviour and treats `UniformScaling` like any other matrix.

Additionally I tried to reduce the amount of redundant code.

### DIFF
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -196,31 +196,24 @@ function MvNormal(μ::AbstractVector, Σ::AbstractPDMat)
     MvNormal(convert(AbstractArray{R}, μ), convert(AbstractArray{R}, Σ))
 end
 
-function MvNormal(Σ::Cov) where {T, Cov<:AbstractPDMat{T}}
-    MvNormal{T,Cov,ZeroVector{T}}(ZeroVector(T, dim(Σ)), Σ)
-end
-
-MvNormal(μ::AbstractVector{<:Real}, Σ::Matrix{<:Real}) = MvNormal(μ, PDMat(Σ))
+# constructor with general covariance matrix
+MvNormal(μ::AbstractVector{<:Real}, Σ::AbstractMatrix{<:Real}) = MvNormal(μ, PDMat(Σ))
 MvNormal(μ::AbstractVector{<:Real}, Σ::Union{Symmetric{<:Real}, Hermitian{<:Real}}) = MvNormal(μ, PDMat(Σ))
 MvNormal(μ::AbstractVector{<:Real}, Σ::Diagonal{<:Real}) = MvNormal(μ, PDiagMat(diag(Σ)))
-MvNormal(μ::AbstractVector{<:Real}, σ::Vector{<:Real}) = MvNormal(μ, PDiagMat(abs2.(σ)))
+MvNormal(μ::AbstractVector{<:Real}, Σ::UniformScaling{<:Real}) =
+    MvNormal(μ, ScalMat(length(μ), Σ.λ))
+
+# constructor with vector of standard deviations
+MvNormal(μ::AbstractVector{<:Real}, σ::AbstractVector{<:Real}) = MvNormal(μ, PDiagMat(abs2.(σ)))
+
+# constructor with scalar standard deviation
 MvNormal(μ::AbstractVector{<:Real}, σ::Real) = MvNormal(μ, ScalMat(length(μ), abs2(σ)))
 
-function MvNormal(μ::AbstractVector{<:Real}, Σ::VecOrMat{<:Real})
-    R = Base.promote_eltype(μ, Σ)
-    MvNormal(convert(AbstractArray{R}, μ), convert(AbstractArray{R}, Σ))
-end
+# constructor without mean vector
+MvNormal(Σ::AbstractVecOrMat{<:Real}) = MvNormal(ZeroVector(eltype(Σ), size(Σ, 1)), Σ)
 
-function MvNormal(μ::AbstractVector{<:Real}, σ::UniformScaling{<:Real})
-    R = Base.promote_eltype(μ, σ.λ)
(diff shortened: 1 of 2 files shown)


src/multivariate/mvnormal.jl
```diff
@@ -196,31 +196,24 @@ function MvNormal(μ::AbstractVector, Σ::AbstractPDMat)
...
 end
 
-function MvNormal(Σ::Cov) where {T, Cov<:AbstractPDMat{T}}
-    MvNormal{T,Cov,ZeroVector{T}}(ZeroVector(T, dim(Σ)), Σ)
-end
-
-MvNormal(μ::AbstractVector{<:Real}, Σ::Matrix{<:Real}) = MvNormal(μ, PDMat(Σ))
+# constructor with general covariance matrix
+MvNormal(μ::AbstractVector{<:Real}, Σ::AbstractMatrix{<:Real}) = MvNormal(μ, PDMat(Σ))
 MvNormal(μ::AbstractVector{<:Real}, Σ::Union{Symmetric{<:Real}, Hermitian{<:Real}}) = MvNormal(μ, PDMat(Σ))
```
mohamed82008: This is included in the previous method?

src/multivariate/mvnormal.jl
```diff
@@ -196,31 +196,24 @@ function MvNormal(μ::AbstractVector, Σ::AbstractPDMat)
...
 
-function MvNormal(μ::AbstractVector{<:Real}, Σ::VecOrMat{<:Real})
-    R = Base.promote_eltype(μ, Σ)
-    MvNormal(convert(AbstractArray{R}, μ), convert(AbstractArray{R}, Σ))
-end
+# constructor with vector of standard deviations
+MvNormal(μ::AbstractVector{<:Real}, σ::AbstractVector{<:Real}) = MvNormal(μ, PDiagMat(abs2.(σ)))
 
-function MvNormal(μ::AbstractVector{<:Real}, σ::UniformScaling{<:Real})
-    R = Base.promote_eltype(μ, σ.λ)
```
mohamed82008: The type promotion seems important here.

src/multivariate/mvnormal.jl
devmotion: Somehow missed that :slightly_smiling_face:

src/multivariate/mvnormal.jl
devmotion: I reduced to the promotion logic to just one implementation in https://github.com/JuliaStats/Distributions.jl/pull/1019/files/43cd8d9512f6429c02ca36f07fc5083441b193f5#diff-4d9faeaf6c64543851cc2242175fac82R194-R196

mohamed82008: Cool.

matbesancon:
Probably stupid question but this is non-breaking right?

devmotion:
It changes the behaviour of `MvNormal(mu, sigma2 * I)`, so it's a breaking change IMO.

matbesancon:
Ok thanks, I'll probably do a patch release before merging this
Is this closing #584 and #826?

devmotion:
No, this PR just removes one inconsistency in the current implementation of `MvNormal` but it neither fixes the inconsistencies between `MvNormal` and `Normal` nor does it allow to define a multivariate standard normal distribution without dimension specified. Without keyword arguments a (simple) solution to the first problem would be to change `Normal` to take the variance instead of the standard deviation and define `MvNormal(v::AbstractVector) = MvNormal(v, I)` it seems, i.e., similar to `Normal` assume that the covariance matrix is `I` if only a vector is specified. In addition one could still define `MvNormal(S::AbstractMatrix) = MvNormal(ZeroVector(eltype(S), size(S, 1)), S)` if a matrix is specified (clearly that's not possible for `Normal`). I'm not sure if it's worth implementing `MvNormal()` without dimension constraints.

matbesancon:
@devmotion could you see to the merge conflict on master? After that we will merge this and bump a breaking change version

devmotion:
Sorry, I forgot to keep the PR up to date! I merged the latest changes on master and resolved the merge conflict.

matbesancon:
thanks a lot, let's get this merged finally :)

src/multivariate/mvnormal.jl
```diff
@@ -195,29 +195,23 @@ function MvNormal(μ::AbstractVector, Σ::AbstractPDMat)
     MvNormal(convert(AbstractArray{R}, μ), convert(AbstractArray{R}, Σ))
 end
 
-MvNormal(Σ::AbstractPDMat) = MvNormal(Zeros{eltype(Σ)}(dim(Σ)), Σ)
```
andreasnoack: @devmotion Was it intentional that you deleted this one or did you expect the `MvNormal(Σ::AbstractVecOrMat{<:Real})` to cover? (It doesn't since the PDMats aren't AsbtractMatrices.)

devmotion: No, I don't remember exactly but I guess I expected it to be covered by the general constructor without mean vector. Unfortunate that I missed this.

Issue #833, opened 2019-02-21
Repalce ZeroVector with FillArrays


oxinabox:
https://github.com/JuliaArrays/FillArrays.jl

Rather than reimplementing all that logic here,
lets just use the FillArrays type for this.
Can also use it for any other scalar constants that get passed in to multivariate distributions

(cc: @nickrobinson251)

simonbyrne:
Can be part of #831.

PR #1020, opened 2019-11-26
Replace ZeroVector with Zeros from FillArrays

devmotion:
Fixes https://github.com/JuliaStats/Distributions.jl/issues/833.


matbesancon:
Could you add a compat entry for FillArrays at 0.8 in the Project.toml?

matbesancon:
Cool, merging as soon as this shows green, thank you for this

PR #1021, opened 2019-11-28
Use native samplers for Poisson distribution

devmotion:
I noticed that `sampler` is not implemented for Poisson distributions although two native samplers `PoissonCountSampler` and `PoissonADSampler` exist. Moreover, currently R is used for sampling from Poisson distributions, [which is quite slow](https://github.com/JuliaDiffEq/PoissonRandom.jl/issues/6). For now I used the same cut-off value of 6 for both samplers, as chosen in [PoissonRandom.jl](https://github.com/JuliaDiffEq/PoissonRandom.jl).

I repeated the benchmarks
```julia
using Distributions, PoissonRandom, StatsFuns
using Plots

function n_count(rng, λ, n)
  tmp = 0
  for i in 1:n
    tmp += PoissonRandom.count_rand(rng,λ)
  end
  tmp
end

function n_pois(rng,λ,n)
  tmp = 0
  for i in 1:n
    tmp += pois_rand(rng,λ)
  end
  tmp
end

function n_ad(rng, λ, n)
  tmp = 0
  for i in 1:n
    tmp += PoissonRandom.ad_rand(rng, λ)
  end
  tmp
end

function n_dist(λ,n)
  tmp = 0
  for i in 1:n
    tmp += rand(Poisson(λ))
  end
  tmp
end

function n_rfunctions(λ, n)
  tmp = 0
  for i in 1:n
    tmp += convert(Int, StatsFuns.RFunctions.poisrand(λ))
  end
  tmp
end

function n_countsampler(rng, λ::Float64, n)
  tmp = 0
  for i in 1:n
    tmp += rand(rng, Distributions.PoissonCountSampler(λ))
  end
  tmp
end

function n_adsampler(rng, λ::Float64, n)
  tmp = 0
  for i in 1:n
    tmp += rand(rng, Distributions.PoissonADSampler(λ))
  end
  tmp
end

function time_λ!(rng, times, λ::Float64, n)
  times[1] = @elapsed n_count(rng, λ, n)
  times[2] = @elapsed n_ad(rng, λ, n)
  times[3] = @elapsed n_pois(rng, λ, n)
  times[4] = @elapsed n_dist(rng, λ, n)
  times[5] = @elapsed n_rfunctions(λ, n)
  times[6] = @elapsed n_countsampler(rng, λ, n)
  times[7] = @elapsed n_adsampler(rng, λ, n)

  nothing
end

function plot_benchmark(rng)
    times = Matrix{Float64}(undef, 7, 20)

    # Compile
    time_λ!(rng, view(times, :, 1), 5, 5_000_000)

    # Run with a bunch of λ
    for λ in 1:20
        time_λ!(rng, view(times, :, λ), float(λ), 5_000_000)
    end

    plot(times',
         labels = ["count_rand" "ad_rand" "pois_rand" "Distributions" "RFunctions" "PoissonCountSampler" "PoissonADSampler"],
         lw = 3)
end
```
from https://github.com/JuliaDiffEq/PoissonRandom.jl/issues/6 with this PR. I get
```julia
using Random
Random.seed!(1234)
plot_benchmark(Random.GLOBAL_RNG)
savefig("global_rng.png")
```
global_rng
and
```julia
using RandomNumbers
plot_benchmark(Xorshifts.Xoroshiro128Plus(1234))
savefig("xoroshiro128plus.png")
```
xoroshiro128plus

Using the native samplers leads to a significant speed-up and a performance which is on par with [PoissonRandom.jl](https://github.com/JuliaDiffEq/PoissonRandom.jl).

### DIFF
--- a/src/samplers/poisson.jl
+++ b/src/samplers/poisson.jl
@@ -102,7 +102,7 @@ function rand(rng::AbstractRNG, sampler::PoissonADSampler)
         c = 0.1069 / μ
 
         # Step H
-        if c * abs(U) <= py * exp(px + E) - fy * exp(fx + E)
+        if c*abs(U) <= py*exp(px + E) - fy*exp(fx + E)
             return K
         end
     end
--- a/src/samplers/poisson.jl
+++ b/src/samplers/poisson.jl
@@ -13,17 +13,20 @@ end
 #
 # Suitable for small μ
 #
-struct PoissonCountSampler <: Sampleable{Univariate,Discrete}
-    μ::Float64
+struct PoissonCountSampler{T<:Real} <: Sampleable{Univariate,Discrete}
+    μ::T
 end
+
+PoissonCountSampler(d::Poisson) = PoissonCountSampler(rate(d))
 
 function rand(rng::AbstractRNG, s::PoissonCountSampler)
     μ = s.μ
+    T = typeof(μ)
     n = 0
-    c = randexp(rng)
+    c = randexp(rng, T)
     while c < μ
         n += 1
-        c += randexp(rng)
+        c += randexp(rng, T)
     end
     return n
 end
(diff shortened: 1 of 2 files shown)


matbesancon:
Samplers are not my area, but do we have a way to test for correctness of the sampler?

devmotion:
It seems they are tested in
https://github.com/JuliaStats/Distributions.jl/blob/802a65b9568f5b5723a204b9767d03aeac0e00cb/test/samplers.jl#L62-L74

matbesancon:
LGTM, I'll wait for another review before merging though, to get more educated opinions

src/samplers/poisson.jl
```diff
@@ -36,72 +39,85 @@ end
...
+    μType = typeof(μ)
 
-function rand(rng::AbstractRNG, s::PoissonADSampler)
     # Step N
-    G = s.μ + s.s*randn(rng)
+    G = μ + s * randn(rng, μType)
 
-    if G >= 0.0
-        K = floor(Int,G)
+    if G >= zero(G)
```
mschauer: You can just compare with `0`?

src/univariate/discrete/poisson.jl
```diff
@@ -141,101 +141,18 @@ fit_mle(::Type{<:Poisson}, ss::PoissonStats) = Poisson(ss.sx / ss.tw)
...
-
-            # Step H
-            if c*abs(U) <= py*exp(px+E)-fy*exp(fx+E)
-                return K
-            end
-        end
-    else # Case B
-        # Ahrens & Dieter use a sequential method for tabulating and looking up quantiles.
-        # TODO: check which is more efficient.
-        return quantile(d,rand(rng))
```
mschauer: So this is the sub approach we will not use anymore?

src/univariate/discrete/poisson.jl
```diff
@@ -141,101 +141,18 @@ fit_mle(::Type{<:Poisson}, ss::PoissonStats) = Poisson(ss.sx / ss.tw)
...
-    c3 = 0.14285714285714285*b1*b2
-    c2 = b2 - 15.0*c3
-    c1 = b1 - 6.0*b2 + 45.0*c3
-    c0 = 1.0 - b1 + 3.0*b2 - 15.0*c3
-
-    if K < 10
-        px = -μ
-        py = μ^K/factorial(K) # replace with loopup?
+function sampler(d::Poisson)
+    if rate(d) < 6
```
mschauer: this magic threshold appears twice, make it a constant?

src/samplers/poisson.jl
```diff
@@ -36,72 +39,85 @@ end
...
-        K = floor(Int,s.μ + s.s*T)
-        px,py,fx,fy = procf(s.μ,K,s.s)
-        c = 0.1069/s.μ
+        K = floor(Int, μ + s * T)
+        px, py, fx, fy = procf(μ, K, s)
+        c = 0.1069 / μ
 
         # Step H
-        if c*abs(U) <= py*exp(px+E)-fy*exp(fx+E)
+        if c * abs(U) <= py * exp(px + E) - fy * exp(fx + E)
```
mschauer:
```suggestion
        if c*abs(U) <= py*exp(px + E) - fy*exp(fx + E)
```
Spacing aligned with order of evaluation

src/univariate/discrete/poisson.jl
devmotion: Yes, instead of the quantile-based generation this PR uses the `PoissonCountSampler` for small rates which just counts exponentially distributed random numbers.

src/samplers/poisson.jl
devmotion: I guess that could be used here as well. I think it makes only a different if one works with non-standard number types such as unitful numbers - e.g., `u"1.0m" > 0` errors whereas `u"1.0m" > zero(u"1.0m")` works as expected.

src/univariate/discrete/poisson.jl
devmotion: I guess it's even better to not define `rand` at all and remove the code duplication.

matbesancon: bump on this? Maybe naming the constant is still a nice idea

src/samplers/poisson.jl
mschauer: Ah, good point, let's be defensive then.

matbesancon:
@mschauer @devmotion good to merge?

devmotion:
Yes 👍 I guess, I adjusted the PR according to your suggestions and comments

matbesancon:
thanks for the PR :)

Issue #1022, opened 2019-12-02
VonMises support


bdeonovic:
I noticed that the support of VonMises is given as:

```
@distr_support VonMises d.μ - π d.μ + π
```

but the pdf/logpdf is defined as:
```
pdf(d::VonMises{T}, x::Real) where T<:Real =
    -π ≤ x ≤ π ? exp(d.κ * (cos(x - d.μ) - 1)) / (twoπ * d.I0κx) : zero(T)
logpdf(d::VonMises{T}, x::Real) where T<:Real =
    -π ≤ x ≤ π ? d.κ * (cos(x - d.μ) - 1) - log(d.I0κx) - log2π : -T(Inf)
```

where `pi` is used as a check. Is this correct? something doesn't seem quite right here.

matbesancon:
just fixed in last release by @bdeonovic

matbesancon:
@JuliaRegistrator register()

PR #1025, opened 2019-12-03
Fix (log)pdf of VonMises

bdeonovic:
Added the correct support check. Fixes https://github.com/JuliaStats/Distributions.jl/issues/1022


matbesancon:
Thanks for the fix, could you add a test, for example with a shifted von Mises, `µ = 10`, that the pdf is > 0 at the two boundaries

bdeonovic:
where is the most appropriate place to put it?

matbesancon:
You can create a vonmises.jl test file, I don't think there is already somewhere appropriate

matbesancon:
ping @bdeonovic :)

matbesancon:
bump

bdeonovic:
heyo! I will finish up this PR this week!

matbesancon:
Great thanks :)

test/vonmises.jl
```diff
@@ -0,0 +1,36 @@
+# Tests for Von-Mises distribution
+
+using Distributions
+using Compat.Test
```
matbesancon:
```suggestion
using Test
```
Do we need legacy Test behavior here?

matbesancon: @bdeonovic this is the last thing before merging I think

bdeonovic:
No I just copy posted from another test file. I'll update after lunch.
Thanks

matbesancon:
Thanks!

PR #975, opened 2019-09-20
added product-function

matbesancon:
Without this in the current situation, one would have to abuse the `Product` constructor to create specialized distributions

PR #1026, opened 2019-12-03
Truncate generic

matbesancon:
Just like https://github.com/JuliaStats/Distributions.jl/pull/975 but for `Truncated`
Sadly Base already exports `truncate` so we had to settle with `truncated(::D, lower, upper)`

### DIFF
--- a/docs/src/truncate.md
+++ b/docs/src/truncate.md
@@ -1,7 +1,7 @@
 # Truncated Distributions
 
 The package provides the `truncated` function which creates the most
-appropriate distribution to represente a truncated version of a given
+appropriate distribution to represent a truncated version of a given
 distribution.
 
 
--- a/docs/src/truncate.md
+++ b/docs/src/truncate.md
@@ -1,27 +1,25 @@
 # Truncated Distributions
 
-The package provides the `Truncated` type to represented truncated distributions, which is defined as below:
+The package provides the `truncated` function which creates the most
+appropriate distribution to represente a truncated version of a given
+distribution.
 
-```julia
-struct Truncated{D<:UnivariateDistribution,S<:ValueSupport} <: Distribution{Univariate,S}
-    untruncated::D      # the original distribution (untruncated)
-    lower::Float64      # lower bound
-    upper::Float64      # upper bound
-    lcdf::Float64       # cdf of lower bound
-    ucdf::Float64       # cdf of upper bound
 
-    tp::Float64         # the probability of the truncated part, i.e. ucdf - lcdf
-    logtp::Float64      # log(tp), i.e. log(ucdf - lcdf)
-end
+A truncated distribution can be constructed using the following signature:
+
+```@docs
+truncated
 ```
 
-A truncated distribution can be constructed using the constructor `Truncated` as follows:
+In the general case, this will create a `Truncated{typeof(d)}`
+structure, defined as follows:
 
 ```@docs
 Truncated
 ```
 
-Many functions, including those for the evaluation of pdf and sampling, are defined for all truncated univariate distributions:
+Many functions, including those for the evaluation of pdf and sampling,
+are defined for all truncated univariate distributions:
 
 - [`maximum(::UnivariateDistribution)`](@ref)
 - [`minimum(::UnivariateDistribution)`](@ref)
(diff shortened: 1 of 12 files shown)


src/truncate.jl
```diff
@@ -44,9 +80,9 @@ isupperbounded(d::Truncated) = isupperbounded(d.untruncated) || isfinite(d.upper
 minimum(d::Truncated) = max(minimum(d.untruncated), d.lower)
 maximum(d::Truncated) = min(maximum(d.untruncated), d.upper)
 
-insupport(d::Truncated{D,Union{Discrete,Continuous}}, x::Real) where {D<:UnivariateDistribution} =
-    d.lower <= x <= d.upper && insupport(d.untruncated, x)
-
+function insupport(d::Truncated{D,<: Union{Discrete,Continuous}}, x::Real) where {D<:UnivariateDistribution}
```
mschauer:
```suggestion
function insupport(d::Truncated{D,<:Union{Discrete,Continuous}}, x::Real) where {D<:UnivariateDistribution}
```

mschauer:
This is how it should be! Were there any tricky parts?

matbesancon:
> Were there any tricky parts?

Yes, I'm still checking the parts of the JSON, which implies parsing + nasty reflection

matbesancon:
One of the things we have to do at some point is clean out those JSON tests into something more standard

matbesancon:
All green, no more warnings. Waiting for one approval before merging

src/truncate.jl
```diff
@@ -1,7 +1,41 @@
+"""
+    truncated(d, l, u):
+
+Truncate a distribution between `l` and `u`.
+Builds the most appropriate distribution for the type of `d`,
+the fallback is constructing a `Truncated` distribution.
```
mschauer: How do we call things like `Symmetric`? Wrapper?

src/truncate.jl
```diff
@@ -1,7 +1,41 @@
...
+    tp = ucdf - lcdf
+    Truncated(d, promote(l, u, lcdf, ucdf, tp, log(tp))...)
+end
+
+truncated(d::UnivariateDistribution, l::Integer, u::Integer) = truncated(d, float(l), float(u))
+
 """
     Truncated(d, l, u):
 
 Construct a truncated distribution.
```
mschauer: Not `construct`, rather create a generic wrapper.

src/truncated/normal.jl
```diff
@@ -6,8 +6,9 @@ The *truncated normal distribution* is a particularly important one in the famil
 We provide additional support for this type with `TruncatedNormal` which calls `Truncated(Normal(mu, sigma), l, u)`.
 Unlike the general case, truncated normal distributions support `mean`, `mode`, `modes`, `var`, `std`, and `entropy`.
 """
```
mschauer: Thought: This doc string is valuable, we should preserve it at the end of the deprecation cycle.

src/truncate.jl
matbesancon: yes

src/truncated/normal.jl
matbesancon: yes, I was thinking of defining a new type `TruncatedNormal`, which would implement these methods and keep the doc

src/truncate.jl
matbesancon: done thanks

src/truncate.jl
matbesancon: I added the wrapper term

src/truncated/normal.jl
mschauer: Let's postpone this problem just not forget it.

docs/src/truncate.md
```diff
@@ -1,27 +1,25 @@
...
-    lower::Float64      # lower bound
-    upper::Float64      # upper bound
-    lcdf::Float64       # cdf of lower bound
-    ucdf::Float64       # cdf of upper bound
-
-    tp::Float64         # the probability of the truncated part, i.e. ucdf - lcdf
-    logtp::Float64      # log(tp), i.e. log(ucdf - lcdf)
-end
+The package provides the `truncated` function which creates the most
+appropriate distribution to represente a truncated version of a given
```
mschauer:
```suggestion
appropriate distribution to represent a truncated version of a given
```

src/truncate.jl
```diff
@@ -1,7 +1,41 @@
...
+end
+
+truncated(d::UnivariateDistribution, l::Integer, u::Integer) = truncated(d, float(l), float(u))
+
 """
     Truncated(d, l, u):
 
-Construct a truncated distribution.
+Create a generic wrapper for a truncated distribution.
+Prefer calling the generic `truncated(d, l, u)`.
```
mschauer:
```suggestion
Prefer calling the function `truncated(d, l, u)`, which can choose the appropriate
representation of the truncated distribution.
```

PR #1027, opened 2019-12-04
Add native logpdf for NegativeBinomial

andreasnoack:
This will allow us to use AD when fitting `NegativeBinomoal`.


matbesancon:
wonderful, one day we'll get rid of delegate_rstats altogether. Good to merge?

andreasnoack:
Yes. I think it should be good to go.

PR #1028, opened 2019-12-04
Fix a type instability issue for logpdf(::Categorical)

andreasnoack:
Found another literal `Inf` in the wild. There was also a redundant `logpdf` method for `Categorical`.

### DIFF
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -341,7 +341,7 @@ Relying on this fallback is not recommended in general, as it is prone to overfl
 """
 logpdf(d::UnivariateDistribution, x::Real) = log(pdf(d, x))
 logpdf(d::DiscreteUnivariateDistribution, x::Integer) = log(pdf(d, x))
-logpdf(d::DiscreteUnivariateDistribution, x::Real) = isinteger(x) ? logpdf(d, round(Int, x)) : -Inf
+logpdf(d::DiscreteUnivariateDistribution, x::Real) = isinteger(x) ? logpdf(d, round(Int, x)) : partype(d)(-Inf)
 
 """
     cdf(d::UnivariateDistribution, x::Real)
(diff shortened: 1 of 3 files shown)


matbesancon:
once this passes CI, we can release a new patch

src/univariates.jl
```diff
@@ -341,7 +341,7 @@ Relying on this fallback is not recommended in general, as it is prone to overfl
 """
 logpdf(d::UnivariateDistribution, x::Real) = log(pdf(d, x))
 logpdf(d::DiscreteUnivariateDistribution, x::Integer) = log(pdf(d, x))
-logpdf(d::DiscreteUnivariateDistribution, x::Real) = isinteger(x) ? logpdf(d, round(Int, x)) : -Inf
+logpdf(d::DiscreteUnivariateDistribution, x::Real) = isinteger(x) ? logpdf(d, round(Int, x)) : partype(d)(-Inf)
```
mschauer: Do we have discrete distributions in which `partype` does not give the type in which probabilities are measured but the type of a location parameter say, so an `eltype` actually? E.g. as it is the case with `Normal`. E.g.

```
julia> partype(Binomial(3,0.2f0))
Float32

julia> logpdf(Binomial(3, 0.2f0), 2)
-2.343407144115017 # sic!

julia> partype(Normal(0//1, 1//1))
Rational{Int64}

julia> logpdf(Normal(0//1, 1//1), 1/1)
-1.4189385332046727
```

matbesancon: Oops, think I merged it a tad too early. Should we re-change it to `eltype`?

andreasnoack: I think the `eltype` isn't relevant for `logpdf` in the discrete case. I think `partype` is correct for the `pdf` but the `log` transformation will make it wrong for `logpdf` when the parameters are e.g. `Rational`. I'll try to come up with a fix.

mschauer: Eltype is also wrong. If we make the convention "`partype` is the type of probabilities for discrete distributions" and check it we might heal this.

PR #1029, opened 2019-12-05
Add varargs constructor for Categorical

andreasnoack:
This will make it easier to use the distribution in broadcasting. It made me wonder if we should deprecate the `Categorical(::Int)` constructor since it's tangentially ambiguous (although not ambiguous to the compiler.) Notice that I made this PR on top of the small fix in #1028.

I also wondered if we eventually should change the internal representation of `Categorical` to use a `Tuple` instead of `Vector` and then use the varargs constructor as the only allowed interface. It would both make the distribution much more efficient and, in my opinion, easier to use. The only issue would be if people often use `Categorical`s with a lot of outcomes. In any case, that is a PR for another day.

### DIFF
--- a/src/univariate/discrete/categorical.jl
+++ b/src/univariate/discrete/categorical.jl
@@ -39,6 +39,8 @@ function Categorical(k::Integer; check_args=true)
     return Categorical{Float64,Vector{Float64}}(Base.OneTo(k), fill(1/k, k), check_args=check_args)
 end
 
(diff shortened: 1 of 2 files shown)


mschauer:
Categorical can in the long run just be parametrised by the type of the probability vector/tuple.

mschauer:
I would deprecate it, it is bound to create confusion. Besides that, luckily `Categorical(1)` and `Categorical(1.0)` coincide on the most important case (1 being both a positive integer and a valid probability.)

PR #1030, opened 2019-12-05
Determine return type of discrete pdf/logpdf based on pdf instead

andreasnoack:
of logpdf and test it on Rational.

This follows up on the discussion in https://github.com/JuliaStats/Distributions.jl/pull/1028#discussion_r354214245. (I like that the fixes seem to be a matter of deleting methods).


src/univariates.jl
```diff
@@ -341,7 +341,6 @@ Relying on this fallback is not recommended in general, as it is prone to overfl
 """
 logpdf(d::UnivariateDistribution, x::Real) = log(pdf(d, x))
 logpdf(d::DiscreteUnivariateDistribution, x::Integer) = log(pdf(d, x))
-logpdf(d::DiscreteUnivariateDistribution, x::Real) = isinteger(x) ? logpdf(d, round(Int, x)) : partype(d)(-Inf)
```
matbesancon: Was this method never called or was it never necessary?

test/categorical.jl
```diff
@@ -6,6 +6,7 @@ using Test
 for p in Any[
     [0.5, 0.5],
     [0.5f0, 0.5f0],
+    [1//2, 1//2],
```
matbesancon: The modified tests don't test the new generic zero
`pdf(d::DiscreteUnivariateDistribution, x::Real) = isinteger(x) ? pdf(d, round(Int, x)) : zero(partype(d))`

src/univariates.jl
andreasnoack: It's sufficient to have the conversion logic to `Int` in the `pdf` version which will then be called by the version in line 342.

test/categorical.jl
andreasnoack: I believe they do. The test in https://github.com/JuliaStats/Distributions.jl/blob/c89d557344591a2c046280b96b2597d019457583/test/categorical.jl#L28 will dispatch to https://github.com/JuliaStats/Distributions.jl/blob/c89d557344591a2c046280b96b2597d019457583/src/univariates.jl#L342 which will then dispatch to the modified version of https://github.com/JuliaStats/Distributions.jl/blob/c89d557344591a2c046280b96b2597d019457583/src/univariates.jl#L333. I wrote the test before the fix, so I think it's exercised.

PR #1031, opened 2019-12-06
Fix call to promote_eltype

sethaxen:
`promote_eltype` is not exported by Julia Base.


ararslan:
Great, thanks! Could you add a test that would have error if not for this change?

matbesancon:
Looks good, is there another test that should be done, a bit more specific on the change done?

matbesancon:
bump @sethaxen

matbesancon:
bump here

matbesancon:
Thanks!

PR #1033, opened 2019-12-09
Allow probability vectors other than Float64 vectors for Multinomial

### LINKED ISSUE #1032: Multinomial distribution rand does not support Float32 probability vectors
Is there a reason `multinom_rand` is [restricted to Float64 arrays only](https://github.com/JuliaStats/Distributions.jl/blob/147a3718f2dba86607926c09741e28ae2cb637d7/src/samplers/multinomial.jl#L2) as the probability vector?

darsnack:
Fixes #1032 

Now any `<:Real` array can be used as a vector and `multinom_rand` will still work.

### DIFF
--- a/src/samplers/multinomial.jl
+++ b/src/samplers/multinomial.jl
@@ -1,6 +1,6 @@
 
-function multinom_rand!(n::Int, p::AbstractVector{Float64},
-                        x::AbstractVector{T}) where T<:Real
+function multinom_rand!(n::Int, p::AbstractVector{PT},
+                        x::AbstractVector{T}) where {T<:Real, PT<:Real}
     k = length(p)
     length(x) == k || throw(DimensionMismatch("Invalid argument dimension."))
 
@@ -39,8 +39,8 @@ function multinom_rand!(n::Int, p::AbstractVector{Float64},
     return x
 end
 
-function multinom_rand!(rng::AbstractRNG, n::Int, p::AbstractVector{Float64},
-                        x::AbstractVector{T}) where T<:Real
+function multinom_rand!(rng::AbstractRNG, n::Int, p::AbstractVector{PT},
+                        x::AbstractVector{T}) where {T<:Real, PT<:Real}
     k = length(p)
     length(x) == k || throw(DimensionMismatch("Invalid argument dimension."))
 
--- a/src/samplers/multinomial.jl
+++ b/src/samplers/multinomial.jl
@@ -1,6 +1,5 @@
-
-function multinom_rand!(n::Int, p::AbstractVector{PT},
-                        x::AbstractVector{T}) where {T<:Real, PT<:Real}
+function multinom_rand!(n::Int, p::AbstractVector{<:Real},
+                        x::AbstractVector{<:Real})
     k = length(p)
     length(x) == k || throw(DimensionMismatch("Invalid argument dimension."))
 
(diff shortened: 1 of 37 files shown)


src/samplers/multinomial.jl
```diff
@@ -1,6 +1,6 @@
 
-function multinom_rand!(n::Int, p::AbstractVector{Float64},
-                        x::AbstractVector{T}) where T<:Real
+function multinom_rand!(n::Int, p::AbstractVector{PT},
```
mschauer: `PT` is not needed 
```suggestion
function multinom_rand!(n::Int, p::AbstractVector{<:Real},
```
Actually also `T` is not needed, replace the lines below by
```
    if i == km1
        @inbounds x[k] = n
    else  # n must have been zero
        for j = i+1 : k
            @inbounds x[j] = 0
        end
    end
```

src/samplers/multinomial.jl
```diff
@@ -1,6 +1,6 @@
...
+                        x::AbstractVector{T}) where {T<:Real, PT<:Real}
```
mschauer:
```suggestion
                        x::AbstractVector{T}) where {T<:Real}
```

mschauer:
There are some other commits in this branch I am not sure where they belong.

darsnack:
Those are from a prior PR that has already been merged into master. I pulled the upstream changes since the last PR I made to Distributions, so I am not sure why those are showing up.

mschauer:
Looking over the code environment, what is the point of this method?

https://github.com/JuliaStats/Distributions.jl/blob/a421b6f1a5d33c455ac38a4c8e1e604b40b1155f/src/samplers/multinomial.jl#L1

darsnack:
> Looking over the code environment, what is the point of this method?
> 
> https://github.com/JuliaStats/Distributions.jl/blob/a421b6f1a5d33c455ac38a4c8e1e604b40b1155f/src/samplers/multinomial.jl#L1

To facilitate direct calls to `rand` without specifying a sampler or RNG? https://github.com/JuliaStats/Distributions.jl/blob/a421b6f1a5d33c455ac38a4c8e1e604b40b1155f/src/multivariate/multinomial.jl#L165

mschauer:
I'm optimistic that we can replace it with a one-liner, either right away or I'll file an issue.

matbesancon:
Could you add a test to validate that it works with other real types (Float32, `ForwardDiff.Dual{Float64}` or anything you like)

matbesancon:
bump

darsnack:
Sorry, yes I can add those test. I’m currently traveling, so it might be a few days before I get to it.

darsnack:
So this doesn't work for `ForwardDiff.Dual{Float64}` because of how the `BinomialGeomSampler` is written. The issue is similar — the method is only defined for `Float64` probabilities, but I am uncertain whether any `Real` is valid in this case (w.r.t. to numerical stability).

Bottomline is that the fix works for `Float32` or any real type that can be automatically converted to `Float64`. Any real type that requires explicit conversion like `Dual{Float64}` will fail due to `BinomialGeomSampler`/`BinomialTPESampler`.

mschauer:
Can you add the test for `Float32` then? Then we can merge and take already a bite out of the problem. 👍

test/multinomial.jl
```diff
@@ -207,3 +207,9 @@ p_v = [0.1, 0.4, 0.3, 0.8]
 @test_throws ArgumentError Multinomial(10, p_v)
 @test_throws ArgumentError Multinomial(10, p_v; check_args=true)
 Multinomial(10, p_v; check_args=false) # should not warn
+
+# check different prob vector types
+p = [0.2, 0.4, 0.3, 0.1]
+@test (rand(Multinomial(10, p)); true)
+@test (rand(Multinomial(10, convert.(Float32, p))); true)
+# @test (rand(Multinomial(10, convert.(ForwardDiff.Dual{Float64}, p))); true)
```
matbesancon: should the commented out test be removed?

matbesancon:
```suggestion
```

darsnack: I had left it in to revisit the issue in a different issue/PR, but yeah makes more sense to take it out.

darsnack:
Bump

darsnack:
I'm not sure whether this PR or the original issue are relevant anymore, but I was going through my old PRs that never got merged and found this one.

Is this something that's missing and we still want? I can rebase and clean it up for merging if so.

matbesancon:
Yes not sure why we forgot this (on my side the summer was crazy), if you can rebase on master it should be good to go, thanks!

PR #1035, opened 2019-12-14
Uniform zeros like syntax

aminya:
```julia
julia> Uniform(Float32,-1,1)
Uniform{Float32}(a=-1.0f0, b=1.0f0)
```

### DIFF
--- a/src/univariate/continuous/uniform.jl
+++ b/src/univariate/continuous/uniform.jl
@@ -38,6 +38,10 @@ Uniform(a::Real, b::Real) = Uniform(promote(a, b)...)
 Uniform(a::Integer, b::Integer) = Uniform(float(a), float(b))
 Uniform() = Uniform(0.0, 1.0, check_args=false)
 
+function Uniform(::Type{T}, a, b) where {T <: Real}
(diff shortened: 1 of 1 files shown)


matbesancon:
The feature is interesting. Wasn't there anything already in the package doing this?
Before merging that, could you add a bunch of tests and docstring to Uniform explaining the behaviour?

andreasnoack:
I think `Uniform{Float32}(-1, 1)` would be more in line with how we typically do this, e.g.
```julia
julia> Complex{Float32}(1,1)
1.0f0 + 1.0f0im
```

aminya:
> 
> 
> I think `Uniform{Float32}(-1, 1)` would be more in line with how we typically do this, e.g.
> 
> ```julia
> julia> Complex{Float32}(1,1)
> 1.0f0 + 1.0f0im
> ```

That is one of the APIs that I specified in this issue. However, we need to first fix the type stability. The issue is bigger than the API

aminya:
Proceeded by https://github.com/JuliaStats/Distributions.jl/pull/1045

PR #1036, opened 2019-12-14
Include LogNormal in the list of distributions for fit_mle

non-Jedi:


matbesancon:
@non-Jedi thanks! Did you check that the documentation renders correctly?

non-Jedi:
I have not. Doing so would honestly involve a lot more effort than I'm willing to put into a trivial doc fix like this.

matbesancon:
Verifying that the changes do what you wanted (not adding a dead link for instance) is mostly for not putting the burden on the maintainer. I checked and it passed

non-Jedi:
Thank you

PR #1038, opened 2019-12-17
Add logdiffcdf function

c123w:
Some time ago I needed to evaluate `logdiffcdf(x, y) = log(cdf(x) - cdf(y))` where `x` and `y` are much less than 0 and `x-y` is very close to 0. I couldn't immediately see an existing libary function to do this, so I wrote one. Example usage:

```julia
a1, b1 = -30.0, -30.001
d = Normal()
println(logdiffcdf(d, a1, b1))
println(log(cdf(d, a1) - cdf(d, b1)))
println(" ")

a2, b2 = -60.0, -60.001
println(logdiffcdf(d, a2, b2))
println(log(cdf(d, a2) - cdf(d, b2)))
println(log(cdf(d, BigFloat(a2)) - cdf(d, BigFloat(b2))))
```

```julia
-457.8416564778862
-457.8416564778832
 
-1807.856543980857
-Inf
-1807.856543980860619228775160138268961556646600503702597980758365928321022814345
```

I suppose this could also be generalized to multivariate distributions (log of the integral of the cdf under the hypercuboid with opposite vertices at the 1D arrays `x`, `y`).

I was wondering what other think of this (or if there's already open source Julia code to do it).


src/univariates.jl
```diff
@@ -397,6 +397,17 @@ logcdf(d::UnivariateDistribution, x::Real) = log(cdf(d, x))
...
 
+"""
+    logdiffcdf(d::Distribution, x::T, y::T) where {T <: Real}
+
+The logarithm of the difference between the cumulative density function at `x` and `y`, i.e. `log(cdf(x) - cdf(y))`.
+"""
+function logdiffcdf(d::Distribution, x::T, y::T) where {T <: Real}
+    x <= y && error("Requires x > y.")
+    u, v = logcdf(d, x), logcdf(d, y)
+    return u + log(convert(T,1.0) - exp(v-u))
```
andreasnoack:
```suggestion
    return u + log1p(-exp(v - u))
```

src/univariates.jl
```diff
@@ -397,6 +397,17 @@ logcdf(d::UnivariateDistribution, x::Real) = log(cdf(d, x))
 logcdf(d::DiscreteUnivariateDistribution, x::Integer) = log(cdf(d, x))
 logcdf(d::DiscreteUnivariateDistribution, x::Real) = logcdf(d, floor(Int,x))
 
+"""
+    logdiffcdf(d::Distribution, x::T, y::T) where {T <: Real}
+
+The logarithm of the difference between the cumulative density function at `x` and `y`, i.e. `log(cdf(x) - cdf(y))`.
+"""
+function logdiffcdf(d::Distribution, x::T, y::T) where {T <: Real}
+    x <= y && error("Requires x > y.")
```
andreasnoack:
```suggestion
    x <= y && throw(ArgumentError("requires x > y."))
```

PR #1042, opened 2019-12-22
use logmvbeta in MatrixBeta and MatrixFDist

johnczito:
JuliaStats/StatsFuns.jl#72 has been merged and released, so it can be used when computing the integrating constants of the Matrix-Beta and Matrix-F distributions.


matbesancon:
Thanks!
The two functions modified are already tested right?

johnczito:
Yes.

Issue #1041, opened 2019-12-21
Uniform is not type stable


aminya:
The current definition of `Uniform` is not type stable. I tried to make a method in https://github.com/JuliaStats/Distributions.jl/pull/1035/files to make it usable, but it doesn't help.

```julia
using Distributions
set = Uniform(Float32(0.0), Float32(2.0))
v = rand(set, 3)
```
```julia
julia> eltype(v)
Float64
```

The definition should not be dependent on the type of `a` and `b`. 
https://github.com/JuliaStats/Distributions.jl/blob/master/src/univariate/continuous/uniform.jl#L26

Instead user should provide the type explicitly. It should use `Float64` if no type is provided.

The following definition is my proposed definition:
for array like constructor `(Uniform{T}(a, b))`
```julia
struct Uniform{T<:Real} <: ContinuousUnivariateDistribution
    a
    b
    Uniform{T}(a, b) where {T <: Real} = new{T}(a, b)
end
```
or zeros like constructor `(Uniform(T, a, b))`
```
function Uniform(::Type{T}, a, b) where {T <: Real}
    return Uniform{T}(a, b)
end
```

andreasnoack:
This isn't really what is usually called type stability. `rand(::Uniform)` is perfectly type stable according the usual definition. The problem is that we hardcode for `Float64`. Historically, the package only supported `Float64` and we have gradually moved towards generic element support. It looks like `Uniform` still relies on `Rmath` which only supports `Float64` but the necessary methods should be fairly simple for `Uniform` so making it fully should be doable without too much effort. However, I don't think the `struct` definition needs to change to achieve this.

aminya:
> 
> 
> This isn't really what is usually called type stability. `rand(::Uniform)` is perfectly type stable according the usual definition. The problem is that we hardcode for `Float64`. Historically, the package only supported `Float64` and we have gradually moved towards generic element support. It looks like `Uniform` still relies on `Rmath` which only supports `Float64` but the necessary methods should be fairly simple for `Uniform` so making it fully should be doable without too much effort. However, I don't think the `struct` definition needs to change to achieve this.

The current definition imposes some restrictions on the API that are not necessary. Removing `::T` anotation from struct defintion opens up the door for a better API. 

In that case, all these redundant defenitions can be removed:
https://github.com/JuliaStats/Distributions.jl/blob/master/src/univariate/continuous/uniform.jl#L37-L38

andreasnoack:
[...]

PR #1045, opened 2019-12-31
Uniform overhaul

aminya:
Fixes https://github.com/JuliaStats/Distributions.jl/issues/1041

+ rand(Uniform) is now type stable 
+ Bounds types are independent of distribution type

Alternative branch: [UniformOverhaul2](https://github.com/aminya/Distributions.jl/blob/UniformOverhaul2/src/univariate/continuous/uniform.jl)

### DIFF
--- a/src/univariate/continuous/uniform.jl
+++ b/src/univariate/continuous/uniform.jl
@@ -118,7 +118,7 @@ end
 
 #### Sampling
 
-rand(rng::AbstractRNG, d::Uniform) = d.a + (d.b - d.a) * rand(rng)
+rand(rng::AbstractRNG, d::Uniform{T}) where {T} =  convert(T, d.a + (d.b - d.a) * rand(rng))
 
 
 #### Fitting
--- a/src/univariate/continuous/uniform.jl
+++ b/src/univariate/continuous/uniform.jl
@@ -24,25 +24,35 @@ External links
 
 """
 struct Uniform{T<:Real} <: ContinuousUnivariateDistribution
-    a::T
-    b::T
-    Uniform{T}(a::T, b::T) where {T <: Real} = new{T}(a, b)
+    a::T1 where {T1 <: Real}
+    b::T2 where {T2 <: Real}
+
+    # Uniform{T}(a, b) constructor
+    function Uniform{T}(a, b; check_args=true) where {T <: Real}
+        check_args && @check_args(Uniform, a < b)
+        new{T}(a, b)
+    end
 end
 
-function Uniform(a::T, b::T; check_args=true) where {T <: Real}
+# zeros() like constructor (Uniform(T, a, b))
+function Uniform(::Type{T}, a, b; check_args=true) where {T <: Real}
     check_args && @check_args(Uniform, a < b)
     return Uniform{T}(a, b)
 end
 
-Uniform(a::Real, b::Real) = Uniform(promote(a, b)...)
-Uniform(a::Integer, b::Integer) = Uniform(float(a), float(b))
+# No type specified constructor:
+function Uniform(a, b; check_args=true)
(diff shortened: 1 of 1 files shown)


aminya:
If you think that we should keep the type of `a` and `b` the same internally, this definition also works:

[UniformOverhaul2](https://github.com/aminya/Distributions.jl/blob/UniformOverhaul2/src/univariate/continuous/uniform.jl)

```julia
struct Uniform{T<:Real} <: ContinuousUnivariateDistribution
    a::T
    b::T

    # Uniform{T}(a::T, b::T) constructor
    function Uniform{T}(a::T, b::T; check_args=true) where {T <: Real}
        check_args && @check_args(Uniform, a < b)
        new{T}(a, b)
    end
end

# Abstract a,b - Uniform{T}(a, b) constructor
function Uniform{T}(a::Real, b::Real; check_args=true) where {T <: Real}
    check_args && @check_args(Uniform, a < b)
    new{T}(T(a), T(b))
end

# zeros() like constructor (Uniform(T, a::T, b::T))
function Uniform(::Type{T}, a::T, b::T; check_args=true) where {T <: Real}
    check_args && @check_args(Uniform, a < b)
    return Uniform{T}(a, b)
end

# Abstract a,b - zeros() like constructor (Uniform(T, a, b))
function Uniform(::Type{T}, a, b; check_args=true) where {T <: Real}
    check_args && @check_args(Uniform, a < b)
    return Uniform{T}(T(a), T(b))
end

# No type specified constructor:
function Uniform(a::Float64, b::Float64; check_args=true)
    check_args && @check_args(Uniform, a < b)
    return Uniform{Float64}(a, b)
end

# Abstract a,b - no type specified constructor:
function Uniform(a, b; check_args=true)
    check_args && @check_args(Uniform, a < b)
    return Uniform{Float64}(Float64(a), Float64(b))
end
```

andreasnoack:
I don't support the changes here. Adding the a `Uniform{T}(a::Real, b::Real) = Uniform(T(a), T(b))` constructor seems fine but I don't see the need for the remaining changes.

#1041 raises a discussion about the type returned by `rand`. I don't think we have made a decision about what the behavior should be but it seems like we currently return `Float64` for most distributions regardless of the parameter type. An exception is the `Normal` distribution where we `rand` uses the type of the parameters. I think the difference is unintended.

We should probably open a separate issue for discussing this or simply change the title of #1041 to cover all distributions. I see four possible behaviors for `rand`

1. a. Unconditionally construct `Float64`s
1. b. Add a type argument to `rand` to specify the variate type but default to `Float64`
2. Return variates of the same type as the parameters
3. Introduce a separate type parameter for the variate type

I think 3. will add too much overhead without a real demand. I think I'd prefer one of the 1. options.

cossio:
I think 1a (always returning `Float64`) might not be the best solution. Sometimes we want to work with `Float32` (e.g. in neural nets this is common because matrix multiplies can be faster). 

2b (adding a type parameter to `rand`) could work.

But it seems that 3., returning types following the parameters, is the most natural solution here. I think this is what users expect (a couple of issues are requesting just this). I don't understand why this would be much more difficult than the 2b. approach?

rafaqz:
My five cents is also that option 3 is the most natural solution. I just attempted to switch a complex simulation over to Float32 for performance and Distributions.jl was the only blocker to that just working with no code changes. 

Having to add an argument specifically to allow Float32 seems unnecessary both for users and devs, when the parameters already contain the type information.

devmotion:
I'm closing this PR since we don't intend to incorporate these changes (see https://github.com/JuliaStats/Distributions.jl/pull/1045#issuecomment-571559677). The `rand`/`eltype` discussion is covered extensively in other issues and PRs such as #1905.

Issue #984, opened 2019-10-01
Type inconsistencies


cscherrer:
There are some serious type inconsistencies here:

```julia
julia> b = Bernoulli()
Bernoulli{Float64}(p=0.5)

julia> rand(b)
false

julia> rand(b,1)
1-element Array{Int64,1}:
 0

julia> eltype(b)
Int64
```

Can we please have `rand(dist) :: T` imply

- `rand(dist, n::Int) :: Array{T}`
- `eltype(dist) == T`
?

matbesancon:
Both fixes are acceptable (moving everything to Bool or to Int)

cscherrer:
Thanks @matbesancon.

`Bool` seems a better fit to me, any objections (from anyone)?

Any thoughts on `rand(dist, n::Int) :: Array{T}`? This would require significant breaking changes for multivariate and matrix-variate distributions, but to me it seems worth it to avoid having to consider special cases.

I can understand the reasoning for the current approach: a regular array-of-vectors is more efficiently stored in a matrix. But the lack of regularity in the interface is a bit painful. It seems like we need a way to store something as a matrix but access it as an array of vectors. I thought I had seen a library that provides this capability, but I'm having trouble finding it right now. Any ideas?

matbesancon:
> Bool seems a better fit to me, any objections (from anyone)?

I'm good with Bool on this, the fact that Bool is an Integer is helpful here.

> Any thoughts on rand(dist, n::Int) :: Array{T}? This would require significant breaking changes for multivariate and matrix-variate distributions, but to me it seems worth it to avoid having to consider special cases.

We have to keep in mind that the standard method is always `rand(RNG, ...)`, which fallbacks to `rand(Random.GLOBAL_RNG, ...)`. The new rand interface would be:
```julia
rand(RNG, dist::Distribution, n::Integer) :: AbstractArray{T}
```

The `AbstractArray` subtype used is specialized for the distribution. 

> It seems like we need a way to store something as a matrix but access it as an array of vectors. I thought I had seen a library that provides this capability, but I'm having trouble finding it right now. Any ideas?

[RecursiveArrayTools.jl](https://github.com/JuliaDiffEq/RecursiveArrayTools.jl) could do the job

cscherrer:
> [RecursiveArrayTools.jl](https://github.com/JuliaDiffEq/RecursiveArrayTools.jl) could do the job

[...]

PR #1046, opened 2020-01-02
Fix type of rand for Bernoulli

juliohm:
Fixes the type of `rand(Bernoulli())` to 1/0 like in other functions. Currently it is inconsistent:

```julia
using Distributions

rand(Bernoulli()) # true/false
rand(Bernoulli(), 1) # 1/0
```

### DIFF
--- a/src/univariate/discrete/bernoulli.jl
+++ b/src/univariate/discrete/bernoulli.jl
@@ -104,7 +104,7 @@ cf(d::Bernoulli, t::Real) = failprob(d) + succprob(d) * cis(t)
 
 #### Sampling
 
-rand(rng::AbstractRNG, d::Bernoulli) = rand(rng) <= succprob(d)
+rand(rng::AbstractRNG, d::Bernoulli) = rand(rng) <= succprob(d) ? 1 : 0
 
 #### MLE fitting
 


matbesancon:
fixes #984
Following the discussion on the issue, I think putting everything to `Bool` is preferred.

juliohm:
I think putting everything to `Int` would be more appealing mathematically. It is somewhat weird to have the concept of a distribution rely on a computer science type. That is why I suggest to pick 1/0 instead.

@cscherrer would you be fine with 1/0 instead of true/false? I understand that Bernoulli specifically is used in many boolean contexts, but I find the decision of adopting `Bool` a little suboptimal for modeling physical phenomena, sequences and series of numbers, etc. One can always do `rand() < p` in an if-else statement to handle the true/false use case.

matbesancon:
yes I get what you mean. In Julia Bool is tricky because it is also a numeric type

juliohm:
Yes, I understand that `Bool` is numeric in the language, but it feels unnatural to define a Bernoulli in terms of it. Moreover, if some hardware is constrained to Int64 or Int32, codes will have to cast the random draws to work. I am not sure this is actually true that hardware is constrained to fixed size integers, but certainly they can be much faster sometimes when the right integer is passed. I am wondering how Bool for example plays with quantum processors, and future technologies.

Anyways, I keep my opinion that 1/0 is much more natural than true/false for a Bernoulli random variable.

matbesancon:
I don't have a strong opinion on the subject, and I think nobody had on the initial issue, fine to go with 1/0. Could you add a test validating the change, something like:
```julia
a = rand(Bernoulli(), 1)
b = rand(Bernoulli())
@test a === b || a === (1-b) 
```

juliohm:
Sure, I am leaving my computer for an hour or so, but will add a test to the PR when I am back.

juliohm:
Closing this in favor of #1048

PR #1048, opened 2020-01-03
Fix type of Bernoulli sample type

juliohm:
Includes #1046 . I messed up with Git again, so creating a new PR from scratch.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -55,6 +55,7 @@ const tests = [
     "chernoff",
     "univariate_bounds",
     "negativebinomial",
+    "bernoulli"
 ]
 
 printstyled("Running tests:\n", color=:blue)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -55,7 +55,7 @@ const tests = [
     "chernoff",
     "univariate_bounds",
     "negativebinomial",
-    "bernoulli"
+    "bernoulli",
 ]
 
 printstyled("Running tests:\n", color=:blue)
--- a/src/univariate/discrete/bernoulli.jl
+++ b/src/univariate/discrete/bernoulli.jl
@@ -104,7 +104,7 @@ cf(d::Bernoulli, t::Real) = failprob(d) + succprob(d) * cis(t)
 
 #### Sampling
 
-rand(rng::AbstractRNG, d::Bernoulli) = rand(rng) <= succprob(d)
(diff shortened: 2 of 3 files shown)


test/runtests.jl
```diff
@@ -55,6 +55,7 @@ const tests = [
     "chernoff",
     "univariate_bounds",
     "negativebinomial",
+    "bernoulli"
```
matbesancon:
```suggestion
    "bernoulli",
```

matbesancon:
Thanks!

MikeInnes:
Just got bitten by this change (which probably should have been treated as breaking).

This seems like a very strange decision to me. `Bool` is useful because it's the narrowest type that represents what `Bernoulli` can output (i.e. `0` or `1`). It's additionally useful because it can be used as a condition, which is pretty frequent in probabilistic models; I'd rather not write `Bool(rand(Bernoulli(p)))` or `rand(Bernoulli(p))) == 1` every time there. (There's no similar downside to emitting `Bool` since that can always be casted to a richer type automatically.)

I'd strongly be in favour of reverting this (the quantum processors can likely look after themselves.)

matbesancon:
Indeed I made the mistake of tagging post-merge as a patch release.

As for the preferred type, as mentioned I had no strong opinion for either,
Bool being more handy but using it as Int is still a very Julian thing.

I'll let Julio give his opinion, but would be ok with switching to Bool, as
long as we maintain consistency (which was broken before this PR)

matbesancon:
cc @juliohm

juliohm:
Sorry for the late reply, just saw this in my inbox today.

It seems strange to me to rely on a computer science type to define
Bernoulli variables. Can you please confirm that performing operations with
large Boolean arrays representing samples of various Bernoulli does not
overflow nor slows down computations in practice? Part of my concern is
that we want to do math with the 0s and 1s in this context. It is not
Boolean semantic that we are doing here. In most cases where Bernoulli is
being used as a check rand(Bernoulli(p)) == 1,  one could actually write
the much simpler rand() < p. I still think that using Bool here is strange,
my vote is in favor of Int.

matbesancon:
> samples of various Bernoulli does not overflow nor slows down computations in practice

any arithmetic on booleans converts them to Int, so no overflow: `true + true == 2`

juliohm:
So this means conversions take place? We loose performance then correct ?

matbesancon:
I don't think the cost here is awful, if there is any

PR #1051, opened 2020-01-07
Add affine transformations for MvNormal.

### LINKED ISSUE #307: linear transformations on MvNormal

tpapp:
Fixes #307.

### DIFF
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -65,6 +65,13 @@ const ZeroMeanIsoNormal{Axes}  = MvNormal{ScalMat,  Zeros{Float64,1,Axes}}
 const ZeroMeanDiagNormal{Axes} = MvNormal{PDiagMat, Zeros{Float64,1,Axes}}
 const ZeroMeanFullNormal{Axes} = MvNormal{PDMat,    Zeros{Float64,1,Axes}}
 ```
+
+Multivariate normal distributions support affine transformations:
+```julia
+d = MvNormal(μ, Σ)
+c + B * d    # == MvNormal(B * μ, c + B * Σ * B')
+dot(b, d)    # == Normal(dot(b, μ), b' * Σ * b)
+```
 """
 abstract type AbstractMvNormal <: ContinuousMultivariateDistribution end
 
@@ -270,6 +277,18 @@ function _rand!(rng::AbstractRNG, d::MvNormal, x::AbstractVector)
     end
     add!(unwhiten!(d.Σ, x), d.μ)
 end
+
+### Affine transformations
+
++(c::AbstractVector, d::MvNormal) = MvNormal(d.μ .+ c, d.Σ)
+
+@inline +(d::MvNormal, c::AbstractVector) = c + d
+
+*(B::AbstractMatrix, d::MvNormal) = MvNormal(B * d.μ, X_A_Xt(d.Σ, B))
+
+dot(b::AbstractVector, d::MvNormal) = Normal(dot(d.μ, b), √quad(d.Σ, b))
+
+@inline dot(d::MvNormal, b::AbstractVector) = dot(b, d)
 
 ###########################################################
 #
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -280,15 +280,15 @@ end
 
 ### Affine transformations
 
-+(c::AbstractVector, d::MvNormal) = MvNormal(d.μ .+ c, d.Σ)
++(d::MvNormal, c::AbstractVector) = MvNormal(d.μ .+ c, d.Σ)
 
-@inline +(d::MvNormal, c::AbstractVector) = c + d
++(c::AbstractVector, d::MvNormal) = d + c
 
 *(B::AbstractMatrix, d::MvNormal) = MvNormal(B * d.μ, X_A_Xt(d.Σ, B))
 
 dot(b::AbstractVector, d::MvNormal) = Normal(dot(d.μ, b), √quad(d.Σ, b))
 
-@inline dot(d::MvNormal, b::AbstractVector) = dot(b, d)
+dot(d::MvNormal, b::AbstractVector) = dot(b, d)
 
 ###########################################################
 #
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -5,13 +5,13 @@ using StatsFuns: logtwo, invsqrt2, invsqrt2π
 
 import QuadGK: quadgk
 import Base: size, length, convert, show, getindex, rand, vec, inv
-import Base: sum, maximum, minimum, extrema, +, -, ==
+import Base: sum, maximum, minimum, extrema, +, -, *, ==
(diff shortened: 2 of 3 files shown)


src/multivariate/mvnormal.jl
```diff
@@ -65,6 +65,13 @@ const ZeroMeanIsoNormal{Axes}  = MvNormal{ScalMat,  Zeros{Float64,1,Axes}}
 const ZeroMeanDiagNormal{Axes} = MvNormal{PDiagMat, Zeros{Float64,1,Axes}}
 const ZeroMeanFullNormal{Axes} = MvNormal{PDMat,    Zeros{Float64,1,Axes}}
 ```
+
+Multivariate normal distributions support affine transformations:
+```julia
+d = MvNormal(μ, Σ)
```
matbesancon: Maybe give more explicit names to c and B to indicate these are constant vectors and matrices respectively

src/multivariate/mvnormal.jl
```diff
@@ -271,6 +278,18 @@ function _rand!(rng::AbstractRNG, d::MvNormal, x::AbstractVector)
     add!(unwhiten!(d.Σ, x), d.μ)
 end
 
+### Affine transformations
+
++(c::AbstractVector, d::MvNormal) = MvNormal(d.μ .+ c, d.Σ)
+
+@inline +(d::MvNormal, c::AbstractVector) = c + d
```
matbesancon: Is the inline needed here?
```julia
@inline +(d::MvNormal, c::AbstractVector) = c + d
```

src/multivariate/mvnormal.jl
```diff
@@ -271,6 +278,18 @@ function _rand!(rng::AbstractRNG, d::MvNormal, x::AbstractVector)
...
+
+*(B::AbstractMatrix, d::MvNormal) = MvNormal(B * d.μ, X_A_Xt(d.Σ, B))
+
```
matbesancon: Where are `X_A_Xt` and `√quad` coming from? Could we use a qualified name instead? `Module.X_A_Xt`

src/multivariate/mvnormal.jl
```diff
@@ -65,6 +65,13 @@ const ZeroMeanIsoNormal{Axes}  = MvNormal{ScalMat,  Zeros{Float64,1,Axes}}
 const ZeroMeanDiagNormal{Axes} = MvNormal{PDiagMat, Zeros{Float64,1,Axes}}
 const ZeroMeanFullNormal{Axes} = MvNormal{PDMat,    Zeros{Float64,1,Axes}}
 ```
+
+Multivariate normal distributions support affine transformations:
+```julia
+d = MvNormal(μ, Σ)
+c + B * d    # == MvNormal(B * μ, c + B * Σ * B')
```
mschauer:
```suggestion
c + B * d    # == MvNormal(B * μ + c, B * Σ * B')
```

src/multivariate/mvnormal.jl
tpapp: I just added it because I was trying to follow the coding style of the package, which uses it pretty liberally. I don't think it is harmful.

src/multivariate/mvnormal.jl
tpapp: It is coming from `PDMats`. No other symbols from that package are qualified in the code, why should these two be an exception?

src/multivariate/mvnormal.jl
tpapp: I think this is pretty obvious from the context, but if you have a suggestion I am happy to change them.

src/multivariate/mvnormal.jl
matbesancon: > No other symbols from that package are qualified in the code

This for a start is bad practice on our side, I wanted to fix it once by explicitly importing things but others contributors were not fond of it.

It is non-trivial that these come from PDMats,.jl unlike, say `PDMat`

src/multivariate/mvnormal.jl
matbesancon: indeed not harmful. I removed a few of them, which were added I believe in a time and age when the compiler was less smart

matbesancon: Also, one thing that I find harder to read: could we have `+(d, c)` be the "main" function and `+(c, d) = +(d, c)`? Having the "important argument first is easier to parse IMO

src/multivariate/mvnormal.jl
tpapp: When I make a PR to a project, I try to follow existing coding style. While I don't necessarily disagree about the style suggestions you are making, I feel that the proper approach is to make these changes in the package first, perhaps clarify coding style (with a link to one of the existing ones like https://github.com/jrevels/YASGuide), and *then* ask for PRs to follow them. 

Conversely, asking for a coding style clearly not followed by a package in PRs is somewhat unfriendly, even when meant well, because it results in endless minor revisions trying to conform to a style that is impossible to predict in advance. I could of course just fix things up, but since I find this tiresome, my policy these days is just to close the PR when this happens.

I am not trying to be difficult here — I am totally fine with people branching and modifying this code, or having it rejected.

matbesancon: Again, this is not an issue with the code in this PR, and this comment is minor enough not to be blocking the PR. What we have to keep in mind is that unlike many packages, Distributions is not "owned" by a very core team. If you look at the last commits for maybe one year, many have been done by one-time contributors, or rare contributors.
Lots of parts are really old and keeping track of it all is hard to keep track of everything, some things that were valid for 5 contributors always working on it is not anymore with the new setting, thus trying to switch to more explicit style.
For what I mentioned earlier, this PR https://github.com/JuliaStats/Distributions.jl/pull/778 was the first attempt

src/multivariate/mvnormal.jl
matbesancon: bump

tpapp: thanks for the bump, done

matbesancon: thanks, I think it's good to merge once the CI passes

matbesancon:
feature released in the last version

PR #1052, opened 2020-01-09
Explicit import

matbesancon:
Following the discussion in #1051, making imports explicit whenever possible, and avoiding the blanket `using`

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -21,7 +21,7 @@ PDMats = "0.9"
 QuadGK = "2"
 SpecialFunctions = "0.8, 0.9"
 StatsBase = "0.32"
-StatsFuns = "0.8, 0.9"
+StatsFuns = ">= 0.9.3"
 julia = "1"
 
 [extras]
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.22.0"
+version = "0.22.1"
 
 [deps]
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
--- a/Project.toml
+++ b/Project.toml
@@ -21,7 +21,7 @@ PDMats = "0.9"
 QuadGK = "2"
 SpecialFunctions = "0.8, 0.9"
 StatsBase = "0.32"
-StatsFuns = ">= 0.9.3"
+StatsFuns = "0.9.3"
 julia = "1"
 
 [extras]
--- a/src/matrix/wishart.jl
+++ b/src/matrix/wishart.jl
@@ -104,7 +104,7 @@ function meanlogdet(d::Wishart)
     df = d.df
     v = logdet(d.S) + p * logtwo
     for i = 1:p
-        v += digamma(0.5 * (df - (i - 1)))
+        v += SFunc.digamma(0.5 * (df - (i - 1)))
     end
     return v
 end
--- a/src/matrix/wishart.jl
+++ b/src/matrix/wishart.jl
@@ -104,7 +104,7 @@ function meanlogdet(d::Wishart)
     df = d.df
     v = logdet(d.S) + p * logtwo
     for i = 1:p
-        v += SFunc.digamma(0.5 * (df - (i - 1)))
+        v += digamma(0.5 * (df - (i - 1)))
     end
     return v
 end
--- a/src/matrix/wishart.jl
+++ b/src/matrix/wishart.jl
@@ -64,12 +64,12 @@ end
 insupport(::Type{Wishart}, X::Matrix) = isposdef(X)
 insupport(d::Wishart, X::Matrix) = size(X) == size(d) && isposdef(X)
 
-dim(d::Wishart) = dim(d.S)
+PDMats.dim(d::Wishart) = dim(d.S)
 size(d::Wishart) = (p = dim(d); (p, p))
 size(d::Wishart, i) = size(d)[i]
 rank(d::Wishart) = dim(d)
 params(d::Wishart) = (d.df, d.S, d.c0)
-@inline partype(d::Wishart{T}) where {T<:Real} = T
+partype(::Wishart{T}) where {T} = T
 
 ### Conversion
 function convert(::Type{Wishart{T}}, d::Wishart) where T<:Real
@@ -138,7 +138,7 @@ end
 #### Sampling
 function _rand!(rng::AbstractRNG, d::Wishart, A::AbstractMatrix)
     _wishart_genA!(rng, dim(d), d.df, A)
-    unwhiten!(d.S, A)
+    PDMats.unwhiten!(d.S, A)
     A .= A * A'
 end
 
@@ -151,7 +151,7 @@ function _wishart_genA!(rng::AbstractRNG, p::Int, df::Real, A::AbstractMatrix)
     #               ~ Normal()                  when i > j
     #
     A .= zero(eltype(A))
-    for i = 1:p
+    for i in 1:p
         @inbounds A[i,i] = rand(rng, Chi(df - i + 1.0))
     end
     for j in 1:p-1, i in j+1:p
--- a/src/multivariate/mvlognormal.jl
+++ b/src/multivariate/mvlognormal.jl
@@ -27,14 +27,17 @@ abstract type AbstractMvLogNormal <: ContinuousMultivariateDistribution end
 
 function insupport(::Type{D},x::AbstractVector{T}) where {T<:Real,D<:AbstractMvLogNormal}
     for i=1:length(x)
-      @inbounds 0.0<x[i]<Inf ? continue : (return false)
+        @inbounds 0.0<x[i]<Inf ? continue : (return false)
     end
-    true
+    return true
 end
+
 insupport(l::AbstractMvLogNormal,x::AbstractVector{T}) where {T<:Real} = insupport(typeof(l),x)
 assertinsupport(::Type{D},m::AbstractVector) where {D<:AbstractMvLogNormal} = @assert insupport(D,m) "Mean of LogNormal distribution should be strictly positive"
 
-###Internal functions to calculate scale and location for a desired average and covariance
+"""
+Internal functions to calculate scale and location for a desired average and covariance
+"""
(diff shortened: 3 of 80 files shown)


Project.toml
```diff
@@ -21,7 +21,7 @@ PDMats = "0.9"
 QuadGK = "2"
 SpecialFunctions = "0.8, 0.9"
 StatsBase = "0.32"
-StatsFuns = "0.8, 0.9"
+StatsFuns = ">= 0.9.3"
```
matbesancon: this is required by the use of a function in StatsFuns

mschauer:
You also added some cosmetic improvements or is this just showing up because of the merge?

matbesancon:
> You also added some cosmetic improvements or is this just showing up because of the merge?

What is showing up?

matbesancon:
if related to the coverage diff, no idea why, all changes done are strictly equivalent

src/matrix/wishart.jl
```diff
@@ -151,7 +151,7 @@ function _wishart_genA!(rng::AbstractRNG, p::Int, df::Real, A::AbstractMatrix)
     #               ~ Normal()                  when i > j
     #
     A .= zero(eltype(A))
-    for i = 1:p
+    for i in 1:p
```
mschauer: @matbesancon Changes like these show up in my diff.

matbesancon: Yes this was added you're right

nalimilan:
I'm in favor of stopping unqualified `using X` statements (since breakage can happen when packages export new functions), but I'm quite reserved about qualifying all uses of functions. That's especially the case for SpecialFunctions, which is used all over the place. If you have a doubt it's very easy to find where it comes from by looking at the `using` or `import` line in src/Distributions.jl: that's another advantage of using explicit imports.

@ararslan Is that your reasoning as well?

Project.toml
nalimilan: Then the new convention is this:
```suggestion
StatsFuns = "0.9.3"
```

CompatHelper is there to bump the dependency when a new version is tagged.

matbesancon:
Thanks for your feedback, indeed for SpecialFunctions it might be better to do `using SF: f1, f2` instead, these are often the same ones coming back. I tried the qualified approach after the feedback on https://github.com/JuliaStats/Distributions.jl/pull/778, where people didn't seem enthusiastic about specified using

nalimilan:
Let's hear what @ararslan and @simonbyrne say then.

matbesancon:
I would be for `using SF: a, b` for things we have all over the place. For  others though, having one or two qualified imports in the code base helps a lot knowing where to look things up

ararslan:
> @ararslan Is that your reasoning as well?

No, I'm a big proponent of simply `using X`. I especially dislike qualifying every call.

nalimilan:
The problem with general `using X` is that if X starts exporting a new function, it can clash with other identifiers. For example, when Base started exporting `eachrow`, DataFrames had to adjust immediately to avoid breakage. That's really too bad as we have a very clear policy that minor releases shouldn't break dependencies... except if they use `using X`.

ararslan:
Yeah there are cases where things like that happen, but I've found they occur sufficiently rarely in practice to not be much of a practical concern, and fixing/working around is quite straightforward.

nalimilan:
Except if due to some complex dependency structure users can update the package which added the export, but not your package which fixes the bug. In that case they're screwed unless you make bugfix releases for all affected minor series. Granted that's theoretical, but with an expanding universe of packages, half-maintained ones will become more common and dependencies won't always be updated quickly.

matbesancon:
The case was striking with the StatsFuns function that was available only in `>= 0.9.3`, with a blanket using, I had to google the function to see where it's supposed to be. With a specific or qualified using, I would have gotten an error message telling me StatsFuns with the current version does not have this or that

matbesancon:
bump on this

src/multivariate/mvnormal.jl
```diff
@@ -187,9 +187,9 @@ const IsoNormal  = MvNormal{Float64,ScalMat{Float64},Vector{Float64}}
 const DiagNormal = MvNormal{Float64,PDiagMat{Float64,Vector{Float64}},Vector{Float64}}
 const FullNormal = MvNormal{Float64,PDMat{Float64,Matrix{Float64}},Vector{Float64}}
 
-const ZeroMeanIsoNormal{Axes}  = MvNormal{Float64,ScalMat{Float64},Zeros{Float64,1,Axes}}
-const ZeroMeanDiagNormal{Axes} = MvNormal{Float64,PDiagMat{Float64,Vector{Float64}},Zeros{Float64,1,Axes}}
-const ZeroMeanFullNormal{Axes} = MvNormal{Float64,PDMat{Float64,Matrix{Float64}},Zeros{Float64,1,Axes}}
+const ZeroMeanIsoNormal{Axes}  = MvNormal{Float64,ScalMat{Float64},FillArrays.FillArrays.Zeros{Float64,1,Axes}}
```
nalimilan: Is the repetition intended?

Anyway, given how often `Zeros` is used, probably better import it.

src/multivariate/mvnormal.jl
```diff
@@ -258,24 +258,25 @@ logdetcov(d::MvNormal) = logdet(d.Σ)
...
-    invquad!(r, d.Σ, x .- d.μ)
+    PDMats.invquad!(r, d.Σ, x .- d.μ)
 
 gradlogpdf(d::MvNormal, x::AbstractVector{<:Real}) = -(d.Σ \ (x .- d.μ))
 
 # Sampling (for GenericMvNormal)
 
-_rand!(rng::AbstractRNG, d::MvNormal, x::VecOrMat) =
-    add!(unwhiten!(d.Σ, randn!(rng, x)), d.μ)
+function _rand!(rng::AbstractRNG, d::MvNormal, x::VecOrMat)
```
nalimilan: The old one-line form seemed OK.

src/multivariate/mvlognormal.jl
```diff
@@ -27,14 +27,17 @@ abstract type AbstractMvLogNormal <: ContinuousMultivariateDistribution end
 
 function insupport(::Type{D},x::AbstractVector{T}) where {T<:Real,D<:AbstractMvLogNormal}
     for i=1:length(x)
-      @inbounds 0.0<x[i]<Inf ? continue : (return false)
+        @inbounds 0.0<x[i]<Inf ? continue : (return false)
     end
-    true
+    return true
```
nalimilan: This doesn't sound useful to me. Do you follow a particular style guide?

src/multivariate/mvnormal.jl
matbesancon: agreed we use it a lot, done in last PR

src/multivariate/mvnormal.jl
matbesancon: Ok, I considered awkward to have a one-liner spread on multiple lines, but not a drama, reverted in last commit

src/multivariate/mvlognormal.jl
matbesancon: both of these recently developed guides follow the convention of always using an explicit `return` keyword with long-form functions
https://github.com/jrevels/YASGuide
https://github.com/invenia/BlueStyle

src/multivariate/mvnormal.jl
nalimilan: I interpret "one-liner" as referring to a single expression (i.e. no `;`). :-)

src/multivariate/mvnormal.jl
```diff
@@ -333,22 +333,23 @@ end
 
 ## MLE estimation with covariance known
 
-fit_mle(g::MvNormalKnownCov{C}, ss::MvNormalKnownCovStats{C}) where {C<:AbstractPDMat} =
-    MvNormal(ss.sx * inv(ss.tw), g.Σ)
+function fit_mle(g::MvNormalKnownCov{C}, ss::MvNormalKnownCovStats{C}) where {C<:AbstractPDMat}
```
nalimilan: Keep one-liner too.

src/multivariate/mvlognormal.jl
```diff
@@ -27,14 +27,17 @@ abstract type AbstractMvLogNormal <: ContinuousMultivariateDistribution end
...
 end
+
 insupport(l::AbstractMvLogNormal,x::AbstractVector{T}) where {T<:Real} = insupport(typeof(l),x)
 assertinsupport(::Type{D},m::AbstractVector) where {D<:AbstractMvLogNormal} = @assert insupport(D,m) "Mean of LogNormal distribution should be strictly positive"
 
-###Internal functions to calculate scale and location for a desired average and covariance
+"""
```
nalimilan: This isn't appropriate as the comment described two functions, not one.

src/multivariate/mvlognormal.jl
nalimilan: OK. I had another convention in mind, which is that `return` must be used when the control flow is complex, i.e. the function doesn't just have a single return path. Not sure it's written anywhere.

Note that you haven't changed all occurrences (see e.g. `suffstats`). I'm not sure it makes sense to enforce this kind of conventino unless we enable an automated formatter like JuliaFormatter.

src/multivariate/mvlognormal.jl
matbesancon: reversed in last commit

src/multivariate/mvnormal.jl
matbesancon: reversed in last commit

src/multivariate/mvnormal.jl
matbesancon: yes got it :)

matbesancon:
bump here

mschauer:
I am not strongly in favour of either way to do it, but I am in favour of not introducing a hard rule. For example I like `using`, but I also see the risk of a new export messing up things with `using`. But then we can keep `using` packages we have somewhat control over, e.g. because they are looked over by the same people as `Distributions` and protect ourself by not `using` small packages from remote corners of the ecosystem. And those are also the packages where it is useful to qualify the function call, because at first sight it might not be clear where that particular function comes from.

matbesancon:
Even for packages that are closer to Distributions, it can get tricky, see the bug caused by
`StatsFuns = "0.9.3"`, first reaction was to wonder where the `logmvbeta` function was coming from in the bug, with no better tool than googling it in hope of finding related packages

matbesancon:
closing this for now, the PR is too old and most changes can be reproduced quickly. I still think blanket `using X` is a debugging burden for a package the size of Distributions

PR #1057, opened 2020-01-16
Nan checks

### LINKED ISSUE #1055: Checking foo() === NaN is brittle
Tests for Distributions.jl fails on Julia master likely due to the type of NaN that is generated. `===` on NaN will do a bitwise check. It is likely the tests should use `isnan`.

matbesancon:
closes #1055

### DIFF
--- a/test/normal.jl
+++ b/test/normal.jl
@@ -1,5 +1,7 @@
 using Test, Distributions, ForwardDiff
 
+isnan_type(::Type{T}, v) where {T} = isnan_type(T, v) && v isa T
+
 @testset "Normal" begin
     @test isa(convert(Normal{Float64}, Float16(0), Float16(1)),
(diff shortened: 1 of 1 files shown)


matbesancon:
the only problem with these new tests is that it is ignoring the check with the specific NaN we want.
If the compiler infers`NaN`, but we expect `NaN32`, this is not validated anymore

ararslan:
You're still using `@inferred` here though.

matbesancon:
yes so the functions have to be type-stable, but they don't have to return the propagated type (if pdf always returned Float64 regardless of the input type, `@inferred` would work

ararslan:
You can check `isnan(x) && x isa T`

matbesancon:
should be fine now

Issue #827, opened 2019-02-06
Numerical issues for variance of truncated normal


pagnani:
Minimal repro

```
julia> var(TruncatedNormal(1000000.0,1.0,0.0,1000.0))
-0.0001220703125
```

pinging @cossio

PR #1058, opened 2020-01-19
Improve evaluation of variance of truncated normal in the tails

cossio:
Fixes https://github.com/JuliaStats/Distributions.jl/issues/827.
Added more tests to mean and variance of truncated normal.
@pagnani

### DIFF
--- a/src/truncated/normal.jl
+++ b/src/truncated/normal.jl
@@ -24,84 +24,85 @@ end
 
 modes(d::Truncated{Normal{T},Continuous}) where {T <: Real} = [mode(d)]
 
-# do not export. Used in mean, var
-function _F1(x::Real, y::Real; thresh=1e-7)
-    @assert 0 < thresh < Inf
-    -Inf < x < Inf && -Inf < y < Inf || throw(DomainError())
-    ϵ = exp(x^2 - y^2)
-    if abs(x) > abs(y)
-        _F1(y,x)
-    elseif abs(x - y) ≤ thresh
-        δ = y - x
-        √π*x + (√π/2 + (-√π*x/6 + (-√π/12 + x*(√π/90 + (√π*x^2)/90)δ)δ)δ)δ
-    elseif max(x,y) < 0
-        (1 - ϵ) / (ϵ * erfcx(-y) - erfcx(-x))
-    elseif min(x,y) > 0
-        (1 - ϵ) / (erfcx(x) - ϵ * erfcx(y))
-    else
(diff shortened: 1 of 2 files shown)


matbesancon:
did you check how fast the new implementation is compared to the former one? I know this is fixing a bug, but just to check how much slowdown is associated with it

cossio:
@matbesancon It's either faster or comparable speed.

Before this PR:

```
julia> @btime var(truncated(Normal(1000000,1), 0,1000));
  88.183 ns (0 allocations: 0 bytes)
```

After this PR:

```
julia> @btime var(truncated(Normal(1000000,1), 0,1000));
  74.911 ns (0 allocations: 0 bytes)
```

matbesancon:
Awesome thanks a lot

magerton:
Hi @cossio , I have been using the `_F1` function to evaluate the derivative of the logcdf of a truncated normal in order to use the derivative of an ordered probit in finding an MLE. This means computing `(normpdf(b) - normpdf(a)) / (normcdf(b) - normcdf(a))`, which generates numerical issues. The `_F1` function allowed me to do this (see below).

PR #1058 removes `_F1`. Has the function simply been renamed, or should I cut & paste from your other repo, say https://github.com/cossio/TruncatedNormal.jl/blob/576d5154006e21b9822c5f9edc88ca6ada6248de/src/F.jl ?

Thanks a bunch!

```julia
# see
# https://github.com/cossio/TruncatedNormal.jl/blob/fc904152f2da11a257e3ccdd3e49ef118b81d437/notes/normal.pdf
# https://stats.stackexchange.com/questions/7200/evaluate-definite-interval-of-normal-distribution/7206#7206

function dlogcdf_trunc(a::Real, b::Real)
    # https://github.com/scipy/scipy/blob/a2ffe09aa751749f2372aa13c19c61b2dec5266f/scipy/stats/_continuous_distns.py
    # https://github.com/JuliaStats/Distributions.jl/blob/master/src/truncated/normal.jl
    # https://github.com/cossio/TruncatedNormal.jl/blob/master/notes/normal.pdf
    a == typemin(a) && return   normpdf(b) / normcdf(b)
    b == typemax(b) && return - normpdf(a) / normccdf(a)
    return - _F1(a*invsqrt2, b*invsqrt2) * invsqrthalfπ
end
```

magerton:
**UPDATE** Ah! I see here. Things just got renamed. Sorry to bother!

```julia
-√(2/π) * _F1(a / √2, b / √2) == _tnmom1(a,b)
```

Issue #456, opened 2016-01-30
LogNormal inconsistent when outside support


KrisDM:
Behaviour for LogNormal pdf and logpdf for values outside its support is inconsistent (see below). 

This is partly because the @distr_support macro gives a closed rather than open interval (which is required here), and partly because the pdf function does not test the support. I will submit a PR with a fix and tests.

Results should be 0.0 for pdf, and -Inf for logpdf, when testing values outside the support. 

Not NaN, and not the DomainError generated by a call to log(-1.0), as below:

```
julia> d = LogNormal(0.0,1.0)
Distributions.LogNormal(μ=0.0, σ=1.0)

julia> pdf(d,0.0)
NaN

julia> pdf(d,-1.0)
ERROR: DomainError:
 in pdf at C:\Users\kris\.julia\v0.4\Distributions\src\univariate\continuous\lognormal.jl:56

julia> logpdf(d,0.0)
NaN

julia> logpdf(d,-1.0)
-Inf
```

KrisDM:
When I started fixing this, the problem was much bigger than I thought: several distributions had similar inconsistencies. PR #457 fixes all of them.

matbesancon:
@JuliaRegistrator register()

PR #1059, opened 2020-01-20
Fix typestability of (log)pdf, (log)cdf, (log)ccdf, and gradlogpdf for LogNormal

devmotion:
I noticed some type-instabilities in the current implementation of (log)pdf for `LogNormal`, and tried to fix them.

I adapted the type inference tests for `Normal` distributions. In these tests degenerate normal distributions (with zero variance) are used quite often, hence to simplify the transfer of the tests and for consistency I also allowed the use of degenerate log normal distributions.

This PR fixes https://github.com/JuliaStats/Distributions.jl/issues/456.


matbesancon:
LGTM, @devmotion good to merge right?

devmotion:
Yes :+1:

Issue #960, opened 2019-08-23
Wishart samples are always Float64, regardless partype


johnczito:
To clarify, this is not the expected behavior, correct?

```julia 
julia> using Distributions, LinearAlgebra

julia> d = Wishart(Float32(4), Matrix{Float32}(I, 2, 2))

julia> partype(d)
Float32

julia> rand(d)
2×2 Array{Float64,2}:
 1.47454  1.53805
 1.53805  6.80249
```

The sample should be `Array{Float32,2}`. If so, I'll submit a PR to fix.

mschauer:
In the fix, also add `eltype` ... or wait for #951

richardreeve:
Yes, it turns out that almost every single distribution fails to return the right type from quantile() and most from rand(). I'm going through fixing them all as part of #951.

johnczito:
@richardreeve Very good. Thanks for that. Most of the matrix-variates rely on the Wishart for sampling at some point. So fixing it will be a big step toward straightening all of them out.

PR #1066, opened 2020-02-11
Add `LKJ`

johnczito:
This PR adds the [LKJ distribution](https://doi.org/10.1016/j.jmva.2009.04.008), which is a distribution over correlation matrices. It can be found in other statistical computing platforms ([here](https://mc-stan.org/docs/2_22/functions-reference/lkj-correlation.html), [here](https://www.rdocumentation.org/packages/rethinking/versions/1.59/topics/dlkjcorr), or [here](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/distributions/lkj.py), for instance).

Note: One of the unit tests runs a hypothesis test, so this PR adds `HypothesisTests.jl` as an extra testing dependency in the `.toml`. Let me know if I have done this incorrectly, or if I should exclude this change for now.

### DIFF
--- a/src/matrix/lkj.jl
+++ b/src/matrix/lkj.jl
@@ -2,24 +2,20 @@
     LKJ(d, η)
 ```julia
 d::Int   dimension
-η::Real  positive scale
+η::Real  positive shape
 ```
 The [LKJ](https://doi.org/10.1016/j.jmva.2009.04.008) distribution is a distribution over
 ``d\\times d`` real correlation matrices (positive-definite matrices with ones on the diagonal).
 If ``\\mathbf{R}\\sim \\textrm{LKJ}_{d}(\\eta)``, then its probability density function is
 
 ```math
-f(\\mathbf{R};\\eta) = c_0 |\\mathbf{R}|^{\\eta-1},
+f(\\mathbf{R};\\eta) = \\left[\\prod_{k=1}^{d-1}\\pi^{\\frac{k}{2}}
+\\frac{\\Gamma\\left(\\eta+\\frac{d-1-k}{2}\\right)}{\\Gamma\\left(\\eta+\\frac{d-1}{2}\\right)}\\right]^{-1}
+|\\mathbf{R}|^{\\eta-1}.
 ```
 
-where (among many equivalent expressions)
-
-```math
-c_0 = \\prod_{k=1}^{d-1}\\pi^{\\frac{k}{2}}
-\\frac{\\Gamma\\left(\\eta+\\frac{d-1-k}{2}\\right)}{\\Gamma\\left(\\eta+\\frac{d-1}{2}\\right)}.
-```
-
-If ``\\eta = 1``, then the LKJ distribution is uniform over the space of correlation matrices.
+If ``\\eta = 1``, then the LKJ distribution is uniform over
+[the space of correlation matrices](https://www.jstor.org/stable/2684832).
 """
 struct LKJ{T <: Real, D <: Integer} <: ContinuousMatrixDistribution
     d::D
@@ -86,12 +82,12 @@ params(d::LKJ) = d.η
 function lkj_logc0(d::Integer, η::Real)
     if isone(η)
         if iseven(d)
-            logc0 = lkj_onion_logc0_uniform_even(d)
+            logc0 = -lkj_onion_loginvconst_uniform_even(d)
         else
-            logc0 = lkj_onion_logc0_uniform_odd(d)
+            logc0 = -lkj_onion_loginvconst_uniform_odd(d)
         end
     else
-        logc0 = lkj_onion_logc0(d, η)
+        logc0 = -lkj_onion_loginvconst(d, η)
     end
     return logc0
 end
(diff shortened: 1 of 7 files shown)


johnczito:
(As I mentioned [here](https://discourse.julialang.org/t/is-distributions-jl-currently-accepting-prs-for-new-distributions/34078), I'm not sure if folks will want to wait on this indefinitely until #951 is sorted out. Hopefully not...)

mschauer:
Hi, could you post a screenshot of the docstring of `LKJ`?

johnczito:
> Hi, could you post a screenshot of the docstring of `LKJ`?

Screen Shot 2020-02-11 at 6 54 44 AM

mschauer:
Do you know in which sense is this "uniform" if eta=1?

src/matrix/lkj.jl
```diff
@@ -0,0 +1,219 @@
...
+
+function _rand!(rng::AbstractRNG, d::LKJ, R::AbstractMatrix)
+    R .= _lkj_onion_sampler(d.d, d.η, rng)
+end
+
+function _lkj_onion_sampler(d::Integer, η::Real, rng::AbstractRNG = Random.GLOBAL_RNG)
+    #  Section 3.2 in LKJ (2009 JMA)
+    #  1. Initialization
+    R = ones(typeof(η), d, d)
+    β = η + 0.5d - 1
```
mschauer: Maybe 
```suggestion
    β = η + d/2 - 1
```
or should this have some `oftype`?

src/matrix/lkj.jl
```diff
@@ -0,0 +1,219 @@
...
+    u = rand(rng, Beta(β, β))
+    R[1, 2] = 2u - 1
+    R[2, 1] = R[1, 2]
+    #  2.
+    for k in 2:d - 1
+        #  (a)
+        β -= 0.5
```
mschauer: Cf 115

mschauer:
One could test that importance sampling is successful, e.g.
```
E f(R) = E f(U) c_0(eta)/c_0(1) | U | (eta-1) 
```
for R ~ LKJ(eta) and U ~ LKJ(1) and some functional f

johnczito:
> Do you know in which sense is this "uniform" if eta=1?

The density is constant in R. The determinant term collapses to 1 and you just have f(R; η) = c₀.

src/matrix/lkj.jl
johnczito: β won't be promoted to whatever η is?

johnczito:
> One could test that importance sampling is successful, e.g.

Fair enough. I can add that.

mschauer:
Yes, I can see that, but I suspect there is something to say what that distribution actually is. Or in other words, what is the reference measure for the densities?

johnczito:
It's just the uniform probability measure on the appropriate subset of [-1, 1]ᵐ, m = d(d-1)/2.

johnczito:
Granted, the notation in the original paper is pretty inconsistent, so what I have here might be off by a sign. Your importance sampling suggestion should clarify that.

johnczito:
@mschauer It's a density with respect to Lebesgue measure, so uniform here just means the density is one over the Lebesgue measure of the support. The m = d(d-1)/2 free elements of a correlation matrix live in a subset of [-1, 1]ᵐ that has finite and strictly positive Lebesgue measure. So not some measure zero submanifold that requires a funky dominating measure to define a density. More [here](https://www.jstor.org/stable/2684832).

Consequently though, the log integrating constant here is currently off by a sign. Fixing that now.

src/matrix/lkj.jl
mschauer:
```
η = 1f0
d = 4

julia> typeof(η), typeof(η + 0.5d - 1)
(Float32, Float64)
```

johnczito: Right, thanks. Since `Beta` suffers from #960 it doesn't end up mattering, unfortunately.

mschauer:
Thank you, this is nice. What is the price we pay for the test in terms of time?

test/lkj.jl
```diff
@@ -86,38 +86,70 @@ end
...
+@testset "importance sampling check" begin
+    d = 3
+    M = 20000
+    f = LKJ(d, 2)
+    g = LKJ(d, 1)
+    R = rand(g, M)
+    fvals = [pdf(f, R[m]) for m in 1:M]
+    gvals = [pdf(g, R[m]) for m in 1:M]
+    h = mean(logdet.(rand(f, M)))
+    ĥ = sum(logdet.(R) .* fvals ./ gvals) / M
```
mschauer:
```suggestion
    ĥ = mean(logdet(R) * pdf(f, R)/pdf(g, R) for R in (rand(g) for i in 1:M))
```

johnczito:
> Thank you, this is nice. What is the price we pay for the test in terms of time?

0.21 seconds

test/lkj.jl
johnczito: Nice, thanks. It's wild that stuff like this works.

src/matrix/lkj.jl
```diff
@@ -0,0 +1,239 @@
+"""
+    LKJ(d, η)
+```julia
```
matbesancon:
```suggestion

```julia
```

matbesancon: newline after the type name

src/matrix/lkj.jl
```diff
@@ -0,0 +1,239 @@
...
+struct LKJ{T <: Real, D <: Integer} <: ContinuousMatrixDistribution
+    d::D
+    η::T
+    logc0::T
+end
+
+#  -----------------------------------------------------------------------------
+#  Constructors
+#  -----------------------------------------------------------------------------
+
```
matbesancon: You could use a keyword `check_args=true` to avoid checking multiple times when creating the distribution, and `@check_args(LKJ, d > 0 && η > 0)`. The latter is up to your preference, both are fine with me

johnczito: > to avoid checking multiple times when creating the distribution

Not sure what you mean.

matbesancon: sorry yes. When creating multiple time the same distribution, or when copying a distribution, it's practical to have a method shortcircuit all the checks

johnczito: Okey-doke. Done.

src/matrix/lkj.jl
johnczito: Done.

johnczito:
Thanks for all of the help on this, as usual. Let me know what else you want to see for this to be considered for merging.

matbesancon:
Looks great to me. @johnczito maybe add a reference in the testset from the last commit to indicate where the stan test set comes from (documentation link or something)

src/matrix/lkj.jl
mschauer: Lets fix it anyway.

matbesancon:
Awesome, the tests were green before, so I'll just merge it here, thanks!

mschauer:
I now believe that it is uniform in the constraint space of correlation matrices!

```
using Distributions
using Makie
K = 50000; ρ = [Point3f0(rand(LKJ(3, 1.0))[[2,3,6]]) for k in 1:K]
scatter(ρ, markersize=0.009, color=(:black, 0.2))
```
lkj

johnczito:
@mschauer Nice! And that's a code snippet I'll definitely be stealing...

PR #1076, opened 2020-02-24
relax type constraints on weights for fitting MVNormal

oxinabox:
`UnitWeights` from StatsBase are by default `AbstractVector{Int}` IIRC.
With them all being `1`

Requiring things to be `AbstractArray{Float64}` or worse `Array{Float64}` is just excessive

### DIFF
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -449,7 +449,7 @@ function fit_mle(D::Type{DiagNormal}, x::AbstractMatrix{Float64})
     MvNormal(mu, PDiagMat(va))
 end
 
-function fit_mle(D::Type{DiagNormal}, x::AbstractMatrix{Float64}, w::AbstractArray)
+function fit_mle(D::Type{DiagNormal}, x::AbstractMatrix{Float64}, w::AbstractVector)
     m = size(x, 1)
     n = size(x, 2)
     length(w) == n || throw(DimensionMismatch("Inconsistent argument dimensions"))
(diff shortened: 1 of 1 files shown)


src/multivariate/mvnormal.jl
```diff
@@ -449,7 +449,7 @@ function fit_mle(D::Type{DiagNormal}, x::AbstractMatrix{Float64})
     MvNormal(mu, PDiagMat(va))
 end
 
-function fit_mle(D::Type{DiagNormal}, x::AbstractMatrix{Float64}, w::AbstractArray{Float64})
+function fit_mle(D::Type{DiagNormal}, x::AbstractMatrix{Float64}, w::AbstractArray)
```
oxinabox:
```suggestion
function fit_mle(D::Type{DiagNormal}, x::AbstractMatrix{Float64}, w::AbstractVector)
```

src/multivariate/mvnormal.jl
```diff
@@ -371,7 +371,7 @@ function suffstats(D::Type{MvNormal}, x::AbstractMatrix{Float64})
     MvNormalStats(s, m, s2, Float64(n))
 end
 
-function suffstats(D::Type{MvNormal}, x::AbstractMatrix{Float64}, w::Array{Float64})
+function suffstats(D::Type{MvNormal}, x::AbstractMatrix{Float64}, w::AbstractVector)
```
oxinabox: This one was really excessive.
No `AbstractArray` had to be a dense `Array`

matbesancon:
it makes sense thanks.

PR #1079, opened 2020-02-25
Make Bernoulli produce Bools

MikeInnes:
This effectively reverts #1048 and also addresses the original issue #984.

I'm not that familiar with this codebase, but it appears that `rand(D, ...)` uses the `eltype` of the distribution to create the right array container. For `Bernoulli`, `eltype` hit a fallback method that incorrectly returned `Int64`, so changing that to `Bool` fixes the issue.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -29,11 +29,10 @@ Calculus = "49dc2e85-a5d0-5ad3-a950-438e2897f1b9"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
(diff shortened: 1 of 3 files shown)


matbesancon:
tests failing at:
https://travis-ci.org/JuliaStats/Distributions.jl/jobs/655002051#L601

MikeInnes:
The `cdf` and `ccdf` functions were previously incorrect (but not tested since the `Int` rather than `Bool` versions got called).

cossio:
After this PR, will `rand(Bernoulli(0.5), 5,5)` produce a `BitArray`?

MikeInnes:
@cossio not currently, since array generation just uses a fallback that creates an array based on the `eltype`. I'm sure that could be changed without too much difficulty later, but that should have its own discussion (I would be hesitant).

@matbesancon anything left to do for this to get merged and tagged?

matbesancon:
yes all good it seems, I wanted a last patch release before this one, which is now on the way

PR #1080, opened 2020-02-26
improve pdmat usage in the matrix-variates

johnczito:
The matrix-variates commit many small sins that make them incapable of handling `AbstractPDMat` subtypes other than `PDMat`. This PR fixes that:

  - Allow the two posdef matrices in `MatrixNormal` and `MatrixTDist` to be different types;
  - Don't implicitly assume a `PDMat` by hard-coding direct access to fields like `.mat` and `.chol.L`;
  - Use the new `kron` methods in `PDMats.jl`;
  - ~~Temporary workaround for the fact that not all `AbstractPDMat` subtypes have `\` defined such that `S \ M'` will work (not everyone's `\` plays nice with `Adjoint`)~~.

(Shout-out to @willtebbutt, who [nailed](https://github.com/JuliaStats/Distributions.jl/pull/917#discussion_r297628281) this issue from the start, but I completely misunderstood what he was asking at the time.)

### DIFF
--- a/src/matrix/inversewishart.jl
+++ b/src/matrix/inversewishart.jl
@@ -1,8 +1,8 @@
 """
     InverseWishart(ν, Ψ)
 ```julia
-ν::Real   degrees of freedom (greater than p - 1)
-Ψ::PDMat  p x p scale matrix
+ν::Real           degrees of freedom (greater than p - 1)
+Ψ::AbstractPDMat  p x p scale matrix
 ```
 The [inverse Wishart distribution](http://en.wikipedia.org/wiki/Inverse-Wishart_distribution)
 generalizes the inverse gamma distribution to ``p\\times p`` real, positive definite
--- a/src/matrix/matrixnormal.jl
+++ b/src/matrix/matrixnormal.jl
@@ -2,8 +2,8 @@
     MatrixNormal(M, U, V)
 ```julia
 M::AbstractMatrix  n x p mean
-U::PDMat           n x n row covariance
-V::PDMat           p x p column covariance
+U::AbstractPDMat   n x n row covariance
+V::AbstractPDMat   p x p column covariance
 ```
 The [matrix normal distribution](https://en.wikipedia.org/wiki/Matrix_normal_distribution) generalizes the multivariate normal distribution to ``n\\times p`` real matrices ``\\mathbf{X}``.
 If ``\\mathbf{X}\\sim MN_{n,p}(\\mathbf{M}, \\mathbf{U}, \\mathbf{V})``, then its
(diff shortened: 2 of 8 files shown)


src/matrix/inversewishart.jl
```diff
@@ -1,8 +1,8 @@
 """
     InverseWishart(ν, Ψ)
 ```julia
-ν::Real   degrees of freedom (greater than p - 1)
-Ψ::PDMat  p x p scale matrix
+ν::Real           degrees of freedom (greater than p - 1)
```
matbesancon: hit tab by error?

src/matrix/matrixnormal.jl
```diff
@@ -109,7 +109,8 @@ end
 
 function logkernel(d::MatrixNormal, X::AbstractMatrix)
     A  = X - d.M
```
matbesancon: In cases where `A'` does play nicely with `\`, this creates an additional copy, could there be a solution that avoids that?

src/matrix/inversewishart.jl
johnczito: not by error, but I can eliminate if you want.

matbesancon: is it to align with the following lines?

src/matrix/matrixnormal.jl
johnczito: The main issue is [here](https://github.com/JuliaStats/PDMats.jl/blob/7ce4a5c609b288ab4b0ebaf3c397c9b46f7ce6a2/src/pdiagmat.jl#L45), for instance. I don't think anyone even remembers why that had to be `StridedVecOrMat`, and it's probably just going to be gotten rid of: [here](https://github.com/JuliaStats/PDMats.jl/pull/89), and [here](https://github.com/JuliaStats/PDMats.jl/pull/90). But for now it's why `S \ A'` doesn't work in general. So I can submit a PR to `PDMats`, wait for merge/release, and then come back here and kill this extra copy. Or else keep it for now, and submit another PR later whenever `PDMats` is updated.

src/matrix/inversewishart.jl
johnczito: yeah

src/matrix/matrixnormal.jl
matbesancon: I'm ok to continue as is, can you post an issue here as a reminder of what has to be done once the change is done in PDMats?

johnczito: Definitely.

matbesancon:
LGTM

johnczito:
Thanks. Let me know if I should do anything else.

matbesancon:
can you create the issues and reference them in this PR, so we can keep track of where they come from

johnczito:
I opened the PDMats [PR](https://github.com/JuliaStats/PDMats.jl/pull/111). If it gets turned over quickly enough, perhaps we can just wait a bit.

matbesancon:
the PR in PDMats seems merged, is there anything more to do here or good to merge as is?

johnczito:
If we wait for a `PDMats` release, I can commit the change that eliminates `At = Matrix(A')`, and then we're really done.

matbesancon:
yes perfect, I asked the maintainers of PDMats if they are ok with making a release, should be done in the day

matbesancon:
@johnczito a release has just been made, see https://github.com/JuliaStats/PDMats.jl/issues/108
If the changes of the last release (basically your PR), can you update the project compat to indicate the minimum version to be `>= v0.9.12`

matbesancon:
the releasing PR is here: https://github.com/JuliaRegistries/General/pull/10786

johnczito:
`At = Matrix(A')` is gone, and the minimum version of `PDMat` is set. Let me know what else you want here.

matbesancon:
this is perfect, merging when CI passes :)

johnczito:
@matbesancon Not sure where the right place to discuss this is, but I just noticed that since #1086 was merged (not blaming that PR), the unit tests have been erroring on the nightly build. Do we know why?

matbesancon:
No I hadn't noticed, thanks for pointing it out, and the error we get is really ugly

matbesancon:
it may be a change in nightly more than in Distributions itself?

johnczito:
Tests are currently passing on the nightly build, so no worries.

matbesancon:
phew

PR #1081, opened 2020-02-26
Update truncate.jl

githubtomtom:
simply added the missing `eltype()` for `Truncated`


src/truncate.jl
```diff
@@ -73,6 +73,8 @@ Truncated(d::UnivariateDistribution, l::Integer, u::Integer) = Truncated(d, floa
 
 params(d::Truncated) = tuple(params(d.untruncated)..., d.lower, d.upper)
 partype(d::Truncated) = partype(d.untruncated)
+Base.eltype(::Type{Truncated{D, S, T} } ) where {D, S, T <: Real} = T
```
matbesancon:
```suggestion
Base.eltype(::Type{Truncated{D, S, T} } ) where {D, S} = T
```
The type constraint is not necessary

githubtomtom: ok

Issue #1089, opened 2020-03-07
Truncated normal with generic float


cossio:
Truncated normal samplers only support `Float64`. It would be nice to relax this to at least cover `Float32` too.

```
julia> tn = truncated(Normal(0f0,1f0), 0f0, Inf32);
julia> rand(tn)
ERROR: MethodError: no method matching randnt(::Random._GLOBAL_RNG, ::Float32, ::Float32, ::Float32)
Closest candidates are:
  randnt(::Random.AbstractRNG, ::Float64, ::Float64, ::Float64) at /home/cossio/.julia/packages/Distributions/6HQ2U/src/truncated/normal.jl:155
Stacktrace:
 [1] rand(::Random._GLOBAL_RNG, ::Truncated{Normal{Float32},Continuous,Float32}) at /home/cossio/.julia/packages/Distributions/6HQ2U/src/truncated/normal.jl:146
 [2] rand(::Truncated{Normal{Float32},Continuous,Float32}) at /home/cossio/.julia/packages/Distributions/6HQ2U/src/genericrand.jl:22
 [3] top-level scope at REPL[10]:1
```

rfarnham:
I ran into the same problem:
```
julia> rand(Normal(0f0, 1f0)) # ok!
2.229807f0

julia> rand(TruncatedNormal(0, 1, -2, 2)) # ok!
-0.023194308156410635

julia> rand(TruncatedNormal(0f0, 1f0, -2f0, 2f0)) # nope
ERROR: MethodError: no method matching randnt(::Random._GLOBAL_RNG, ::Float32, ::Float32, ::Float32)
```

PR #1090, opened 2020-03-07
Generic truncated normal & exponential

cossio:
Fixes https://github.com/JuliaStats/Distributions.jl/issues/1089.

Since the sampler for truncated normal takes samples from `Exponential`, I also modified the `Exponential` sampler to return values following the parameter type.


matbesancon:
This looks good to me, but I don't remember what was the decision on return types for sampling so I'll let @andreasnoack give his opinion before moving this forward

matbesancon:
ping @andreasnoack

src/truncated/normal.jl
```diff
@@ -12,6 +12,8 @@ TruncatedNormal
 
 ### statistics
 
+Base.eltype(::Type{<:Truncated{Normal{T},Continuous}}) where {T <: Real} = T
```
devmotion, line 15: I'm wondering why this is needed and the generic implementation in https://github.com/JuliaStats/Distributions.jl/blob/3f4856c901b79d183eaf9f1ab469500310fd0182/src/truncate.jl#L76 is not sufficient.

rfarnham:
I'm coming from https://github.com/JuliaStats/Distributions.jl/issues/1089
I've also run into the same problem. Any blockers for merging this?

burmecia:
I'm also coming from #1089. Any reasons why this PR is still pending?

devmotion:
I still wonder why/if the definition of `eltype` is needed (https://github.com/JuliaStats/Distributions.jl/pull/1090/files#r467471768).

In general, as far as I remember, the general policy is that there are no guarantees that `eltype(d) === typeof(rand(d))`, i.e., that the type of the samples is the same as of the parameters. There are many issues and discussions regarding this topic, e.g., https://github.com/JuliaStats/Distributions.jl/issues/1252 and https://github.com/JuliaStats/Distributions.jl/issues/1071.

cossio:
In general the parameter type doesn't have to match the type of samples. E.g. a Poisson, has a float parameter but the samples are integers.

The decision that `eltype(dist)` should mean the type of parameters seems strange to me. The parameters can also be of different types.

devmotion:
> The decision that eltype(dist) should mean the type of parameters seems strange to me. The parameters can also be of different types.

I am very unsure if this is the right approach but it is definitely what is done currently (and also what you implement in this PR when you define `eltype`). In many distributions parameters are promoted to a common type and hence there is no problem with parameters of different types.

devmotion:
BTW I just checked and already without this PR, as expected, we have
```julia
julia> eltype(truncated(Normal(0f0,1f0), 0f0, Inf32))
Float32
```
so the `eltype` definition above is not needed. In fact, I think it is wrong since it does not respect the types of the truncated distribution: with this PR also `eltype(truncated(Normal(0f0,1f0), 0, Inf)) === Float32` even though all calculations of the truncated distribution are performed with `Float64` (since `logcdf(Normal(0f0, 1f0), Inf) isa Float64`).

src/truncated/normal.jl
burmecia: I agree, that seems not necessary for me.

cossio:
> I am very unsure if this is the right approach but it is definitely what is done currently (and also what you implement in this PR when you define `eltype`). In many distributions parameters are promoted to a common type and hence there is no problem with parameters of different types.

Can we be more flexible about this? Seems like an unnecessary restriction.

PR #1091, opened 2020-03-19
Matrix Gaussian Distribution

san-soucie:
Implements the more general matrix-valued normal distribution as presented in [this preprint](https://arxiv.org/pdf/1804.11010.pdf)


johnczito:
Thanks for the PR! This looks good. The main thing I'm wondering is if it would be better to just wrap an `MvNormal` and inherit everything from it:

```julia
struct MatrixGaussian
    n::Integer
    m::Integer
    N::MvNormal
end

MatrixGaussian(M, Σ) = MatrixGaussian(size(M)..., MvNormal(vec(M), Σ))

_logpdf(d, A) = logpdf(d.N, vec(A))

function _rand!(rng::AbstractRNG, d::MatrixGaussian, Y::AbstractMatrix)
    Y .= reshape(rand(d.N), (m, n))
end
```
But thinking in those terms, there's nothing special about `MvNormal`. You could use this construction with any `MultivariateDistribution`: `MvTDist`, `Product`, etc.

san-soucie:
> But thinking in those terms, there's nothing special about `MvNormal`. You could use this construction with any `MultivariateDistribution`: `MvTDist`, `Product`, etc.

I copied the structure of the MatrixNormal distribution, but your suggestion would be an elegant way to wrap both distributions under the hood.

johnczito:
True. But if we said
```julia
MatrixNormal(M, U, V) = MatrixGaussian(size(M)..., MvNormal(vec(M), kron(V, U)))
```
we wouldn't be able to separately retrieve U and V from the distribution if we wanted, so that's a reason to keep it separate.

matbesancon:
+1 for wrapping the MvNormal, less code is better code

matbesancon:
@johnczito I'll let you give your opinion on the PR, I'm not super-educated on matrix distributions

matbesancon:
Also, should there be a note on the difference from matrix normal, or is the distinction strong enough in the literature?

johnczito:
@matbesancon Okay. The `MatrixGaussian` here is not a "known" distribution that you'd find in books or papers other than the linked [preprint](https://arxiv.org/pdf/1804.11010.pdf). And imo a designated implementation for `MatrixGaussian` is a bit much. There's nothing special about the "Gaussian" part. The construction here is generic: draw from a `MultivariateDistribution` and reshape the vector into a matrix. Everything about the `MatrixDistribution` (sampling, density evaluation) is directly inherited from the underlying `MultivariateDistribution` you're reshaping, so it's really just a wrapper. Maybe this could be something like `ReshapedDistribution` that wraps any `MultivariateDistribution` and provides generic methods like `logpdf(d, A) = logpdf(d.vecD, vec(A))`, `rand(d) = reshape(rand(d.vecD), size(d)...)`, etc?

matbesancon:
I like the idea of a generalized (matrix-)reshaped distribution. Should it be an abstract type sub-typed by e.g. MatrixGaussian or a concrete implementation, with possibly an alias `const MatrixGaussian = ReshapedDistribution{<:MvNormal}`

johnczito:
I'd go with a concrete implementation.

matbesancon:
@san-soucie would you be ok with generalizing this PR to define a `MatrixReshapedDistribution` struct?
 I'd see the design as
```julia
struct MatrixReshapedDistribution{S <: ValueSupport, D <: MultivariateDistribution{S}} <: MatrixDistribution{S}
    d::D
    num_cols::Int
    num_rows::Int
end
```

matbesancon:
If you don't feel like it, we can do it, just let us know

san-soucie:
@johnczito @matbesancon This discussion gets into something I was hoping to avoid in this PR, namely that Distributions.jl does not currently have a struct for N-D array-valued distributions. A MatrixReshapedDistribution like that would not need to be limited to 2-D arrays. Given the restriction to matrix-valued functions, I do think this is a good idea. Even if there was a more general reshaped distribution, it might make sense to have matrix-valued reshaped distributions be treated separately. I can take care of this this change.

matbesancon:
Thank you. I agree that there is currently no definition for N-D arrays, because nobody has found it interesting for their work & research to implement it for now, while matrix distributions have active research going on, and particular semantics attached, beyond being a 2-D array.

I think we can keep the PR in the realm of matrix distributions, which means most of the code you wrote still applies

sethaxen:
This comment is partially relate to `MatrixGaussian`, partially related to `MatrixNormal`. Would be nice to have options like in `MvNormal` to pass a zero matrix for the mean and/or a scalar or `UniformScaling` as the covariance that don't just get converted to dense matrices. There are a number of applications where one wants a standard matrix normal distribution, and it's quite a bit more efficient to sample from this.

johnczito:
If we had `MatrixReshapedDistribution` like above, it's just a wrapper for an `MvNormal`, which would already be doing the nice `zeromean`, `UniformScaling` stuff. 

As for the `MatrixNormal`, point taken on the mean. For the covariances, you can pass `PDMats.ScalMat` and they won't be converted to dense matrices. Does that work?

sethaxen:
> As for the `MatrixNormal`, point taken on the mean. For the covariances, you can pass `PDMats.ScalMat` and they won't be converted to dense matrices. Does that work?

It does, and I see now I can actually pass a `Zeros` for the mean. Probably the shape constructor should call these internally instead of creating dense arrays as it currently does.

johnczito:
Agreed. I'll PR it soon.

johnczito:
Your merge erased the changes it was conflicting with, but you want to keep them. So in `runtests.jl` you brought back all the individual test files for the matrix-variates, but they don't exist anymore. There should just be `matrixreshaped.jl` and `matrixvariates.jl`.

docs/src/matrix.md
```diff
@@ -35,6 +35,8 @@ vec(d::MatrixDistribution)
 MatrixNormal
 Wishart
 InverseWishart
+MatrixNormal
```
johnczito, line 38:
```suggestion

```

src/Distributions.jl
```diff
@@ -315,9 +316,10 @@ Supported distributions:
     Frechet, FullNormal, FullNormalCanon, Gamma, GeneralizedPareto,
     GeneralizedExtremeValue, Geometric, Gumbel, Hypergeometric,
     InverseWishart, InverseGamma, InverseGaussian, IsoNormal,
-    IsoNormalCanon, Kolmogorov, KSDist, KSOneSided, Laplace, Levy, LKJ,
-    Logistic, LogNormal, MatrixBeta, MatrixFDist, MatrixNormal, MatrixTDist, MixtureModel,
-    Multinomial, MultivariateNormal, MvLogNormal, MvNormal, MvNormalCanon,
+    IsoNormalCanon, Kolmogorov, KSDist, KSOneSided, Laplace, Levy,
```
johnczito, line 319:
```suggestion
    IsoNormalCanon, Kolmogorov, KSDist, KSOneSided, Laplace, Levy, LKJ
```

matbesancon:
sorry this got forgotten, merging in few hours if no objection. @johnczito good on your side?

docs/src/matrix.md
```diff
@@ -35,6 +35,7 @@ vec(d::MatrixDistribution)
 MatrixNormal
 Wishart
 InverseWishart
+MatrixReshaped
```
johnczito, line 38: Make this the last thing listed in the docs.

src/matrix/matrixreshaped.jl
```diff
@@ -0,0 +1,75 @@
+"""
+    MatrixReshaped(D, n, p)
+```julia
+D::MultivariateDistribution  base distribution
+n::Integer   number of rows
+p::Integer   number of columns
+```
+Reshapes a multivariate distribution into a matrix distribution with n rows and
```
johnczito, line 8:
```suggestion
Reshapes a multivariate distribution column-wise into a matrix distribution with n rows and
```

johnczito:
Just some minor things. Otherwise looks great.

docs/src/matrix.md
johnczito, line 38: I would move this to the bottom of the list here.

PR #1092, opened 2020-03-22
Added gradlogpdf for Arcsine

matbesancon:
gradlog pdf tested against ForwardDiff

### DIFF
--- a/src/univariate/continuous/arcsine.jl
+++ b/src/univariate/continuous/arcsine.jl
@@ -56,7 +56,7 @@ end
 location(d::Arcsine) = d.a
 scale(d::Arcsine) = d.b - d.a
 params(d::Arcsine) = (d.a, d.b)
-@inline partype(d::Arcsine{T}) where {T<:Real} = T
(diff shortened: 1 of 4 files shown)


src/univariate/continuous/arcsine.jl
```diff
@@ -67,8 +67,8 @@ mode(d::Arcsine) = d.a
 modes(d::Arcsine) = [d.a, d.b]
 
 var(d::Arcsine) = abs2(d.b - d.a) / 8
-skewness(d::Arcsine{T}) where {T<:Real} = zero(T)
-kurtosis(d::Arcsine{T}) where {T<:Real} = -T(3/2)
```
mschauer: Yeah, this shouldn't be constrained here at all. If you can make an `::Arcsine{T}`, `kurtosis` has no business checking that `T` is "valid".

matbesancon: yup this was a common repeated mistake in the code base

matbesancon:
I restarted the job which was failing

PR #1097, opened 2020-04-06
Add singular branch of the Wishart

johnczito:
This PR adds the [singular branch](https://en.wikipedia.org/wiki/Wishart_distribution#The_range_of_the_shape_parameter) of the Wishart distribution. That is, for integer `df < p`, the Wishart is a distribution over rank-`df` positive semidefinite `p x p` matrices, defined in the [usual way](https://en.wikipedia.org/wiki/Wishart_distribution#Definition).

In this case the Wishart does not have a Lebesgue density, but it does have a density with respect to an appropriate dominating measure on rank-`df` positive semidefinite matrices, which is given in Theorem 6 [here](https://projecteuclid.org/euclid.aos/1176325375). The usual density in the full rank case is a special case of this "generalized" density.

I am currently implementing this by extending the behavior of `Wishart` instead of defining a separate distribution. I prefer this because it's what Matlab [does](https://stats.stackexchange.com/questions/24288/how-to-sample-from-a-wishart-distribution), and because I think that a hard distinction between singular/nonsingular Wishart is ultimately artificial and could cause headache for the user. For example, there are certain time series models where the degrees of freedom of a Wishart are time-varying, and may or may not be singular from period to period. A user implementing this might prefer that we supply one Wishart that covers both cases instead of supplying two and forcing them to manually switch between them.

Let me know what you think!

### DIFF
--- a/src/matrix/wishart.jl
+++ b/src/matrix/wishart.jl
@@ -1,58 +1,68 @@
-# Wishart distribution
-#
-#   following the Wikipedia parameterization
-#
 """
     Wishart(ν, S)
 ```julia
-ν::Real           degrees of freedom (greater than p - 1)
+ν::Real           degrees of freedom in {1, 2, ..., p - 1} ∪ (p - 1, ∞)
 S::AbstractPDMat  p x p scale matrix
 ```
 The [Wishart distribution](http://en.wikipedia.org/wiki/Wishart_distribution)
-generalizes the gamma distribution to ``p\\times p`` real, positive definite
-matrices ``\\mathbf{H}``. If ``\\mathbf{H}\\sim \\textrm{W}_p(\\nu,\\mathbf{S})``,
-then its probability density function is
+generalizes the gamma distribution to ``p\\times p`` real, positive semidefinite
+matrices ``\\mathbf{H}``.
+
+If ``\\nu>p-1``, then ``\\mathbf{H}\\sim \\textrm{W}_p(\\nu, \\mathbf{S})``
+has rank ``p`` and its probability density function is
 
 ```math
 f(\\mathbf{H};\\nu,\\mathbf{S}) = \\frac{1}{2^{\\nu p/2} \\left|\\mathbf{S}\\right|^{\\nu/2} \\Gamma_p\\left(\\frac {\\nu}{2}\\right ) }{\\left|\\mathbf{H}\\right|}^{(\\nu-p-1)/2} e^{-(1/2)\\operatorname{tr}(\\mathbf{S}^{-1}\\mathbf{H})}.
 ```
 
-If ``\\nu`` is an integer, then a random matrix ``\\mathbf{H}`` given by
+If ``\\nu\\leq p-1``, then ``\\mathbf{H}`` has rank ``\\nu`` and
+a density is given [here](https://doi.org/10.1214/aos/1176325375).
+
+For integer ``\\nu``, a random matrix given by
 
 ```math
-\\mathbf{H} = \\mathbf{X}\\mathbf{X}^{\\rm{T}}, \\quad\\mathbf{X} \\sim \\textrm{MN}_{p,\\nu}(\\mathbf{0}, \\mathbf{S}, \\mathbf{I}_{\\nu})
+\\mathbf{H} = \\mathbf{X}\\mathbf{X}^{\\rm{T}},
+\\quad\\mathbf{X} \\sim \\textrm{MN}_{p,\\nu}(\\mathbf{0}, \\mathbf{S}, \\mathbf{I}_{\\nu})
 ```
 
-has ``\\mathbf{H}\\sim \\textrm{W}_p(\\nu, \\mathbf{S})``. For non-integer degrees of freedom,
-Wishart matrices can be generated via the [Bartlett decomposition](https://en.wikipedia.org/wiki/Wishart_distribution#Bartlett_decomposition).
+has ``\\mathbf{H}\\sim \\textrm{W}_p(\\nu, \\mathbf{S})``.
+For non-integer ``\\nu``, Wishart matrices can be generated via the
+[Bartlett decomposition](https://en.wikipedia.org/wiki/Wishart_distribution#Bartlett_decomposition).
 """
-struct Wishart{T<:Real, ST<:AbstractPDMat} <: ContinuousMatrixDistribution
-    df::T     # degree of freedom
-    S::ST     # the scale matrix
-    logc0::T  # the logarithm of normalizing constant in pdf
+struct Wishart{T<:Real, ST<:AbstractPDMat, R<:Integer} <: ContinuousMatrixDistribution
+    df::T          # degree of freedom
+    S::ST          # the scale matrix
+    logc0::T       # the logarithm of normalizing constant in pdf
+    rank::R        # rank of a sample
+    singular::Bool # singular of nonsingular wishart?
 end
 
 #  -----------------------------------------------------------------------------
 #  Constructors
 #  -----------------------------------------------------------------------------
 
-function Wishart(df::T, S::AbstractPDMat{T}) where T<:Real
+function Wishart(df::T, S::AbstractPDMat{T}, warn::Bool = true) where T<:Real
     p = dim(S)
-    df > p - 1 || throw(ArgumentError("df should be greater than dim - 1."))
-    logc0 = wishart_logc0(df, S)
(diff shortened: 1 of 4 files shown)


johnczito:
Bump.

src/matrix/wishart.jl
```diff
@@ -1,58 +1,69 @@
-# Wishart distribution
-#
-#   following the Wikipedia parameterization
-#
 """
     Wishart(ν, S)
 ```julia
-ν::Real           degrees of freedom (greater than p - 1)
+ν::Real           degrees of freedom in {1, 2, ..., p - 1} ∪ (p - 1, ∞)
```
mschauer, line 4:
```suggestion
ν::Real           degrees of freedom (whole number or a real number greater than p - 1)
```

src/matrix/wishart.jl
```diff
@@ -1,58 +1,69 @@
...
+If ``\\nu>p-1``, then ``\\mathbf{H}\\sim \\textrm{W}_p(\\nu, \\mathbf{S})``
+has rank ``p`` and its probability density function is
 
 ```math
 f(\\mathbf{H};\\nu,\\mathbf{S}) = \\frac{1}{2^{\\nu p/2} \\left|\\mathbf{S}\\right|^{\\nu/2} \\Gamma_p\\left(\\frac {\\nu}{2}\\right ) }{\\left|\\mathbf{H}\\right|}^{(\\nu-p-1)/2} e^{-(1/2)\\operatorname{tr}(\\mathbf{S}^{-1}\\mathbf{H})}.
 ```
 
-If ``\\nu`` is an integer, then a random matrix ``\\mathbf{H}`` given by
+If ``\\nu\\leq p-1``, then ``\\mathbf{H}`` has rank ``\\nu`` and
+a density is given [here](https://doi.org/10.1214/aos/1176325375).
```
mschauer, line 19: A density with respect to which measure? This is always important to state.

src/matrix/wishart.jl
```diff
@@ -1,58 +1,69 @@
...
+    rank::R        # rank of a sample
+    singular::Bool # singular of nonsingular wishart?
 end
 
 #  -----------------------------------------------------------------------------
 #  Constructors
 #  -----------------------------------------------------------------------------
 
-function Wishart(df::T, S::AbstractPDMat{T}) where T<:Real
+function Wishart(df::T, S::AbstractPDMat{T}, warn::Bool = true) where T<:Real
```
mschauer, line 44: a warning for expected functionality is not so nice

src/matrix/wishart.jl
```diff
@@ -66,27 +77,37 @@ show(io::IO, d::Wishart) = show_multline(io, d, [(:df, d.df), (:S, Matrix(d.S))]
...
 
 dim(d::Wishart) = dim(d.S)
 size(d::Wishart) = (p = dim(d); (p, p))
-rank(d::Wishart) = dim(d)
+rank(d::Wishart) = d.rank
 params(d::Wishart) = (d.df, d.S)
 @inline partype(d::Wishart{T}) where {T<:Real} = T
 
-mean(d::Wishart) = d.df * Matrix(d.S)
+function mean(d::Wishart)
```
mschauer, line 107: The mean of a singular Wishart is just nV also in the singular case

src/matrix/wishart.jl
mschauer: Tricky to specify, you could write "with respect to a suitable chosen volume element on the space of...".

src/matrix/wishart.jl
mschauer: To see this: It is just using the additivity of the covariance of independent centred Gaussians.

src/matrix/wishart.jl
johnczito: Done.

src/matrix/wishart.jl
johnczito: Fair, and I'm happy to get rid of this.

I guess I had in mind that this functionality is more unexpected for most users, so I wanted to give them a heads up.

src/matrix/wishart.jl
johnczito: Shouldn't the expectation over semidefinite matrices be a semidefinite matrix?

mschauer: The sum of only semi-definite matrices can be positive definite. Consider diagonal semi-definite matrices and their sums to see this.

mschauer: The sum of only semi-definite matrices can be positive definite. Consider diagonal semi-definite matrices and their sums to see this.

johnczito: Sure, but you wouldn't report that the mean of a k = 0 von Mises-Fisher distribution is 0, even though that's what sample averages converge to.

mschauer: Hehe, I would :-)

johnczito: Fair enough.

src/matrix/wishart.jl
```diff
@@ -46,7 +46,8 @@ function Wishart(df::T, S::AbstractPDMat{T}, warn::Bool = true) where T<:Real
     rnk = p
     singular = df <= p - 1
     if singular
-        !isinteger(df) && throw(ArgumentError("singular df must be an integer. got $(df)."))
+        df > 0 || throw(ArgumentError("df must be positive. got $(df)."))
```
mschauer, line 50: Put this outside the if statement

src/matrix/wishart.jl
```diff
@@ -179,7 +238,7 @@ end
 
 function _rand_params(::Type{Wishart}, elty, n::Int, p::Int)
     n == p || throw(ArgumentError("dims must be equal for Wishart"))
-    ν = elty( n + 1 + abs(10randn()) )
+    ν = elty( n - 1 + abs(10randn()) )
```
mschauer, line 241: Can you explain?

src/matrix/wishart.jl
```diff
@@ -105,18 +125,21 @@ function meanlogdet(d::Wishart)
...
 function entropy(d::Wishart)
+    d.singular && throw(ArgumentError("entropy not defined for singular Wishart."))
     p = dim(d)
     df = d.df
     -d.logc0 - 0.5 * (df - p - 1) * meanlogdet(d) + 0.5 * df * p
 end
 
 #  Gupta/Nagar (1999) Theorem 3.3.15.i
 function cov(d::Wishart, i::Integer, j::Integer, k::Integer, l::Integer)
+    d.singular && throw(ArgumentError("cov not defined for singular Wishart."))
```
mschauer, line 136: Can you check numerically if this holds for singular too, like for the mean?

src/matrix/wishart.jl
johnczito: Done.

src/matrix/wishart.jl
johnczito: `_rand_params ` should return random nonsingular parameters. You just need `ν > n - 1` for that.

src/matrix/wishart.jl
johnczito: Yeah, the derivation is the same, and it checks out in simulations.

johnczito:
Thanks for the help @mschauer. I guess if there are no objections I'll merge this in 24 hours?

src/matrix/wishart.jl
devmotion: Is it possible to remove the warning? I agree that it's probably not good to warn the users if Distributions supports this functionality, maybe it would be better to just mention it clearly in `show(::Wishart)`? Additionally, it is inconsistent with the behaviour of `Normal(4.0, 0.0)` and other special cases. An additional benefit of not printing warnings would be that the constructor would be compatible with Zygote, as noted in https://github.com/TuringLang/DistributionsAD.jl/pull/84 (although probably that is not a good argument since Distributions does not try to support Zygote in the first place).

johnczito: > maybe it would be better to just mention it clearly in show(::Wishart)?

Yes, great idea. Will do that soon. The warning was really intended more in the spirit of a deprecation: a temporary heads up about functionality people might not be expecting.

> although probably that is not a good argument since Distributions does not try to support Zygote in the first place

Agreed, but I am in favor of thinking through reasonable changes that `Distributions` could make that would allow `Turing` to simply add methods to `Distributions.Wishart` and so on instead of reimplementing it. I'll maybe open an issue about that someplace.

Issue #1064, opened 2020-02-05
Visualizations of distributions in documentation


carstenbauer:
I think it would be nice to have a little plot of the probability density function of each distribution in the documentation, for example here: https://juliastats.org/Distributions.jl/stable/univariate/#Continuous-Distributions-1.

devmotion:
Fixed by https://github.com/JuliaStats/Distributions.jl/pull/1403.

PR #1115, opened 2020-05-16
PR: Visualizations of distributions in documentation

BMasinde:


BMasinde:
Visualizations of distributions in documentation #1064 

Added basic density plots to Arcsine, Rayleigh, LogNormal and Weibull.

Package used: Gladfly. Dependency added in docs/Project.toml

mschauer:
Ref:
Closes#1064

johnczito:
Can you post a screenshot of what the docs look like when you build them?

BMasinde:
After building docs I found that @example does not work within @docs so moved examples 
to univariate.md

Attached are screenshots
Screenshot 2020-05-17 at 08 01 16
Screenshot 2020-05-17 at 08 01 26
Screenshot 2020-05-17 at 08 01 45
Screenshot 2020-05-17 at 08 15 33
Screenshot 2020-05-17 at 08 55 53

andreasnoack:
I like this. Any objections? If not then it would be great to have all univariate distributions covered.

johnczito:
I like it too. I think the plots should be a little smaller, and it would be nice if they were inside the box with the docstring

BMasinde:
> I like it too. I think the plots should be a little smaller, and it would be nice if they were inside the box with the docstring

About the doctstring, I'm facing a weird issue... changes that  I make to the function documentation (for example Arcsine) do not reflect when I build the docs.

After editing function documentation, I run julia make.jl to build the docs. 
Am I missing a step here?

johnczito:
Depending on how you've set things up, you might need to commit your changes and then `]up Distributions` before running `make.jl`?

BMasinde:
Tried `] up Distributions` the running `make.jl` . Still no luck; changes do not reflect in docs hosted by local server. 

Attached are screenshots showing this problem.
Screenshot 2020-05-22 at 22 41 28
Screenshot 2020-05-22 at 22 42 18

So this is the reason the plots are not within the docstrings. 
Moreover, I haven't seen any package documentation with plots inside docstrings.

matbesancon:
this looks good. I also think we should keep the docstrings concise and let the plots in the doc docs

BMasinde:
Added density plots to univariate distributions except for Kolmogorov distributions, NoncentralT,
NoncentralF, NormalCanon,  and Chernoff.

I made the plot smaller as requested.
Screenshot 2020-05-30 at 13 09 13
Screenshot 2020-05-30 at 13 09 24
Screenshot 2020-05-30 at 13 09 35
Screenshot 2020-05-30 at 13 10 01
Screenshot 2020-05-30 at 13 10 19
Screenshot 2020-05-30 at 13 10 29
Screenshot 2020-05-30 at 13 10 45
Screenshot 2020-05-30 at 13 11 00

.vscode/settings.json
```diff
@@ -0,0 +1,10 @@
+{
```
matbesancon, line 1: sorry I hadn't seen that, I don't think we should add editor dotfiles to the project

docs/Project.toml
```diff
@@ -1,6 +1,7 @@
 [deps]
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+Gadfly = "c91e804a-d5a3-530f-b6f0-dfbca275c004"
```
matbesancon, line 4: could we add a compat on Gadfly = "1"

.vscode/settings.json
BMasinde: Oh, my mistake... should I push a new commit to the brach with this file deleted?

docs/Project.toml
BMasinde: you mean edit this file by changing line 4?
```suggestion
Gadfly = "1"
```

matbesancon: no not this one, you can add `Gadfly = "1"` in the compat section of the project file

.vscode/settings.json
matbesancon: yes, I wanted to do it as a suggestion but it does not seem possible

docs/Project.toml
BMasinde: Right, thanks

docs/Project.toml
```diff
@@ -4,3 +4,4 @@ Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 
 [compat]
 Documenter = "~0.22"
+Gadfly = "1"
```
matbesancon, line 7:
```suggestion
Gadfly = "1"

```

BMasinde: should I go ahead commit this suggestion?

matbesancon: yes thank you, I don't think I can commit directly to your PR

.vscode/settings.json
matbesancon: nevermind it works, I committed it

mschauer:
Wait, the density plots look like smoothed density estimates estimated from a sample. Don't we want to show the true density, e.g. a Uniform density piecewise constant as 1 inside [0,1] and as 0 outside. 

Minor thing: the van Mises should only be plotted on [-pi, pi].

BMasinde:
> Wait, the density plots look like smoothed density estimates estimated from a sample. Don't we want to show the true density, e.g. a Uniform density piecewise constant as 1 inside [0,1] and as 0 outside.
> 
> Minor thing: the van Mises should only be plotted on [-pi, pi].

BMasinde:
Sorry closed by mistake.

Yes, a few of the density plots are smoothed (e.g uniform and Semicircle). I'm going to change this as suggested. 

I will also update the VonMises x grid. Thanks.

BMasinde:
@mschauer you also meant that I shouldn't  sample but instead evaluate the PDF over a fine grid
e.g for Gaussian:
`xgrid = -4:0.1:4`
`d = Normal(0,1)`
`plot(x = xgrid, y = pdf.(d, xgrid), Geom.line)`
This way results in this beautiful plot:
Screenshot 2020-06-01 at 19 20 23

mschauer:
That is a good idea

BMasinde:
I have made changes, now there's no sampling but the PDF is evaluated over a fine grid.

See the results as attached.
Screenshot 2020-06-05 at 12 37 15
Screenshot 2020-06-05 at 12 37 24
Screenshot 2020-06-05 at 12 37 50
Screenshot 2020-06-05 at 12 42 59
Screenshot 2020-06-05 at 12 43 36
Screenshot 2020-06-05 at 12 43 49
Screenshot 2020-06-05 at 12 43 58
Screenshot 2020-06-05 at 12 44 08
Screenshot 2020-06-05 at 12 44 17
Screenshot 2020-06-05 at 12 44 25

mschauer:
Nice. Final small thing: for the uniform and the van Mises we need to give an explicit range for the y axis.

BMasinde:
@mschauer see the attached figures with explicit y range. If these are ok I will push changes.
Screenshot 2020-06-05 at 14 57 44
Screenshot 2020-06-05 at 14 57 53

mschauer:
Sorry, I was reading the table wrong, apparently something is not right about the TriangularDist(1,1)

BMasinde:
Right problem was lower and upper limit were both set to 1.

Corrected this. Thanks.
Screenshot 2020-06-05 at 16 59 19

devmotion:
The last time I checked Gadfly was not well maintained (still does not support recent versions of Distributions: https://github.com/GiovineItalia/Gadfly.jl/issues/1551). This causes problems eg when building the documentation of MCMCChains. So I don't think it is a good idea to depend and rely on Gadfly for plotting.

mschauer:
I think Gadfly is not a bad choice: beautiful, actively developed, mature, doesn’t bother with displaying which creates the most friction in terms of stability

mschauer:
Yes, that’s true, but all those are more brittle because they  _plot_ and Gadfly just writes svg pictures and calls your browser to display

devmotion:
CairoMakie also writes svg pictures, and IIRC also some Plots backends can.

mschauer:
That’s important, right. But in the end, we just want pictures in the docs and a way to generate them as plots such that we can update them. Let’s separate picture generation from the docs and add the pictures as assets with a script to generate them?

st--:
I'd be in favour of generating the plots on-the-fly instead of separately storing them as assets, and to use a library that doesn't depend on Distributions.jl itself - it would be very easy to change the plotting from Gadfly to Plots, say, and given that it doesn't use any fancy features, there shouldn't be any significant issues with stability. @BMasinde are you still around? If not I'd be happy to make these changes.

devmotion:
>  If not I'd be happy to make these changes.

Just a heads up, I'm currently playing around with GR in a local branch (IMO Plots is an unnecessarily large dependency given that these are only simple line plots) and plan to open a PR with some additional changes (based on this PR here) within the next few days.

devmotion:
> (if I understood correctly he would have a PR soon)

Yes, I'll open a PR that is based on your PR here and replaces Gadfly with GR but includes some additional changes as well.

devmotion:
The PR: https://github.com/JuliaStats/Distributions.jl/pull/1403

Issue #624, opened 2017-06-21
TruncatedNormal with Inf mean hangs


cossio:
Just try this example:

```
rand(Distributions.TruncatedNormal(Inf, 1., 0., 1.))
```

It hangs forever. I am using Julia 0.5.

It should raise an error somewhere instead.

matbesancon:
Oh man this hasn't been touched in ages. The fault is on `randnt` here, untested. The algorithm seems not to have termination guarantees, we may had a timeout or maxiter

cossio:
In this case I think a simple check `@assert -Inf < mu < Inf` would suffice.

matbesancon:
yes sure, although it is very likely to hang for large non-Inf numbers too right?

matbesancon:
Anyhow PR welcome

cossio:
> yes sure, although it is very likely to hang for large non-Inf numbers too right?

No. It just hangs because when the mean is infinite we end up calling `randnt(Inf, Inf)` which never terminates.

PR #1119, opened 2020-05-21
fix truncnorm with infinite mean

cossio:
fix https://github.com/JuliaStats/Distributions.jl/issues/624

### DIFF
--- a/src/truncated/normal.jl
+++ b/src/truncated/normal.jl
@@ -141,10 +141,14 @@ function rand(rng::AbstractRNG, d::Truncated{Normal{T},Continuous}) where T <: R
     d0 = d.untruncated
     μ = mean(d0)
     σ = std(d0)
-    a = (d.lower - μ) / σ
-    b = (d.upper - μ) / σ
-    z = randnt(rng, a, b, d.tp)
-    return μ + σ * z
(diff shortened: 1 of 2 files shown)


cossio:
The CI fails look unrelated to this PR.

cossio:
Merge?

cossio:
CC: @matbesancon

src/truncated/normal.jl
```diff
@@ -141,10 +141,14 @@ function rand(rng::AbstractRNG, d::Truncated{Normal{T},Continuous}) where T <: R
     d0 = d.untruncated
     μ = mean(d0)
     σ = std(d0)
-    a = (d.lower - μ) / σ
-    b = (d.upper - μ) / σ
-    z = randnt(rng, a, b, d.tp)
-    return μ + σ * z
+    if isfinite(μ)
+        a = (d.lower - μ) / σ
```
matbesancon, line 145: if isfinite()
...
return ..
end
return end

no need for an else with early return

matbesancon: no that's alright by me, not worth bothering too much, the main concern is keeping the code homogeneous

Issue #1117, opened 2020-05-19
Should distributions have `==` defined?


nickrobinson251:
This took me by surprise, so i wondered if this was intentioned behaviour or if it was a missing feature?
```julia
julia> mv = MvNormal(rand(3), let x=rand(3,3); x*x'+I; end);

julia> mv2 = deepcopy(mv);

julia> mv == mv2
false
```

johnczito:
It's intended in the sense that it's a general phenomenon not specific to distributions: [[1]](https://github.com/JuliaLang/julia/issues/4648), [[2]](https://stackoverflow.com/questions/59274263/why-do-structs-and-a-mutable-structs-have-different-default-equality-operators), [[3]](https://discourse.julialang.org/t/surprising-struct-equality-test/4890).

nickrobinson251:
right, makes sense, so i think we file it under "missing feature"? 

I presume that we want `mv == mv2` in the example above, and it's just that this is a case where Julia doesn't give us it for free.

nickrobinson251:
xref https://github.com/JuliaStats/Distributions.jl/issues/696

PR #1122, opened 2020-06-02
Define equality for Sampleables

morris25:
Closes https://github.com/JuliaStats/Distributions.jl/issues/1117.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ authors = ["JuliaStats"]
 version = "0.23.3"
 
 [deps]
+AutoHashEquals = "15f4f7f2-30c1-5605-9d31-71845cf9641f"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 PDMats = "90014a1f-27ba-587c-ab20-58faa44d9150"
@@ -17,7 +18,7 @@ StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 
 [compat]
 FillArrays = "0.8"
-PDMats = "0.9.12"
+PDMats = "0.10"
 QuadGK = "2"
 SpecialFunctions = "0.8, 0.9, 0.10"
 StatsBase = "0.32, 0.33"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.23.3"
+version = "0.23.4"
 
 [deps]
 AutoHashEquals = "15f4f7f2-30c1-5605-9d31-71845cf9641f"
--- a/Project.toml
+++ b/Project.toml
@@ -17,6 +17,7 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 
 [compat]
+AutoHashEquals = "0.2"
 FillArrays = "0.8"
 PDMats = "0.10"
 QuadGK = "2"
--- a/Project.toml
+++ b/Project.toml
@@ -4,7 +4,6 @@ authors = ["JuliaStats"]
 version = "0.23.4"
 
 [deps]
-AutoHashEquals = "15f4f7f2-30c1-5605-9d31-71845cf9641f"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 PDMats = "90014a1f-27ba-587c-ab20-58faa44d9150"
@@ -17,7 +16,6 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 
 [compat]
-AutoHashEquals = "0.2"
 FillArrays = "0.8"
 PDMats = "0.10"
 QuadGK = "2"
--- a/src/common.jl
+++ b/src/common.jl
@@ -74,9 +74,9 @@ nsamples(::Type{D}, x::Array{Matrix{T}}) where {D<:Sampleable{Matrixvariate},T<:
 
 for func in (:(==), :isequal, :isapprox)
     @eval function Base.$func(s1::A, s2::B; kwargs...) where {A<:Sampleable, B<:Sampleable}
-        nameof(A) == nameof(B) || return false
+        nameof(A) === nameof(B) || return false
         fields = fieldnames(A)
-        fields == fieldnames(B) || return false
+        fields === fieldnames(B) || return false
 
         for f in fields
             isdefined(s1, f) && isdefined(s2, f) || return false
@@ -87,6 +87,15 @@ for func in (:(==), :isequal, :isapprox)
     end
 end
 
+function Base.hash(s::S, h::UInt) where S <: Sampleable
+    hashed = hash(nameof(S), h)
+
+    for f in fieldnames(S)
+        hashed = hash(getfield(s, f), hashed)
+    end
+
+    return hashed
+end
 
 """
     Distribution{F<:VariateForm,S<:ValueSupport} <: Sampleable{F,S}
--- a/src/common.jl
+++ b/src/common.jl
@@ -72,30 +72,19 @@ nsamples(::Type{D}, x::AbstractMatrix) where {D<:Sampleable{Multivariate}} = siz
 nsamples(::Type{D}, x::Number) where {D<:Sampleable{Matrixvariate}} = 1
 nsamples(::Type{D}, x::Array{Matrix{T}}) where {D<:Sampleable{Matrixvariate},T<:Number} = length(x)
 
-function Base.:(==)(s1::A, s2::B) where {A<:Sampleable, B<:Sampleable}
-    nameof(A) == nameof(B) || return false
-    fields = fieldnames(A)
-    fields == fieldnames(B) || return false
+for func in (:(==), :isequal, :isapprox)
+    @eval function Base.$func(s1::A, s2::B; kwargs...) where {A<:Sampleable, B<:Sampleable}
+        nameof(A) == nameof(B) || return false
+        fields = fieldnames(A)
+        fields == fieldnames(B) || return false
 
-    for f in fields
-        isdefined(s1, f) && isdefined(s2, f) || return false
-        getfield(s1, f) == getfield(s2, f) || return false
+        for f in fields
+            isdefined(s1, f) && isdefined(s2, f) || return false
+            $func(getfield(s1, f), getfield(s2, f); kwargs...) || return false
+        end
+
+        return true
     end
-
-    return true
-end
-
-function Base.isequal(s1::A, s2::B) where {A<:Sampleable, B<:Sampleable}
-    nameof(A) == nameof(B) || return false
-    fields = fieldnames(A)
-    fields == fieldnames(B) || return false
-
-    for f in fields
-        isdefined(s1, f) && isdefined(s2, f) || return false
-        isequal(getfield(s1, f), getfield(s2, f)) || return false
-    end
-
-    return true
 end
 
 
--- a/src/common.jl
+++ b/src/common.jl
@@ -72,8 +72,32 @@ nsamples(::Type{D}, x::AbstractMatrix) where {D<:Sampleable{Multivariate}} = siz
 nsamples(::Type{D}, x::Number) where {D<:Sampleable{Matrixvariate}} = 1
 nsamples(::Type{D}, x::Array{Matrix{T}}) where {D<:Sampleable{Matrixvariate},T<:Number} = length(x)
 
-Base.:(==)(d1::T, d2::T) where {T<:Sampleable} = all(i -> getfield(d1, i) == getfield(d2, i), 1:fieldcount(T))
-Base.:(==)(d1::Sampleable, d2::Sampleable) = false
+function Base.:(==)(s1::A, s2::B) where {A<:Sampleable, B<:Sampleable}
+    nameof(A) == nameof(B) || return false
+    fields = fieldnames(A)
+    fields == fieldnames(B) || return false
+
+    for f in fields
+        isdefined(s1, f) && isdefined(s2, f) || return false
+        getfield(s1, f) == getfield(s2, f) || return false
+    end
+
+    return true
+end
+
+function Base.isequal(s1::A, s2::B) where {A<:Sampleable, B<:Sampleable}
+    nameof(A) == nameof(B) || return false
+    fields = fieldnames(A)
+    fields == fieldnames(B) || return false
+
+    for f in fields
+        isdefined(s1, f) && isdefined(s2, f) || return false
+        isequal(getfield(s1, f), getfield(s2, f)) || return false
+    end
+
+    return true
+end
+
 
 """
     Distribution{F<:VariateForm,S<:ValueSupport} <: Sampleable{F,S}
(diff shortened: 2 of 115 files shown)


Project.toml
```diff
@@ -1,9 +1,10 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.23.3"
+version = "0.23.4"
 
 [deps]
+AutoHashEquals = "15f4f7f2-30c1-5605-9d31-71845cf9641f"
```
nickrobinson251, line 7: let's add compat for this?

Project.toml
```diff
@@ -17,7 +18,7 @@ StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 
 [compat]
 FillArrays = "0.8"
-PDMats = "0.9.12"
+PDMats = "0.10"
```
nickrobinson251, line 21: why did this one need to change? can we still support `0.9`?

src/edgeworth.jl
```diff
@@ -9,7 +9,7 @@ abstract type EdgeworthAbstract <: ContinuousUnivariateDistribution end
 skewness(d::EdgeworthAbstract) = skewness(d.dist) / sqrt(d.n)
 kurtosis(d::EdgeworthAbstract) = kurtosis(d.dist) / d.n
 
-struct EdgeworthZ{D<:UnivariateDistribution} <: EdgeworthAbstract
+@auto_hash_equals struct EdgeworthZ{D<:UnivariateDistribution} <: EdgeworthAbstract
```
nickrobinson251, line 12: perhaps add something to the dev docs recommending that `isequal` and `hash` are defined to (and perhaps recommending AutoHasEquals.jl)? https://juliastats.org/Distributions.jl/latest/extends/#Create-New-Samplers-and-Distributions-1

Project.toml
morris25: `0.9` causes the equality tests for the `MvNormal` Distributions to fail.

morris25:
@matbesancon is there anyone in particular I should ask to review this?

matbesancon:
@mschauer?

mschauer:
Are there any drawbacks of going this route?

matbesancon:
One I can see is that it adds yet another layer to the code. Also
precompilation is already super long for the package.

Is there a reason we need this? Hashing is rarely used for distributions

nickrobinson251:
> Is there a reason we need this?

the current behaviour in #1117 seems odd to me i.e. distributions are sometime equal to one another and i think Distributions.jl should support that

mschauer:
I think this makes sense, as evidenced by
https://github.com/mschauer/GaussianDistributions.jl/blob/6f3ac9c329c1e832c414f940bf6c8da91e34a16a/src/GaussianDistributions.jl#L68

Regarding the increase computation time, could you measure it? Then we know.

One question I have: AutoHashEquals uses `isequal` and not `==`. Is this the right thing? E.g. should `NaN`-means of `Normal`s compare unequal?

In that case I should also change my code perhaps.

nickrobinson251:
> Regarding the increase computation time, could you measure it?

What's the best way to measure compile time? Here's what i tried:

This branch:
```julia
$ git checkout invenia/sm/equality
Previous HEAD position was feae3ba9 Patch bump
HEAD is now at 4e646d0d Adds compat and note

$ julia --project=@. -q
(Distributions) pkg> up
   Updating registry at `~/.julia/registries/General`
   Updating git-repo `git@github.com:JuliaRegistries/General.git`
   Updating registry at `~/.julia/registries/Invenia`
   Updating git-repo `git@gitlab.invenia.ca:invenia/PackageRegistry.git`
Updating `~/repos/Distributions.jl/Project.toml`
  [15f4f7f2] + AutoHashEquals v0.2.0
  [90014a1f] ↑ PDMats v0.9.12 ⇒ v0.10.0
Updating `~/repos/Distributions.jl/Manifest.toml`
  [7d9fca2a] - Arpack v0.4.0
  [68821587] - Arpack_jll v3.5.0+3
  [15f4f7f2] + AutoHashEquals v0.2.0
  [4536629a] - OpenBLAS_jll v0.3.9+4
  [90014a1f] ↑ PDMats v0.9.12 ⇒ v0.10.0

julia> @timed using Distributions
[ Info: Precompiling Distributions [31c24e10-a181-5473-b8eb-7969acd0382f]
(value = nothing, time = 5.987775708, bytes = 164154461, gctime = 0.03380193, gcstats = Base.GC_Diff(164154461, 32, 0, 2748228, 522, 0, 33801930, 3, 0))
```
Master:
```julia
$ git checkout origin/master
Previous HEAD position was 4e646d0d Adds compat and note
HEAD is now at feae3ba9 Patch bump

$ julia --project=@. -q
(Distributions) pkg> up
   Updating registry at `~/.julia/registries/General`
   Updating git-repo `git@github.com:JuliaRegistries/General.git`
   Updating registry at `~/.julia/registries/Invenia`
   Updating git-repo `git@gitlab.invenia.ca:invenia/PackageRegistry.git`
Updating `~/repos/Distributions.jl/Project.toml`
  [90014a1f] ↓ PDMats v0.10.0 ⇒ v0.9.12
Updating `~/repos/Distributions.jl/Manifest.toml`
  [7d9fca2a] + Arpack v0.4.0
  [68821587] + Arpack_jll v3.5.0+3
  [15f4f7f2] - AutoHashEquals v0.2.0
  [4536629a] + OpenBLAS_jll v0.3.9+4
  [90014a1f] ↓ PDMats v0.10.0 ⇒ v0.9.12

julia> @timed using Distributions
[ Info: Precompiling Distributions [31c24e10-a181-5473-b8eb-7969acd0382f]
(value = nothing, time = 6.000390004, bytes = 171419321, gctime = 0.027515303, gcstats = Base.GC_Diff(171419321, 38, 0, 2832029, 536, 0, 27515303, 3, 0))
```

morris25:
I was just about to say the same thing. Mine results were ~7 seconds with this branch also being slightly faster.

morris25:
My use case is just to add some tests to one of my packages checking that a couple of functions produce equivalent results.

nickrobinson251:
> One question I have: AutoHashEquals uses isequal and not ==. Is this the right thing? E.g. should NaN-means of Normals compare unequal?

Good question! I think i agree with this issue https://github.com/andrewcooke/AutoHashEquals.jl/issues/18, that `==` should use `==` recursively rather than `isequals`. If that were the case then we'd be able to use `==` to have NaN-means of Normals be `!=` but still have them be `isequal`, which i think is the behaviour we'd want?

mschauer:
In concur that `==` should recursively call `==`.

morris25:
> In concur that == should recursively call ==.

Is that suggestion for this PR or for AutoHashEquals? Should we move forward with this method for now?

ararslan:
Do we actually need hashing, or can we just define something like
```julia
Base.:(==)(d1::T, d2::T) where {T<:Distribution} = all(i -> getfield(d1, i) == getfield(d2, i), 1:fieldcount(T))
Base.:(==)(d1::Distribution, d2::Distribution) = false
```
? That is, recursively compare all internal fields of two distributions of the same type and declare distributions of different types to be unequal.

nickrobinson251:
I don't know... I'd have guessed it's acceptable to define `==` without `hash`. 

It seems [having `hash` defined usually implies having `==` defined too ](https://docs.julialang.org/en/v1/base/base/#Base.hash) but [having `==` defined does not imply `hash` is also defined](https://docs.julialang.org/en/v1/base/math/#Base.:==)

So if we don't need/want `hash` (which neither Sam nor I need for our uses), then I think we could implement this as Alex suggests if folks prefer?

ararslan:
I'd personally prefer that over taking on another dependency if nobody needs `hash`.

matbesancon:
I would also favour the solution with only == and no external dependency

nickrobinson251:
Cool :) I guess we just want to change `Distribution` to `Sampleable` in the code Alex suggests above?

Shall we add `isequal` too (calling `isequal` on fields)?

matbesancon:
yes for both points

morris25:
The downside to the changed code is that now `MvNormal(map(Float64,ones(2))) != MvNormal(map(Float32,ones(2)))`.

nickrobinson251:
Hmm... How many distributions have a similar issue? If it's just MvNormal we can special case it?

ararslan:
Actually, that would indeed affect any distribution with a type parameter.

ararslan:
So, we could do something like this:
```julia
function Base.:(==)(d1::Distribution, d2::Distribution)
    nameof(d1) === nameof(d2) || return false
    names = fieldnames(typeof(d1))
    names == fieldnames(typeof(d2)) || return false
    for name in names
        (isdefined(d1, name) && isdefined(d2, name)) || return false
        getfield(d1, name) == getfield(d2, name) || return false
    end
    return true
end
```
I kind of hate it though. Particularly the `nameof` check, though without it, you'd be able to equate `Normal` and `LogNormal`, for example.

nickrobinson251:
If AutoHashEquals had the desired behaviour (via fixing https://github.com/andrewcooke/AutoHashEquals.jl/issues/18) would that be preferable? Or would it still be preferred to define `==` and `isequal` directly (via something like Alex's latest suggestion above - thanks for that btw!)?

src/common.jl
```diff
@@ -72,6 +72,33 @@ nsamples(::Type{D}, x::AbstractMatrix) where {D<:Sampleable{Multivariate}} = siz
...
+    fields = fieldnames(A)
+    fields == fieldnames(B) || return false
+
+    for f in fields
+        isdefined(s1, f) && isdefined(s2, f) || return false
+        isequal(getfield(s1, f), getfield(s2, f)) || return false
+    end
+
+    return true
+end
```
nickrobinson251, line 99: rather than maintain this somewhat complicated code in two places, we could write it like
```julia
for func in (:(==), :isequal)
    @eval function Base.$func(s1::A, s2::B) where {A<:Sampleable, B<:Sampleable}
        nameof(A) == nameof(B) || return false
        fields = fieldnames(A)
        fields == fieldnames(B) || return false

        for f in fields
            isdefined(s1, f) && isdefined(s2, f) || return false
            isequal(getfield(s1, f), getfield(s2, f)) || return false
        end

        return true
    end
end
```	
to my eyes that's clearer that it's the same logic in both cases

src/common.jl
```diff
@@ -72,6 +72,33 @@ nsamples(::Type{D}, x::AbstractMatrix) where {D<:Sampleable{Multivariate}} = siz
...
+
+    for f in fields
+        isdefined(s1, f) && isdefined(s2, f) || return false
+        getfield(s1, f) == getfield(s2, f) || return false
+    end
+
+    return true
+end
+
+function Base.isequal(s1::A, s2::B) where {A<:Sampleable, B<:Sampleable}
```
nickrobinson251, line 88: is wonder if we should also define `isapprox` the same way (just also allowing `kwargs...`)?
```julia
function Base.isapprox(s1::A, s2::Bl kwargs...) where {A<:Sampleable, B<:Sampleable}
```

src/common.jl
nickrobinson251: also defining `isapprox` would then just be adding it to the list of funs we loop over (and change the signature to allow `kwargs`)

src/multivariate/mvnormal.jl
```diff
@@ -197,6 +197,11 @@ function MvNormal(μ::AbstractVector{T}, Σ::AbstractPDMat{T}) where {T<:Real}
     MvNormal{T,typeof(Σ), typeof(μ)}(μ, Σ)
 end
 
+function MvNormal(μ::AbstractVector{<:Real}, Σ::AbstractPDMat)
+    R = Base.promote_eltype(μ, Σ)
+    MvNormal(convert(AbstractArray{R}, μ), convert(AbstractArray{R}, Σ))
+end
```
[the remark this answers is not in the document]
morris25: Nope. It removes a breaking ambiguity
```
MethodError: MvNormal(::Array{Float64,1}, ::PDMat{Float32,Array{Float32,2}}) is ambiguous. Candidates:
    (::Type{MvNormal})(μ::AbstractArray{T,1} where T, Σ::AbstractPDMat) in Distributions at /Users/sam/.julia/dev/Distributions/src/multivariate/mvnormal.jl:206
    (::Type{MvNormal})(μ::AbstractArray{#s156,1} where #s156<:Real, Σ::AbstractArray{#s155,2} where #s155<:Real) in Distributions at /Users/sam/.julia/dev/Distributions/src/multivariate/mvnormal.jl:211
  Possible fix, define
    MvNormal(::AbstractArray{#s156,1} where #s156<:Real, ::AbstractPDMat)
  Stacktrace:
   [1] top-level scope at /Users/sam/.julia/dev/Distributions/test/mvnormal.jl:153
   [2] top-level scope at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Test/src/Test.jl:1113
   [3] top-level scope at /Users/sam/.julia/dev/Distributions/test/mvnormal.jl:149
```

morris25:
Added `isapprox` and did the suggested cleanup. Is this good to merge now?

rofinn:
LGTM, unless there are any outstanding objections, I'll play to merge this on Monday.

src/common.jl
```diff
@@ -72,6 +72,22 @@ nsamples(::Type{D}, x::AbstractMatrix) where {D<:Sampleable{Multivariate}} = siz
 nsamples(::Type{D}, x::Number) where {D<:Sampleable{Matrixvariate}} = 1
 nsamples(::Type{D}, x::Array{Matrix{T}}) where {D<:Sampleable{Matrixvariate},T<:Number} = length(x)
 
+for func in (:(==), :isequal, :isapprox)
+    @eval function Base.$func(s1::A, s2::B; kwargs...) where {A<:Sampleable, B<:Sampleable}
+        nameof(A) == nameof(B) || return false
+        fields = fieldnames(A)
+        fields == fieldnames(B) || return false
```
nalimilan, line 79: Could help the compiler determine this statically for differing types:
```suggestion
        nameof(A) === nameof(B) || return false
        fields = fieldnames(A)
        fields === fieldnames(B) || return false
```

nalimilan:
My reading of https://docs.julialang.org/en/v1/base/math/#Base.:== implies that it's better to define hash if you define `==`. Otherwise the type is going to behave incorrectly in dicts. But that docstring is weird since it's hard to know in advance whether a type will be used as a dict key. Indeed https://docs.julialang.org/en/v1/base/base/#Base.hash says that "`isequal(x,y)` implies `hash(x)==hash(y)`", which isn't the case in the current state of this PR.

One simple solution is to define hash to always return the same value: that's inefficient but correct.

nickrobinson251:
I believe we already discussed this, the docs state it's optional, and the decision was not to implement `hash` until someone had a use-case for it.

mschauer:
What kind of hash computes `hash` currently for Distributions?

nalimilan:
The fallback `hash` method uses `objectid` so it's unique to each instance.

mschauer:
That's not so nice, because this way `isequal` and `hash` do not match and there is no easy way to opt out.

mschauer:
On the other hand, exactly the same logic that defines `isequal` can define the `hash`, so let's just do that.

morris25:
Sorry for the delay. I was off last week. Hashing has been added

morris25:
All good now @nalimilan?

src/common.jl
```diff
@@ -72,6 +72,31 @@ nsamples(::Type{D}, x::AbstractMatrix) where {D<:Sampleable{Multivariate}} = siz
...
+            isdefined(s1, f) && isdefined(s2, f) || return false
+            $func(getfield(s1, f), getfield(s2, f); kwargs...) || return false
+        end
+
+        return true
+    end
+end
+
+function Base.hash(s::S, h::UInt) where S <: Sampleable
+    hashed = hash(nameof(S), h)
```
nalimilan, line 91: I think it would be better to add a hash which is specific to the family of types that are being hashed, to avoid collisions with other types (even if it's unlikely). This is inspired by how `hash` works for `AbtractArray`:
```suggestion
    hashed = hash(Sampleable, h)
    hashed = hash(nameof(S), hashed)
```

Otherwise, looks good!

PR #1124, opened 2020-06-06
Maxwell boltzmann

bojohnson02:
I've added the Maxwell-Boltzmann distribution, something I've been wanting to do for a while. I'm still really new to Julia and contributing to any package. Please let me know how I can make this work. I couldn't get things working on my local machine: the docs and the tests. I'm not sure what I was doing wrong. If you could point me in the direction to run those locally, I can do that and make a new pull request.

### DIFF
--- a/src/univariate/continuous/maxwellboltzmann.jl
+++ b/src/univariate/continuous/maxwellboltzmann.jl
@@ -4,7 +4,7 @@
 The *Maxwell-Boltzmann distribution* has probability density function
 
 ```math
-f(x) = \\sqrt{\\frac{2}{\\pi}}\\frac{x^2e^{-x^2/\\left(2a^2\\right)}}{a^3}, \\quad x \\in [0, \\infty] \\
+f(x; a) = \\sqrt{\\frac{2}{\\pi}}\\frac{x^2e^{-x^2/\\left(2a^2\\right)}}{a^3}, \\quad x \\in [0, \\infty] \\
 ```
 where the parameter
 ```math
--- a/src/univariate/continuous/maxwellboltzmann.jl
+++ b/src/univariate/continuous/maxwellboltzmann.jl
@@ -1,0 +1,50 @@
+"""
+    MaxwellBoltzmann(a)
+The *Maxwell-Boltzmann distribution* is the chi distribution with three degrees of freedom and scale parameter `a`. The scale parameter is related to the mass and temperature of particles a = √(kT/m) where k is Boltzmann's constant, T is the temperature of the gas in Kelvin, and m is the mass of the particle in kilograms.
+
+External links
+
+* [Maxwell-Boltzmann distribution on Wikipedia](https://en.wikipedia.org/wiki/Maxwell-Boltzmann_distribution)
+"""
+struct MaxwellBoltzmann{T<:Real} <: ContinuousUnivariateDistribution
+    a::T
+    MaxwellBoltzmann{T}(a::T) where {T} = new{T}(a)
+end
+
+function MaxwellBoltzmann(a::T; check_args=true) where {T<:Real}
+    check_args && @check_args(MaxwellBoltzmann, a > zero(a))
+    return MaxwellBoltzmann{T}(a)
+end
+
+MaxwellBoltmann() = MaxwellBoltzmann(1.0)
+
+MaxwellBoltzmann(a::Integer) = MaxwellBoltzmann(float(a))
+
+@distr_support MaxwellBoltzmann 0.0 Inf
+
+#### Parameters
+
+dof(d::MaxwellBoltzmann) = 3
+params(d::MaxwellBoltzmann) = (d.a,)
+@inline partype(d::MaxwellBoltzmann{T}) where {T<:Real} = T
+
+#### Conversions
+
+convert(::Type{MaxwellBoltzmann{T}}, a::Real) where {T<:Real} = MaxwellBoltzmann(T(a))
+convert(::Type{MaxwellBoltzmann{T}}, d::MaxwellBoltzmann{S}) where {T<:Real, S<:Real} = MaxwellBoltzmann(T(d.a))
+
+#### Statistics
+
+mean(d::MaxwellBoltzmann) = 2d.a * sqrt(2 / π)
+var(d::MaxwellBoltzmann) = d.a^2(3π - 8) / π
+skewness(d::MaxwellBoltzmann) = 2 * sqrt(2) * (16 - 5π) / (3π - 8)^(3/2)
+kurtosis(d::MaxwellBoltzmann) = 4 * (-96 + 40π - 3π^2) / (3π - 8)^2
+mode(d::MaxwellBoltzmann) = sqrt(2) * d.a
+entropy(d::MaxwellBoltzmann) = log(d.a * sqrt(2π)) + (-digamma(1)) - 0.5
+
+#### Evaluation
+
+@_delegate_statsfun MaxwellBoltzmann maxwellboltzmann a
+
+pdf(d::MaxwellBoltzmann, x::Real) = sqrt(2 / π) * x^2 * exp(-x^2 / (2d.a^2)) / d.a^3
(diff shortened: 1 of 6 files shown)


johnczito:
Thanks! This would be a good add. The tests are failing for two reasons:

- You've introduced a new dependence `PhysicalConstants` without listing its `uuid` under `[deps]` in `Distributions.jl/Project.toml`. We may not ultimately want to introduce that dependency here, but you can still do it locally just to get the tests running.
- When you add `"maxwellboltzmann"` in `test/runtests.jl`, it's expecting to find a file `test/maxwellboltzmann.jl` which does not exist. So you should create that and add some tests.

You can build the docs locally by running `Distributions.jl/docs/make.jl` in the REPL.

bojohnson02:
I've added a line to `test/maxwellboltzmann.jl` and run the tests. When running the tests locally in the REPL the MaxwellBoltzmann ones didn't show up. And when running `docs/make.jl` in the RELP I get an error saying `undefined binding 'MaxwellBoltzmann' in `@docs` block in src/univariate.md:94-144`. Is there any other place I need to be putting references to my new code?

johnczito:
It's probably not using the new branch of your fork. You can to do something like 
```
 pkg> add https://github.com/bojohnson02/Distributions.jl.git#MaxwellBoltzmann
```
so that when you call `using Distributions` it's using your new stuff. Or instead of the GitHub link you can give the file path to your local fork, with the `#MaxwellBoltzmann` on the end to indicate the correct branch.

bojohnson02:
Thanks so much for helping to point me in the right direction to get these fixed. I've added my fork with Pkg and am now getting the documentation working. But in running the tests I'm getting an error from `undefined binding 'params' in `@docs` block in src/multivariate.md:89-95`. Did I inadvertently add something to a file to change this?

bojohnson02:
I was able to fix whatever error was showing up before. By removing and then adding back my branch in Pkg, the errors went away. The documentation is looking good now and it looks like it has passed all the tests on my machine. I'm still pretty new to this contributing thing, so if there's anything else I can do, please let me know.

src/univariate/continuous/maxwellboltzmann.jl
```diff
@@ -0,0 +1,63 @@
...
+    a::T
+    MaxwellBoltzmann{T}(a::T) where {T} = new{T}(a)
+end
+
+function MaxwellBoltzmann(a::T; check_args=true) where {T<:Real}
+    check_args && @check_args(MaxwellBoltzmann, a > zero(a))
+    return MaxwellBoltzmann{T}(a)
+end
+
+MaxwellBoltmann() = MaxwellBoltzmann(1.0)
```
jamesgardner1421, line 33:
```suggestion
MaxwellBoltzmann() = MaxwellBoltzmann(1.0)
```
There's a typo here.

src/univariate/continuous/maxwellboltzmann.jl
```diff
@@ -0,0 +1,63 @@
...
+```math
+f(x; a) = \\sqrt{\\frac{2}{\\pi}}\\frac{x^2}{a^3}\\exp\\left(-\\frac{x^2}{2a^2}\\right)
+```
+where the parameter
+```math
+a = \\sqrt{\\frac{kT}{m}}
+```
+
+```julia
+MaxwellBoltamann()        # Maxwell-Boltzmann distribution with a = 1
```
jamesgardner1421, line 14:
```suggestion
MaxwellBoltzmann()        # Maxwell-Boltzmann distribution with a = 1
```
Another typo here, less important this time.

bojohnson02:
Thanks for the feedback @jamesgardner1421. I'm actually thinking to close this pull request. I want to get things working on my machine and clean things up and then open up a new pull request where I've done things correctly. Is that a good plan?

bojohnson02:
@jamesgardner1421 I'm pretty new to this contributing thing. What are the best next steps for me to take with the PR?

src/univariate/continuous/maxwellboltzmann.jl
```diff
@@ -0,0 +1,84 @@
...
+where
+```math
+a = \\sqrt{\\frac{kT}{m}}
+```
+`k` is the Boltzmann constant and `m` is the particle mass.
+
+```julia
+MaxwellBoltzmann()        # Maxwell-Boltzmann distribution with a = 1
+MaxwellBoltzmann(a)       # Maxwell-Boltzmann distribution with parameter a
+MaxwellBoltzmann(T, m)    # Maxwell-Boltzmann distribution with temperature T and mass m
```
lmiq, line 17: I suggest fixing this to some explicit unit (kcal/(mol/K), preferably):

```
MaxwellBoltzmann(T, m)    # Maxwell-Boltzmann distribution with temperature T and mass m with k in kcal mol⁻¹ K⁻¹
```

and adding a fourth method:

```
MaxwellBoltzmann(T, m, k) # Maxwell-Boltzmann distribution with temperature T and mass m given k 
```

src/univariate/continuous/maxwellboltzmann.jl
```diff
@@ -0,0 +1,84 @@
...
+end
+
+function MaxwellBoltzmann(a::T; check_args=true) where {T<:Real}
+    check_args && @check_args(MaxwellBoltzmann, a > zero(a))
+    return MaxwellBoltzmann{T}(a)
+end
+
+MaxwellBoltzmann() = MaxwellBoltzmann(1.0)
+MaxwellBoltzmann(temp::T, mass::T) where {T<:Real} = MaxwellBoltzmann(sqrt(ustrip(k_B) * temp / mass))
+
```
lmiq, line 38: `MaxwellBoltzmann(temp, mass, k) = MaxwellBoltzmann(sqrt(k * temp / mass))
`

src/univariate/continuous/maxwellboltzmann.jl
```diff
@@ -0,0 +1,84 @@
...
+    MaxwellBoltzmann{T}(a::T) where {T} = new{T}(a)
+end
+
+function MaxwellBoltzmann(a::T; check_args=true) where {T<:Real}
+    check_args && @check_args(MaxwellBoltzmann, a > zero(a))
+    return MaxwellBoltzmann{T}(a)
+end
+
+MaxwellBoltzmann() = MaxwellBoltzmann(1.0)
+MaxwellBoltzmann(temp::T, mass::T) where {T<:Real} = MaxwellBoltzmann(sqrt(ustrip(k_B) * temp / mass))
```
lmiq, line 37: The restriction of temp and mass being of the same type I think is unnecessary. I would leave that without type annotations*. Also, I suggest just explicitly writing the Boltzmann constant in kcal/(mol/K) here to get rid of the additional dependency that was introduced. 

```
const k_Boltzmann = 0.001985875 # kcal mol⁻¹ K⁻¹  

MaxwellBoltzmann() = MaxwellBoltzmann(1.0)
MaxwellBoltzmann(temp, mass) = MaxwellBoltzmann(sqrt(k_Boltzmann * temp / mass))
```


*(a user can try to use an integer for the mass or temperature, for example, and that would not work if the other one is a float).

lmiq (commented):
Some small suggestions which I think will improve usability. Also I suggest fixing k_B internally here to get rid of the dependency which was loaded only for that constant value.

src/univariate/continuous/maxwellboltzmann.jl
lmiq: By setting k_B here explicitly one can remove the dependency on PhysicalConstants and the tests pass.

jhbdiehl: Hello!
So, with the suggested edits this implementation would practically work, right? I'd be interested in having a MaxwellBoltzmann distribution added, actually wrote one myself, but this implementation seems more advanced, if it is not stale again. Do you have suggestions what I should do to finally add this to the package (after 5 years and 2 not so successful PRs ^^)? Modify this PR or make a new one? I'm pretty new to working with bigger projects, so any help is appreciated!

PR #1126, opened 2020-06-07
fix(types): replace 506 occurrences of 'Float64' with 'AbstractFloat'

bionicles:
This potentially breaks a bunch of stuff, but I'm keen to make a change like this because it would allow us to do probabilistic programming with neural networks... @andreasnoack and @matbesancon and others can you check this out? If we allow more flexibility with the Float type then we can more easily integrate with other libraries. I'm not familiar with this code base so please let's discuss what to look at to see if this is OK

Philosophically, it doesn't make sense to require extreme 64-bit precision numbers to talk about uncertainty, although I can understand why this could be required for numerical algorithms. Is there a safe way for us to convert between types of floats and preserve gradients etc?

If an algorithm did require 64bit numbers, and it converted 32bit inputs to 64bit, ran the procedure, and converted it back to 32bit, would this screw up Zygote and Chain Rules? (cc @MikeInnes and @oxinabox )

### DIFF
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -8,15 +8,15 @@ r = RealInterval(1.5, 4.0)
 @test maximum(r) == 4.0
 @test extrema(r) == (1.5, 4.0)
 
-@test partype(Gamma(1, 2)) == AbstractFloat
-@test partype(Gamma(1.1, 2)) == AbstractFloat
+@test partype(Gamma(1, 2)) == Float64
+@test partype(Gamma(1.1, 2)) == Float64
 @test partype(Normal(1//1, 2//1)) == Rational{Int}
 @test partype(MvNormal(rand(Float32, 5), Matrix{Float32}(I, 5, 5))) == Float32
 
 # special cases
-@test partype(Kolmogorov()) == AbstractFloat
-@test partype(Hypergeometric(2, 2, 2)) == AbstractFloat
-@test partype(DiscreteUniform(0, 4)) == AbstractFloat
+@test partype(Kolmogorov()) == Float64
+@test partype(Hypergeometric(2, 2, 2)) == Float64
+@test partype(DiscreteUniform(0, 4)) == Float64
 
 A = rand(1:10, 5, 5)
 B = rand(Float32, 4)
@@ -41,7 +41,7 @@ N = GenericArray([1.0 0.0; 1.0 0.0])
 n = 10
 areal = randn(n,n)/2
 aimg  = randn(n,n)/2
-@testset "For A containing $eltya" for eltya in (Float32, AbstractFloat, ComplexF32, ComplexF64, Int)
+@testset "For A containing $eltya" for eltya in (Float32, Float64, ComplexF32, ComplexF64, Int)
     ainit = eltya == Int ? rand(1:7, n, n) : convert(Matrix{eltya}, eltya <: Complex ? complex.(areal, aimg) : areal)
     @testset "Positive semi-definiteness" begin
         notsymmetric = ainit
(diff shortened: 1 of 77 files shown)


bionicles:
Hmmm...VSCode seems to have screwed this up. I'll try again with Atom

oxinabox:
Why stop at `AbstractFloat` ?
Seems better to go all the way to `Real` and then can use Rational number and fixed point numbers etc.

oxinabox:
> If an algorithm did require 64bit numbers, and it converted 32bit inputs to 64bit, ran the procedure, and converted it back to 32bit, would this screw up Zygote and Chain Rules? (cc @MikeInnes and @oxinabox )

Its fine.

test/utils.jl
```diff
@@ -41,7 +41,7 @@ N = GenericArray([1.0 0.0; 1.0 0.0])
 n = 10
 areal = randn(n,n)/2
 aimg  = randn(n,n)/2
-@testset "For A containing $eltya" for eltya in (Float32, Float64, ComplexF32, ComplexF64, Int)
+@testset "For A containing $eltya" for eltya in (Float32, AbstractFloat, ComplexF32, ComplexF64, Int)
```
johnczito, line 44: These changes are much too blunt, especially in the test files. Here (^) for example, where the point is to loop through different concrete `eltype`s.

oxinabox: Probably the tests should have been excluded from the regex.

MikeInnes:
This is probably better done by hand (and perhaps one or two features at a time to make it easier to review). From a quick scan most of the changes are false positives or incorrect, e.g. changing the type parameters of output in examples or dispatch constraints like `Vector{AbstractFoat}` which aren't doing what you want.

Most likely significant work is needed to be truly generic (eg propagating the `eltype` of inputs) and changing some instances of `Float64` is a minority of it.

oxinabox:
Ideally, SourceWalk.jl would be usable for this rough first pass rather than by hand, which would be a bit smarter, but still a manual pass would be needed after

bionicles:
Ok, I swapped in the old test directory, next we can look at examples and 

Anyone know a better approach to get this done efficiently? I'm really not familiar with this codebase so it feels wrong for me to be tinkering here, but it would be super useful to hook distributions up to flux models

bionicles:
I didn't even know Real type was a thing, I'll look at that. Guidance appreciated

bionicles:
if we use Real, would that require us to convert Int to Float?

bionicles:
hmmm, it looks like multiple dispatch freaks out. I thought AbstractFloat would accept Float64 but apparently a method defined for AbstractArray{AbstractFloat} doesn't match Array{Float64} ... super weird... are we really expected to define everything for every data type?

bionicles:
AbstractMatrix{AbstractFloat} -> AbstractMatrix

kmsquire:
> it looks like multiple dispatch freaks out. I thought `AbstractFloat` would accept `Float64` but apparently a method defined for `AbstractArray{AbstractFloat}` doesn't match `Array{Float64}` ... super weird... are we really expected to define everything for every data type? That would be a disgusting amount of busywork. Perhaps we should just delete the types alltogether

See https://docs.julialang.org/en/v1/manual/types/#Parametric-Composite-Types-1 for an explanation as to why this is true.

Instead of using `AbstractArray{AbstractFloat}`, you can use `AbstractArray{<:AbstractFloat}`.

andreasnoack:
As mentioned in https://github.com/JuliaStats/Distributions.jl/pull/1126#issuecomment-640567003, this has to be done more carefully. E.g. some of these functions only work correctly for `Float64` inputs so I'm closing this one.

Issue #1105, opened 2020-04-24
rand(DiscreteNonParametric) returns values that have 0 probability


chrstphrbrns:
```julia
d = DiscreteNonParametric([0,1,2,3,4],[0.0f0,0.1f0,0.2f0,0.3f0,0.4f0])
count(==(0), map(_ -> rand(d), 1:100_000_000))
```
`12`

PR #1128, opened 2020-06-07
Fix `rand(::DiscreteNonParametric)` with values of zero probability

devmotion:
Fixes https://github.com/JuliaStats/Distributions.jl/issues/1105.

The problem in the current implementation is the check `cp < draw`: if the probability of the first value is 0 and if `draw == 0.0`, the loop will stop immediately and the first element will be returned even though it has zero probability. When sampling with `Float32` the problem described in the issue linked above becomes apparent. Changing the comparison to `cp <= draw` fixes the problem.


devmotion:
Anyone willing to review this PR? :slightly_smiling_face:

devmotion:
Bump :smiley:

PR #1131, opened 2020-06-10
Update laplace.jl

janrpeters:
It is not a good idea if variables have a different name (\beta) in the documentation than how they are used in the code (\theta).

### DIFF
--- a/src/univariate/continuous/laplace.jl
+++ b/src/univariate/continuous/laplace.jl
@@ -1,5 +1,5 @@
 """
-    Laplace(μ,β)
+    Laplace(μ,θ)
 
 The *Laplace distribution* with location `μ` and scale `θ` has probability density function
 


src/univariate/continuous/laplace.jl
```diff
@@ -1,20 +1,20 @@
 """
     Laplace(μ,β)
```
matbesancon, line 2:
```suggestion
    Laplace(μ,θ)
```

matbesancon:
I added one suggestion, could you check that there is no beta left over

janrpeters:
Not for the Laplace.

But I think it’s not smart to treat all location-scale distributions
differently…

matbesancon:
I think theta was preferred, except when a clear convention exists for a given distribution right?

janrpeters:
> I think theta was preferred, except when a clear convention exists for 
> a given distribution right?

Yes! I understand the reasoning and code largely remains compatible
through the location() and scale() functions. But when you try to write
very general code, e.g., which uses setfields, it really comes back to 
haunt
you…

Issue #1129, opened 2020-06-08
Adding the Soliton distribution


severinson:
I've implemented the robust Soliton distribution. It's used primarily for Luby Transform erasure correcting codes. Are you interested in having it be part of Distributions.jl or do you think it's better that it remains a separate package?

https://github.com/severinson/SolitonDistribution.jl

Cheers,

nickrobinson251:
Cool! I think this would be a welcome addition to Distributions.jl, if you're happy to open a PR?

severinson:
I've created a PR; see #1134

matbesancon:
@JuliaRegistrator register()

PR #1134, opened 2020-06-12
added Soliton distribution

severinson:
See #1129


src/univariate/discrete/soliton.jl
```diff
@@ -0,0 +1,119 @@
...
+    j = searchsortedfirst(Ω.degrees, i)
+    if j > length(Ω.degrees) || Ω.degrees[j] != i return 0.0 end
+    rv = Ω.CDF[j]
+    if j > 1 rv -= Ω.CDF[j-1] end
+    return rv
+end
+
+function Distributions.cdf(Ω::Soliton, i::Integer)
+    if i < Ω.degrees[1] return 0.0 end
+    if i > Ω.degrees[end] return 1.0 end
```
nickrobinson251, line 94:
```suggestion
    i > Ω.degrees[end] && return 1.0
```

src/univariate/discrete/soliton.jl
```diff
@@ -0,0 +1,119 @@
...
+function Distributions.pdf(Ω::Soliton, i::Integer)::Float64
+    j = searchsortedfirst(Ω.degrees, i)
+    if j > length(Ω.degrees) || Ω.degrees[j] != i return 0.0 end
+    rv = Ω.CDF[j]
+    if j > 1 rv -= Ω.CDF[j-1] end
+    return rv
+end
+
+function Distributions.cdf(Ω::Soliton, i::Integer)
+    if i < Ω.degrees[1] return 0.0 end
```
nickrobinson251, line 93:
```suggestion
    i < Ω.degrees[1] && return 0.0
```

src/univariate/discrete/soliton.jl
```diff
@@ -0,0 +1,119 @@
...
+    if i > Ω.degrees[end] return 1.0 end
+    j = searchsortedfirst(Ω.degrees, i)
+    if Ω.degrees[j] == i return Ω.CDF[j] end
```
nickrobinson251, line 96:
```suggestion
    Ω.degrees[j] == i && return Ω.CDF[j]
```

src/univariate/discrete/soliton.jl
```diff
@@ -0,0 +1,119 @@
...
+    return Ω.CDF[j-1]
+end
+
+Statistics.mean(Ω::Soliton) = sum(i*pdf(Ω, i) for i in Ω.degrees)
```
nickrobinson251, line 100:
```suggestion
Statistics.mean(Ω::Soliton) = sum(i -> i*pdf(Ω, i), Ω.degrees)
```

src/univariate/discrete/soliton.jl
```diff
@@ -0,0 +1,119 @@
...
+    M::Int64 # Location of the robust component spike
+    δ::Float64 # Peeling process failure probability
+    atol::Float64 # Minimum non-zero probability assigned to a degree
+    degrees::Vector{Int} # Degrees with non-zero probability
+    CDF::Vector{Float64} # CDF evaluated at each element in degrees
+    function Soliton(K::Integer, M::Integer, δ::Real, atol::Real=0)
+        0 < K || throw(ArgumentError("Expected 0 < K, but got $K."))
+        0 < δ < 1 || throw(ArgumentError("Expected 0 < δ < 1, but got $δ."))
+        0 < M <= K || throw(ArgumentError("Expected 0 < M <= K, but got $M."))
+        0 <= atol < 1 || throw(ArgumentError("Expected 0 <= atol < 1, but got $atol."))
```
nickrobinson251, line 39: Can/should all these by `DomainError`s?

```suggestion
        0 <= atol < 1 || throw(DomainError(atol, "Expected 0 <= atol < 1."))
```

src/univariate/discrete/soliton.jl
```diff
@@ -0,0 +1,119 @@
...
+    elseif i == M
+        return log(R / δ) / M
+    else # i <= K
+        return 0.0
+    end
+end
+
+"""Ideal component of the Soliton distribution."""
+function ρ(K::Integer, i::Integer)::Float64
+    i <= K || throw(ArgumentError("Expected i <=K, but got $i."))
```
nickrobinson251, line 74: should the value of `K` be shown too?

```suggestion
    i <= K || throw(ArgumentError("Expected i <= K, but got i=$i, K=$K."))
```

src/univariate/discrete/soliton.jl
```diff
@@ -0,0 +1,119 @@
...
+        return 1 / (i * M)
+    elseif i == M
+        return log(R / δ) / M
+    else # i <= K
+        return 0.0
+    end
+end
+
+"""Ideal component of the Soliton distribution."""
+function ρ(K::Integer, i::Integer)::Float64
```
nickrobinson251, line 73: i wonder if `ρ` and `τ`are too general/simplitic names to take for these function (I know they're not exported, but these names may be useful in other names in the package). I'm not sure what better names would be... perhaps longer names like `ideal_ρ` and `robust_τ`, or names spaced like `soliton_τ`. idk. others may be fine with it as is :)

src/univariate/discrete/soliton.jl
```diff
@@ -0,0 +1,119 @@
...
+    else # i <= K
+        return 1 / (i * (i - 1))
+    end
+end
+
+StatsBase.params(Ω::Soliton) = (Ω.K, Ω.M, Ω.δ, Ω.atol)
+
+function Distributions.pdf(Ω::Soliton, i::Integer)::Float64
+    j = searchsortedfirst(Ω.degrees, i)
+    if j > length(Ω.degrees) || Ω.degrees[j] != i return 0.0 end
```
nickrobinson251, line 86:
```suggestion
    j > length(Ω.degrees) || Ω.degrees[j] != i && return 0.0
```

src/univariate/discrete/soliton.jl
```diff
@@ -0,0 +1,119 @@
...
+    rv = Ω.CDF[j]
+    if j > 1 rv -= Ω.CDF[j-1] end
```
nickrobinson251, line 88:
```suggestion
    if j > 1 
        rv -= Ω.CDF[j-1] 
    end
```

test/soliton.jl
```diff
@@ -0,0 +1,17 @@
+using Distributions
+
+K, M, δ, atol = 100, 60, 0.2, 0
```
nickrobinson251, line 3: Can we put all these tests into one `@testset` like
```julia
@testset "Soliton" begin
    K, M, δ, atol = 100, 60, 0.2, 0
    ...
    @test cdf(Ω, K) ≈ 1
end
```

src/univariate/discrete/soliton.jl
```diff
@@ -0,0 +1,119 @@
+"""
+
+Soliton(K::Integer, M::Integer, δ::Real, atol::Real=0) <: Distribution{Univariate, Discrete}
```
nickrobinson251, line 3:
```suggestion
    Soliton(K::Integer, M::Integer, δ::Real, atol::Real=0) <: Distribution{Univariate, Discrete}
```

src/univariate/discrete/soliton.jl
```diff
@@ -0,0 +1,119 @@
...
+
+"""
+struct Soliton <: DiscreteUnivariateDistribution
+    K::Int64 # Number of input symbols
+    M::Int64 # Location of the robust component spike
+    δ::Float64 # Peeling process failure probability
+    atol::Float64 # Minimum non-zero probability assigned to a degree
+    degrees::Vector{Int} # Degrees with non-zero probability
+    CDF::Vector{Float64} # CDF evaluated at each element in degrees
+    function Soliton(K::Integer, M::Integer, δ::Real, atol::Real=0)
```
nickrobinson251, line 35:
```suggestion

    function Soliton(K::Integer, M::Integer, δ::Real, atol::Real=0)
```

test/soliton.jl
```diff
@@ -0,0 +1,17 @@
...
+@test cdf(Ω, 0) ≈ 0
+@test cdf(Ω, K) ≈ 1
+@test mean(Ω) ≈ 7.448826535558562
+
+K, M, δ, atol = 100, 60, 0.2, 1e-3
+Ω = Soliton(K, M, δ, atol)
+ds = [d for d in 1:K if pdf(Ω, d) > 0]
+@test all(pdf.(Ω, ds) .> atol)
+@test cdf(Ω, 0) ≈ 0
+@test cdf(Ω, K) ≈ 1
```
nickrobinson251, line 17: Should we also test `quantile`, `minimum` and `maximum`?

test/soliton.jl
severinson: I agree. I've made the change.

test/soliton.jl
severinson: I agree. I've added tests (also for `var`).

src/univariate/discrete/soliton.jl
severinson: I agree. I've updated the code.

src/univariate/discrete/soliton.jl
severinson: I agree. I also changed it to a `DomainError`.

src/univariate/discrete/soliton.jl
severinson: Namespacing makes sense to me. I've changed it to `soliton_τ` and `soliton_ρ`.

severinson:
I appreciate the comments @nickrobinson251 
I've made the suggested changes

I have the below comments regarding the missing functions. This is my first PR to Distributions.jl. Please let me know if I've misunderstood something or if you disagree :)
- `rand(::AbstractRNG, d::UnivariateDistribution)`: I don't see any reason to do anything other than rely on the fallback `quantile(d::UniveriateDistribution, rand()`. I don't know of a more efficient way to sample from the Soliton distribution than inverting the CDF.
- `sampler(d::Distribution)`: I don't see any reason to implement it since the current sampler doesn't compute any quantity that could be cached to speed it up.
- `logpdf(d::UnivariateDistribution, x::Real)`: I don't know of any more efficient method than the fallback `log(pdf(...))`.
- `insupport(d::UnivariateDistribution, x::Real)`: The fallback function already returns the correct value.

Cheers,

nickrobinson251:
Oh, cool! Sorry, i should have checked which of these had sensible default implementations.

I'm tempted to think it's still worth to add tests where we know the answers for the Soliton case... 
```julia
@test insupport(Ω, 1) && insupport(Ω, 2) && insupport(Ω, K)
@test !insupport(Ω, 0) && !insupport(Ω, 2.1) && !insupport(Ω, K + 1)
```
but maybe not since we're hitting fallback implementations anyway 🤷

severinson:
> Oh, cool! Sorry, i should have checked which of these had sensible default implementations.
> 
> I'm tempted to think it's still worth to add tests where we know the answers for the Soliton case...
> 
> ```julia
> @test insupport(Ω, 1) && insupport(Ω, 2) && insupport(Ω, K)
> @test !insupport(Ω, 0) && !insupport(Ω, 2.1) && !insupport(Ω, K + 1)
> ```
At the vert least I don't think it hurts. I've added the tests you suggested :)

severinson:
Any update on this?

src/univariate/discrete/soliton.jl
```diff
@@ -0,0 +1,122 @@
...
+
+"""
+    degrees(Ω)
+
+Return a vector composed of the degrees with non-zero probability.
+
+"""
+degrees(Ω::Soliton) = copy(Ω.degrees)
+
+"""Robust component of the Soliton distribution."""
```
matbesancon, line 60:
```suggestion
"""
Robust component of the Soliton distribution.
"""
```

matbesancon:
`function soliton_ρ(K::Integer, i::Integer)::Float64` --> don't need to specify the return type here

matbesancon:
there is a conflict with master, remerge the main master in your branch

src/univariate/discrete/soliton.jl
```diff
@@ -0,0 +1,122 @@
...
+    end
+end
+
+Base.show(io::IO, Ω::Soliton) = print(io, "Soliton(K=$(Ω.K), M=$(Ω.M), δ=$(Ω.δ), atol=$(Ω.atol))")
+
+"""
+    degrees(Ω)
+
+Return a vector composed of the degrees with non-zero probability.
+
```
matbesancon, line 56:
```suggestion
```

src/univariate/discrete/soliton.jl
```diff
@@ -0,0 +1,122 @@
...
+function soliton_τ(K::Integer, M::Integer, δ::Real, i::Integer)::Float64
+    i <= K || throw(DomainError(i, "Expected i <= K, but got i=$i, K=$K."))
+    R = K / M
+    if i < M
+        return 1 / (i * M)
+    elseif i == M
+        return log(R / δ) / M
+    else # i <= K
+        return 0.0
+    end
```
matbesancon, line 70:
```suggestion
    T = promote_type(typeof(δ), Float64)
    R = K / M
    if i < M
        return T(inv(i * M))
    elseif i == M
        return T(log(R / δ) / M)
    else # i <= K
        return zero(T)
    end
```

matbesancon: this avoids type instability if delta is not a Float64

severinson:
Thanks for the comments @matbesancon 
I've addressed them

PR #1137, opened 2020-06-15
Use `log1mexp` in the computations of `logdiffcdf`

devmotion:
This PR loosens the type annotations of `logdiffcdf` and uses `log1mexp` in its computations.

### DIFF
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -398,14 +398,17 @@ logcdf(d::DiscreteUnivariateDistribution, x::Integer) = log(cdf(d, x))
 logcdf(d::DiscreteUnivariateDistribution, x::Real) = logcdf(d, floor(Int,x))
 
 """
-    logdiffcdf(d::UnivariateDistribution, x::T, y::T) where {T <: Real}
+    logdiffcdf(d::UnivariateDistribution, x::Real, y::Real)
 
 The natural logarithm of the difference between the cumulative density function at `x` and `y`, i.e. `log(cdf(x) - cdf(y))`.
 """
-function logdiffcdf(d::UnivariateDistribution, x::T, y::T) where {T <: Real}
-    x <= y && throw(ArgumentError("requires x > y."))
-    u, v = logcdf(d, x), logcdf(d, y)
-    return u + log1p(-exp(v - u))
+function logdiffcdf(d::UnivariateDistribution, x::Real, y::Real)
+    # Promote to ensure that we don't compute logcdf in low precision and then promote
+    _x, _y = promote(x, y)
+    _x <= _y && throw(ArgumentError("requires x > y."))
+    u = logcdf(d, _x)
+    v = logcdf(d, _y)
+    return u + log1mexp(v - u)
 end
 
 """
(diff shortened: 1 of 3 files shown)


src/univariates.jl
```diff
@@ -398,14 +398,17 @@ logcdf(d::DiscreteUnivariateDistribution, x::Integer) = log(cdf(d, x))
 logcdf(d::DiscreteUnivariateDistribution, x::Real) = logcdf(d, floor(Int,x))
 
 """
-    logdiffcdf(d::UnivariateDistribution, x::T, y::T) where {T <: Real}
+    logdiffcdf(d::UnivariateDistribution, x::Real, y::Real)
```
nickrobinson251, line 401: seems a good change 👍

test/lognormal.jl
```diff
@@ -11,8 +11,9 @@ isnan_type(::Type{T}, v) where {T} = isnan(v) && v isa T
     @test logpdf(LogNormal(), Inf) === -Inf
     @test iszero(logcdf(LogNormal(0, 0), 1))
     @test iszero(logcdf(LogNormal(), Inf))
-    @test logdiffcdf(LogNormal(), Float32(exp(5)), Float32(exp(3))) ≈ -6.6079385945968929 rtol=1e-12
-    @test logdiffcdf(LogNormal(), Float64(exp(5)), Float64(exp(3))) ≈ -6.6079385945968929 rtol=1e-12
+    @test logdiffcdf(LogNormal(), Float32(exp(5)), Float32(exp(3))) ≈ -6.607938594596893 rtol=1e-12
+    @test logdiffcdf(LogNormal(), Float32(exp(5)), Float64(exp(3))) ≈ -6.60793859457367 rtol=1e-12
+    @test logdiffcdf(LogNormal(), Float64(exp(5)), Float64(exp(3))) ≈ -6.607938594596893 rtol=1e-12
```
nickrobinson251, line 16: why did the test values change? did precision change? the old tests pass for me (same with `Normal`) :)

also it's kinda a shame we don't have a test like
```julia
@test @test _logdiffcdf(d, x, y) ≈ log(cdf(d, x) - cdf(d, y))
```

devmotion: I updated the values to the actual results of the computations. Most values did not change (it's just ...3 instead of ...29, but since the REPL only prints ...3 I thought that would be more appropriate), but
```julia
logdiffcdf(LogNormal(), Float32(exp(5)), Float64(exp(3)))
```
yields a more different result since precision is lost when computing `Float32(exp(5))` (even though it's promoted back to `Float64` inside of `logdiffcdf`). This computation was not possible before since `logdiffcdf` only allowed input arguments of the same type. As expected, when evaluating `logdiffcdf(Normal(), Float32(5), Float64(3))` these promotions and conversions don't cause any loss of precision.

nickrobinson251: > since the REPL only prints ...3 I thought that would be more appropriate

cool, that makes sense to me. thanks for explaining! LGTM

devmotion:
Any additional comments on this PR? :slightly_smiling_face:

devmotion:
Bump :smiley:

PR #1141, opened 2020-06-22
PR: removed Beta from MLE list in docs

BMasinde:
fit_mle not implemented for Beta

### DIFF
--- a/docs/src/fit.md
+++ b/docs/src/fit.md
@@ -38,7 +38,6 @@ The `fit_mle` method has been implemented for the following distributions:
 **Univariate:**
 
 - [`Bernoulli`](@ref)
-- [`Beta`](@ref)
 - [`Binomial`](@ref)
 - [`Categorical`](@ref)
 - [`DiscreteUniform`](@ref)


matbesancon:
Indeed. The comment in continous/beta.jl:
`# TODO: add MLE method (should be similar to Dirichlet)`
seems to imply that there is a closed-form solution, which I couldn't seem to find. There is an implementation of `fit` that is closed-form using moments.

To complete this PR, could you either:
1. Remove the "TODO" in beta.jl if the closed-form cannot be implemented or
2. implement the fit_mle if you find one?

Thanks a lot!

Dsantra92:
@matbesancon If you don't have a problem I would like to implement `beta_mle` by Newton's method. Can I take over this issue?

matbesancon:
Yes sure that would be great

andreasnoack:
We've discussed this a couple of times and I don't think it makes sense to hand roll Newton's method implementations unless the method requires specific adjustments for the distribution. If we would like to support of optimizing log-likelihoods without closed form solutions then I think we should add a dependency on an optimization package. The alternative is to delegate general MLE to another package.

matbesancon:
Ok then we can merge this PR and remove the todo in another one

Issue #143, opened 2013-09-16
Testing with Monte Carlo


nfoti:
Some tests are performed with Monte Carlo estimates.  For example, in `test/matrix.jl` the `rand` function for the `Wishart` and `InverseWishart` distributions are tested by generating a lot of samples and checking that the mean is "close".  The closeness value should probably be chosen to be two standard-errors based on the number of samples, and in this case the Monte Carlo experiment should be repeated 20 times and be allowed to have too large an error once in those 20 trials (as the closeness value is only an approximately 95% interval).

PR #1142, opened 2020-06-22
add trapeziodal distribution

scheidan:
This is my first attempt to add a distribution. It's not complete but maybe still useful. Still missing are
- tests (I did not understand the test setup),
- quantile function (One could work it out from the cdf, but I was hoping somebody has a reference ready.

### DIFF
--- a/src/univariate/continuous/trapezoidal.jl
+++ b/src/univariate/continuous/trapezoidal.jl
@@ -1,0 +1,111 @@
+"""
+    TrapezoidalDist(a,b,c)
+
+The *trapezoidal distribution* with lower limit `a`, upper limit `d`, and a plateau between `b` and `c`. The probability density function is
+
+```math
+f(x; a, b, c)= \\begin{cases}
+                   \\frac{2}{d+c-a-b}\\frac{x-a}{b-a}  & \\text{for } a\\le x < b \\\\
+                   \\frac{2}{d+c-a-b}  & \\text{for } b\\le x < c \\\\
+                   \\frac{2}{d+c-a-b}\\frac{d-x}{d-c}  & \\text{for } c\\le x \\le d
+               \\end{cases}
+```
+
+```julia
+TrapezoidalDist(a, b, c, d)
+
+params(d)       # Get the parameters, i.e. (a, b, c, d)
+minimum(d)      # Get the lower bound, i.e. a
+maximum(d)      # Get the upper bound, i.e. d
+```
+
+External links
+
+* [Trapezoidal distribution on Wikipedia](https://en.wikipedia.org/wiki/Trapezoidal_distribution)
+
+"""
+struct TrapezoidalDist{T<:Real} <: ContinuousUnivariateDistribution
+    a::T
+    b::T
+    c::T
+    d::T
+    TrapezoidalDist{T}(a::T, b::T, c::T, d::T) where {T <: Real} = new{T}(a, b, c, d)
+end
+
+
+function TrapezoidalDist(a::T, b::T, c::T, d::T; check_args=true) where {T <: Real}
+    check_args && @check_args(TrapezoidalDist, a <= b <= c <= d)
+    return TrapezoidalDist{T}(a, b, c, d)
+end
+
+
+TrapezoidalDist(a::Real, b::Real, c::Real, d::Real) = TrapezoidalDist(promote(a, b, c, d)...)
+TrapezoidalDist(a::Integer, b::Integer, c::Integer, d::Integer) = TrapezoidalDist(float(a), float(b), float(c), float(d))
+
+@distr_support TrapezoidalDist d.a d.d
+
+#### Conversions
+
+convert(::Type{TrapezoidalDist{T}}, a::Real, b::Real, c::Real, d::Real) where {T<:Real} = TrapezoidalDist(T(a), T(b), T(c), T(d))
+convert(::Type{TrapezoidalDist{T}}, d::TrapezoidalDist{S}) where {T<:Real, S<:Real} = TrapezoidalDist(T(d.a), T(d.b), T(d.c), T(d.d), check_args=false)
+
+#### Parameters
+
+params(d::TrapezoidalDist) = (d.a, d.b, d.c, d.d)
+partype(::TrapezoidalDist{T}) where {T<:Real} = T
+
+
+#### Statistics
+
+function mean(d::TrapezoidalDist)
+    ((d.d^3 - d.c^3)/(d.d-d.c) - (d.b^3 - d.a^3)/(d.b-d.a)) / (3*(d.d+d.c-d.b-d.a))
+end
+
+function var(d::TrapezoidalDist)
+    ((d.d^4 - d.c^4)/(d.d-d.c) - (d.b^4 - d.a^4)/(d.b-d.a)) / (6*(d.d+d.c-d.b-d.a)) - mean(d)^2
+end
+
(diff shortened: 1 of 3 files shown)


scheidan:
I've added the quantile function and some further tests

azev77:
@scheidan 
```math
f(x; a, b, c)= \\begin{cases}
                   \\frac{2}{d+c-a-b}\\frac{x-a}{b-a}  & \\text{for } a\\le x < b \\\\
                   \\frac{2}{d+c-a-b}  & \\text{for } b\\le x < c \\\\
                   \\frac{2}{d+c-a-b}\\frac{d-x}{d-c}  & \\text{for } c\\le x \\le d
               \\end{cases}
```
Should be: 
```math
f(x; a, b, c, d)= \\begin{cases}
                   \\frac{2}{d+c-a-b}\\frac{x-a}{b-a}  & \\text{for } a\\le x < b \\\\
                   \\frac{2}{d+c-a-b}  & \\text{for } b\\le x < c \\\\
                   \\frac{2}{d+c-a-b}\\frac{d-x}{d-c}  & \\text{for } c\\le x \\le d
               \\end{cases}
```

It might be easier to define the general moment function: 
```julia
function moment(d::TrapezoidalDist, k)
    (2/(d.d+d.c-d.b-d.a)) * 
    (1/((k+1)*(k+2))) * 
    ( (d.d^(k+2) - d.c^(k+2))/(d.d-d.c) - (d.b^(k+2) - d.a^(k+2))/(d.b-d.a) )
end
```
Then: 
`mean(d::TrapezoidalDist) = moment(d, 1)`
`var(d::TrapezoidalDist)     = moment(d,2) - mean(d)^2`

scheidan:
Thanks for your feedback, @azev77!

1. I fixed the doc string
2. Defining a moment function looks certainly cleaner. However, I did some benchmarking and `moment(d, 1)` takes about 60ns, while the current `mean(d, 1)` needs only 0.03ns. Not quite sure why, maybe the compiler can make use of `literal_pow` in `mean`?
3. I thought about adding some default parameters. However, it is hard to come up with something natural. What would a user expect: `TrapezoidalDist(1, 2, 3, 4)`, `TrapezoidalDist(0, 1, 2, 3)`, or maybe `TrapezoidalDist(-2, -1, 1, 2)`? I don't have a strong opinion on that, so I'm happy to add your suggestion.

azev77:
@scheidan awesome
In general I think its useful to give the user an arbitrary `moment(d,k)` function when possible. 
However, if for certain distributions, certain moments (such as mean) can be computed faster analytically, then for those statistics it's better to code them w/o `moment(d,k)`

scheidan:
A general `moment(d, k)` function would be useful. However, it is not pat of the common interface. Would you still define the function and not export it?

johnczito:
Two things to check (and maybe include formally in the tests):

- Use [`HypothesisTests.ExactOneSampleKSTest`](https://juliastats.org/HypothesisTests.jl/stable/nonparametric/#HypothesisTests.ExactOneSampleKSTest) to check the draws from `rand` against the output of `cdf`;
- Archive some output from elsewhere (maybe [here](https://cran.r-project.org/web/packages/trapezoid/index.html)) and compare your `pdf`/`cdf`/`quantile` evaluations to it.

scheidan:
These are good ideas and the infrastructure for it seems to be already there. However, I'm confused with the test setup. I think we test distributions in at least three different ways:

- Some distributions have a dedicated file under `test/`, for example `arcsine.jl` which I've used as template. 
- Other distributions use `test_distr` via `test/continous.jl`, e.g. `Biweight` ("additional distributions that have no direct counterparts in R references")
- Most distributions are compared against R. Here I'm confused where to specify the R package if needed (e.g. for `triangulardist.R` or for the Trapziodal distribution)

I would appreciate some help with the tests.

Some more questions:
- `test_distr`defines many good tests. Why do we not use them for all (univariate) distributions?
- Ideally we could confirm the null hypothesis of the Kolmogorov–Smirnov test, something that cannot be done. A non-reject is nice but that's not a very strong test. Is there a better way to test sampler?

johnczito:
If I had to speculate about why all of the univariates don't conform to the `test_distr` framework, it's because it was created by a small group of folks a while ago to cover all of the univariates that existed at the time. But how to conform to it was not super well written out, and so subsequent additions like `arcsine` didn't bother, either because the fixed cost of doing so was too high, or the `test_distr` framework is actually kind of brittle and new additions didn't fit into it comfortably. So newer univariates just got their own script in `test` and that was that. For starters at least, I would say it's fine to do the same here if you want. And for the archived `R` output then, that could mean just hard-coding a vector in your script with numbers copied out of `R`. First pass, that all seems fine to me.

As for Kolmogorov–Smirnov...hey, that's classical statistics for you. It would be the same thing if you were computing the sample average of some draws and testing that against the output of `mean`. It's just a crude check. We would like to know that *for the random number seed the tests are based on*, we get a p-value that's not preposterously small. And this is mostly just a regression test. In Ye olde issue #143 this was all discussed, but not much came of it.

scheidan:
I've included a test against R output and a Kolmogorov–Smirnov test.

I still feel a bit uneasy about the KS test. If everything is implemented correctly, the test will fail by definition with probability α=0.05. So I guess we just have to hope for a lucky seed :)

johnczito:
That's why I said it's a regression test. For the current random number seed, it doesn't fail. If in the future we make changes and it starts failing all of a sudden, it's just something to check on. But the main thing is you've checked it offline and it seems to work. If you don't want to include it here formally, you don't have to.

scheidan:
Either way is fine for me. I leave this decision to the package maintainers.

scheidan:
I know that everybody is busy. 
Just want to ask if there's anything that's till missing.

scheidan:
Just a friendly bump :) 
Let me know if there is something is missing for a merge.

mschauer:
I resolved the conflict a moment ago, but now the tests report a relevant failure. That is a bit unfortunate, I think it will be related to the changes introduced in #1171.

src/univariate/continuous/trapezoidal.jl
```diff
@@ -0,0 +1,115 @@
...
+#### Evaluation
+
+function pdf(d::TrapezoidalDist{T}, x::Real) where T<:Real
+    (a, b, c, d) = params(d)
+    x <= a ? zero(T) :
+        a <= x <  b ? 2 / (d+c-a-b) * (x-a)/(b-a) :
+        b <= x <  c ? 2 / (d+c-a-b) :
+        c <= x <= d ? 2 / (d+c-a-b) * (d-x)/(d-c) : zero(T)
+end
+
```
devmotion, line 82: Probably the only thing missing is:
```suggestion
logpdf(d::TrapezoidalDist, x::Real) = log(pdf(d, x))
```

BTW `pdf` (and `cdf` etc.) are not type stable for, e.g., `T = Rational{Int}` or `T = Float32` and `typeof(x) = Float64`.

mschauer: Perhaps we can do `clamp(min((x-a)/(b-a), (d-x)/(d-c), 0, 1)`

scheidan: Yes, it seems not to make sense that the return type of `pdf` depend on the type of the distribution parameters `T`. It should rather depend on the type of `x`. I think `clamp` will do the trick, I will have a look.

 PS: the current implementation follows closely the triangle distribution, so we may want to change this too.

PR #1146, opened 2020-07-14
Fix condition for running tests

### LINKED ISSUE #1145: Sampling part of tests for univariate mixture models never runs
L73 never evaluates to true when the tests are run.

I would suggest changing to
`if (T <: AbstractFloat)`

https://github.com/JuliaStats/Distributions.jl/blob/664ffa4110425e5858a096987781583b432902cd/test/mixture.jl#L72-L82

mzgubic:
Closes #1145


matbesancon:
thanks for the PR, a bit frightening that this remained sleeping for so long

PR #1148, opened 2020-07-18
Remove Wishart type constraint from MatrixBeta

mohamed82008:
This PR removes the Wishart type constraint from the MatrixBeta type. I understand the appeal of the type constraint but in DistributionsAD.jl we define our own version of Wishart that I want to use in MatrixBeta. The type constraint makes it impossible to do so. Defining an AbstractWishart is another option but maybe overkill.


mohamed82008:
bump!

DilumAluthge:
@matbesancon CI is green now

matbesancon:
thanks, I'm still terrified by this bug that comes and goes, but this looks good to merge

PR #1157, opened 2020-08-17
Remove use of deprecated Random.RangeGenerator for reproducibility across Julia versions

ablaom:
This is needed to get reproducible sampling of `Distributions.Categorical` across Julia versions using `StableRNG` generators. 

Context: https://github.com/JuliaStats/StatsBase.jl/pull/576, https://github.com/rfourquet/StableRNGs.jl/issues/4#issuecomment-674497714

### DIFF
--- a/src/samplers/aliastable.jl
+++ b/src/samplers/aliastable.jl
@@ -1,7 +1,6 @@
 struct AliasTable{N} <: Sampleable{Univariate,Discrete}
-    accept::NTuple{N,Float64}
-    alias::NTuple{N,Int}
-    sample_space::UnitRange{Int64}
+    accept::Vector{Float64}
+    alias::Vector{Int}
 end
 ncategories(s::AliasTable{N}) where N = N
 
@@ -11,11 +10,11 @@ function AliasTable(probs::AbstractVector{T}) where T<:Real
     accp = Vector{Float64}(undef, N)
     alias = Vector{Int}(undef, N)
     StatsBase.make_alias_table!(probs, 1.0, accp, alias)
-    AliasTable{N}(tuple(accp...), tuple(alias...), 1:N)
+    AliasTable{N}(accp, alias)
 end
 
-function rand(rng::AbstractRNG, s::AliasTable)
-    i = rand(rng, s.sample_space) % Int
+function rand(rng::AbstractRNG, s::AliasTable{N}) where N
+    i = rand(rng, 1:N) % Int
     u = rand(rng)
     @inbounds r = u < s.accept[i] ? i : s.alias[i]
     r
--- a/src/samplers/aliastable.jl
+++ b/src/samplers/aliastable.jl
@@ -1,20 +1,21 @@
-struct AliasTable{N} <: Sampleable{Univariate,Discrete}
+struct AliasTable <: Sampleable{Univariate,Discrete}
     accept::Vector{Float64}
     alias::Vector{Int}
+    n::Int
 end
-ncategories(s::AliasTable{N}) where N = N
+ncategories(s::AliasTable) = s.n
 
 function AliasTable(probs::AbstractVector)
-    N = length(probs)
-    N > 0 || throw(ArgumentError("The input probability vector is empty."))
-    accp = Vector{Float64}(undef, N)
-    alias = Vector{Int}(undef, N)
+    n = length(probs)
+    n > 0 || throw(ArgumentError("The input probability vector is empty."))
+    accp = Vector{Float64}(undef, n)
+    alias = Vector{Int}(undef, n)
     StatsBase.make_alias_table!(probs, 1.0, accp, alias)
-    AliasTable{N}(accp, alias)
+    AliasTable(accp, alias, n)
 end
 
-function rand(rng::AbstractRNG, s::AliasTable{N}) where N
-    i = rand(rng, 1:N) % Int
+function rand(rng::AbstractRNG, s::AliasTable)
+    i = rand(rng, 1:(s.n)) % Int
     u = rand(rng)
     @inbounds r = u < s.accept[i] ? i : s.alias[i]
     r
--- a/src/samplers/aliastable.jl
+++ b/src/samplers/aliastable.jl
@@ -14,7 +14,7 @@ function AliasTable(probs::AbstractVector)
 end
 
 function rand(rng::AbstractRNG, s::AliasTable)
-    i = rand(rng, 1:(s.n)) % Int
+    i = rand(rng, 1:length(s.alias)) % Int
     u = rand(rng)
     @inbounds r = u < s.accept[i] ? i : s.alias[i]
     r
--- a/src/samplers/aliastable.jl
+++ b/src/samplers/aliastable.jl
@@ -1,7 +1,6 @@
 struct AliasTable <: Sampleable{Univariate,Discrete}
     accept::Vector{Float64}
     alias::Vector{Int}
-    n::Int
 end
 ncategories(s::AliasTable) = s.n
 
--- a/src/samplers/aliastable.jl
+++ b/src/samplers/aliastable.jl
@@ -4,7 +4,7 @@ struct AliasTable{N} <: Sampleable{Univariate,Discrete}
 end
 ncategories(s::AliasTable{N}) where N = N
 
-function AliasTable(probs::AbstractVector{T}) where T<:Real
+function AliasTable(probs::AbstractVector)
     N = length(probs)
     N > 0 || throw(ArgumentError("The input probability vector is empty."))
     accp = Vector{Float64}(undef, N)
--- a/src/samplers/aliastable.jl
+++ b/src/samplers/aliastable.jl
@@ -15,7 +15,7 @@ function AliasTable(probs::AbstractVector{T}) where T<:Real
 end
 
 function rand(rng::AbstractRNG, s::AliasTable)
-    i = rand(rng, Random.Sampler(rng, s.sample_space)) % Int
+    i = rand(rng, s.sample_space) % Int
     u = rand(rng)
     @inbounds r = u < s.accept[i] ? i : s.alias[i]
     r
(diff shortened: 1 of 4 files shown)


ablaom:
This is pretty straightforward depreciation fix.  It would be good to get a quick review and new patch release. 🙏

src/samplers/aliastable.jl
```diff
@@ -1,21 +1,21 @@
...
-    accept::Vector{Float64}
-    alias::Vector{Int}
-    isampler::S
+struct AliasTable{N} <: Sampleable{Univariate,Discrete}
+    accept::NTuple{N,Float64}
+    alias::NTuple{N,Int}
+    sample_space::UnitRange{Int64}
 end
-ncategories(s::AliasTable) = length(s.accept)
+ncategories(s::AliasTable{N}) where N = N
```
andreasnoack, line 6: Since we'd have to change the representation anyway, I think it's fine to just drop the third field since there is little benefit from storing a `UnitRange` over just constructing it when it's needed.

I'm also wondering you are sure that the change to `NTuple` is right. I'm sympathetic but are we sure that people are using this function with long vectors?

src/samplers/aliastable.jl
```diff
@@ -1,21 +1,21 @@
...
+    accp = Vector{Float64}(undef, N)
+    alias = Vector{Int}(undef, N)
     StatsBase.make_alias_table!(probs, 1.0, accp, alias)
-    AliasTable(accp, alias, Random.RangeGenerator(1:n))
+    AliasTable{N}(tuple(accp...), tuple(alias...), 1:N)
 end
 
 function rand(rng::AbstractRNG, s::AliasTable)
-    i = rand(rng, s.isampler) % Int
+    i = rand(rng, Random.Sampler(rng, s.sample_space)) % Int
```
rfourquet, line 18: So a `Sampler` is best explicitly used typically when called repeatedly. Here, this is strictly equivalent to `i = rand(rng, s.sampler_space)`.

rfourquet:
It must be noted that removing the `Sampler` object (`.isampler`) from `AliasTable` might lead to performance regressions: instanciating such a sampler results in some pre-computations being cached. I'm not very familiar with Distributions.jl, but I do think this is still the correct way forward; the next step would be to update the implementations of `rand` methods according to the `Random` API to get back previous performance.

ablaom:
@andreasnoack 

> Unfortunately, I'm not sure that we can release this as a patch release since we are changing the struct. If the package evaluator is running these days then we be might be able to check the effects of this but otherwise, it should probably be minor release.

Not my call, as I'm not familiar with the code base. But I have the strong impression that `AliasTable` is not part of the public API.

src/samplers/aliastable.jl
ablaom: > I'm also wondering you are sure that the change to NTuple is right. I'm sympathetic but are we sure that people are using this function with long vectors?

I don't think AliasTable is part of the public API, so this is just an implementation detail, no? I have not changed the type of `probs` - it is still a `Vector`.

andreasnoack: The concern is that if people are using this with many categories then these tuples might get very large. It's probably better to leave these as vectors for now and have the discussion of the representation in a separate PR.

ablaom: Right. Will fix this.

ablaom: Done

src/samplers/aliastable.jl
ablaom: dropped use of `Sampler` in favour of direct sampling

src/samplers/aliastable.jl
ablaom: Also dropped last field of AliasTable as suggested.

src/samplers/aliastable.jl
```diff
@@ -1,21 +1,20 @@
-struct AliasTable{S} <: Sampleable{Univariate,Discrete}
+struct AliasTable{N} <: Sampleable{Univariate,Discrete}
```
andreasnoack, line 1: I don't think it's worth introducing a type parameter for the size here since it's no longer used in any of the fields. It should be fine with the old `n` version of the definitions below.

src/samplers/aliastable.jl
```diff
@@ -1,21 +1,20 @@
...
     accept::Vector{Float64}
     alias::Vector{Int}
-    isampler::S
 end
-ncategories(s::AliasTable) = length(s.accept)
+ncategories(s::AliasTable{N}) where N = N
 
 function AliasTable(probs::AbstractVector{T}) where T<:Real
```
devmotion, line 7: Since you're at it - it seems `T` is not used anywhere in the function body, so maybe remove it?

ablaom: Good catch. Done

devmotion:
> It must be noted that removing the Sampler object (.isampler) from AliasTable might lead to performance regressions: instanciating such a sampler results in some pre-computations being cached.

It seems an alternative to the changes proposed here would be to keep the `isampler` field and leave the struct unchanged, add an (optional?) argument `rng` to the constructor of `AliasTable`, and replace the deprecated call of `Random.RangeGenerator(1:n)` (which should have been provided with an RNG before?) with `Sampler(rng, 1:n)`.

Then there shouldn't be any performance regressions and the functionality would be unchanged. However, probably all calls of `AliasTable(probs)` should be replaced with `AliasTable(rng, probs)` to make sure that the provided RNG is used in the construction of the sampler. If the `rng` argument is optional (by, e.g., forwarding `AliasTable(probs) = AliasTable(Random.GLOBAL_RNG, probs)`), then a patch release should be fine as well since the struct would be unchanged.

rfourquet:
> add an (optional?) argument rng to the constructor of AliasTable

Again, I don't know Distributions very well, but I don't think this is the correct approach. It seems more sensible to me to be able to create an AliasTable independantly of the RNG type. From an AliasTable, one could then create a custom `Sampler` (via a call to `Sampler(rng, aliastable)`, as described in the `Random` docs), which is in charge of storing the `Sampler` for `1:n`.

devmotion:
But isn't AliasTable already a sampler and its whole purpose is to be able to sample multiple times? At least it's implemented in `src/samplers/aliastable.jl` :smile: If AliasTable is already viewed as a sampler, then it would actually be more consistent with the Random API to add the `rng` argument to the constructor, wouldn't it?

ablaom:
Let me clarify the API as I understand it

- AliasTable is completely private object (no mention in docs - I have checked this with a `grep`).

- It's raison d'etre is to support the implementation of `rand([rng, ] d, [n])` where `d` is an instance of `Distribtutions.Categorical`, which *is* public; see https://juliastats.org/Distributions.jl/latest/univariate/#Discrete-Distributions-1 .

- In the current API *no* `Distribution` object is a "sampler" in the sense @devmotion  appears to be using: these are never bound to any random number generator. Rather, the interface point for specifying a RNG is in the `rand([::AbstractRNG,] ::Distribution, [::Int])` call. 

In view of this I would agree with  @rfourquet, unless there is a complete design change (ie change in the interface point for RNG's) the existing proposal is probably optimal the Distributions side.

andreasnoack:
I agree that the current implementation should be ([almost](https://github.com/JuliaStats/Distributions.jl/pull/1157#discussion_r472953764)) fine and I'm also okay with considering this an internal change. The struct is used directly in a couple of packages but it's only used for `rand` calls.

At some point, we might want to think through if the changes to `Random` should have any consequences for how samplers are set up here but that is a bigger task beyond the scope of this PR.

src/samplers/aliastable.jl
ablaom: Made n a field instead of a type parameter

ablaom:
The windows fail does appear to be related to the changes in this PR. It relates to a test of Poisson samplers:

**test/samplers.jl: 64**
```julia
## Poisson samplers

for (S, paramlst) in [
    (PoissonCountSampler, [0.2, 0.5, 1.0, 2.0, 5.0, 10.0, 15.0, 20.0, 30.0]),
    (PoissonADSampler, [5.0, 10.0, 15.0, 20.0, 30.0])]
    local S

    println("    testing $S")
    for μ in paramlst
        test_samples(S(μ), Poisson(μ), n_tsamples)
        test_samples(S(μ), Poisson(μ), n_tsamples, rng=rng)
    end
end
```

I am not familiar enough with the code to say for sure. @andreasnoack Any ideas?

src/samplers/aliastable.jl
```diff
@@ -1,21 +1,21 @@
-struct AliasTable{S} <: Sampleable{Univariate,Discrete}
+struct AliasTable <: Sampleable{Univariate,Discrete}
     accept::Vector{Float64}
     alias::Vector{Int}
-    isampler::S
+    n::Int
```
andreasnoack, line 4:
```suggestion
```

src/samplers/aliastable.jl
```diff
@@ -1,21 +1,21 @@
...
 end
-ncategories(s::AliasTable) = length(s.accept)
+ncategories(s::AliasTable) = s.n
```
andreasnoack, line 6:
```suggestion
ncategories(s::AliasTable) = length(s.alias)
```

src/samplers/aliastable.jl
```diff
@@ -1,21 +1,21 @@
...
     accp = Vector{Float64}(undef, n)
     alias = Vector{Int}(undef, n)
     StatsBase.make_alias_table!(probs, 1.0, accp, alias)
-    AliasTable(accp, alias, Random.RangeGenerator(1:n))
+    AliasTable(accp, alias, n)
 end
 
 function rand(rng::AbstractRNG, s::AliasTable)
-    i = rand(rng, s.isampler) % Int
+    i = rand(rng, 1:(s.n)) % Int
```
andreasnoack, line 18:
```suggestion
    i = rand(rng, 1:length(s.alias)) % Int
```

andreasnoack (commented):
Just a last minor suggestion: the `n` is implied by the vectors in the struct so it can be left out.

> The windows fail does appear to be related to the changes in this PR.

Yes. I can see similar failures for other PRs so we can just ignore that.

ablaom:
Thanks ❤️

johnczito:
Sorry I'm a bit behind. Do we know why some of the builds on AppVeyor haven't been passing since this was merged?

PR #1159, opened 2020-08-17
Undeprecate quantile() for arrays

### LINKED ISSUE #1150: Undeprecate no-dot syntax for quantile() and nquantile()
This issue picks up where JuliaStats/StatsBase.jl#586 left off. At issue is the following snippet in the Distributions package:

https://github.com/JuliaStats/Distributions.jl/blob/20c91d9efcc5f96913bf8e38be2e3fb14b21942b/src/deprecates.jl#L39

The macro affects the `quantile()` and `nquantile()` functions, of which the former is used by StatsBase in [this line](https://github.com/JuliaStats/StatsBase.jl/blob/54cc6fe2e2f709623b9c75275884f9b295fdbd35/src/scalarstats.jl#L340).

According to @nalimilan, StatsBase relies on the deprecated behavior. This is because when `quantile()` is broadcasted over the given weights/probabilities (second param, `p::Union{AbstractArray, Tuple{Vararg{Real}}}`), `_quantilesort!` is run twice on the same parameter (first param, `v::AbstractVector`). See [this code in the Statistics package](https://github.com/JuliaLang/Statistics.jl/blob/b384104d35ff0e7cf311485607b177223ed72b9a/src/Statistics.jl#L858-L1053) where the sorting function would be broadcast over using the new syntax requested by Distributions. In any case, requiring broadcasting would mean requiring the first parameter to be sorted twice.

Would it be best to un-deprecate this functionality, or is there another course? For now, using Statistics as per its own documentation gives the user a deprecation warning, which is… not ideal, especially for first-time users, of which I’m one.

ron-wolf:
Fixes #1150 by undeprecating omission of the broadcast operator `.` when using `quantile`, `quantile!`, and `_quantile!`.

For reason, see [this `StatsBase` comment](https://github.com/JuliaStats/StatsBase.jl/pull/586#issuecomment-653936354). For full discussion, see JuliaStats/StatsBase.jl#586 and [JuliaStats/Distributions.jl#1150](JuliaStats/Distributions.jl/issues/1150).

### DIFF
--- a/test/quantile_collection.jl
+++ b/test/quantile_collection.jl
@@ -1,0 +1,14 @@
+using Test
+using StatsBase, Statistics, Distributions
+
+# tests that quantile(collection) still works by way of iqr(collection)
+
+@testset "Quantile of Collection" begin
+    numbers, weights = 0:9, repeat(0.1, 10)
+    avg_num = mean(numbers, weights)
+    
+    stddev = std(numbers, weights, mean=avg_num)
+    num_dist = Normal(avg_num, stdev)
+    
+    range = iqr(num_dist)
+end


ron-wolf:
@nalimilan Sorry, I only just saw this. Where would I add a test? Also, are there any basic methods/tools or best practices for tests in Julia that I might not be aware of?

nalimilan:
Sorry, I meant testing `quantile` on a distribution, not on a collection (collections are tested in StatsBase). I'm not sure where to put it, maybe in test/utils.jl?

andreasnoack:
@ron-wolf Bump. Would you be able to add a tests as suggested in https://github.com/JuliaStats/Distributions.jl/pull/1159#issuecomment-702589660?

nalimilan:
Bump.

ron-wolf:
@nalimilan

> Sorry, I meant testing `quantile` on a distribution, not on a collection (collections are tested in StatsBase).

Looking back on this I was testing it on a distribution after all, not a collection. So would this just require a rename of the file and the test case?

ron-wolf:
Actually nevermind, I have completely lost track of what has to be changed and why. Why can't the definition for `iqr` just be changed to use `quantile.` instead of `quantile`? Why are we only undeprecating `quantile` and not the other implicitly broadcasted operators? In any case, what are the concrete steps involved? “I don’t know” is an acceptable answer.

ron-wolf:
I think I may have found a better solution to the point brought up in [this comment](JuliaStats/StatsBase.jl/pull/586#issuecomment-653936354)?:

```julia
# we specify AbstractVector{<:Real} as the type for a discrete 
# distribution because it is the most generic type that describes 
# a sortable  sample: a discrete sequence of ordinal values that 
# can be sampled at certain positions

quantile(x::AbstractVector{<:Real}, p::Real; sorted=false) = let
    # ensure x is sorted, then sample it at the location/input p
    sorted || x = sort!(collect(x))
    ... # do your sampling magic!
end
quantile(x::Any, p::Real) = ...

# dispatch on discrete distributions, both sorted & unsorted
quantile(x::AbstractVector{<:Real}, p::AbstractWeights; sorted=false) = let
    # ensure x is sorted, then sample it at each of the locations/inputs in p
    sorted || x = sort!(collect(x))
    quantile.(Ref(x), p; sorted=true)
end

# a quick re-dispatch to broadcasting; instead of deprecating 
# this, we just silently correct to the right behavior so 
# end-users don't have to worry about the distinction
@inline quantile(x::Any, p::AbstractWeights) = quantile.(Ref(x), p)

# TODO: is there some way to deprecate 
#   quantile.(::AbstractVector{<:Real}, p::AbstractWeights),
# which is now radically inefficient?
```

which would fix our efficiency woes, while also allowing this simple definition to remain in place:

```julia
iqr(x) = let
    q = quantile(x, [.25, .75])
    q[2] - q[1]
end
```

Are there any other functions which, like you said, have this optimization of sorting and then sampling at each point? If so, perhaps it would be best to enumerate them all in this issue, and then collaborate to write a define-them-all-in-one-go metaprogramming loop.

matbesancon:
what's the status here?

devmotion:
It seems in Distributions there is no performance improvement if `quantile` is not broadcasted but would support arrays (correct me if I am wrong). Thus it seems the PR would fix an inconsistency with the use of `quantile` in StatsBase by introducing an inconsistency with other functions such as `pdf`, `cdf`, etc. in Distributions without any real benefit for Distributions. Wouldn't be the most natural solution just be that StatsBase supports arrays (since it improves performance) but Distributions doesn't?

nalimilan:
> Thus it seems the PR would fix an inconsistency with the use of `quantile` in StatsBase by introducing an inconsistency with other functions such as `pdf`, `cdf`, etc. in Distributions without any real benefit for Distributions. Wouldn't be the most natural solution just be that StatsBase supports arrays (since it improves performance) but Distributions doesn't?

Actually there would be a benefit for Distribution, as currently a deprecation warning is printed when calling `iqr` on a distribution object. See https://github.com/JuliaStats/StatsBase.jl/pull/586. The problem is that to be efficient on both arrays and distributions, `iqr` has to call the vectorized method `quantile(x, p)` rather than `quantile.(x, p)`. One alternative would be to override the definition of `broacast` for `quantile` in StatsBase to call the optimized method under the hood (i.e. to sort data only once).

devmotion:
> One alternative would be to override the definition of broacast for quantile in StatsBase to call the optimized method under the hood (i.e. to sort data only once).

Another possible approach would be to add something like
```julia
function StatsBase.iqr(d::Distribution)
    q = map(Base.Fix1(quantile, d), (0.25, 0.75))
    return q[2] - q[1]
end
```
to Distributions, it seems?

nalimilan:
Yeah that would be simpler fix. Not sure whether other functions would need the same change.

ron-wolf:
> Wouldn't be the most natural solution just be that StatsBase supports arrays (since it improves performance) but Distributions doesn't?

In the interest of getting this done, I think Step 1 is to figure out what state we want the code to end up in—in a module-agnostic fashion. Step 2, then, would be figuring out what belongs to Distributions vs. StatsBase; this can be considered a secondary concern.

Regarding the actual code to be written, how do we feel about the merits of the two outlines we’ve developed thus far? To reiterate, this code snippet communicates the main thrust of @devmotion’s sketch:

```julia
StatsBase.iqr(d::Distribution) = quantile(d, 0.75) - quantile(d, 0.25)
```

And this was my sketch:

```julia
# we specify AbstractVector{<:Real} as the type for a discrete 
# distribution because it is the most generic type that describes 
# a sortable  sample: a discrete sequence of ordinal values that 
# can be sampled at certain positions

quantile(x::AbstractVector{<:Real}, p::Real; sorted=false) = let
    # ensure x is sorted, then sample it at the location/input p
    sorted || x = sort!(collect(x))
    ... # do your sampling magic!
end
quantile(x::Any, p::Real) = ...

# dispatch on discrete distributions, both sorted & unsorted
quantile(x::AbstractVector{<:Real}, p::AbstractWeights; sorted=false) = let
    # ensure x is sorted, then sample it at each of the locations/inputs in p
    sorted || x = sort!(collect(x))
    quantile.(Ref(x), p; sorted=true)
end

# a quick re-dispatch to broadcasting; instead of deprecating 
# this, we just silently correct to the right behavior so 
# end-users don't have to worry about the distinction
@inline quantile(x::Any, p::AbstractWeights) = quantile.(Ref(x), p)

# TODO: is there some way to deprecate 
#   quantile.(::AbstractVector{<:Real}, p::AbstractWeights),
# which is now radically inefficient?
```

Any thoughts on whether one of these sketches, or a different sketch, would be the best way forward?

nalimilan:
> Regarding the actual code to be written, how do we feel about the merits of the two outlines we’ve developed thus far? To reiterate, this code snippet communicates the main thrust of @devmotion’s sketch:
> 
> ```julia
> StatsBase.iqr(d::Distribution) = quantile(d, 0.75) - quantile(d, 0.25)
> ```
AFAICT that's the best solution.

ron-wolf:
Okay. Can someone author a PR [over here](/ron-wolf/Distributions.jl/tree/patch-1) for me to accept, so as to have the full changeset show up in this PR?

PR #1162, opened 2020-08-24
Optimize vMF sampling

### LINKED ISSUE #1161: RFC: Performance Improvements for the VonMisesFisher sampler
I've noticed that the samplers for the vMF distribution can be optimized further.

There's two parts which I believe can be improved:

1. In the case of a vMF distribution on the 2-sphere (which is perhaps the most common case), the CDF for the W random variable is analytically invertible, meaning we can completely skip the rejection sampling step. See section 3 of: http://www.mitsuba-renderer.org/~wenzel/files/vmf.pdf

2. The orthogonal matrix used to re-align the drawn samples to have the correct mean does not need to be explicitly constructed. As it is now, the sampler builds a matrix by performing a QR decomposition. However, we can create a different orthogonal matrix which serves the same purpose: a Householder reflection defined by the vector \mu - e_1 (I can give more details on the proof if necessary). The advantage of this approach is that the final transformation applied to our samples now takes linear time and space in the length of the vector. 

The disadvantage with both changes is breaking backwards reproducibility, though for (2) the sampled vectors before and after this change would still be related by some orthogonal transformation which leaves the mean vector invariant.

I'm open to implementing both of these changes, just want some feedback about whether the maintainers are OK with these.

emerali:
Resolves #1161 

Had some free time so I just implemented it right away lol

### DIFF
--- a/src/samplers/vonmisesfisher.jl
+++ b/src/samplers/vonmisesfisher.jl
@@ -18,10 +18,10 @@ function VonMisesFisherSampler(μ::Vector{Float64}, κ::Float64)
     VonMisesFisherSampler(p, κ, b, x0, c, v)
 end
 
-@inline function _vmf_rot!(spl::VonMisesFisherSampler, x::AbstractVector)
+@inline function _vmf_rot!(v::AbstractVector, x::AbstractVector)
     # rotate
-    scale = 2.0 * (spl.v' * x)
-    @. x -= (scale * spl.v)
+    scale = 2.0 * (v' * x)
+    @. x -= (scale * v)
     return x
 end
 
@@ -42,7 +42,7 @@ function _rand!(rng::AbstractRNG, spl::VonMisesFisherSampler, x::AbstractVector)
         x[i] *= r
     end
 
-    return _vmf_rot!(spl, x)
+    return _vmf_rot!(spl.v, x)
 end
 
 
--- a/src/samplers/vonmisesfisher.jl
+++ b/src/samplers/vonmisesfisher.jl
@@ -36,8 +36,8 @@ function _rand!(rng::AbstractRNG, spl::VonMisesFisherSampler,
     end
 
     # rotate
-    scale = 2.0 * (spl.v' * t)
     copyto!(x, t)
+    scale = 2.0 * (spl.v' * t)
     @. x -= (scale * spl.v)
     return x
 end
@@ -84,8 +84,19 @@ end
 _vmf_genw(rng::AbstractRNG, s::VonMisesFisherSampler) =
     _vmf_genw(rng, s.p, s.b, s.x0, s.c, s.κ)
 
-function _vmf_householder_vec(u::Vector{Float64})
-    u = normalize(u)
-    u[1] -= 1.0
-    return normalize!(u)
+function _vmf_householder_vec(μ::Vector{Float64})
+    # assuming μ is a unit-vector (which it should be)
+    #  can compute v in a single pass over μ
+
+    p = length(μ)
+    v = zeros(p)
+    v[1] = μ[1] - 1.0
+    s = sqrt(-2*v[1])
+    v[1] /= s
+
+    @inbounds for i in 2:p
+        v[i] = μ[i] / s
+    end
+
+    return v
 end
--- a/src/samplers/vonmisesfisher.jl
+++ b/src/samplers/vonmisesfisher.jl
@@ -58,18 +58,13 @@ end
 
 _vmf_bval(p::Int, κ::Real) = (p - 1) / (2.0κ + sqrt(4 * abs2(κ) + abs2(p - 1)))
 
-function _vmf_genw(rng::AbstractRNG, p, b, x0, c, κ)
-    # generate the W value -- the key step in simulating vMF
-    #
-    #   following movMF's document for the p > 3 case
-    #   and Wenzel Jakob's document for the p == 3 case
-    #
-    if p == 3
-        ξ = rand(rng)
-        w = 1.0 + (log(ξ + (1.0 - ξ)*exp(-2κ))/κ)
-        return w::Float64
-    end
+function _vmf_genw3(rng::AbstractRNG, p, b, x0, c, κ)
+    ξ = rand(rng)
+    w = 1.0 + (log(ξ + (1.0 - ξ)*exp(-2κ))/κ)
+    return w::Float64
+end
 
+function _vmf_genwp(rng::AbstractRNG, p, b, x0, c, κ)
     r = (p - 1) / 2.0
     betad = Beta(r, r)
     z = rand(rng, betad)
@@ -80,6 +75,13 @@ function _vmf_genw(rng::AbstractRNG, p, b, x0, c, κ)
     end
     return w::Float64
 end
+
+# generate the W value -- the key step in simulating vMF
+#
+#   following movMF's document for the p != 3 case
+#   and Wenzel Jakob's document for the p == 3 case
+_vmf_genw(rng::AbstractRNG, p, b, x0, c, κ) = (p == 3) ? _vmf_genw3(rng, p, b, x0, c, κ) : _vmf_genwp(rng, p, b, x0, c, κ)
+
 
 _vmf_genw(rng::AbstractRNG, s::VonMisesFisherSampler) =
     _vmf_genw(rng, s.p, s.b, s.x0, s.c, s.κ)
--- a/src/samplers/vonmisesfisher.jl
+++ b/src/samplers/vonmisesfisher.jl
@@ -6,7 +6,7 @@ struct VonMisesFisherSampler
     b::Float64
     x0::Float64
     c::Float64
-    Q::Matrix{Float64}
+    v::Vector{Float64}
 end
 
 function VonMisesFisherSampler(μ::Vector{Float64}, κ::Float64)
@@ -14,8 +14,8 @@ function VonMisesFisherSampler(μ::Vector{Float64}, κ::Float64)
     b = _vmf_bval(p, κ)
     x0 = (1.0 - b) / (1.0 + b)
     c = κ * x0 + (p - 1) * log1p(-abs2(x0))
-    Q = _vmf_rotmat(μ)
-    VonMisesFisherSampler(p, κ, b, x0, c, Q)
+    v = _vmf_householder_vec(μ)
+    VonMisesFisherSampler(p, κ, b, x0, c, v)
 end
 
 function _rand!(rng::AbstractRNG, spl::VonMisesFisherSampler,
@@ -36,7 +36,9 @@ function _rand!(rng::AbstractRNG, spl::VonMisesFisherSampler,
     end
 
     # rotate
-    mul!(x, spl.Q, t)
+    scale = 2.0 * (spl.v' * t)
+    copyto!(x, t)
+    @. x -= (scale * spl.v)
     return x
 end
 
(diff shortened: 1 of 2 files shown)


emerali:
# Benchmarks:

## master:
### p = 3
```
julia> vmf = VonMisesFisher(rand(3));

julia> @benchmark rand($vmf)
BenchmarkTools.Trial: 
  memory estimate:  1.22 KiB
  allocs estimate:  10
  --------------
  minimum time:     3.841 μs (0.00% GC)
  median time:      4.130 μs (0.00% GC)
  mean time:        4.639 μs (0.65% GC)
  maximum time:     157.598 μs (95.62% GC)
  --------------
  samples:          10000
  evals/sample:     8

julia> @benchmark rand($vmf, 1000)
BenchmarkTools.Trial: 
  memory estimate:  24.63 KiB
  allocs estimate:  11
  --------------
  minimum time:     319.272 μs (0.00% GC)
  median time:      332.700 μs (0.00% GC)
  mean time:        340.943 μs (0.15% GC)
  maximum time:     1.573 ms (73.92% GC)
  --------------
  samples:          10000
  evals/sample:     1

```

### p = 1000
```
julia> vmf = VonMisesFisher(rand(1000));

julia> @benchmark rand($vmf)
BenchmarkTools.Trial: 
  memory estimate:  31.08 MiB
  allocs estimate:  16
  --------------
  minimum time:     102.893 ms (0.00% GC)
  median time:      121.886 ms (0.63% GC)
  mean time:        162.252 ms (5.46% GC)
  maximum time:     377.688 ms (0.67% GC)
  --------------
  samples:          31
  evals/sample:     1

julia> @benchmark rand($vmf, 1000)
BenchmarkTools.Trial: 
  memory estimate:  38.70 MiB
  allocs estimate:  17
  --------------
  minimum time:     449.606 ms (0.27% GC)
  median time:      675.868 ms (0.15% GC)
  mean time:        657.465 ms (4.89% GC)
  maximum time:     839.365 ms (0.09% GC)
  --------------
  samples:          8
  evals/sample:     1

```

## this PR

### p = 3
```
julia> vmf = VonMisesFisher(rand(3));

julia> @benchmark rand($vmf)
BenchmarkTools.Trial: 
  memory estimate:  336 bytes
  allocs estimate:  3
  --------------
  minimum time:     242.661 ns (0.00% GC)
  median time:      256.136 ns (0.00% GC)
  mean time:        282.311 ns (1.80% GC)
  maximum time:     2.801 μs (82.71% GC)
  --------------
  samples:          10000
  evals/sample:     413

julia> @benchmark rand($vmf, 1000)
BenchmarkTools.Trial: 
  memory estimate:  23.73 KiB
  allocs estimate:  4
  --------------
  minimum time:     114.304 μs (0.00% GC)
  median time:      116.085 μs (0.00% GC)
  mean time:        123.298 μs (0.32% GC)
  maximum time:     1.100 ms (81.23% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

### p = 1000
```
julia> vmf = VonMisesFisher(rand(1000));

julia> @benchmark rand($vmf)
BenchmarkTools.Trial: 
  memory estimate:  23.81 KiB
  allocs estimate:  3
  --------------
  minimum time:     13.781 μs (0.00% GC)
  median time:      14.682 μs (0.00% GC)
  mean time:        15.868 μs (3.30% GC)
  maximum time:     1.195 ms (92.74% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark rand($vmf, 1000)
BenchmarkTools.Trial: 
  memory estimate:  7.64 MiB
  allocs estimate:  4
  --------------
  minimum time:     10.918 ms (0.00% GC)
  median time:      11.292 ms (0.00% GC)
  mean time:        12.387 ms (3.96% GC)
  maximum time:     83.618 ms (83.11% GC)
  --------------
  samples:          404
  evals/sample:     1

```

emerali:
This PR (48c4245):

```
julia> for p in [2, 3, 4, 1000]
           vmf = VonMisesFisher(rand(p)); @show p
           show(stdout, MIME"text/plain"(), @benchmark rand($vmf))
           println()
           show(stdout, MIME"text/plain"(), @benchmark rand($vmf, 1000))
           println()
       end
p = 2
BenchmarkTools.Trial: 
  memory estimate:  288 bytes
  allocs estimate:  3
  --------------
  minimum time:     318.882 ns (0.00% GC)
  median time:      353.138 ns (0.00% GC)
  mean time:        389.258 ns (2.49% GC)
  maximum time:     8.583 μs (91.68% GC)
  --------------
  samples:          10000
  evals/sample:     246
BenchmarkTools.Trial: 
  memory estimate:  15.94 KiB
  allocs estimate:  3
  --------------
  minimum time:     207.156 μs (0.00% GC)
  median time:      226.899 μs (0.00% GC)
  mean time:        237.945 μs (0.10% GC)
  maximum time:     1.189 ms (71.64% GC)
  --------------
  samples:          10000
  evals/sample:     1
p = 3
BenchmarkTools.Trial: 
  memory estimate:  336 bytes
  allocs estimate:  3
  --------------
  minimum time:     241.591 ns (0.00% GC)
  median time:      254.943 ns (0.00% GC)
  mean time:        280.737 ns (1.98% GC)
  maximum time:     3.008 μs (86.80% GC)
  --------------
  samples:          10000
  evals/sample:     421
BenchmarkTools.Trial: 
  memory estimate:  23.73 KiB
  allocs estimate:  4
  --------------
  minimum time:     114.325 μs (0.00% GC)
  median time:      119.033 μs (0.00% GC)
  mean time:        124.773 μs (0.34% GC)
  maximum time:     1.250 ms (88.97% GC)
  --------------
  samples:          10000
  evals/sample:     1
p = 4
BenchmarkTools.Trial: 
  memory estimate:  336 bytes
  allocs estimate:  3
  --------------
  minimum time:     355.152 ns (0.00% GC)
  median time:      406.272 ns (0.00% GC)
  mean time:        441.461 ns (1.19% GC)
  maximum time:     4.822 μs (86.08% GC)
  --------------
  samples:          10000
  evals/sample:     204
BenchmarkTools.Trial: 
  memory estimate:  31.55 KiB
  allocs estimate:  4
  --------------
  minimum time:     247.321 μs (0.00% GC)
  median time:      275.635 μs (0.00% GC)
  mean time:        285.842 μs (0.17% GC)
  maximum time:     1.177 ms (75.45% GC)
  --------------
  samples:          10000
  evals/sample:     1
p = 1000
BenchmarkTools.Trial: 
  memory estimate:  23.81 KiB
  allocs estimate:  3
  --------------
  minimum time:     13.957 μs (0.00% GC)
  median time:      15.014 μs (0.00% GC)
  mean time:        16.490 μs (3.78% GC)
  maximum time:     1.668 ms (80.91% GC)
  --------------
  samples:          10000
  evals/sample:     1
BenchmarkTools.Trial: 
  memory estimate:  7.64 MiB
  allocs estimate:  4
  --------------
  minimum time:     10.936 ms (0.00% GC)
  median time:      11.199 ms (0.00% GC)
  mean time:        11.792 ms (3.50% GC)
  maximum time:     83.431 ms (86.34% GC)
  --------------
  samples:          424
  evals/sample:     1
```

emerali:
Added a couple more benchmarks for p=2 and p=4. You can see that there's a dip in the computation time for p=3, since we avoid rejection sampling there. Hence the reduction in computation time for p=3 doesn't *just* come from the improved rotation operation.

Additionally, I noticed that the vMF distribution errors for p=1. In this case the distribution technically reduces to a discrete distribution on {+1, -1} (basically a post-processed Bernoulli), should vMF have a special case for that, or should the constructor error out and tell the user to use a Bernoulli distribution instead? I'd assume that anyone using the vMF distribution likely won't make use of this case, but it's probably a good idea for the sampling function to not error out in that scenario regardless.

There's a similar issue for kappa = 0 (vMF becomes a uniform spherical distribution), though currently the vMF constructor rejects that value.

johnczito:
Thanks for this! Looks good to me. My only suggestion is to take this opportunity to make the unit tests more convincing. Might include:

- Test the two `w` samplers against one another in the `p = 3` case;
- Check that the `x` that's constructed at the end of `rand!` is the same as if you naively computed `H(μ - e₁ ) * t`.

I am in favor of things collapsing to the special cases when `p = 1` or `κ = 0` (especially the latter), but I think that should be handled in a separate PR.

emerali:
I've added the tests for the w sampler. It now compares the sample statistics of the p=3 case against those of the rejection sampling method. In addition, the sample statistics for the p=3 case are compared against analytical formulas for the mean/variance.

> Check that the x that's constructed at the end of rand! is the same as if you naively computed `H(μ - e₁) * t`.

I'm having trouble reading this notation, do you mean construct the Householder matrix explicitly and apply it as a comparison?

emerali:
I agree with dealing with those special cases in a separate PR, particularly the `κ = 0` case, since I feel that a `UniformSpherical` distribution deserves an implementation in this package.

Another optimization I believe should be possible would be to defer the sampling for the p=2 case to the univariate `VonMises` distribution, and then post process the result to produce a vector (via `sincos`), though again, that might be better suited for a separate PR.

emerali:
Also, quick question, is the temporary variable `t` in `_rand!` necessary anymore? AFAICT, we can skip it now, since the only reason it was needed before was because the old implementation needed to apply the matrix Q, but we can do this matrix application fully in-place now. Just need a sanity check on this.

johnczito:
> I'm having trouble reading this notation, do you mean construct the Householder matrix explicitly and apply it to the vector `t` as a comparison?

Exactly. I believe the correctness of it, but no reason not to document it.

> I agree with dealing with those special cases in a separate PR, particularly the `κ = 0` case, since I feel that a `UniformSpherical` distribution deserves an implementation in this package.

Yeah, if `Normal(0, 0)` and `TDist(Inf)` work without error, then these special cases should too.

> is the temporary variable `t` in `_rand!` necessary anymore?

Not sure.

src/samplers/vonmisesfisher.jl
```diff
@@ -6,47 +6,49 @@ struct VonMisesFisherSampler
...
 end
 
-function _rand!(rng::AbstractRNG, spl::VonMisesFisherSampler,
-                x::AbstractVector, t::AbstractVector)
+@inline function _vmf_rot!(spl::VonMisesFisherSampler, x::AbstractVector)
+    # rotate
+    scale = 2.0 * (spl.v' * x)
+    @. x -= (scale * spl.v)
+    return x
+end
```
johnczito, line 26:
```suggestion
@inline function _vmf_rot!(v::AbstractVector, x::AbstractVector)
    # rotate
    scale = 2.0 * (v' * x)
    @. x -= (scale * v)
    return x
end
```

src/samplers/vonmisesfisher.jl
```diff
@@ -6,47 +6,49 @@ struct VonMisesFisherSampler
...
-    for i = 2:p
-        t[i] *= r
+    @inbounds for i = 2:p
+        x[i] *= r
     end
 
-    # rotate
-    mul!(x, spl.Q, t)
-    return x
+    return _vmf_rot!(spl, x)
```
johnczito, line 45:
```suggestion
    return _vmf_rot!(spl.v, x)
```

test/vonmisesfisher.jl
```diff
@@ -22,6 +22,54 @@ function gen_vmf_tdata(n::Int, p::Int,
...
+        μ = randn(rng, p)
+        x = randn(rng, p)
+    end
+    κ = norm(μ)
+    μ = μ ./ κ
+
+    s = Distributions.VonMisesFisherSampler(μ, κ)
+    H = I - 2*s.v*s.v'
+
+    @test Distributions._vmf_rot!(s, copy(x)) ≈ (H*x)
```
johnczito, line 39:
```suggestion
    @test Distributions._vmf_rot!(s.v, copy(x)) ≈ (H*x)
```

test/vonmisesfisher.jl
```diff
@@ -22,6 +22,54 @@ function gen_vmf_tdata(n::Int, p::Int,
...
+        x = randn(p)
+    else
+        μ = randn(rng, p)
+        x = randn(rng, p)
+    end
+    κ = norm(μ)
+    μ = μ ./ κ
+
+    s = Distributions.VonMisesFisherSampler(μ, κ)
+    H = I - 2*s.v*s.v'
```
johnczito, line 37:
```suggestion
    s = Distributions.VonMisesFisherSampler(μ, κ)
    v = μ - vcat(1, zeros(p-1))
    H = I - 2*v*v'/(v'*v)
```

johnczito: So we're also checking `_vmf_householder_vec`.

johnczito (approved):

emerali:
Nice, once this gets merged I'll get started on the UniformSpherical/UniformBall distributions, and then the special cases for vMF will come after.

johnczito:
I'll merge this in 24 hours if there is no other feedback.

PR #1164, opened 2020-08-26
Enable Windows and x86 testing on Travis. Remove appveyor.yml.

andreasnoack:
To consolidate the testing. Travis also has a better account management system. @simonbyrne It looks like you have the AppVeyor account running the current Windows tests so you probably want to disable them if this is merged.


andreasnoack:
It looks like the `PoissonADSampler` now fails on 32 bit on AppVeyor which only tested on 1.5 and nightly. The sampler fails on 32 bit for all versions of Julia when testing on Travis. The failure only happens for `μ=5` and I noticed https://github.com/JuliaStats/Distributions.jl/blob/253c3a0813e0a08bf41fdf22f83ea540cc89a4c5/src/samplers/poisson.jl#L40 so maybe we just shouldn't test that sampler with `μ=5`.

andreasnoack:
So it looks like mle fitting of Dirichlet doesn't work on 32 bit Linux or the random variates are not right. Any ideas?

devmotion:
The error about a non-positive parameter vector of the Dirichlet distribution in https://github.com/JuliaStats/Distributions.jl/blob/82120fcb8b51e0e3f5789b7ae78a9f9aa8ad2375/src/multivariate/dirichlet.jl#L33-L34 seems weird. It is caused by the call in https://github.com/JuliaStats/Distributions.jl/blob/82120fcb8b51e0e3f5789b7ae78a9f9aa8ad2375/src/multivariate/dirichlet.jl#L384 - however, lines https://github.com/JuliaStats/Distributions.jl/blob/82120fcb8b51e0e3f5789b7ae78a9f9aa8ad2375/src/multivariate/dirichlet.jl#L365-L368 ensure that the values in the parameter vector are at least `1.0e-12`. I would assume that `1.0e-12 > 0` should not fail on 32bit machines?

andreasnoack:
I'm not sure what to do here. I just tried to install a 32 bit Debian in VirtualBox to see if I could reproduce but it was unsuccessful. Do we care about 32 anymore? I noticed that 32 bit Ubuntu is longer available.

andreasnoack:
I've now pushed a commit that disables testing on 32 bit. I don't think it makes sense anymore since we are getting spurious errors and nobody is able to debug this because none of us have access to a 32 bit system. I tried running a 32 bit Debian in VirtualBox but that wasn't sufficient to trigger the error so it probably requires an actual 32 bit system. Unless somebody with a 32 bit system steps forward and volunteers to debug this, I suggest that we merge this to get CI working properly again.

andreasnoack:
So @ararslan summarized the arch settings for Travis in a Slack thread and I realized that Travis is testing 32 bit Julia on a 64 bit Linux whereas I'd installaed a 32 bit Debian to try to reproduce the test failure but unsuccessfully. Now I just tried to install 32 bit Julia on a 64 bit Linux and I can reproduce the failure. I'm not sure what to conclude based on that. Now that I can reproduce the failure, I'll try to debug a little further.

andreasnoack:
With https://github.com/JuliaStats/Distributions.jl/pull/1191 merged, I've been able to reenable 32 bit testing in this PR so I think we should get this one to consolidate the CI runs on Travis.

PR #1168, opened 2020-09-03
Update extends.md

severinson:
I think there's a typo in the multivariate sampler section.

Implementing `function _rand!(s::Spl, x::AbstractVector{T}) where T<:Real` results in `MethodError: no method matching _rand!(::Random._GLOBAL_RNG, ...`. I think the correct function definition is `function _rand!(rng::AbstractRNG, s::Spl, x::AbstractVector{T}) where T<:Real`.

I suspect that it's the same for the remaining functions associated with multivariate samplers. Could someone confirm this?

### DIFF
--- a/docs/src/extends.md
+++ b/docs/src/extends.md
@@ -42,36 +42,36 @@ This function can assume that the dimension of `x` is correct, and doesn't need 
 The package implements both `rand` and `rand!` as follows (which you don't need to implement in general):
 
 ```julia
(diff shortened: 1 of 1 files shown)


matbesancon:
this sounds about right, thanks for pointing it out. Do you want to merge this PR or correct the other calls before?

severinson:
I've corrected the other calls now. Please have a look.

matbesancon:
Thanks a lot!

Issue #1170, opened 2020-09-04
Inconsistent fallback behaviour of `logpdf` and `pdf`


phipsgabler:
We had at least two issues in Turing where calling `pdf` on custom discrete univariate distributions lead to a stack overflow: https://github.com/TuringLang/Turing.jl/issues/1397, https://github.com/TuringLang/Turing.jl/pull/1089.  This was because someone expected `pdf` to fall back to `logpdf` -- as is already the case with multivariate distributions -- while it's [actually the other way round](https://github.com/JuliaStats/Distributions.jl/blob/09f2bbdf9faca686d72ff8d30d0299137d601842/src/univariates.jl#L342), and `pdf` for discrete univariate distributions falls back to itself:

```julia
pdf(d::DiscreteUnivariateDistribution, x::Real) = isinteger(x) ? pdf(d, round(Int, x)) : zero(partype(d))
```

Is there a reasonable way to make these behaviours consistent (i.e., everything always works if you implement `logpdf`)?  

I mean, it's documented already, but it's just not what I expect for many univariate distributions.  Or has it already been explicitely decided to do it this way for better reasons?

PR #1171, opened 2020-09-04
[Breaking] Fix inconsistent fallback behaviour of logpdf and pdf

devmotion:
This PR fixes https://github.com/JuliaStats/Distributions.jl/issues/1170 by also defining the fallback `pdf(d, x) = exp(logpdf(d, x))` for univariate distributions instead of the other way around.

Moreover, it removes the default behaviour of forwarding `logpdf(d, x)` to `logpdf(d, round(Int, x))` (or `pdf` before) in case of discrete distributions and instead lets distributions deal with it. This avoids method ambiguity issues for DiscreteNonParametric and is often not needed, it seems.


matbesancon:
is this change breaking in some way?

src/univariate/discrete/binomial.jl
```diff
@@ -110,6 +110,18 @@ end
 
 @_delegate_statsfuns Binomial binom n p
 
+function pdf(d::Binomial, x::Real)
+    _insupport = insupport(d, x)
+    s = pdf(d, _insupport ? round(Int, x) : 0)
```
matbesancon, line 115: is this not recursing without ending (calling pdf again and again)?

src/univariate/discrete/geometric.jl
```diff
@@ -69,17 +69,8 @@ entropy(d::Geometric) = (-xlogx(succprob(d)) - xlogx(failprob(d))) / d.p
...
-        p = d.p
-        return p < one(p) / 10 ? p * exp(log1p(-p) * x) : d.p * (one(p) - p)^x
-    else
-        return zero(T)
-    end
-end
-
-function logpdf(d::Geometric{T}, x::Int) where T<:Real
-    x >= 0 ? log(d.p) + log1p(-d.p) * x : -T(Inf)
+function logpdf(d::Geometric, x::Real)
```
mschauer, line 72: This gives nonsensical results for non-integer x

src/univariate/discrete/categorical.jl
```diff
@@ -80,7 +80,10 @@ function cdf(d::Categorical{T}, x::Int) where T<:Real
     return c
 end
 
-pdf(d::Categorical{T}, x::Int) where {T<:Real} = insupport(d, x) ? probs(d)[x] : zero(T)
+function pdf(d::Categorical, x::Real)
+    ps = probs(d)
+    return insupport(d, x) ? ps[round(Int, x)] : zero(eltype(ps))
```
mschauer, line 85: This should be convert?

src/univariate/discrete/binomial.jl
devmotion: It shouldn't since `@_delegate_statsfuns` defines `pdf(d::Binomial, ::Int)`. I'll check it again though.

src/univariate/discrete/categorical.jl
devmotion: It's not completely consistent throughout the package, but I got the impression that (more) often `round` is used in Distributions for such cases (https://github.com/JuliaStats/Distributions.jl/search?q=round%28Int).

src/univariate/discrete/geometric.jl
devmotion: Good catch! Seems I forgot to change the check `x >= 0` to `insupport(d, x)`.

src/univariate/discrete/geometric.jl
```diff
@@ -69,17 +69,8 @@ entropy(d::Geometric) = (-xlogx(succprob(d)) - xlogx(failprob(d))) / d.p
...
+    x >= 0 ? log(d.p) + log1p(-d.p) * x : log(zero(d.p))
```
devmotion, line 73:
```suggestion
    insupport(d, x) ? log(d.p) + log1p(-d.p) * x : log(zero(d.p))
```

devmotion:
> is this change breaking in some way?

I guess it is debatable, but IMO changing the order of documented fallback definitions is breaking and hence I bumped the version number to 0.24.0. Everyone who relied on `logpdf` being defined automatically if `pdf` is implemented for a univariate distribution will get errors when trying to call `logpdf` with this PR.

src/univariate/discrete/geometric.jl
devmotion: I fixed it.

devmotion:
Hmm tests fail since HypothesisTests does not support Distributions 0.24.0 (understandably) - should I reduce the version number again and you update it after the PR is merged? Or how do you want to proceed here?

src/univariate/discrete/binomial.jl
devmotion: It works:
```julia
julia> b = Binomial(10, 0.5)
Binomial{Float64}(n=10, p=0.5)

julia> pdf(b, 0)
0.0009765625

julia> pdf(b, 0.0)
0.0009765625

julia> pdf(b, 1.0)
0.009765625000000002

julia> pdf(b, 1)
0.009765625000000002

julia> pdf(b, 0.5)
0.0
```

devmotion:
I reverted the version bump so that CI tests actually run (and hopefully pass) :slightly_smiling_face: 

One should keep in mind though that IMO the changes are breaking (as discussed above).

devmotion:
Test errors on Windows 32bit are unrelated (see https://github.com/JuliaStats/Distributions.jl/pull/1164#issuecomment-680913155).

mschauer:
What's your own assessment of the status of the PR?

devmotion:
I think I addressed the first round of reviews and tests pass. It also seems to fix at least two other open issues, I noticed. IMO it is ready to be merged, but one should make sure to update the version number to 0.24.0 before making a new release.

PR #1172, opened 2020-09-07
Revert "Return an Int from `invlogccdf` on Geometric"

mkborregaard:
Really sorry @andreasnoack but I think my suggested solution here was the wrong one. This implementation breaks type stability of the method.

I'm actually not even sure Distributions does the wrong thing here, so I suggest simply reverting. I've handled the StatsPlots issue downstream.

This reverts commit 82120fcb8b51e0e3f5789b7ae78a9f9aa8ad2375.

### DIFF
--- a/src/univariate/discrete/geometric.jl
+++ b/src/univariate/discrete/geometric.jl
@@ -132,7 +132,7 @@ function invlogccdf(d::Geometric{T}, lp::Real) where T<:Real
     elseif lp == zero(d.p)
         return zero(T)
     end
-    max(ceil(Int, lp/log1p(-d.p)) - 1, zero(T))
+    max(ceil(lp/log1p(-d.p)) - 1, zero(T))
 end
 
 function mgf(d::Geometric, t::Real)


andreasnoack:
Indeed, I didn't read the PR carefully enough. However, I'm not sure that returning an `Int` is wrong here. That is what we do for e.g. `Binomial`. We'd have to handle the other branches though and I'm not sure what to do with the `NaN` branch.

mkborregaard:
Yes exactly. I also think this should return an Int but unsure how to deal with it.

devmotion:
A different take on this problem would be to be generally less restrictive with definitions of `pdf`, `logpdf`, `cdf`, `logcdf`, `quantile` etc. for discrete distributions and always allow arguments and outputs of type `Real` instead of `Int` only. Even if the support of a discrete distribution is integer-valued (which doesn't have to be the case BTW) and internal implementations are based on integers, one can support evaluations with floating point numbers. And if evaluations can be performed with floating point numbers it should be fine to return floating point numbers for `quantile` as well.

andreasnoack:
Let's merge this and figure out what to do elsewhere.

I think we have some work to do here to clarify the interface: what should be the types returned by functions like `pdf` and `cdf`, what should be types returned by functions that returns values from the sample space like `rand` and `quantile` and how do these two categories relate to the types of the parameters. https://github.com/JuliaStats/Distributions.jl/pull/951 is an attempt to make some progress here.

PR #1176, opened 2020-09-14
Fix for Expectation Bug

### LINKED ISSUE #1175: Error calling `expectation`
Calling `expectation` on a `ContinuousUnivariateDistribution` raises the following error:
```julia
julia> Distributions.expectation(Normal(0.0, 1.0), identity, 1e-3)
ERROR: UndefVarError: integrate not defined
Stacktrace:
 [1] expectation(::Normal{Float64}, ::typeof(identity), ::Float64) at /Users/saxen/.julia/packages/Distributions/5UNuk/src/functionals.jl:11
 [2] top-level scope at REPL[6]:1
```

I'm no certain where `integrate` is supposed to have been defined.

Dsantra92:
Fixed #1175


andreasnoack:
The fix is correct. However, I didn't know we had this function defined and it's kind of odd that we define `expectation` when we also define `mean`.

Dsantra92:
The `expectation` function computes the expectation of function g(x) over the distribution  i.e. E[g(x)], mean is the expectation for only a unique case when g(x) is `identity`.

andreasnoack:
...i.e. the mean of `g(X)`.

Dsantra92:
" For a given distribution (let's say a Normal Distribution) , we can define the expectation of a function g(x) as **E[g(x)]**.The mean of the distribution is for g(x) = x *i.e.* E[x]. "

This the terminology I have been using. I find it difficult to see how to find expectation of function like `cos(x)` over any distribution using `mean`.

andreasnoack:
> I presume that mean and expectation are defined separately because various distributions have better way to computing mean that provides more accurate results than doing numerical integration over it.

You are probably right but it's hard to know. I can see that the `expectation` function was introduced seven years ago but it has never been documented. Again, your fix is correct so I'll merge this but I do think we should consider deleting this function.

Issue #1178, opened 2020-09-17
Suggestion for logdiffcdf return value when x == y


MFairley:
Currently `logdiffcdf` will return an error if `x == y` but since `cdf(x) - cdf(x) = 0`, I think it would make more sense to return `-Inf` rather than an error. Otherwise I need to add if statements to my code to deal with cases when `x` is equal to `y`.

```
julia> d = Beta(1, 1)
Beta{Float64}(α=1.0, β=1.0)

julia> logdiffcdf(d, 0.5, 0.5)
ERROR: ArgumentError: requires x > y.
Stacktrace:
 [1] logdiffcdf(::Beta{Float64}, ::Float64, ::Float64) at /Users/mfairley/.julia/packages/Distributions/5UNuk/src/univariates.jl:408
 [2] top-level scope at REPL[7]:1
```

matbesancon:
that would make sense on my end. pinging @mschauer for a second opinion

mschauer:
Yes, that makes sense.

MFairley:
#1180 closes this

PR #1180, opened 2020-09-21
return -Inf from logdiffcdf if x == y

MFairley:
#1178

### DIFF
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -405,7 +405,8 @@ The natural logarithm of the difference between the cumulative density function 
 function logdiffcdf(d::UnivariateDistribution, x::Real, y::Real)
     # Promote to ensure that we don't compute logcdf in low precision and then promote
     _x, _y = promote(x, y)
-    _x <= _y && throw(ArgumentError("requires x > y."))
+    _x == _y && return -Inf
+    _x < _y && throw(ArgumentError("requires x >= y."))
     u = logcdf(d, _x)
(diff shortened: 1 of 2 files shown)


mschauer:
This is of course our trick to draw people in but could you consider also adding a test to perhaps here:     
https://github.com/JuliaStats/Distributions.jl/blob/bb977ce300e8daa5f8a620d4ce2c4568c180984a/test/lognormal.jl#L14

src/univariates.jl
```diff
@@ -405,7 +405,8 @@ The natural logarithm of the difference between the cumulative density function
 function logdiffcdf(d::UnivariateDistribution, x::Real, y::Real)
     # Promote to ensure that we don't compute logcdf in low precision and then promote
     _x, _y = promote(x, y)
-    _x <= _y && throw(ArgumentError("requires x > y."))
+    _x == _y && return -Inf
```
mschauer, line 408:
```suggestion
```
Perhaps just wait for `u + log1mexp(v - u)` to produce an `Inf` (because `iszero(v-u)` which will be promoted to the right type?

mschauer (approved):

mschauer:
Any further onlookers/comments?

matbesancon:
reopening to see how CI goes on this

devmotion:
We should merge master and rerun tests before merging it. There were recent changes to `logcdf` etc.

PR #1181, opened 2020-09-22
dirname(@__FILE__) can be replaced with @__DIR__

petershintech:
Just replacement with a simpler expression.


andreasnoack:
Now that you are at it, it's actually better to resolve this dynamically with `pathof(Distributions)` since it works better if you want to relocate as is sometimes the case when you compile system images.

matbesancon:
merging this for now, we can improve to pathof(D) later

Issue #824, opened 2019-01-31
Add Gompertz distribution?


Deduction42:
This distribution is used extensively in the area of vital statistics. I love this package but I'm quite surprised that the Gompertz distribution isn't included here. Is it because the moments are so hard to calculate?

matbesancon:
@Deduction42 this would be fine to add, there is an analytical computation of major things, you can give it a try :)

tomguillon:
I'm working on reliability theory and I use a lot the Weibul and Gompertz distribution. I'd like to contribute (I want to switch from python), but for the Gompertz dsitribution we need some numerical integration to compute the moments (except for the mean value if the E1 special function exists in julia). @matbesancon What about implementing some generic methods like in scipy.stats :
- generic moment base on the numerical integration of x^m * pdf(x)
- generic fit maximum likelihood estimation using the pdf(x,params) or a reduce function and the Nelder-Mead method. We could also do better by computing the gradient wrt the parameters.

matbesancon:
The general rule we have tried to keep in Distributions is avoiding to have non-closed forms implemented here, which would rely on non-guaranteed or expensive methods (local optimization, numerical integration). That said, having something in the Distributions.jl package is not necessary to have it efficient or "first-class". You could create a package GompertzDistribution.jl package using Distributions

tomguillon:
Thank you @matbesson, I understand this choice but it seems quite restrictive for some distributions. For example, computing the moments of the distribution of Gompertz needs the [generalized integro-exponential function](https://www.ams.org/journals/mcom/1985-44-170/S0025-5718-1985-0777276-4/S0025-5718-1985-0777276-4.pdf) which are not implemented in most of scientific libraries. The same is true for the maximum likelihood estimator of the Weibull and Gompertz distributions. We can get robust and efficient numerical results with some works (computing gradient and hessian wrt of the negative log likelihood and using box constrained optimization algorithms), we do it all the days with my team due to the presence of censored data !
Keep me informed if this directive is likely to change.

matbesancon:
This could be generalized by regrouping all distributions depending on optimization schemes as you mention. In any case the trend for distributions is to keep it as light as possible for people to extend

tomguillon:
Ok, Here are some thoughts (I'm sorry if it's not the right place to discuss about that !):
The context: we are a team of 10 people working on survival analysis using python or R and I wanted to give a chance at julia because we have to implement many things for censored data either in R or Python.
[...]

PR #1182, opened 2020-09-22
Add Burr distribution

sparktseung:
I was looking for an implementation of Burr distribution in Julia but found none. Since Burr distribution is quite different from existing heavy-tail distributions included in Distribution.jl, I thought it may be helpful to just add it to this package.

### DIFF
--- a/src/univariate/continuous/burr.jl
+++ b/src/univariate/continuous/burr.jl
@@ -9,9 +9,6 @@ Burr()                  # Burr distribution with k = 1, c = 1 and λ = 1, i.e. B
 Burr(k, c)              # Burr distribution with λ = 1, i.e. Burr(k, c, 1)
 Burr(k, c, λ)           # Burr distribution with shape1 k, shape2 c and scale λ
 params(d)        # Get the parameters, i.e. (k, c, λ)
-shape1(d)        # Get the shape1 parameter, i.e. k
-shape2(d)        # Get the shape2 parameter, i.e. c
-scale(d)         # Get the scale parameter, i.e. λ
 ```
 External links
 * [Burr distribution on Wikipedia](https://en.wikipedia.org/wiki/Burr_distribution)
@@ -50,7 +47,7 @@ end
 # shape1 and shape2 not defined in the package -> ignore these. params is sufficient
 # shape1(d::Burr) = d.k
 # shape2(d::Burr) = d.c
-scale(d::Burr) = d.λ
+# scale(d::Burr) = d.λ
 params(d::Burr) = (d.k, d.c, d.λ)
 partype(::Burr{T}) where {T<:Real} = T
 
--- a/src/univariate/continuous/burr.jl
+++ b/src/univariate/continuous/burr.jl
@@ -47,8 +47,9 @@ end
 
 #### Parameters
 
-shape1(d::Burr) = d.k
-shape2(d::Burr) = d.c
+# shape1 and shape2 not defined in the package -> ignore these. params is sufficient
+# shape1(d::Burr) = d.k
+# shape2(d::Burr) = d.c
 scale(d::Burr) = d.λ
 params(d::Burr) = (d.k, d.c, d.λ)
 partype(::Burr{T}) where {T<:Real} = T
@@ -130,7 +131,9 @@ end
 #### Evaluation
 
 function pdf(d::Burr{T}, x::Real) where T<:Real
-    if x > 0
+    if isinf(x)
+        return zero(T)
+    elseif x > 0
         (k, c, λ) = params(d)
         return k*c/λ * (x/λ)^(c-1) * ( 1 + (x/λ)^c )^(-k-1)
     else
@@ -139,7 +142,9 @@ function pdf(d::Burr{T}, x::Real) where T<:Real
 end
 
 function logpdf(d::Burr{T}, x::Real) where T<:Real
-    if x > 0
+    if isinf(x)
+        return -T(Inf)
+    elseif x > 0
         (k, c, λ) = params(d)
         return log(k) + log(c) - log(λ) + (c-1)*log(x) - (c-1)*log(λ) - (k+1)*log1p( (x/λ)^c )
     else
--- a/src/univariate/continuous/burr.jl
+++ b/src/univariate/continuous/burr.jl
@@ -55,11 +55,11 @@ partype(::Burr{T}) where {T<:Real} = T
 
 #### Statistics
 
-function m(d::Burr, g::Real) # The calculating moments: this should not be exported.
+function m(d::Burr{T}, g::Real) where T<:Real # The calculating moments: this should not be exported.
     (k, c, λ) = params(d)
 
     if (g-(-c))*(g-k*c) < 0
-        λ^g * gamma(1 + g*c) * gamma(k - g/c) / gamma(k)
+        λ^g * gamma(1 + g/c) * gamma(k - g/c) / gamma(k)
     else
         return T(Inf)
     end
--- a/src/univariate/continuous/burr.jl
+++ b/src/univariate/continuous/burr.jl
@@ -1,0 +1,192 @@
+"""
+    Burr(k, c, λ)
+The *Burr distribution*  with shape1 `k`, shape2 `c` and scale `λ` has probability density function
+```math
+f(x; k, c, \\lambda) = \\frac{ck}{\\lambda} \\left(\\frac{x}{\\lambda}\\right)^{c-1} \\left[ 1+\\left(\\frac{x}{\\lambda}\\right)^{c} \\right]^{-k-1}, \\quad x > 0
+```
+```julia
+Burr()                  # Burr distribution with k = 1, c = 1 and λ = 1, i.e. Burr(1, 1, 1)
+Burr(k, c)              # Burr distribution with λ = 1, i.e. Burr(k, c, 1)
+Burr(k, c, λ)           # Burr distribution with shape1 k, shape2 c and scale λ
+params(d)        # Get the parameters, i.e. (k, c, λ)
+shape1(d)        # Get the shape1 parameter, i.e. k
+shape2(d)        # Get the shape2 parameter, i.e. c
+scale(d)         # Get the scale parameter, i.e. λ
+```
+External links
+* [Burr distribution on Wikipedia](https://en.wikipedia.org/wiki/Burr_distribution)
+"""
+
+
+struct Burr{T<:Real} <: ContinuousUnivariateDistribution
+    k::T # shape1
+    c::T # shape2
+    λ::T # scale
+end
+
+function Burr(k::T, c::T, λ::T; check_args=true) where {T}
+    check_args && @check_args(Burr, k > zero(k) && c > zero(c) && λ > zero(λ))
+    return Burr{T}(k, c, λ)
+end
+
+Burr(k::Real, c::Real, λ::Real) = Burr(promote(k, c, λ)...)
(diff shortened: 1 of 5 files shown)


azev77:
@sparktseung this is awesome!
Is it possible to add options to fit this distribution?

sparktseung:
Sure, I can also implement a `fit_mle` function. There is one concern:

The MLE requires Newton's method on 3 equations simultaneously. I see an implementation of Newton's method in `fit_mle` for `Gamma`, so I assume computation cost at this level is acceptable, is that correct?

I am asking because I read [here](https://github.com/JuliaStats/Distributions.jl/issues/824#issuecomment-512175048) about avoiding certain expensive numerical procedures.

azev77:
@sparktseung  
I think this is acceptable, especially since it's done in [Gamma](https://github.com/JuliaStats/Distributions.jl/blob/master/src/univariate/continuous/gamma.jl).jl.

We might start encountering challenges if you need to add dependencies to do complicated numerical integration etc...

azev77:
@sparktseung 
1. you may enjoy @gecrooks [Field Guide](https://threeplusone.com/pubs/FieldGuide.pdf) 
2. By adding a `shift` parameter `a`, the Burr distribution can be [slightly generalized](https://threeplusone.com/pubs/FieldGuide.pdf#244) 
image
3. By adding parameter `α`, the shifted-Burr distribution can be greatly generalized to the [Generalized Beta Prime](https://threeplusone.com/pubs/FieldGuide.pdf#1ca) 
image
This automatically gives 16+ distributions as a special case ;)
image

sparktseung:
@azev77 Thanks for the reference! I will write a more general version. I will also get the `fit_mle` done sometime.

mschauer:
Perhaps we were too greedy... Should we settle for the Burr distribution for now - what are your plans, @sparktseung ?

src/univariate/continuous/burr.jl
```diff
@@ -0,0 +1,194 @@
+"""
+    Burr(k, c, λ)
```
devmotion, line 2:
```suggestion
    Burr(k, c, λ)

```

src/univariate/continuous/burr.jl
```diff
@@ -0,0 +1,194 @@
...
+The *Burr distribution*  with shape1 `k`, shape2 `c` and scale `λ` has probability density function
```
devmotion, line 3: Maybe
```suggestion
The *Burr distribution*  with shapes `k` and `c` and scale `λ` has probability density function
```

src/univariate/continuous/burr.jl
```diff
@@ -0,0 +1,194 @@
...
+    λ::T # scale
+end
+
+function Burr(k::T, c::T, λ::T; check_args=true) where {T}
+    check_args && @check_args(Burr, k > zero(k) && c > zero(c) && λ > zero(λ))
+    return Burr{T}(k, c, λ)
+end
+
+Burr(k::Real, c::Real, λ::Real) = Burr(promote(k, c, λ)...)
+Burr(k::Integer, c::Integer, λ::Integer) = Burr(float(k), float(c), float(λ))
```
devmotion, line 30: I don't think this is needed, integer-valued parameters should work as well:
```suggestion
```

src/univariate/continuous/burr.jl
```diff
@@ -0,0 +1,194 @@
...
+Burr(k::T, c::T) where {T <: Real} = Burr(k, c, one(T))
```
devmotion, line 31: It could be useful to allow also to specify `k` and `c` of different types:
```suggestion
Burr(k::T, c::T) where {T <: Real} = Burr(k, c, one(T))
Burr(k::Real, c::Real) = Burr(promote(k, c)...)
```

src/univariate/continuous/burr.jl
```diff
@@ -0,0 +1,194 @@
...
+Burr() = Burr(1.0, 1.0, 1.0, check_args=false)
```
devmotion, line 32: One could maybe even define

```suggestion
Burr() = Burr(1, 1, 1; check_args=false)
```

I also wonder if one should define a constructor with only `k`.

src/univariate/continuous/burr.jl
```diff
@@ -0,0 +1,194 @@
...
+
+@distr_support Burr 0.0 Inf
+
+#### Conversions
+
+function convert(::Type{Burr{T}}, k::S, c::S, λ::S) where {T <: Real, S <: Real}
```
devmotion, line 38: I guess it is not necessary that the parameters are of the same type:

```suggestion
function convert(::Type{Burr{T}}, k::Real, c::Real, λ::Real) where {T<:Real}
```

src/univariate/continuous/burr.jl
```diff
@@ -0,0 +1,194 @@
...
+function convert(::Type{Burr{T}}, d::Burr{S}) where {T <: Real, S <: Real}
+    Burr(T(d.k), T(d.c), T(d.λ), check_args=false)
+end
+
+#### Parameters
+
+# shape1 and shape2 not defined in the package -> ignore these. params is sufficient
+# shape1(d::Burr) = d.k
+# shape2(d::Burr) = d.c
+# scale(d::Burr) = d.λ
```
devmotion, line 50: `scale` could be defined though:

```suggestion
scale(d::Burr) = d.λ
```
And one could even define `shape`:

```suggestion
shape(d::Burr) = (d.k, d.c)
scale(d::Burr) = d.λ
```

src/univariate/continuous/burr.jl
```diff
@@ -0,0 +1,194 @@
...
+params(d::Burr) = (d.k, d.c, d.λ)
+partype(::Burr{T}) where {T<:Real} = T
+
+#### Statistics
+
+function m(d::Burr{T}, g::Real) where T<:Real # The calculating moments: this should not be exported.
```
devmotion, line 56: Maybe use a more descriptive name? Also it would be good to not base the result on `T` but instead always perform as many computations as possible and then just return `oftype(result, Inf)` in the other branch.

src/univariate/continuous/burr.jl
```diff
@@ -0,0 +1,194 @@
...
+    (k, c, λ) = params(d)
```
devmotion, line 57:
```suggestion
    k, c, λ = params(d)
```

src/univariate/continuous/burr.jl
```diff
@@ -0,0 +1,194 @@
...
+
+    if (g-(-c))*(g-k*c) < 0
+        λ^g * gamma(1 + g/c) * gamma(k - g/c) / gamma(k)
+    else
+        return T(Inf)
+    end
+
```
devmotion, line 64: Eg here it would be better to write

```suggestion
    result = λ^g * gamma(1 + g/c) * gamma(k - g/c) / gamma(k)
    return if (g + c)*(g - k*c) < 0
        result
    else
        oftype(result, Inf)
    end
```

src/univariate/continuous/burr.jl
```diff
@@ -0,0 +1,194 @@
...
+end
+
+
+function median(d::Burr{T}) where T<:Real
+    (k, c, λ) = params(d)
+
```
devmotion, line 70:
```suggestion
function median(d::Burr)
    k, c, λ = params(d)
```

src/univariate/continuous/burr.jl
```diff
@@ -0,0 +1,194 @@
...
+    return λ * (2^(1/k) - 1)^(1/c)
+end
+
+function mean(d::Burr{T}) where T<:Real
+    return m(d, 1)
+end
```
devmotion, line 76:
```suggestion
mean(d::Burr) = m(d, 1)
```
As said above, a more descriptive name for `m` would be good.

src/univariate/continuous/burr.jl
```diff
@@ -0,0 +1,194 @@
...
+
+function mode(d::Burr{T}) where T<:Real
+    (k, c, λ) = params(d)
+
```
devmotion, line 80:
```suggestion
function mode(d::Burr)
    k, c, λ = params(d)
```

src/univariate/continuous/burr.jl
```diff
@@ -0,0 +1,194 @@
...
+    if c > 1.0
+        return λ * ( (c-1) / (k*c+1) )^(1/c)
+    else
+        return 0.0
+    end
```
devmotion, line 85:
```suggestion
    cmax = max(c, 1)
    result = λ * ((cmax - 1)/(k*c + 1))^(1/c)
    return if c > 1
        result
    else
        zero(result)
    end
```

src/univariate/continuous/burr.jl
```diff
@@ -0,0 +1,194 @@
...
+end
+
+function var(d::Burr{T}) where T<:Real
```
devmotion, line 88: Similar here and in the functions below.

sparktseung:
@mschauer  Apologies for the delay. I've been sidetracked by other projects. Yes, I agree settling for Burr for the moment,

@devmotion Thanks for the comments! I will make the changes in a few days, and then commit again.

src/univariate/continuous/burr.jl
sparktseung: Changed as commented.

src/univariate/continuous/burr.jl
sparktseung: Changed as commented.

src/univariate/continuous/burr.jl
sparktseung: Changed as commented.

src/univariate/continuous/burr.jl
sparktseung: Changed as commented. I also added a constructor with only `k`.

src/univariate/continuous/burr.jl
sparktseung: Changed as commented.

src/univariate/continuous/burr.jl
sparktseung: Changed as commented.

src/univariate/continuous/burr.jl
sparktseung: Changed as commented.

src/univariate/continuous/burr.jl
sparktseung: Changed as commented.

src/univariate/continuous/burr.jl
sparktseung: Changed as commented. 

Calculating `result = λ^g * gamma(1 + g/c) * gamma(k - g/c) / gamma(k)` would sometimes result in `NaN`, e.g. `k-g/c<0` then the gamma function gets an error. I used `max(0, k-g/c)` first.

src/univariate/continuous/burr.jl
sparktseung: Changed as commented.

src/univariate/continuous/burr.jl
sparktseung: Changed as commented.

src/univariate/continuous/burr.jl
sparktseung: Changed as commented.

src/univariate/continuous/burr.jl
sparktseung: Changed as commented.

src/univariate/continuous/burr.jl
sparktseung: Changed as commented.

Issue #807, opened 2019-01-08
fit_mle for Laplace with weights


yakir12:
In [the docs](http://juliastats.github.io/Distributions.jl/latest/fit.html#Applicable-distributions-1) it looks like fitting Laplace to a dataset with weights would work, but: 
```julia
n = 100
x = rand(n)
w = rand(n)
fit_mle(Normal, x, w) # works
fit_mle(Laplace, x, w) # doesn't work
```
errors out with:
```julia
ERROR: suffstats is not implemented for (Laplace, Array{Float64,1}, Array{Float64,1}).                 
Stacktrace:
 [1] suffstats(::Type{Laplace}, ::Array{Float64,1}, ::Array{Float64,1}) at /home/yakir/.julia/packages/Distributions/yz3Bc/src/genericfit.jl:5
 [2] fit_mle(::Type{Laplace}, ::Array{Float64,1}, ::Array{Float64,1}) at /home/yakir/.julia/packages/Distributions/yz3Bc/src/genericfit.jl:28
 [3] top-level scope at none:0
```
I didn't check all the other distributions but there are probably more that don't work with weights. A quick fix might be to change the docs to "only the Normal distribution can be fitted with weights".

matbesancon:
PR welcome if this is a simple doc change :)

PR #1187, opened 2020-10-01
Removed Laplace from the list of fit.md

Dsantra92:
As requested in https://github.com/JuliaStats/Distributions.jl/issues/807#, I have removed the Laplace from the list of distribution from the list in fit.md .

### DIFF
--- a/docs/src/fit.md
+++ b/docs/src/fit.md
@@ -72,9 +72,9 @@ fit_mle(Categorical, k, x, w)
 fit_mle(Categorical, x)        # equivalent to fit_mle(Categorical, max(x), x)
 fit_mle(Categorical, x, w)
 ```
-!!! note
+!!! warning
 
-    Laplace distribution currently does not support weighted fit_mle.
+     Laplace distribution currently does not support weighted fit_mle.
 ## Sufficient Statistics
 
 For many distributions, estimation can be based on (sum of) sufficient statistics computed from a dataset. To simplify implementation, for such distributions, we implement `suffstats` method instead of `fit_mle` directly:
--- a/docs/src/fit.md
+++ b/docs/src/fit.md
@@ -75,6 +75,7 @@ fit_mle(Categorical, x, w)
 !!! warning
 
      Laplace distribution currently does not support weighted `fit_mle`.
+     
 ## Sufficient Statistics
 
 For many distributions, estimation can be based on (sum of) sufficient statistics computed from a dataset. To simplify implementation, for such distributions, we implement `suffstats` method instead of `fit_mle` directly:
--- a/docs/src/fit.md
+++ b/docs/src/fit.md
@@ -74,7 +74,7 @@ fit_mle(Categorical, x, w)
 ```
 !!! warning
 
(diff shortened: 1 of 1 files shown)


Dsantra92:
On another note Laplace supports `fit_mle(dist, x)` and does not support `fit_mle(dist, x, w)`. I would like to ask if I could add the `fit_mle(dist,x,w)` for Laplace in a future PR.

andreasnoack:
I don't think removing it from the list is the way to go here. As you mention, `fit_mle(dist, x)` works so the right thing seems to be to clarify which of the distributions support the weight option in `fit_mle`

Dsantra92:
@andreasnoack  Can I just make a PR to add weighted_fit_mle for Laplacian ? I guess that will resolve the issue.

andreasnoack:
Yes. That would be even better.

Dsantra92:
Is there any weighted MAD in Julia? 
Laplace weighted `fit_mle` requires weighted MAD and I found no implementation in StatsBase.jl. I implemented my own MAD and realized it's too messy for a single function to holdv(which is not even intended to it). I actually see 3 options now :

1. Implement weighted MAD in StatsBase.jl and then use it.
2. Implement weighted map in the `fit_mle` function itself.
3. Or as you suggested, just mention that the weighted `fit_mle` is not implemented in the package.

docs/src/fit.md
```diff
@@ -72,7 +72,9 @@ fit_mle(Categorical, k, x, w)
 fit_mle(Categorical, x)        # equivalent to fit_mle(Categorical, max(x), x)
 fit_mle(Categorical, x, w)
 ```
+!!! warning
 
+     Laplace distribution currently does not support weighted `fit_mle`.
```
matbesancon, line 77:
```suggestion
     Laplace distribution currently does not support weighted `fit_mle`.
     
```

Dsantra92: Pardon, but aren't both the lines same?

matbesancon: this is just adding a new line before the next header

docs/src/fit.md
```diff
@@ -47,7 +47,7 @@ The `fit_mle` method has been implemented for the following distributions:
 - [`Normal`](@ref)
 - [`Gamma`](@ref)
 - [`Geometric`](@ref)
-- [`Laplace`](@ref)
+- [`Laplace`](@ref)
```
matbesancon, line 50:
```suggestion
- [`Laplace`](@ref)
```

devmotion:
IIRC this was fixed recently.

devmotion:
Ah no, the unweighted `fit` was fixed (https://github.com/JuliaStats/Distributions.jl/pull/1309) and a PR for the weighted `fit` was closed (https://github.com/JuliaStats/Distributions.jl/pull/1310) but is planned to be updated (possibly in a new PR) after https://github.com/JuliaStats/StatsBase.jl/pull/687 is fixed. So depending on how timely the weighted `fit` is implemented, the warning in this PR might not be needed.

PR #1191, opened 2020-10-07
Stop testing PoissonADSampler for μ=5.0

andreasnoack:
since it's not accurate for too small values.

Initialize arrays in fit_mle for Dirichlet.

Add debug keyword to fit_mle for Dirichlet and fix debug branches
in `fit_dirichlet!`

The initialization bug was the one that caused the test failure in https://github.com/JuliaStats/Distributions.jl/pull/1164 on 32 bit. It's a bit surprising that we haven't hit it more frequently since the initial values were often completely off because of the this, e.g.
```julia
julia> Distributions.dirichlet_mle_init(x)
3-element Array{Float64,1}:
 -0.0
 -0.0
 -0.0
julia> Distributions.dirichlet_mle_init(x)
3-element Array{Float64,1}:
  5.158988734525853e104
  8.769340308733536e104
 -0.16535356495305348
julia> Distributions.dirichlet_mle_init(x)
3-element Array{Float64,1}:
 5.040668036024078e103
 8.579050171199378e103
 3.165512062017993e197
julia> Distributions.dirichlet_mle_init(x)
3-element Array{Float64,1}:
 -1.9279331031482295e199
 -1.9280754428163168e199
  0.9840394381466419
```
It looks like the bug has been there fore seven years so I don't think we have many users of `fit_mle` for `Dirichlet`.

### DIFF
--- a/src/multivariate/dirichlet.jl
+++ b/src/multivariate/dirichlet.jl
@@ -266,22 +266,12 @@ function dirichlet_mle_init(P::AbstractMatrix{Float64})
     K = size(P, 1)
     n = size(P, 2)
 
-    μ = Vector{Float64}(undef, K)  # E[p]
-    γ = Vector{Float64}(undef, K)  # E[p^2]
-
-    for i = 1:n
-        for k = 1:K
-            @inbounds pk = P[k, i]
-            @inbounds μ[k] += pk
-            @inbounds γ[k] += pk * pk
-        end
-    end
+    μ = vec(sum(P, dims=2))       # E[p]
+    γ = vec(sum(abs2, P, dims=2)) # E[p^2]
 
     c = 1.0 / n
-    for k = 1:K
-        μ[k] *= c
-        γ[k] *= c
-    end
+    μ .*= c
+    γ .*= c
 
     _dirichlet_mle_init2(μ, γ)
 end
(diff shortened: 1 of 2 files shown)


src/multivariate/dirichlet.jl
```diff
@@ -265,23 +265,13 @@ end
...
-            @inbounds γ[k] += pk * pk
-        end
-    end
-
-    c = 1.0 / n
-    for k = 1:K
-        μ[k] *= c
-        γ[k] *= c
-    end
+    μ .*= c
```
mschauer, line 273: change to
```
 μ .*= c
```
also below for consistency?

src/multivariate/dirichlet.jl
```diff
@@ -381,26 +371,32 @@ function fit_dirichlet!(elogp::Vector{Float64}, α::Vector{Float64};
         converged = gnorm < tol
     end
 
+    if !converged
+        throw(ErrorException("no convergence after $maxiter (maxiter) iterations"))
```
mschauer, line 375:
```suggestion
        throw(ErrorException("No convergence after $maxiter (maxiter) iterations."))
```

andreasnoack: In Julia Base, error messages typically don't start will a capital and the don't end with a period. It looks like we are pretty inconsistent here and I'll file an issues about it.

mschauer: I just saw the style of the one below

andreasnoack: Fair point. Also one like that above it. Within function error message consistency seems like a pretty low requirement.

PR #1193, opened 2020-10-08
MvNormalCanon constructor for sparse precision matrix

ElOceanografo:
Just a simple convenience constructor--I got tired of typing `MvNormalCanon(PDSparseMat(J))` : )


src/multivariate/mvnormalcanon.jl
```diff
@@ -112,10 +113,14 @@ function MvNormalCanon(h::AbstractVector{T}, J::P) where {T<:Real, P<:AbstractPD
     MvNormalCanon{eltype(hh),typeof(JJ),typeof(hh)}(JJ \ hh, hh, JJ)
 end
 
+MvNormalCanon(μ::AbstractVector{<:Real}, h::AbstractVector{<:Real}, J::AbstractSparseMatrix{<:Real}) = MvNormalCanon(μ, PDSparseMat(J))
```
andreasnoack, line 116: These definitions would have to go behind a `isdefined` flag to avoid breaking builds without the sparse libraries.

PR #1194, opened 2020-10-11
Add UniformSpherical and UniformBall distributions and samplers

emerali:
Pretty self-explanatory. 

One thing I'm unsure about though is the `partype` function. Since these two distributions don't actually have parameters, what should I be returning? Or should I just leave it undefined?


johnczito:
Do we want `UniformSpherical` to be its own separate thing, or should we just be extending `VonMisesFisher ` to behave in the `κ = 0` case? I think I prefer the latter, but it's open.

emerali:
I feel like they're useful enough on their own to warrant a separate type.

PR #1195, opened 2020-10-13
Implement quantile for MixtureModel{Univariate,Continuous}

nignatiadis:
This implements `quantile` for `MixtureModel{Univariate,Continuous}`. I am using the `quantile_bisect` function with starting range determined based on the component distributions.

Right now on master e.g., the following fails
```julia
julia> mix = MixtureModel([Normal(0.0, 1.0); Normal(0.0, 2.0 )], [0.4;0.6])
julia> median(mix)
ERROR: MethodError: no method matching iterate(::MixtureModel{Univariate,Continuous,Normal{Float64},Float64})
```
After this PR:
```julia
julia> median(mix)
0.0
```

### DIFF
--- a/src/mixtures/mixturemodel.jl
+++ b/src/mixtures/mixturemodel.jl
@@ -500,6 +500,9 @@ componentwise_logpdf(d::MultivariateMixture, x::AbstractVector) = componentwise_
 componentwise_logpdf(d::MultivariateMixture, x::AbstractMatrix) = componentwise_logpdf!(Matrix{eltype(x)}(undef, size(x,2), ncomponents(d)), d, x)
 
 
+_default_tol(p::Real) = 1e-12
+_default_tol(p::AbstractFloat) = 10eps(eltype(p))
+
 function quantile(d::UnivariateMixture{Continuous}, p::Real)
     ps = probs(d)
     min_q, max_q = extrema(quantile(component(d, i), p) for (i, pi) in enumerate(ps) if pi > 0)
@@ -510,7 +513,15 @@ function quantile(d::UnivariateMixture{Continuous}, p::Real)
         return(max_q)
     end
 
-    quantile_bisect(d, p, min_q - 0.01*abs(min_q), max_q + 0.01*abs(max_q), 1.0e-12)
+    tol =  _default_tol(p)
+
+    if abs(cdf(d, min_q) - p) < tol
+        return(min_q)
+    elseif abs(cdf(d, max_q) - p) < tol
+        return(max_q)
+    end
+
+    quantile_bisect(d, p, min_q, max_q, tol)
 end
 
 ## Sampling
--- a/src/mixtures/mixturemodel.jl
+++ b/src/mixtures/mixturemodel.jl
@@ -500,6 +500,21 @@ componentwise_logpdf(d::MultivariateMixture, x::AbstractVector) = componentwise_
 componentwise_logpdf(d::MultivariateMixture, x::AbstractMatrix) = componentwise_logpdf!(Matrix{eltype(x)}(undef, size(x,2), ncomponents(d)), d, x)
 
 
+function quantile(d::UnivariateMixture{Continuous}, p::Real)
+    ps = probs(d)
+    cs = [component(d,i) for i in eachindex(ps) if ps[i] > 0.0]
+    qs = quantile.(cs, p)
+    min_q, max_q = extrema(qs)
+
+    if iszero(p)
+        return(min_q)
+    elseif isone(p)
+        return(max_q)
+    end
+
+    quantile_bisect(d, p, min_q, max_q, 1.0e-12)
+end
+
 ## Sampling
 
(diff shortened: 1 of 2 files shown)


src/mixtures/mixturemodel.jl
```diff
@@ -500,6 +500,21 @@ componentwise_logpdf(d::MultivariateMixture, x::AbstractVector) = componentwise_
 componentwise_logpdf(d::MultivariateMixture, x::AbstractMatrix) = componentwise_logpdf!(Matrix{eltype(x)}(undef, size(x,2), ncomponents(d)), d, x)
 
 
+function quantile(d::UnivariateMixture{Continuous}, p::Real)
+    ps = probs(d)
+    cs = [component(d,i) for i in eachindex(ps) if ps[i] > 0.0]
+    qs = quantile.(cs, p)
+    min_q, max_q = extrema(qs)
```
devmotion, line 507: An alternative without allocating `cs` and `qs`:
```suggestion
    min_q, max_q = extrema(quantile(component(d, i), p) for i in 1:ncomponents(d) if ps[i] > 0)
```

src/mixtures/mixturemodel.jl
```diff
@@ -500,6 +500,21 @@ componentwise_logpdf(d::MultivariateMixture, x::AbstractVector) = componentwise_
...
+
+    if iszero(p)
+        return(min_q)
+    elseif isone(p)
+        return(max_q)
+    end
+
+    quantile_bisect(d, p, min_q - 0.01*abs(min_q), max_q + 0.01*abs(max_q), 1.0e-12)
```
devmotion, line 515: It seems this could be type unstable?

nignatiadis:
@devmotion Thank you a lot; that was super helpful!! I cannot believe how much I learned about Julia from your review of a 10 line PR.

Regarding type instability, I went back to the code to make sure the following returns a `Float32`:
```julia
mix_ffff = MixtureModel([Normal(0.0f0, 1.0f0); Normal(0.0f0, 2.0f0 )], [0.4f0;0.6f0])
quantile(mix_ffff, 0.1f0)
```
After my first attempt at fixing the type instability you pointed out, I realized that the problem is more fundamental: The bisection code went into an infinite loop, because `tol=1e-12` could never be achieved with `Float32`. Let me know if you think the new approach is reasonable or if there is a better way to handle it.

nignatiadis:
@devmotion I just updated this again. Pinging just in case you have time to (re)review it! Thanks :D

src/mixtures/mixturemodel.jl
```diff
@@ -500,6 +500,27 @@ componentwise_logpdf(d::MultivariateMixture, x::AbstractVector) = componentwise_
...
+    ps = probs(d)
+    min_q, max_q = extrema(quantile(component(d, i), p) for (i, pi) in enumerate(ps) if pi > 0)
+
+    if iszero(p)
+        return(min_q)
+    elseif isone(p)
+        return(max_q)
+    end
+
+    tol =  Base.rtoldefault(typeof(p)) * abs(p)
```
devmotion, line 513: The tolerance refers to the support of the distribution, not the quantile level (see https://github.com/JuliaStats/Distributions.jl/blob/d6de7214e6d84df77d1eb3ec7a98d23862d1232d/src/quantilealgs.jl#L10). I.e., it should be based on `min_q` and `max_q` (e.g. `max(rtoldefault(typeof(min_q)),rtoldefault(typeof(max_q)))` as in Base).

If one wants to use a type-dependent relative tolerance (which IMO seems reasonable since it will give more sensible tolerances for e.g. Float32), it would be even better if it is not converted to an absolute tolerance here - the values of the initial bracket might be ridiculously large which would lead to an unnecessarily large tolerance. Instead the comparison in https://github.com/JuliaStats/Distributions.jl/blob/d6de7214e6d84df77d1eb3ec7a98d23862d1232d/src/quantilealgs.jl#L10 could be based on the scale of `lx` and `rx`.

src/mixtures/mixturemodel.jl
```diff
@@ -500,6 +500,27 @@ componentwise_logpdf(d::MultivariateMixture, x::AbstractVector) = componentwise_
...
+
+    if abs(cdf(d, min_q) - p) < tol
+        return(min_q)
+    elseif abs(cdf(d, max_q) - p) < tol
+        return(max_q)
+    end
```
devmotion, line 519: I guess it would be reasonable to handle this in the general algorithm in https://github.com/JuliaStats/Distributions.jl/blob/d6de7214e6d84df77d1eb3ec7a98d23862d1232d/src/quantilealgs.jl#L3-L22.

devmotion:
I'll close this PR in favour of https://github.com/JuliaStats/Distributions.jl/pull/1389 which is an updated and slightly modified version of your PR.

nignatiadis:
That is awesome @devmotion! Thank you for pushing this forward, as well as reviewing my code twice. I learned a lot!

PR #1199, opened 2020-10-16
Make new unwhiten_winv! conditional on PDSparseMat

andreasnoack:
The `PDSparseMat` struct is defined behind a `HAVE_CHOLMOD` flag which is false when people build Julia without GPL libraries, see https://github.com/JuliaStats/PDMats.jl/blob/d50e35a37d218f0e6ac552512bc4c79b102cc930/src/PDMats.jl#L51-L53. Hence, https://github.com/JuliaStats/Distributions.jl/pull/855 broke Distributions when Julia has been built without GPL libraries.

cc @ElOceanografo


ElOceanografo:
Looks okay to me...so if I understand correctly, if someone built Julia without CHOLMOD, trying to sample from a sparse `MvNormalCanon` would make `unwhiten_winv!` fall back on the method for an `AbstractPDMat` (and thus throw an error when it tries to invert the precision matrix, cf. https://github.com/JuliaStats/PDMats.jl/pull/119)?

andreasnoack:
> ...so if I understand correctly, if someone built Julia without CHOLMOD, trying to sample from a sparse MvNormalCanon would make unwhiten_winv! fall back on the method for an AbstractPDMat (and thus throw an error when it tries to invert the precision matrix

Yes, I'd think so. It's a bit unfortunate that we don't have a CI setup for Julia builds without SuiteSparse (CHOLMOD) but on the other hand, it's quite niche since this package is GPL because of the Rmath dependency and people usually disable SuiteSparse to get a GPL free Julia build.

PR #1204, opened 2020-10-23
Add Zipf-Mandelbrot law

mahdisarikhani:
Add Zipf-Mandelbrot law based on:
[Zipf-Mandelbrot law on Wikipedia](https://en.wikipedia.org/wiki/Zipf%E2%80%93Mandelbrot_law)

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -163,6 +163,7 @@ export
     ZeroMeanDiagNormalCanon,
     ZeroMeanFullNormal,
     ZeroMeanFullNormalCanon,
+    ZipfMandelbrot,
 
     # auxiliary types
     RealInterval,
@@ -330,7 +331,7 @@ Supported distributions:
     VonMises, VonMisesFisher, WalleniusNoncentralHypergeometric, Weibull,
     Wishart, ZeroMeanIsoNormal, ZeroMeanIsoNormalCanon,
     ZeroMeanDiagNormal, ZeroMeanDiagNormalCanon, ZeroMeanFullNormal,
-    ZeroMeanFullNormalCanon
+    ZeroMeanFullNormalCanon, ZipfMandelbrot
 
 """
 Distributions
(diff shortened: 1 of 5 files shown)


azev77:
@mahdisarikhani thanks for this PR! 
A few comments you may (or may not) wanna consider:
1. currently `ZipfMandelbrot(N, q, s)` requires N to be an integer-type. 
`ZipfMandelbrot(1., 0., 1.)` would give an error
`ZipfMandelbrot(Int(1.0), 0., 1.)` works
2. The support is currently missing. I think it's something like
`@distr_support ZipfMandelbrot 1 d.N`
support(d)
3. If  N<Inf & q=0 Zipf–Mandelbrot law becomes Zipf's law. Which was [previously requested](https://github.com/JuliaStats/Distributions.jl/issues/124).
Perhaps something like:
`Zipf(N::Int, s::Real) = ZipfMandelbrot(N::Int, 0.0, s::Real)`
You prob also wanna create a new type etc, the maintainers can guide you
4. If N=Inf & q = 0  it becomes a Zeta distribution. This uses the function `zeta()` part of SpecialFunctions.jl, which other distributions here already use. 
Maybe something like:
`Zeta(s::Real) = ZipfMandelbrot(Inf, 0.0, s::Real)`
`H(N::Int, q, s) = sum(1 / (i+q)^s for i = 1:N)`
`H(N, q, s) = zeta(s)`
5. Using Mathematica, I can get the variance & other properties in closed form:
`var(d::ZipfMandelbrot) = H(d.N, d.q, d.s-1) / H(d.N, d.q, d.s) - d.q`
```julia
var(d::ZipfMandelbrot) = 
    (
    -(zeta(-1+s,1+q)-zeta(-1+s,1+N+q))^2
    +(zeta(-2+s,1+q)-zeta(-2+s,1+N+q)) 
    *(zeta(s,1+q)-zeta(s,1+N+q))
    )/(zeta(s,1+q)-zeta(s,1+N+q))^2
# 
var(ZipfMandelbrot(10, 0.0, 5.0))
```

mahdisarikhani:
@azev77 thanks for your comments!
I will consider your comments for the next commits. There is one concern, should I implement `Zipf` and `Zeta` as new types or implement as functions that point to `ZipfMandelbrot`?

azev77:
Zipf & Zeta should prob have their own types. 
Check w/ @johnczito @matbesancon @mschauer

azev77:
@mahdisarikhani why did you close this issue?

PR #1206, opened 2020-10-23
relax qqbuild signature

### LINKED ISSUE #1196: widen qqbuild signature?
I've just noticed that `qqbuild` has a [very tight signature](https://github.com/JuliaStats/Distributions.jl/blob/master/src/qq.jl#L6) and errors with e.g. a `SubArray`. Would it be possible to widen it? (Happy to make a PR if it is.)

I'm asking because, from what I understood, it is the interface to create `QQPair` and AbstractPlotting and StatsPlots use it for QQPlots, but it currently errors if one doesn't pass a concrete `Array`.

piever:
Fixes #1196. I'm not completely sure whether the type parameter of `QQPair` should be widened to `AbstractVector` or `AbstractVector{<:Real}`. I chose the first option for simplicity.

### DIFF
--- a/src/qq.jl
+++ b/src/qq.jl
@@ -1,9 +1,9 @@
-struct QQPair
-    qx::Vector{Float64}
-    qy::Vector{Float64}
+struct QQPair{U<:AbstractVector, V<:AbstractVector}
+    qx::U
+    qy::V
 end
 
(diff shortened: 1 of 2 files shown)


piever:
Bump? (Do let me know if this needs anything else to be merged)

PR #1207, opened 2020-10-26
Bump patch and FillArrays minor version

willtebbutt:
Supercedes #1205 so that patch version is bumped and new release can be made. @mschauer

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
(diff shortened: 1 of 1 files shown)


mschauer (approved):

willtebbutt:
@matbesancon any chance you could merge this? Don't think that either of us have commit rights.

mschauer:
Ah, sorry, I have commit rights but I avoid tagging new versions because I don't have a good mental image of all the things one has to keep in mind.

matbesancon:
there is not much to keep, if the project.toml version is already changed, one can tag the new release

PR #1212, opened 2020-10-31
Update skewnormal.jl

azev77:
Adds CDF for SkewNormal() using `quadgk` which is currently used by other distributions in Distributions.jl.


azev77:
@devmotion are we good?

PR #1215, opened 2020-11-08
bump patch number

piever:
Would it be possible to merge this and tag a new patch release? I would like to use the fix from #1206 in another package.


andreasnoack:
Do you know if the struct is used downstream? There is a small risk that the change could be breaking but I'd prefer not to make this a minor release.

piever:
The only uses I know are in StatsPlots / AbstractPlotting, but they are unaffected by this. The only thing that would break is code that relies on the fact that `QQPair` is concretely typed, but I'm not aware of any example of that in practice. I'd also say it's OK to make this a patch release.

Issue #1214, opened 2020-11-08
Feature request: Add  a shifted (or 3-parameter) LogNormal


briochemc:
I would like to be able to get the `logpdf` and `gradlogpdf` of a random variable *X* with a shifted log-normal distribution, i.e., such that log(*X* + *λ*) ~ *N*(*μ*,*σ*). Maybe adding a 3-argument LogNormal could work?

I saw that there have been discussions about generic transformations (e.g., https://github.com/JuliaStats/Distributions.jl/issues/69), but it does not look like that came to fruition.

I also saw that some similar features have been added for, e.g., multivariate normal distributions (https://github.com/JuliaStats/Distributions.jl/pull/1051), but this was not implemented for other distributions, not even *univariate* normal distributions, AFAIK. (This seems like an oversight to me, but maybe I'm missing something.)

Finally, I guess that some external packages (maybe Bijectors.jl, DistributionsAD.jl, TransformVariables.jl or others), could help, but I have no idea how to use them in my specific use case.

I'm not a stats expert by any means, so I'm not sure this request makes sense 😅 but I have a concrete use case for it and I would love to be able to keep using just Distributions.jl for it. Also, please let me know if there is a workaround or a solution involving other packages!

JHonaker:
The gradient of a shifted distribution is just the gradient of the unshifted distribution evaluated at the point. So, you already have the gradient of any shifted distribution, and actually you have the gradient of any shifted or scaled pretty easily (as well as any function). This follows directly from the chain rule of Calculus:

![image](https://user-images.githubusercontent.com/3012597/98483356-98c91e00-21d5-11eb-9120-f284bceb5a27.png)

So, if `u(x) = ax + b` then:

![image](https://user-images.githubusercontent.com/3012597/98483433-36245200-21d6-11eb-8fe4-e3b59d791205.png)

This isn't an issue, and should probably be closed.

briochemc:
This is not an issue, it is a feature request. I can do the maths, it's not what this is about... I have a complicated-ish pipeline which I feed `Distribution`s to and which calls, e.g., `logpdf` and `gradlogpdf` on those, so I'm looking for a way to manipulate `Distribution`s into other, transformed `Distribution`s, such that `logpdf` and `gradlogpdf` are defined on them.  

For example, you can manipulate `MvNormal`s this way:
```julia
julia> n = 2
2

julia> μ = rand(n)
2-element Array{Float64,1}:
 0.7348050451884647
[...]

PR #1133, opened 2020-06-11
Add gradlogpdf to Uniform

briochemc:

PR #1217, opened 2020-11-10
Add syntactic sugar for LocationScale

briochemc:
In line with how affine transformations can be applied to `MvNormal` (PR https://github.com/JuliaStats/Distributions.jl/pull/1051), this PR uses the already available machinery of `LocationScale` to allow affine transformations of `ContinuousUnivariateDistribution`s, so that one can do, for example:

```julia
julia> d = LogNormal(0.0, 1.0)
LogNormal{Float64}(μ=0.0, σ=1.0)

julia> 2 * d + 1
LocationScale{Float64,LogNormal{Float64}}(
μ: 1.0
σ: 2.0
ρ: LogNormal{Float64}(μ=0.0, σ=1.0)
)
```

This PR follows the discussions in issue #1214.

On top of the syntactic sugar for affine transformations, this PR also specializes `LocationScale(μ,σ,ρ::LocationScale)` so that composed calls to `LocationScale` don't return nested `LocationScales` (as suggested in issue #1214).

The `LocationScale` tests are the same up to a tiny difference (see below) but are now written in a more functional way so that hopefully adding more distributions will be more straightforward (as suggested in issue #1214).

I think the added tests cover all the new functions. But note the tiny difference: I had to change the `atol=1e-100` of the `ccdf` test to `atol=1e-15` for it to pass, since this was off by the last `Float64` digit for the composed `ScaleLocation` test distribution...

Note there is a leftover small fix for `Uniform` from PR https://github.com/JuliaStats/Distributions.jl/pull/1133 ~and another change in `LogNormal` that I guess comes from me manipulating my local repo the wrong way~ [EDIT: This has been reverted in the last commit]. Please advise on what to do about those!


briochemc:
Gentle bump!

andreasnoack (commented):
Wouldn't it be a good idea to add methods for `LocationScale` that would fuse the location and scale instead of nesting?

briochemc:
Yes, I think that's what 

https://github.com/JuliaStats/Distributions.jl/blob/4e026ee9204b3d6f8826028d58c20dfd1628d363/src/univariate/continuous/locationscale.jl#L42

is supposed to do in this PR IIRC, but maybe there's a bug there?

andreasnoack:
Ah. I misread that definition. Then I think this one is good to go.

PR #1218, opened 2020-11-10
Use un-pivoted Cholesky triangle to sample from sparse MvNormalCanon

### LINKED ISSUE #1200: PDMats doesn't account for pivoting in Cholesky of sparse matrices
See JuliaStats/PDMats.jl#120.

This results in incorrect realizations from `rand(::MvNormalCanon)` when the provided precision matrix is factored by CHOLMOD. It may affect other methods as well.

ElOceanografo:
Fixes #1200.

Note I also rewrote the tests--the SuiteSparse Cholesky decomposition returns a different decomposition than the `Base.LinearAlgebra` function (since the Cholesky factorization is not in general unique).  As a result, the random samples from an `MvNormalCanon` distribution with a sparse precision matrix are not in general identical to those from an `MvNormalCanon` or `MvNormal`, even if the seeds are identical.  As a result, these tests only check for approximate statistical equality, rather than strict numerical equality of the samples.  Not sure if there's a better way to handle that.


test/mvnormal.jl
```diff
@@ -1,6 +1,6 @@
 # Tests on Multivariate Normal distributions
 
-import PDMats: ScalMat, PDiagMat, PDMat
+import PDMats: ScalMat, PDiagMat, PDMat, PDSparseMat
```
andreasnoack, line 3: You'd have to put uses of `PDSparseMat` behind a `if isdefined(PDMats, :PDSparseMat)` since it's not defined when Julia is built without GPL libraries.

andreasnoack (commented):
Looks good except for a small issue when using the non-GPL version of Julia

ElOceanografo:
Bump...@andreasnoack, does this need anything else before it can merge?

andreasnoack:
No. Looks good.

PR #1220, opened 2020-11-12
Update show.jl

briochemc:
This changes the `show` method to match the criteria describing it: "if total number of values is greater than 8, or there are matrix-valued params, we use multi-line format". Tests passed locally but maybe there was another reason behind the `!isa(pv, Number)` condition that the tests don't capture?  

Why this PR, then? Because the current show method is not working great for `LocationScale`, and is even worse for composed `LocationScale`s:
```julia
julia> d = LocationScale(1,2,LocationScale(3,4,Normal()))
LocationScale{Float64,LocationScale{Float64,Normal{Float64}}}(
μ: 1.0
σ: 2.0
ρ: LocationScale{Float64,Normal{Float64}}(
μ: 3.0
σ: 4.0
ρ: Normal{Float64}(μ=0.0, σ=1.0)
)

)
```

PR https://github.com/JuliaStats/Distributions.jl/pull/1217 would improve things a bit because it specializes `LocationScale` on itself:
```julia
julia> 1 + 2 * (3 + 4 * Normal())
LocationScale{Float64,Normal{Float64}}(
μ: 7.0
σ: 8.0
ρ: Normal{Float64}(μ=0.0, σ=1.0)
)
```

but it would be even better combined with this PR:
```julia
julia> 1 + 2 * (3 + 4 * Normal())
LocationScale{Float64,Normal{Float64}}(μ=7.0, σ=8.0, ρ=Normal{Float64}(μ=0.0, σ=1.0))
```

and even better with adding a special
```julia
distrname(::LocationScale) = "LocationScale"
```

So that one would get the most readable output IMHO:
```julia
julia> 1 + 2 * (3 + 4 * Normal())
LocationScale(μ=7.0, σ=8.0, ρ=Normal{Float64}(μ=0.0, σ=1.0))
```

### DIFF
--- a/src/show.jl
+++ b/src/show.jl
@@ -33,7 +33,7 @@ function _use_multline_show(d::Distribution, pnames)
     tlen = 0
     for (i, p) in enumerate(pnames)
         pv = getfield(d, p)
-        if !(isa(pv, Number) || isa(pv, NTuple) || isa(pv, AbstractVector))
+        if isa(pv, NTuple) || isa(pv, Array)
             multline = true
         else
             tlen += length(pv)


briochemc:
Gentle bump! 😁

src/show.jl
```diff
@@ -33,7 +33,7 @@ function _use_multline_show(d::Distribution, pnames)
     tlen = 0
     for (i, p) in enumerate(pnames)
         pv = getfield(d, p)
-        if !(isa(pv, Number) || isa(pv, NTuple) || isa(pv, AbstractVector))
+        if isa(pv, NTuple) || isa(pv, Array)
```
devmotion, line 36: @bdeonovic I am a bit confused by this change - it seems it should rather be
```suggestion
         if pv isa AbstractMatrix
```
to correspond exactly to the criterion mentioned in the comment above. But in general, it seems that this hardcoded criterion is a bit unflexible - I assume that there are many other more complex parameter types which should rather be printed in multiple lines. I guess this motivated the previous implementation which instead defined a whitelist of types that can be printed in one line.

BTW I came across this PR because it seems this change leads to errors in https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/80.

bdeonovic: probably meant to tag @briochemc

devmotion: Oh sorry, indeed this was my intention :grimacing:

briochemc: I agree? 🤷 FWIW, I think I merely meant to remove the !isa(pv, Number)...

I guess your suggestion is cleaner and simpler, too!

devmotion: IMO it would be better to change the comment and revert the implementation. The change was quite breaking for more complex distributions (such as `AbstractGPs.FiniteGP`) since now the default is that `length` has to be implemented for all fields that are not a tuple or an array. This would not be resolved by checking `pv isa AbstractMatrix`.

briochemc: I think I get it. Now that I think of it, I think this only worked here with `LocationScale` because univariate distributions come with a length of 1. I guess a better solution would be to specialize `show` for `LocationScale` instead?

devmotion: The problem is that any distribution with a field that is neither a `NTuple` nor an `Array` and does not implement `length(::typeof(field))` (which is called in the other branch) is broken currently. It can even happen for univariate distributions. So I think it would be best to just revert this change and release a new version as soon as possible.

If the update for `LocationScale` can't be integrated in a non-breaking way, it might be best to specialize `show`, as you suggest.

briochemc: Oh yes, I meant that in a LocationScale, one of the fields is a univariate distribution, of which the length is 1. So that worked by chance as I don’t think that was the intended use.

Issue #1221, opened 2020-11-12
incorrect gradlogpdf for LogitNormal


briochemc:
I noticed because it created buggy behaviour in variable transforms for an optimization problem of mine... It struck me when I noticed it is not even symmetric for the default parameters `LogitNormal()`, although it should be.
 
<img width=70% alt="Screen Shot 2020-11-13 at 12 58 44 am" src="https://user-images.githubusercontent.com/4486578/98949110-627bfd80-254b-11eb-9b23-faafc3caebc5.png">

<img width=34% alt="Screen Shot 2020-11-13 at 12 50 58 am" src="https://user-images.githubusercontent.com/4486578/98948308-61969c00-254a-11eb-8b92-91c00736a4d9.png"> <img width=34% alt="Screen Shot 2020-11-13 at 12 50 41 am" src="https://user-images.githubusercontent.com/4486578/98948314-62c7c900-254a-11eb-9a80-e65aa2ea0cdd.png">

Here is the culprit (I guess nobody checked until now!)

https://github.com/JuliaStats/Distributions.jl/blob/ae2d6c5c68abc0aeafdc249ca60df058c19279cb/src/univariate/continuous/logitnormal.jl#L139-L143

This is not symmetric around 0.5 because of the `log(x)` on the left.

Maybe someone with Mathematica or sagemath can do the symbolic derivation from the pdf/logpdf, and correct it?

andreasnoack:
Fixed by https://github.com/JuliaStats/Distributions.jl/pull/1223

PR #1223, opened 2020-11-13
Fix gradlogpdf for LogitNormal

briochemc:
This PR fixes `gradlogpdf` for `LogitNormal`. Comparison with ForwardDiff:

```julia
using ForwardDiff, Distributions, Plots
plot(x -> gradlogpdf(LogitNormal(), x), xlims=(-0.1,1.1), lab="gradlogpdf", lw=10, ylims=(-500,500))
plot!(x -> ForwardDiff.derivative(y -> logpdf(LogitNormal(), y), x), xlims=(-0.1,1.1), lab="AD(logpdf)", lw=4)
```

produces

Screen Shot 2020-11-13 at 12 16 38 pm

And for non-standard parameters:

```julia
μ, σ = 1.0, 2.0
plot(x -> gradlogpdf(LogitNormal(μ, σ), x), xlims=(-0.1,1.1), lab="gradlogpdf", lw=10, ylims=(-50,50))
plot!(x -> ForwardDiff.derivative(y -> logpdf(LogitNormal(μ, σ), y), x), xlims=(-0.1,1.1), lab="AD(logpdf)", lw=4)
```

produces

Screen Shot 2020-11-13 at 12 19 03 pm

### DIFF
--- a/src/univariate/continuous/logitnormal.jl
+++ b/src/univariate/continuous/logitnormal.jl
@@ -136,10 +136,12 @@ cquantile(d::LogitNormal, q::Real) = logistic(norminvccdf(d.μ, d.σ, q))
 invlogcdf(d::LogitNormal, lq::Real) = logistic(norminvlogcdf(d.μ, d.σ, lq))
 invlogccdf(d::LogitNormal, lq::Real) = logistic(norminvlogccdf(d.μ, d.σ, lq))
 
-function gradlogpdf(d::LogitNormal{T}, x::Real) where T<:Real
-    #TODO check
-    (μ, σ) = params(d)
-    0 < x < 1 ? (μ - logit(x)) / (σ^2 * x * (1-x)) - 1/(x-1) - 1/x : zero(T)
+function gradlogpdf(d::LogitNormal, x::Real)
+    μ, σ = params(d)
+    _insupport = insupport(d, x)
+    _x = _insupport ? x : zero(x)
+    z = (μ - logit(_x) + σ^2 * (2 * _x - 1)) / (σ^2 * _x * (1 - _x))
+    return _insupport ? z : oftype(z, NaN)
 end
 
 # mgf(d::LogitNormal)
--- a/src/univariate/continuous/logitnormal.jl
+++ b/src/univariate/continuous/logitnormal.jl
@@ -139,7 +139,7 @@ invlogccdf(d::LogitNormal, lq::Real) = logistic(norminvlogccdf(d.μ, d.σ, lq))
 function gradlogpdf(d::LogitNormal{T}, x::Real) where T<:Real
     #TODO check
     (μ, σ) = params(d)
-    0 < x < 1 ? - ((log(x) - μ) / ((σ^2) + 1) * x * (1-x)) : zero(T)
+    0 < x < 1 ? (μ - logit(x)) / (σ^2 * x * (1-x)) - 1/(x-1) - 1/x : zero(T)
 end
 
 # mgf(d::LogitNormal)


briochemc:
Fixes #1221 BTW

briochemc:
Maybe @bgctw could you review this? (Pinging you because you added `LogitNormal` a while back and I guess you left those `# to check` comments...)

bgctw:
Dear @briochemc, I did not touch Julia for several months and would probably currently introduce more issues in trying to fix one. I plan to look into this beginning of next year but will not take the time to look into this soon.

briochemc:
Can someone else review this?

src/univariate/continuous/logitnormal.jl
```diff
@@ -139,7 +139,7 @@ invlogccdf(d::LogitNormal, lq::Real) = logistic(norminvlogccdf(d.μ, d.σ, lq))
 function gradlogpdf(d::LogitNormal{T}, x::Real) where T<:Real
     #TODO check
     (μ, σ) = params(d)
-    0 < x < 1 ? - ((log(x) - μ) / ((σ^2) + 1) * x * (1-x)) : zero(T)
+    0 < x < 1 ? (μ - logit(x)) / (σ^2 * x * (1-x)) - 1/(x-1) - 1/x : zero(T)
```
devmotion, line 142: This implementation seems prone to type instability issues - maybe it would be a good opportunity to fix them?

briochemc: Should `x` be converted to `T`?

briochemc: Or should it be
```julia
 function gradlogpdf(d::LogitNormal{T}, x::Tx) where {T<:Real, Tx<:Real}
     (μ, σ) = params(d) 
     0 < x < 1 ? (μ - logit(x)) / (σ^2 * x * (1-x)) - 1/(x-1) - 1/x : zero(promote(T,Tx))
end
```
instead?

briochemc: or something else?

briochemc: Gentle bump! @devmotion Could you comment if the proposed change (with the `promote(T,Tx)`) two comments above would be OK?

devmotion: Oh sorry, somehow I missed your comments. Usually I try to avoid type computations since it is to miss specific cases. For instance, in your suggestion the output would still be type unstable if `x` is an integer:
```julia
julia> using StatsFuns

julia> function f(μ::T, σ::T, x::Tx) where {T<:Real, Tx<:Real}
           0 < x < 1 ? (μ - logit(x)) / (σ^2 * x * (1-x)) - 1/(x-1) - 1/x : zero(promote_type(T,Tx))
       end

julia> @code_warntype f(0f0, 1f0, 3)
Variables
  #self#::Core.Compiler.Const(f, false)
  μ::Float32
  σ::Float32
  x::Int64
  @_5::Bool

Body::Union{Float32, Float64}
...
```

Therefore usually I just perform all computations, in particular if these special cases are not common and hence it does not affect performance. For instance:
```julia
julia> function g(μ::T, σ::T, x::Tx) where {T<:Real, Tx<:Real}
           insupport = 0 < x < 1
           # usually I would just use `x` but `logit` errors for values outside of [0, 1]
           _x = insupport ? x : zero(x)
           z = (μ - logit(_x)) / (σ^2 * _x * (1 - _x)) - 1 / (_x - 1) - 1 / _x
           return insupport ? z : zero(z)
       end

julia> @code_warntype g(0f0, 1f0, 3)
Variables
  #self#::Core.Compiler.Const(g, false)
  μ::Float32
  σ::Float32
  x::Int64
  insupport::Bool
  _x::Int64
  z::Float64
  @_8::Bool
  @_9::Int64

Body::Float64
...
```

briochemc: oh I see because `1/x` with `x` an `Int64` converts to `Float64` :) OK I'll check these cases and edit the PR!

briochemc: Actually, now that I think about it, I feel like using just `x` (instead of `_x`) and letting `gradlogpdf` error for `x` outside of `[0,1]` might be better? Since the `logpdf` of `LogitNormal` is `Inf` everywhere outside of `[0,1]`, its gradient should not be defined there, right? 

So maybe the right behavior is simply with

```julia
function gradlogpdf(d::LogitNormal{T}, x::Real) where T<:Real
    μ, σ = params(d)
    (μ - logit(x)) / (σ^2 * x * (1-x)) - 1/(x-1) - 1/x
end
```

What do you think @devmotion

briochemc: From a slack suggestion (Michael Abbott + Moritz Schauer) suggested that the derivative of `f(x) = x -> Inf` should maybe return `NaN` instead because `(f(x + h) - f(x)) / h` is `(Inf - Inf) / h`, which is a `NaN`, so what about something like:

```julia
function gradlogpdf(d::LogitNormal{T}, x::Tx) where {T<:Real, Tx<:Real}
    μ, σ = params(d)
    _x = insupport(d, x) ? x : Tx(NaN)
    (μ - logit(_x)) / (σ^2 * _x * (1 - _x)) - 1 / (_x - 1) - 1 / _x
end
```

This prevents throwing an error but rightfully propagates a `NaN`?

devmotion: I just wanted to suggest that it is probably better to return `NaN` instead of erroring. The conversion to NaN happens to early - this won't work for the example above since `Int(NaN)` will error. IMO it is still better to use my suggestion above and just return `oftype(z, NaN)` instead of `zero(z)` for inputs `x` outside of the support.

briochemc: Actually, and apologies for dragging this, but it seems wrong to get a `Float64` when the input was `Float32` and `Int64` IMHO, no? What about using multiple dispatch and converting the possible `x::Int` to the right `Float` type instead? if that makes sense? E.g.,
```julia
function gradlogpdf(d::LogitNormal{Td}, x::Tx) where {Td<:Real, Tx<:Real}
    μ, σ = params(d)
    insupport(d, x) ? (μ - logit(x)) / (σ^2 * x * (1 - x)) - 1 / (x - 1) - 1 / x : promote_type(Td,Tx)(NaN)
end
gradlogpdf(d::LogitNormal{T}, x::Int) where T<:Real = gradlogpdf(d, T(x))
```
This looks type stable to me, as I see no red with
```julia
xs = Any[1, 1.0, 1f0, Float16(1)]
for d in LogitNormal.(0,x), x in [xs; 2xs]
       @code_warntype gradlogpdf(d,x)
end
```
but maybe that's not how to one checks type stability? And the output seems correct with, e.g.,
```julia
for d in LogitNormal.(0,x), x in [xs; 2xs]
       @show d, x, gradlogpdf(d,x)
end
```
What do you think @devmotion ?

devmotion: I still think that these type computations are not a perfect solution and would break for other types than `Int` (what about other Integers? special number types etc. Rational?). So I think the best approach is to just compute the result in all cases, and use `zero(x)` to avoid errors in `logit`.

If you don't like the Float64 output for this example, I would rather suggest to use
```julia
(μ - logit(x) + σ^2 * (2 * x - 1)) / (σ^2 * x * (1 - x))
```
instead.

briochemc: Again, sorry for the long drag, but I'm learning in the process! So I now realize I got `Int` confused with `Integer`, but I see your point about rationals. But now also realize this type does not need to be prescribed here, right? So what about replacing the last line with  
```julia
gradlogpdf(d::LogitNormal{T}, x) where T<:Real = gradlogpdf(d, T(x))
```
instead?

briochemc: Ah no that does not work, sorry!

briochemc: So would this do?
```julia
function gradlogpdf(d::LogitNormal{<:Real}, x::Real)
    μ, σ = params(d)
    _insupport = insupport(d, x)
    x = _insupport ? x : zero(x)
    z = (μ - logit(x)) / (σ^2 * x * (1 - x)) - 1 / (x - 1) - 1 / x
    return _insupport ? z : oftype(z, NaN)
end
```

devmotion: Yes, I would only avoid using the same variable name and you can also drop the type constraint on `LogitNormal`. Alternatively,
```julia
function gradlogpdf(d::LogitNormal, x::Real)
    μ, σ = params(d)
    _insupport = insupport(d, x)
    _x = _insupport ? x : zero(x)
    z = (μ - logit(_x) + σ^2 * (2 * _x - 1)) / (σ^2 * _x * (1 - _x))
    return _insupport ? z : oftype(z, NaN)
end
```
If you want to avoid all the underscores you could also write something like
```julia
function gradlogpdf(d::LogitNormal, x::Real)
    μ, σ = params(d)
    _insupport = insupport(d, x)
    z = logitnormal_gradlogpdf(μ, σ, _insupport ? x : zero(x))
	return _insupport ? z : oftype(z, NaN)
end
function logitnormal_gradlogpdf(μ, σ, x)
    return μ - logit(x) + σ^2 * (2 * x - 1)) / (σ^2 * x * (1 - x)
end
```

briochemc: OK I put this into a new commit! Thanks again @devmotion!

PR #1227, opened 2020-11-18
Addition of heslogpdf of cont. univariate distributions

### LINKED ISSUE #1225: Addition of heslogpdf
Hi,

I'm one of the main developers of [PowerModelsDistributionStateEstimation.jl](https://github.com/Electa-Git/PowerModelsDistributionStateEstimation.jl). For its version v0.2, we are planning to extend the state-estimator to incorporate non-Gaussian maximum likelihood estimation. To enhance the computational speed, it is preferable to give the solver both the first and second derivative of the logpdf. The first derivative: gradlogpdf is provided in this package, however the second: 'heslogpdf' is not. 

Currently, we are providing the latter in our package. However, given that it might benefit other applications, would it be of interest to this package to add the 'heslogpdf'?

with kind regards,

Tom Van Acker

timmyfaraday:
Closes #1225

Addition of the heslogpdf, i.e., the second derivative of the logpdf, of continuous univariate distributions, associated tests and documentation. 

Included distributions: 
- arcsine, 
- beta, 
- chi, 
- chisq, 
- exponential, 
- frechet, 
- gamma, 
- gumbel, 
- laplace, 
- locationscale, 
- logistic, 
- logitnormal, 
- lognormal, 
- normal, 
- tdist, and 
- weibull.


MartaVanin:
Hello,
this is just to say that I also would like to have this functionality in Distributions.jl. 
I wondered if there is any plan of merging this?

src/univariate/continuous/arcsine.jl
```diff
@@ -99,3 +99,13 @@ function gradlogpdf(d::Arcsine{T}, x::R) where {T, R <: Real}
...
+function heslogpdf(d::Arcsine{T}, x::R) where {T, R <: Real}
+    TP = promote_type(T, R)
+    (a, b) = extrema(d)
+    # on the bounds, we consider the gradient limit inside the domain
+    # right side for the left bound,
+    # left side for the right bound
+    a < x <= b || return TP(-Inf)
+    x == b && return TP(Inf)
+    return TP(0.5 * (inv((x - a)^2) + inv((x - b)^2)))
+end
```
devmotion, line 111: It would be better to not base the return values on `T` and `R` explicitly. Instead one should perform as many calculations as possible and only return then `oftype(result, Inf)` (and similar things) in the cases where it is needed. This is currently not done for `gradlogpdf` it seems but since `heslogpdf` is a new addition it would be good to get it right from the start.

E.g., here one could write:
```suggestion
function heslogpdf(d::Arcsine, x::Real)
    a, b = extrema(d)
    result = (inv((x - a)^2) + inv((x - b)^2)) / 2
    # on the bounds, we consider the gradient limit inside the domain
    # right side for the left bound,
    # left side for the right bound
    return if x <= a
       oftype(result, -Inf)
    elseif x >= b
       oftype(result, Inf)
    else
       result
    end
end
```

timmyfaraday: Thank you for your feedback, I'll adjust the code accordingly asap.

PR #1231, opened 2020-11-27
Implement Dirac distribution (updated PR)

devmotion:
This is an update of https://github.com/JuliaStats/Distributions.jl/pull/861, according to the discussion over there.

The main differences are:
- Merged master and fixed merge conflicts
- Extended documentation
- Removed `kurtosis` and `skewness`
- Added `logpdf`
- Implement `eltype` for type instead of instance
- Use `cis` in the implementation of `cf`

I took some inspiration from `bernoulli.jl`.

Two questions:
- Would you like me to bump the version?
- I guess it is the convention in Distributions to return Float64 (at least it was suggested in the previous PR) - it seems in cases where we only return `0` or `1` this causes promotions in code of users that could be avoided?


devmotion:
Travis tests don't run due to negative credit balance.

mschauer:
Tests passing now with the new setup

devmotion:
I think this PR is ready for review :slightly_smiling_face:

devmotion:
Gentle bump :slightly_smiling_face:

src/univariate/discrete/dirac.jl
```diff
@@ -0,0 +1,54 @@
...
+Base.eltype(::Type{Dirac{T}}) where {T} = T
+
+insupport(d::Dirac, x::Real) = x == d.value
+minimum(d::Dirac) = d.value
+maximum(d::Dirac) = d.value
+
+#### Properties
+
+mean(d::Dirac) = d.value
+var(d::Dirac) = 0.0
```
matbesancon, line 34:
```suggestion
var(d::Dirac{T}) where {T} = zero(T)
```

src/univariate/discrete/dirac.jl
```diff
@@ -0,0 +1,54 @@
...
+
+mode(d::Dirac) = d.value
+
+entropy(d::Dirac) = 0.0
```
matbesancon, line 38:
```suggestion
entropy(d::Dirac{T}) where {T} = zero(T)
```

matbesancon:
few minor comments, good to merge when these are addressed

matbesancon:
thanks a lot for the PR!

devmotion:
Thanks! Is it possible to make a new release?

matbesancon:
sure yes, doing it now

devmotion:
Thank you!

PR #1236, opened 2020-12-05
Fix sampling from DiscreteNonParametric with integer-valued probs

devmotion:
Fixes https://github.com/JuliaStats/Distributions.jl/issues/1111.

PR #1242, opened 2020-12-18
Fix #1241; correct the behavior of logpdf(d::Dirichlet, x)

### LINKED ISSUE #1241: Incorrect pdf for Dirichlet distribution
Hello,

I was using Julia to explore statistical distributions and came across what seems to me to be an error in the `pdf` of the Dirichlet distribution.

```julia
julia> using Distributions

julia> beta = Beta(1, 1);

julia> pdf(beta, 0)
1.0

julia> dirichlet = Dirichlet([1, 1]);

julia> pdf(dirichlet, [0, 1])
NaN
```

My understanding is that the two distributions above are [mathematically equivalent](https://stats.stackexchange.com/questions/78618/what-is-the-dirichlet-equivalent-of-a-beta-1-1-distribution), and therefore the results should be the same for both of the `pdf`s above. The second case uses `[0, 1]` because the Dirichlet distribution is defined on a simplex.

I believe the behavior of `pdf(beta, 0)` is correct.

I also noticed a difference in behavior on points outside of the support: the Beta distribution returns zero density, while the Dirichlet function can error or report incredibly large values:

```julia
julia> pdf(beta, -0.5)
0.0

julia> pdf(dirichlet, [-0.5, 1.5])
ERROR: DomainError with -0.5:
log will only return a complex result if called with a complex argument. Try log(Complex(x)).

julia> pdf(Dirichlet([1000, 100]), [1, 1])
1.293606521475018e146
```

yurivish:
This fixes `pdf(d::Dirchlet, x)` at the boundary of and outside of the support, and replaces hard-coded Float64 values with more general types.

### DIFF
--- a/src/multivariate/dirichlet.jl
+++ b/src/multivariate/dirichlet.jl
@@ -170,11 +170,14 @@ function insupport(d::Dirichlet, x::AbstractVector{T}) where T<:Real
     return true
 end
 
-function _logpdf(d::Dirichlet, x::AbstractVector{T}) where T<:Real
+function _logpdf(d::Dirichlet{S}, x::AbstractVector{T}) where {S, T<:Real}
+    if !insupport(d, x)
+        return convert(promote_type(S, T), -Inf)
+    end
     a = d.alpha
-    s = 0.
+    s = zero(promote_type(S, T))
     for i in 1:length(a)
-        @inbounds s += (a[i] - 1.0) * log(x[i])
+        @inbounds s += xlogy(a[i] - one(S), x[i])
     end
     return s - d.lmnB
 end
--- a/test/dirichlet.jl
+++ b/test/dirichlet.jl
@@ -22,6 +22,12 @@ d = Dirichlet(3, 2.0)
 @test cov(d)  ≈ [8 -4 -4; -4 8 -4; -4 -4 8] / (36 * 7)
 @test var(d)  ≈ diag(cov(d))
 
+@test pdf(Dirichlet([1, 1]), [0, 1]) ≈ 1.0
+@test pdf(Dirichlet([1f0, 1f0]), [0f0, 1f0]) ≈ 1.0f0
+@test typeof(pdf(Dirichlet([1f0, 1f0]), [0f0, 1f0])) == Float32
+
+@test pdf(d, [-1, 1, 0])         ≈ 0.0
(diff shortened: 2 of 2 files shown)


mschauer:
I doubt one can get it really perfect as function of `n` argument vectors
```
julia> mylogpdf(Dirichlet([1, 1.0]), [1.0 + sqrt(eps()), 0.])
-Inf

julia> mylogpdf(Dirichlet([1, 1.0]), [1.0 + 0.5sqrt(eps()), 0.])
0.0
```

mschauer (approved):

yurivish:
That seems to be a bug in `insupport`:

```julia
julia> insupport(d, [1.0 + 0.5sqrt(eps()), 0.])
true
```

andreasnoack:
Should `insupport` be adjusted as part of this PR?

mschauer:
I think not, if there are any later improvements to insupport they likely require no changes to the PR, and improving insupport is tricky without going to an proper n-1 dimensional representation of probability vectors anyway

devmotion:
I suspect this PR, and in particular the type promotions, broke DistributionsAD: https://github.com/TuringLang/DistributionsAD.jl/pull/143/checks?check_run_id=1592162436

Maybe they are fixed by https://github.com/JuliaStats/Distributions.jl/pull/1243.

mschauer:
The error `no method matching Float64(::Tracker.TrackedReal{Float64})` suggest 
`raw = rand(rng, float(eltype(p)))` in https://github.com/JuliaStats/Distributions.jl/pull/1236/files#diff-532294b375dcec6c10dc02b6b34da3851007768391e96b6d02c2a87eb74a1626R76 could also be a culprit?

devmotion:
This change is not released yet, is it?

mschauer:
Hm, no, but perhaps it will be a problem?

devmotion:
The error occurs only for `Dirichlet` whereas the other change affected `DiscreteNonParametric`. Additionally, we do not differentiate through `rand` in DistributionsAD (well, apart from normal distributions...) and I checked that `rand(float(eltype(Tracker.params(x)))` works fine.

mschauer:
Thanks for checking.

devmotion:
It took me a while but in the end I figured out that the latest release of NNlib is responsible for the test failures: https://github.com/FluxML/NNlib.jl/issues/251

PR #1243, opened 2020-12-21
Allow general vectors in `Dirichlet` [BREAKING]

devmotion:
This PR generalizes `Dirichlet` to general `AbstractVector`s of parameters.

This change allows to use `Fill` instead of `fill` for the special case of equal alpha parameter in every dimension, which also simplifies and improves efficiency of computations of `var` and `entropy` as well as sampling. In some places hardcoded `Float64` had to be removed.


devmotion:
I also added a missing `check_args` keyword argument.

devmotion:
Bump :slightly_smiling_face:

devmotion:
Bump :slightly_smiling_face:

PR #1244, opened 2020-12-22
Revert update of `show`

devmotion:
This PR reverts https://github.com/JuliaStats/Distributions.jl/pull/1220 which leads to errors for all distributions with fields that are neither of type `NTuple` nor `Array` and do not implement `length` (see https://github.com/JuliaStats/Distributions.jl/pull/1220#discussion_r547174197).

For instance, this leads to errors for `AbstractGPs.FiniteGP` which has a field of type `AbstractGPs.GP` that does not implement `length`.

It would be great if a version with this bugfix could be released soon.

### DIFF
--- a/src/show.jl
+++ b/src/show.jl
@@ -26,14 +26,15 @@ function _use_multline_show(d::Distribution, pnames)
     # decide whether to use one-line or multi-line format
     #
     # Criteria: if total number of values is greater than 8, or
-    # there are matrix-valued params, we use multi-line format
+    # there are params that are neither numbers, tuples, or vectors,
+    # we use multi-line format
     #
     namevals = _NameVal[]
     multline = false
     tlen = 0
     for (i, p) in enumerate(pnames)
         pv = getfield(d, p)
-        if isa(pv, NTuple) || isa(pv, Array)
+        if !(isa(pv, Number) || isa(pv, NTuple) || isa(pv, AbstractVector))
             multline = true
         else
             tlen += length(pv)


src/show.jl
```diff
@@ -26,14 +26,15 @@ function _use_multline_show(d::Distribution, pnames)
...
+    # there are params that are neither numbers, tuples, or vectors,
+    # we use multi-line format
     #
     namevals = _NameVal[]
     multline = false
     tlen = 0
     for (i, p) in enumerate(pnames)
         pv = getfield(d, p)
-        if isa(pv, NTuple) || isa(pv, Array)
+        if !(isa(pv, Number) || isa(pv, NTuple) || isa(pv, AbstractVector))
```
mschauer, line 37: Yes, this was restricting the `else` branch to `Number`, `NTuple` and `AbstractVectors` which all have length

mschauer (approved):

mschauer:
Can you run the tests of the `DistributionsAD` dependency locally to check?

devmotion:
This is not related to DistributionsAD (only the other PR for `Dirichlet` which was merged and released recently is). I noticed the bug introduced by the change that is reverted by this PR in AbstractGPs. It causes the errors in https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/82/checks?check_run_id=1592211377 and https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/runs/1591243441?check_suite_focus=true.

mschauer:
Ok, will merge later if nobody objects

devmotion:
Bump :slightly_smiling_face:

devmotion:
Another bump. CI of AbstractGPs is still broken due to this issue, it would be great to get a new release of Distributions with this fix :slightly_smiling_face:

devmotion:
Thanks for merging the PR! Is it possible to get a new release with the fix?

PR #1245, opened 2020-12-29
Add Landau distribution

Moelf:
I don't know what's the standard for numerical methods, I'd happy to split into a separate package if this is not suitable for Distributions.jl


src/univariate/continuous/landau.jl
```diff
@@ -0,0 +1,191 @@
...
+
+pdf(d::Landau, x::Real) = _landau_pdf((x-d.μ)/d.θ)
+
+cdf(d::Landau, x::Real) = _landau_cdf((x-d.μ)/d.θ)
+
+function cf(d::Landau, t::Real)
+    exp(im*t*d.μ - 2im*d.θ*t/π*log(abs(t)) - d.θ*abs(t))
+end
+
+# https://root.cern.ch/doc/v622/PdfFuncMathCore_8cxx_source.html
```
devmotion, line 72: I'm not an expert but I think this is problematic since root uses the LGPL license: https://root.cern/about/license/

src/univariate/continuous/landau.jl
```diff
@@ -0,0 +1,191 @@
...
+end
+
+
+function Landau(μ::T, θ::T; check_args=true) where {T <: Real}
+    check_args && @check_args(Landau, θ > zero(θ))
+    return Landau{T}(μ, θ)
+end
+
+Landau(μ::Real, θ::Real) = Landau(promote(μ, θ)...)
+Landau(μ::Integer, θ::Integer) = Landau(float(μ), float(θ))
```
devmotion, line 39: Is it necessary to convert integers to floating point numbers?
```suggestion
```

src/univariate/continuous/landau.jl
```diff
@@ -0,0 +1,191 @@
...
+Landau(μ::T) where {T <: Real} = Landau(μ, one(T))
+Landau() = Landau(0.0, 1.0, check_args=false)
```
devmotion, line 41:
```suggestion
Landau(μ::Real = 0) = Landau(μ, zero(μ); check_args=false)
```

src/univariate/continuous/landau.jl
```diff
@@ -0,0 +1,191 @@
...
+    Landau{T}(µ::T, θ::T) where {T} = new{T}(µ, θ)
+end
+
+
+function Landau(μ::T, θ::T; check_args=true) where {T <: Real}
+    check_args && @check_args(Landau, θ > zero(θ))
+    return Landau{T}(μ, θ)
+end
+
+Landau(μ::Real, θ::Real) = Landau(promote(μ, θ)...)
```
devmotion, line 38:
```suggestion
Landau(μ::Real, θ::Real; check_args=true) = Landau(promote(μ, θ)...; check_args=check_args)
```

src/univariate/continuous/landau.jl
```diff
@@ -0,0 +1,191 @@
...
+
+External links
+
+* [Landau distribution on Wikipedia](http://en.wikipedia.org/wiki/Landau_distribution)
+
+"""
+struct Landau{T<:Real} <: ContinuousUnivariateDistribution
+    μ::T
+    θ::T
+    Landau{T}(µ::T, θ::T) where {T} = new{T}(µ, θ)
```
devmotion, line 29: AFAICT this is not needed since it is the default definition:
```suggestion
```

src/univariate/continuous/landau.jl
```diff
@@ -0,0 +1,191 @@
+using StaticArrays
```
devmotion, line 1: I assume dependencies should be loaded in `src/Distributions.jl`. I'm a bit surprised though that Distributions depends on StaticArrays but StaticArrays doesn't seem to be imported.

src/univariate/continuous/landau.jl
```diff
@@ -0,0 +1,191 @@
...
+
+#### Evaluation
+
+pdf(d::Landau, x::Real) = _landau_pdf((x-d.μ)/d.θ)
+
+cdf(d::Landau, x::Real) = _landau_cdf((x-d.μ)/d.θ)
+
+function cf(d::Landau, t::Real)
+    exp(im*t*d.μ - 2im*d.θ*t/π*log(abs(t)) - d.θ*abs(t))
+end
```
devmotion, line 70: The following definition uses `cis(x) = exp(im*x)` and avoids NaN values for `t = 0`:
```suggestion
function cf(d::Landau, t::Real)
    z = t * (d.μ + d.θ * (sign(t) * im - log(abs(t)) * 2 / π))
    return iszero(t) ? cis(zero(z)) : cis(z) 
end
```

src/univariate/continuous/landau.jl
```diff
@@ -0,0 +1,191 @@
...
+
+# https://root.cern.ch/doc/v622/PdfFuncMathCore_8cxx_source.html
+function _landau_pdf(x, xi=1, x0=0)
+    # landau pdf : algorithm from cernlib g110 denlan
+    p1 = @SVector [0.4259894875,-0.1249762550, 0.03984243700, -0.006298287635,   0.001511162253]
```
devmotion, line 75: In general, I think it might be easier to just use `evalpoly` or `@evalpoly`.

src/univariate/continuous/landau.jl
```diff
@@ -0,0 +1,191 @@
...
+    q6 = @SVector [1.0            , 0.1339887843e+3, 0.1055990413e+4, 0.5532224619e+3]
+
+    a1 = @SVector [0              ,-0.4583333333e+0, 0.6675347222e+0,-0.1641741416e+1]
+    a2 = @SVector [0              , 1.0            ,-0.4227843351e+0,-0.2043403138e+1]
+
+    v = (x - x0)/xi
+
+    if (v < -5.5)
+        u   = exp(v+1)
+        lan = 0.3989422803*exp(-1. /u)*sqrt(u)*(1+(a1[2]+(a1[3]+a1[4]*u)*u)*u)
```
devmotion, line 168: Same here, I guess using `evalpoly` or `@evalpoly` would be easier to read.

src/univariate/continuous/landau.jl
Moelf: yes, someone pointed out on Slack as well, it is true we cannot put derived code directly. so this may call for a new pkg

src/univariate/continuous/landau.jl
Moelf: I think it only uses them in test, so technically we should change that in Project.toml to reflect this

devmotion: Yes, I thought so as well.

devmotion: Ref: https://github.com/JuliaStats/Distributions.jl/pull/1246

PR #1250, opened 2021-01-06
MvNormal and MvTDist 32-bit logpdf fixes

rofinn:
logpdf should probably return 32-bit results if only 32-bit inputs are provided.
This only fixes it for MvNormal and MvTDist since that's what I'm working with.
Not sure if my tests fit with the current organizational system.


andreasnoack:
It would be great to also have the other methods for `MvNormal` and `MvTDist` made generic but this is fine as a start.

rofinn:
My plan was to address these as they come up to avoid an overly large PR. If you'd prefer I can probably make some time to do a full scan over those Distributions types? Also, should I make a PR to tag a patch release?

Issue #1252, opened 2021-01-07
Inconsistent sampling type for Uniform{BigFloat}


yurivish:
While playing around with the `Uniform` distribution I noticed a seeming inconsistency between the results of generating a single sample from a high-precision uniform distribution using `rand` and generating multiple samples. 

For a `Uniform{BigFloat}` distribution, a single sample using `rand(dist)` yields a `BigFloat` value while multiple samples using `rand(dist, n)` yield `Float64` values.

```julia
julia> using Distributions

julia> dist = Uniform(BigFloat(0), BigFloat(1))
Uniform{BigFloat}(a=0.0, b=1.0)

julia> rand(dist)
0.518440823260704508612661811639554798603057861328125

julia> typeof(rand(dist))
BigFloat

julia> rand(dist, 2)
2-element Array{Float64,1}:
 0.8587350116719372
 0.6892639654172594

julia> typeof(rand(dist, 2))
Array{Float64,1}

julia> versioninfo()
Julia Version 1.5.3
Commit 788b2c77c1 (2020-11-09 13:37 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin18.7.0)
  CPU: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)
```

PR #1257, opened 2021-01-11
Remove pre-allocation of outputs for `logpdf` and `pdf` + deprecate implementations for multiple samples

devmotion:
This PR removes the pre-allocation of output arrays for `logpdf` and `pdf`.

There seems little reason to fall back to the in-place methods since it does neither reduce allocations nor lines of code. Instead, I'd argue that the pre-allocation causes problems since the computation of the output type is only a heuristic and it seems easy to come up with examples where it fails since typically `logpdf` and `pdf` do not ensure that their output is of this specific type. Additionally, the pre-allocations lead to calls of mutating functions which break AD with e.g. Zygote. Moreover, it is consistent with `logpdf` and `pdf` implementations for univariate distributions that do not fall back to the mutating functions but use broadcasting (one could use the same implementation with `map` and `Base.Fix1` there as well) - however, it seems a bit strange that the use of `(log)pdf` with arrays is deprecated for univariate distributions but not for multi- and matrixvariate distributions.

### DIFF
--- a/src/matrixvariates.jl
+++ b/src/matrixvariates.jl
@@ -197,14 +197,12 @@ function pdf!(r::AbstractArray, d::MatrixDistribution, X::AbstractArray{M}) wher
     _pdf!(r, d, X)
 end
 
-function logpdf(d::MatrixDistribution, X::AbstractArray{M}) where M<:Matrix
-    T = promote_type(partype(d), eltype(M))
-    _logpdf!(Array{T}(undef, size(X)), d, X)
+function logpdf(d::MatrixDistribution, X::AbstractArray{<:AbstractMatrix{<:Real}})
+    map(Base.Fix1(logpdf, d), X)
 end
 
-function pdf(d::MatrixDistribution, X::AbstractArray{M}) where M<:Matrix
-    T = promote_type(partype(d), eltype(M))
-    _pdf!(Array{T}(undef, size(X)), d, X)
+function pdf(d::MatrixDistribution, X::AbstractArray{<:AbstractMatrix{<:Real}})
+    map(Base.Fix1(pdf, d), X)
 end
 
 """
--- a/src/matrixvariates.jl
+++ b/src/matrixvariates.jl
@@ -145,72 +145,53 @@ end
 
 # pdf & logpdf
 
+"""
+    _logpdf(d::MatrixDistribution, X::AbstractMatrix)
+
+Return the logarithm of the probability density of distribution `d` evaluated at `X`.
+
+This function does not need to perform dimension checking. Its default implementation is
+`logkernel(d, X) + d.logc0`.
+"""
 _logpdf(d::MatrixDistribution, X::AbstractMatrix) = logkernel(d, X) + d.logc0
 
-_pdf(d::MatrixDistribution, x::AbstractMatrix{T}) where {T<:Real} = exp(_logpdf(d, x))
+"""
+    _pdf(d::MultivariateDistribution, X::AbstractMatrix)
+
+Return the probability density of distribution `d` evaluated at `X`.
+
+This function does not need to perform dimension checking. It does not need to be
+implemented and falls back to `exp(_logpdf(d, X))`.
+
+See also: [`_logpdf`](@ref)
+"""
+_pdf(d::MatrixDistribution, X::AbstractMatrix) = exp(_logpdf(d, X))
 
 """
-    logpdf(d::MatrixDistribution, AbstractMatrix)
+    logpdf(d::MatrixDistribution, X::AbstractMatrix)
 
-Compute the logarithm of the probability density at the input matrix `x`.
+Compute the logarithm of the probability density at the input matrix `X`.
+
+The default implementation performs dimension checking and falls back to [`_logpdf`](@ref).
 """
-function logpdf(d::MatrixDistribution, x::AbstractMatrix{T}) where T<:Real
+function logpdf(d::MatrixDistribution, x::AbstractMatrix)
     size(x) == size(d) ||
         throw(DimensionMismatch("Inconsistent array dimensions."))
     _logpdf(d, x)
 end
 
 """
-    pdf(d::MatrixDistribution, x::AbstractArray)
+    pdf(d::MatrixDistribution, X::AbstractMatrix)
 
-Compute the probability density at the input matrix `x`.
+Compute the probability density at the input matrix `X`.
+
+The default implementation performs dimension checking and falls back to [`_pdf`](@ref).
 """
-function pdf(d::MatrixDistribution, x::AbstractMatrix{T}) where T<:Real
+function pdf(d::MatrixDistribution, x::AbstractMatrix)
     size(x) == size(d) ||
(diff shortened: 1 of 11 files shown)


src/multivariates.jl
```diff
@@ -230,15 +230,13 @@ end
 function logpdf(d::MultivariateDistribution, X::AbstractMatrix)
     size(X, 1) == length(d) ||
         throw(DimensionMismatch("Inconsistent array dimensions."))
-    T = promote_type(partype(d), eltype(X))
-    _logpdf!(Vector{T}(undef, size(X,2)), d, X)
+    map(i -> _logpdf(d, view(X, :, i)), axes(X, 2))
```
devmotion, line 233: It would be nice to just use
```suggestion
    map(Base.Fix1, eachcol(X))
```
here but `eachcol` requires Julia >= 1.1 or Compat.

src/multivariates.jl
```diff
@@ -230,15 +230,13 @@ end
...
 end
 
 function pdf(d::MultivariateDistribution, X::AbstractMatrix)
     size(X, 1) == length(d) ||
         throw(DimensionMismatch("Inconsistent array dimensions."))
-    T = promote_type(partype(d), eltype(X))
-    _pdf!(Vector{T}(undef, size(X,2)), d, X)
+    map(i -> _pdf(d, view(X, :, i)), axes(X, 2))
```
devmotion, line 239: Same here, with `eachcol` one could write
```suggestion
    map(Base.Fix1(_pdf, d), eachcol(X))
```

devmotion:
Similar changes could (should?) be applied to `rand` to avoid inconsistencies such as reported in https://github.com/JuliaStats/Distributions.jl/issues/1252.

src/matrixvariates.jl
```diff
@@ -197,14 +197,12 @@ function pdf!(r::AbstractArray, d::MatrixDistribution, X::AbstractArray{M}) wher
     _pdf!(r, d, X)
 end
 
-function logpdf(d::MatrixDistribution, X::AbstractArray{M}) where M<:Matrix
-    T = promote_type(partype(d), eltype(M))
-    _logpdf!(Array{T}(undef, size(X)), d, X)
+function logpdf(d::MatrixDistribution, X::AbstractArray{<:AbstractMatrix{<:Real}})
```
andreasnoack, line 200: It seems to me that we shouldn't have such an implicitly broadcasting method at all. We generally deprecated these a long time ago. Is there a reason we kept this one?

devmotion: I don't know why it exists but it is definitely inconsistent that currently the application of `(log)pdf(!)` to multiple samples is deprecated for univariate distributions but not for multi- and matrixvariate distributions. In general, I think it would be good to be consistent and either allow this application for all distributions or remove all implementations. I lean towards the former, mainly because it is difficult to write generic code that computes the logpdf for multiple samples for arbitrary distributions without these `logpdf` definitions since one cannot use broadcasting to evaluate the `logpdf` for a matrix of multivariate samples in the same way.

andreasnoack: Couldn't you just write
```julia
julia> logpdf.(Ref(InverseWishart(2, Matrix{Float64}(I, 2, 2))), fill(Matrix{Float64}(I, 2, 2), 2))
2-element Array{Float64,1}:
 -3.5310242469692907
 -3.5310242469692907
```

devmotion: Sure, you can do this in this case. And you can do it in the same way for univariate distributions:
```julia
julia> logpdf.(Ref(Normal()), rand(Normal(), 2))
2-element Array{Float64,1}:
 -2.413956289132882
 -1.0116727704190984
```
Unsurprisingly though, the following fails:
```julia
julia> logpdf.(Ref(MvNormal(2, 1)), rand(MvNormal(2, 1), 2))
ERROR: MethodError: no method matching logpdf(::MvNormal{Float64,PDMats.ScalMat{Float64},FillArrays.Zeros{Float64,1,Tuple{Base.OneTo{Int64}}}}, ::Float64)
...
```
So it is not possible to generically use `logpdf.(dist, samples)` to compute logpdfs of samples but instead one has to dispatch on the type of distributions and samples and either add a custom implementation for these cases or call
```julia
julia> logpdf(MvNormal(2, 1), rand(MvNormal(2, 1), 2))
2-element Array{Float64,1}:
 -2.0814343848408745
 -3.5913383903349345
```

As I mentioned above, I would be fine with either removing all of these `logpdf` implementations for multiple samples or adding them for all distributions, as long as it is done in a consistent way. However, I think it would be easier if Distributions would implement `logpdf` for multiple samples for all distributions since otherwise every downstream package that wants to handle logpdf calculations for all distribution types has to add these implementations and make sure that broadcasting is used for univariate and matrixvariate but not for multivariate distributions + matrices of samples.

andreasnoack: I really think we should avoid implicit broadcasting. We've spent quite some time teaching people to be explicit about broadcasting and it's also the only way to hope for composability of broadcasting operations. IMO, the issue here is the return type of the multivariate distributions which doesn't fit well into the general interface. Until we have a general solution for that, I think the better approach is to ask people to use `eachrow` explicitly for the multivariates, i.e. something like
```julia
julia> logpdf.(Ref(MvNormal(ones(2))), eachrow(randn(10,2)))
10-element Array{Float64,1}:
 -3.2672567129269323
 -2.6944444370392286
 -4.083857499332683
 -2.221300852511519
 -1.9903141948994234
 -3.009577938574812
 -4.12761318469939
 -3.0240385144021076
 -2.0975290149847705
 -2.5841749740831976
 ```

devmotion: OK, so should all these `(log)pdf(!)` be deprecated in this PR? Or better in a separate PR? I think regardless of these deprecations the in-place calls are not "correct" and should be avoided.

andreasnoack: I'm fine with either doing it as part of this PR or in a follow-up you believe it simplifies things.

src/multivariates.jl
andreasnoack: Indeed. I had forgotten that `eachcol`/`row`  weren't available in 1.0.

devmotion:
I deprecated all `(log)pdf(!)` methods for multiple samples, as discussed in https://github.com/JuliaStats/Distributions.jl/pull/1257#discussion_r555740393. However, maybe it would be better to separate this in a different PR since it implies that the optimization for `_logpdf!` in some distributions such as `MvNormal` are not possible anymore (since these functions don't exist anymore/are deprecated).

devmotion:
I think this requires some additional discussion - maybe it would be better to revert the deprecations that were suggested by @andreasnoack and just fix the allocating calls for now? What's the general opinion?

Ping @matbesancon @mschauer

andreasnoack:
I'd be fine with splitting this into two PRs: one similar to your original version which could then be a bugfix release and then a PR similar to the current version that would require a breaking release but I still think it would be good to deprecate these methods.

st--:
@devmotion should this be closed now?

devmotion:
Seems it was addressed in the linked PRs :+1:

PR #1262, opened 2021-01-21
Fix sampling from distributions with integer-valued parameters (e.g. `MvNormal` and `Dirichlet`)

### LINKED ISSUE #1004: Sampling from MvNormal with integer arrays fails
When an `MvNormal` is created using integer arrays, sampling from the distribution fails.
```julia
julia> using Distributions

julia> rand(MvNormal([1,1],[1,1]), 10)
ERROR: MethodError: no method matching randn(::Random._GLOBAL_RNG, ::Type{Int64})
Closest candidates are:
  randn(::AbstractRNG, ::Type{T}, ::Tuple{Vararg{Int64,N}} where N) where T at /home/fredrikb/julia-1.3.0-rc4/share/julia/stdlib/v1.3/Random/src/normal.jl:181
  randn(::AbstractRNG, ::Type{T}, ::Integer, ::Integer...) where T at /home/fredrikb/julia-1.3.0-rc4/share/julia/stdlib/v1.3/Random/src/normal.jl:184
  randn(::AbstractRNG) at /home/fredrikb/julia-1.3.0-rc4/share/julia/stdlib/v1.3/Random/src/normal.jl:39
  ...
Stacktrace:
 [1] randn!(::Random._GLOBAL_RNG, ::Array{Int64,2}) at /home/fredrikb/julia-1.3.0-rc4/share/julia/stdlib/v1.3/Random/src/normal.jl:173
 [2] _rand! at /home/fredrikb/.julia/packages/Distributions/2i7UF/src/multivariate/mvnormal.jl:269 [inlined]
 [3] rand at /home/fredrikb/.julia/packages/Distributions/2i7UF/src/multivariates.jl:77 [inlined]
 [4] rand(::MvNormal{Int64,PDMats.PDiagMat{Int64,Array{Int64,1}},Array{Int64,1}}, ::Int64) at /home/fredrikb/.julia/packages/Distributions/2i7UF/src/multivariates.jl:76
 [5] top-level scope at REPL[10]:1

(v1.3) pkg> st Distributions
    Status `~/.julia/environments/v1.3/Project.toml`
  [31c24e10] Distributions v0.21.5
  [1fd47b50] QuadGK v2.0.3
  [2913bbd2] StatsBase v0.32.0

julia> VERSION
v"1.3.0-rc4.1"
```

devmotion:
This PR fixes and tests sampling from `Dirichlet` with integer-valued parameters which is broken in 0.24.11 and not tested currently:
```julia
julia> rand(Dirichlet(3, 4))
ERROR: InexactError: Int64(2.581230906529874)
Stacktrace:
 [1] Int64 at ./float.jl:710 [inlined]
 [2] convert at ./number.jl:7 [inlined]
 [3] setindex! at ./array.jl:847 [inlined]
 [4] rand!(::Random._GLOBAL_RNG, ::Gamma{Float64}, ::Array{Int64,1}) at /home/david/.julia/packages/Distributions/KDTM8/src/univariates.jl:165
 [5] _rand!(::Random._GLOBAL_RNG, ::Dirichlet{Int64,FillArrays.Fill{Int64,1,Tuple{Base.OneTo{Int64}}},Float64}, ::Array{Int64,1}) at /home/david/.julia/packages/Distributions/KDTM8/src/multivariate/dirichlet.jl:162
 [6] rand at /home/david/.julia/packages/Distributions/KDTM8/src/multivariates.jl:76 [inlined]
 [7] rand(::Dirichlet{Int64,FillArrays.Fill{Int64,1,Tuple{Base.OneTo{Int64}}},Float64}) at /home/david/.julia/packages/Distributions/KDTM8/src/genericrand.jl:22
 [8] top-level scope at REPL[4]:1
```

IMO the main underlying problem is that `rand` falls back to the in-place method `rand!` and uses a heuristic to determine the type of the samples. This problem seems similar to the one described (and fixed) for `(log)pdf` in https://github.com/JuliaStats/Distributions.jl/pull/1257. However, I did not want to make any major to Distributions in this PR and therefore it is just a fix for the problem with `Dirichlet`.

*Edit:* This PR also fixes #1004 and similar issues for other distributions by using containers with floating point numbers for sampling from continuous distributions in the default implementation of `rand`. E.g, the PR fixes the following example as well:
```julia
julia> d = MvNormal([0, 0], [1, 1]);

julia> rand(d)
ERROR: MethodError: no method matching randn(::Random._GLOBAL_RNG, ::Type{Int64})
Closest candidates are:
  randn(::Random.AbstractRNG, ::Type{T}, ::Tuple{Vararg{Int64,N}} where N) where T at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Random/src/normal.jl:201
  randn(::Random.AbstractRNG, ::Type{T}, ::Integer, ::Integer...) where T at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Random/src/normal.jl:204
  randn(::Random.AbstractRNG) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Random/src/normal.jl:38
  ...
Stacktrace:
 [1] randn!(::Random._GLOBAL_RNG, ::Array{Int64,1}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Random/src/normal.jl:178
 [2] _rand! at /home/david/.julia/packages/Distributions/7Ifyw/src/multivariate/mvnormal.jl:275 [inlined]
 [3] rand at /home/david/.julia/packages/Distributions/7Ifyw/src/multivariates.jl:76 [inlined]
 [4] rand(::MvNormal{Int64,PDMats.PDiagMat{Int64,Array{Int64,1}},Array{Int64,1}}) at /home/david/.julia/packages/Distributions/7Ifyw/src/genericrand.jl:22
 [5] top-level scope at REPL[5]:1
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.24.11"
+version = "0.24.12"
 
 [deps]
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
--- a/src/matrixvariates.jl
+++ b/src/matrixvariates.jl
@@ -130,10 +130,14 @@ end
 # multiple matrix-variates, must allocate array of arrays
 rand(rng::AbstractRNG, s::Sampleable{Matrixvariate}, dims::Dims) =
     rand!(rng, s, Array{Matrix{eltype(s)}}(undef, dims), true)
+rand(rng::AbstractRNG, s::Sampleable{Matrixvariate,Continuous}, dims::Dims) =
+    rand!(rng, s, Array{Matrix{float(eltype(s))}}(undef, dims), true)
 
 # single matrix-variate, must allocate one matrix
 rand(rng::AbstractRNG, s::Sampleable{Matrixvariate}) =
     _rand!(rng, s, Matrix{eltype(s)}(undef, size(s)))
+rand(rng::AbstractRNG, s::Sampleable{Matrixvariate,Continuous}) =
+    _rand!(rng, s, Matrix{float(eltype(s))}(undef, size(s)))
 
 # single matrix-variate with pre-allocated matrix
 function rand!(rng::AbstractRNG, s::Sampleable{Matrixvariate},
--- a/src/multivariate/dirichlet.jl
+++ b/src/multivariate/dirichlet.jl
@@ -46,7 +46,7 @@ end
 
 length(d::DirichletCanon) = length(d.alpha)
 
-Base.eltype(::Type{<:Dirichlet{T}}) where {T} = T
+Base.eltype(::Type{<:Dirichlet{T}}) where {T} = float(T)
 
 #### Conversions
 convert(::Type{Dirichlet{T}}, cf::DirichletCanon) where {T<:Real} =
--- a/src/multivariate/dirichlet.jl
+++ b/src/multivariate/dirichlet.jl
@@ -46,7 +46,7 @@ end
 
 length(d::DirichletCanon) = length(d.alpha)
 
-Base.eltype(::Type{<:Dirichlet{T}}) where {T} = float(T)
+Base.eltype(::Type{<:Dirichlet{T}}) where {T} = T
 
 #### Conversions
 convert(::Type{Dirichlet{T}}, cf::DirichletCanon) where {T<:Real} =
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -278,7 +278,7 @@ _rand!(rng::AbstractRNG, d::MvNormal, x::VecOrMat) =
 # Workaround: randn! only works for Array, but not generally for AbstractArray
 function _rand!(rng::AbstractRNG, d::MvNormal, x::AbstractVector)
     for i in eachindex(x)
-        @inbounds x[i] = randn(rng,eltype(d))
+        @inbounds x[i] = randn(rng, eltype(x))
     end
     add!(unwhiten!(d.Σ, x), d.μ)
 end
--- a/src/multivariates.jl
+++ b/src/multivariates.jl
@@ -69,12 +69,18 @@ end
 rand(s::Sampleable{Multivariate}, n::Int) = rand(GLOBAL_RNG, s, n)
 rand(rng::AbstractRNG, s::Sampleable{Multivariate}, n::Int) =
     _rand!(rng, s, Matrix{eltype(s)}(undef, length(s), n))
+rand(rng::AbstractRNG, s::Sampleable{Multivariate,Continuous}, n::Int) =
+    _rand!(rng, s, Matrix{float(eltype(s))}(undef, length(s), n))
 rand(rng::AbstractRNG, s::Sampleable{Multivariate}, dims::Dims) =
-    rand(rng, s, Array{Vector{eltype(s)}}(undef, dims), true)
+    rand!(rng, s, Array{Vector{eltype(s)}}(undef, dims), true)
+rand(rng::AbstractRNG, s::Sampleable{Multivariate,Continuous}, dims::Dims) =
+    rand!(rng, s, Array{Vector{float(eltype(s))}}(undef, dims), true)
 
 # single multivariate, must allocate vector
 rand(rng::AbstractRNG, s::Sampleable{Multivariate}) =
     _rand!(rng, s, Vector{eltype(s)}(undef, length(s)))
+rand(rng::AbstractRNG, s::Sampleable{Multivariate,Continuous}) =
+    _rand!(rng, s, Vector{float(eltype(s))}(undef, length(s)))
 
 ## domain
 
--- a/src/univariate/continuous/normal.jl
+++ b/src/univariate/continuous/normal.jl
@@ -240,7 +240,7 @@ cf(d::Normal, t::Real) = exp(im * t * d.μ - d.σ^2 / 2 * t^2)
 
 #### Sampling
 
-rand(rng::AbstractRNG, d::Normal{T}) where {T} = d.μ + d.σ * randn(rng, T)
+rand(rng::AbstractRNG, d::Normal{T}) where {T} = d.μ + d.σ * randn(rng, float(T))
 
 #### Fitting
 
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -157,6 +157,8 @@ end
 # multiple univariate, must allocate array
 rand(rng::AbstractRNG, s::Sampleable{Univariate}, dims::Dims) =
     rand!(rng, sampler(s), Array{eltype(s)}(undef, dims))
+rand(rng::AbstractRNG, s::Sampleable{Univariate,Continuous}, dims::Dims) =
+    rand!(rng, sampler(s), Array{float(eltype(s))}(undef, dims))
 
 # multiple univariate with pre-allocated array
 function rand!(rng::AbstractRNG, s::Sampleable{Univariate}, A::AbstractArray)
--- a/test/dirichlet.jl
+++ b/test/dirichlet.jl
@@ -16,7 +16,7 @@ rng = MersenneTwister(123)
         d = Dirichlet(3, T(2))
 
         @test length(d) == 3
-        @test eltype(d) === Float64
+        @test eltype(d) === T
         @test d.alpha == [2, 2, 2]
         @test d.alpha0 == 6
 
@@ -46,7 +46,7 @@ rng = MersenneTwister(123)
         v = [2, 1, 3]
         d = Dirichlet(T.(v))
 
-        @test eltype(d) === Float64
+        @test eltype(d) === T
         @test Dirichlet([2, 1, 3]).alpha == d.alpha
 
         @test length(d) == length(v)
--- a/test/dirichlet.jl
+++ b/test/dirichlet.jl
@@ -12,100 +12,104 @@ rng = MersenneTwister(123)
     Dict("rand(...)" => [rand, rand],
          "rand(rng, ...)" => [dist -> rand(rng, dist), (dist, n) -> rand(rng, dist, n)])
 
-d = Dirichlet(3, 2.0)
+    for T in (Int, Float64)
+        d = Dirichlet(3, T(2))
 
-@test length(d) == 3
-@test d.alpha == [2.0, 2.0, 2.0]
-@test d.alpha0 == 6.0
+        @test length(d) == 3
+        @test eltype(d) === Float64
+        @test d.alpha == [2, 2, 2]
+        @test d.alpha0 == 6
 
-@test mean(d) ≈ fill(1.0/3, 3)
-@test cov(d)  ≈ [8 -4 -4; -4 8 -4; -4 -4 8] / (36 * 7)
-@test var(d)  ≈ diag(cov(d))
+        @test mean(d) ≈ fill(1/3, 3)
+        @test cov(d)  ≈ [8 -4 -4; -4 8 -4; -4 -4 8] / (36 * 7)
+        @test var(d)  ≈ diag(cov(d))
 
-@test pdf(Dirichlet([1, 1]), [0, 1]) ≈ 1.0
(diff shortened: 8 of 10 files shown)


andreasnoack:
I think this one is a bit different than the `logpdf` case. For the `logpdf` case, I believe the result type should be possible to compute from the inputs. However, we've struggled with `rand` since it's now clear what should determine the type of the random variates. I think we have a couple of open issues and PRs related to that problem. I believe almost all the continuous distributions return `Float64` except for `Normal`.

In any case, I don't think this PR is the right fix since it "lies" about the element type. The element type usually refers to the type of the parameters and not the type of the sample space. I think the simplest fix would be to promote to floats in https://github.com/JuliaStats/Distributions.jl/blob/0da0bcb622cdb567237be6377ecbca4401ba0eb3/src/multivariate/dirichlet.jl#L28-L35. Alternative, we should change https://github.com/JuliaStats/Distributions.jl/blob/0da0bcb622cdb567237be6377ecbca4401ba0eb3/src/multivariates.jl#L77 to versions for continuous and discrete distributions and hardcode for `Float64` and `Int` respectively. I believe the former is the less controversial solution.

devmotion:
Thanks for your comment! I think you touch on different issues, I'll try to respond as briefly as possible.

>  However, we've struggled with rand since it's now clear what should determine the type of the random variates.

IMO the natural approach is to let users specify the type of the samples of the underlying RNG, which would be `Float64` by default but could be changed to e.g. `Float32` or `BigFloat`, analogously to `rand(rng, T, size...)` in base. AFAICT every sampling procedure at some point starts by sampling from the definitions of `rand`, `randexp`, or `randn` in Random. The samples from the distributions are obtained by transforming these most basic samples, and therefore the type of the random variates follows automatically from the type of these basic samples and the function that transforms them and involves the parameters of the distribution. Viewed this way, it seems also the result type of `rand` could be computed, similar to `(log)pdf`.

A while ago I mentioned this view also in a discussion about MeasureTheory in which AFAIK the support of a measure is encoded in its type. I think it can be difficult to know it in advance and could be avoided by specifying the type of the most basic sampling calls.

> In any case, I don't think this PR is the right fix since it "lies" about the element type. The element type usually refers to the type of the parameters and not the type of the sample space.

I think it really boils down to the question what exactly `eltype` specifies in Distributions. In Random, the convention is that `eltype` specifies the type of the samples (see, e.g., https://docs.julialang.org/en/v1/stdlib/Random/#An-optimized-sampler-with-pre-computed-data). However, if one sticks with the interpretation that in Distributions `eltype` should specify the type of the parameters, then clearly you are right.

If it should be the type of the parameters (which is e.g. not enforced in the type of [`MvNormal`](https://github.com/JuliaStats/Distributions.jl/blob/01ba56b213ec965c96980312fae85f5e867beca1/src/multivariate/mvnormal.jl#L179) but only its outer constructors), the main problem is that both use cases are mixed in Distributions since the implementation of `rand` uses `eltype` as a heuristic for the type of the samples.

I am aware of the many issues and discussions in Distributions regarding `eltype` and therefore tried to keep everything specific to `Dirichlet` to avoid any more general discussions :slightly_smiling_face: 

> I think the simplest fix would be to promote to floats in 

This would be a simple fix but the allocations seem wasteful and only necessary to fix the incorrect heuristic in `rand`.

> Alternative, we should change to versions for continuous and discrete distributions and hardcode for Float64 and Int respectively.

I think this will break many use cases with non-standard parameter types that are handled by the current heuristic.

Taken together, I think the most natural and possibly appropriate fix would be to change the heuristic of the type of the random variates in the default implementation of `rand` from `eltype(d)` to `float(eltype(d))`.

andreasnoack:
I'm not that worried about the extra allocation. If it matters for a user then it would be easy for a user to change their code to construct the parameter vector with floating-point elements.

> I think this will break many use cases with non-standard parameter types that are handled by the current heuristic.

I'd be surprised if that is really the case since you can't really rely on such behavior for `Distribution`s in general. I'm curious how that would look like.

There is the problem with pretending that the variates follow the `eltype` that you might think the computation has been fully based on the element type. That would be an issue for Dirichlet where the sampling is based on the Gamma samplers where you have
```julia
julia> typeof(rand(Gamma(0.5f0)))
Float64

julia> typeof(rand(Gamma(0.5)))
Float64

julia> typeof(rand(Gamma(big(0.5))))
BigFloat
```
but all these numbers are based on the `Float64` variates in https://github.com/JuliaStats/Distributions.jl/blob/01ba56b213ec965c96980312fae85f5e867beca1/src/samplers/gamma.jl#L180. I'm not sure we should pretend that random number generation is generic.

devmotion:
If `alpha` is the integer-valued output of another function that the user can't control, it is not possible to avoid the allocations. In some experiments, I had to construct and evaluate thousands of different Dirichlet distributions on a 1000-dimensional probability simplex in a loop. It seems surprising to convert the parameters at construction if everything but sampling works fine and the only problem is the heuristic in
https://github.com/JuliaStats/Distributions.jl/blob/01ba56b213ec965c96980312fae85f5e867beca1/src/multivariates.jl#L76-L77

An example that would be broken by hardcoding `Int` and `Float64` instead of using `eltype(s)` in this line would be:
```julia
julia> using Distributions, ForwardDiff

julia> s = MvNormal(ForwardDiff.Dual.(ones(10)));

julia> rand(s)
10-element Array{ForwardDiff.Dual{Nothing,Float64,0},1}:
  Dual{Nothing}(0.1666028548436855)
 Dual{Nothing}(-0.43914145214951167)
 Dual{Nothing}(-1.0691807830172526)
  Dual{Nothing}(1.8454427275951966)
  Dual{Nothing}(0.39196945140339634)
 Dual{Nothing}(-1.046448411408746)
 Dual{Nothing}(-0.7625542381972902)
  Dual{Nothing}(0.5384999518634571)
 Dual{Nothing}(-1.266646900649526)
 Dual{Nothing}(-0.37844015490701083)

julia> eltype(s)
ForwardDiff.Dual{Nothing,Float64,0}
```
Basically, anything where `eltype(s)` would be not `Float64`.

Instead I would suggest adding
```julia
 rand(rng::AbstractRNG, s::Sampleable{Multivariate,Continuous}) = 
     _rand!(rng, s, Vector{float(eltype(s))}(undef, length(s))) 
```
(and similarly for other `rand` calls). This seems like a minor change that
- would solve the example in the OP,
- would allow to use integer-valued parameters without any conversions and allocations at construction,
- would allow to use `eltype(::Type{<:Dirichlet{T}}) where T = T` that actually reports the type of the parameters,
- would avoid any substantial changes and general discussions of the `eltype` setup and sampling.

> There is the problem with pretending that the variates follow the eltype that you might think the computation has been fully based on the element type. 

I'm sorry, isn't this exactly what currently Distributions assumes in the implementation of `rand`? As I  wrote before, clearly this is only a heuristic that fails e.g. for integer-valued parameters but it seems it could be fixed for the problem at hand by using `float(eltype(s))` for continuous distributions.

> but all these numbers are based on the Float64 variates in . I'm not sure we should pretend that random number generation is generic.

Sorry, I am not sure what these comments refer to.

https://github.com/JuliaStats/Distributions.jl/blob/01ba56b213ec965c96980312fae85f5e867beca1/src/samplers/gamma.jl#L180

is exactly one of these "base methods" that I referred to above. My suggestion was to allow to specify the type of exactly (and only) these base variates, i.e., to change such calls to
```julia
    x = randn(rng, T)
```
where `T` can be specified by the user. The type of the random variates of the distribution of interest would then be determined both by `T`, the parameters of the distribution, and the transformations needed to turn `x` into a sample from the distribution. More concretely, the example above would correspond to the default case `T = Float64` and other examples would be
```julia
julia> typeof(rand(Gamma(0.5f0), Float32))
Float32

julia> typeof(rand(Gamma(0.5), Float32))
Float64

julia> typeof(rand(Gamma(big(0.5)), Float32))
BigFloat

julia> typeof(rand(Gamma(0.5f0), BigFloat)) # requires https://github.com/JuliaLang/julia/pull/35111
BigFloat

julia> typeof(rand(Gamma(0.5f), BigFloat)) # requires https://github.com/JuliaLang/julia/pull/35111
BigFloat

julia> typeof(rand(Gamma(0.5f), BigFloat)) # requires https://github.com/JuliaLang/julia/pull/35111
BigFloat
```
But it seems this discussion is a bit too general here and not needed to fix the problem in the OP.

Coming back to the PR, would it be OK to revert the `eltype` definition and instead add something like
```julia
 rand(rng::AbstractRNG, s::Sampleable{Multivariate,Continuous}) = 
     _rand!(rng, s, Vector{float(eltype(s))}(undef, length(s))) 
```
?

devmotion:
BTW I just checked, the suggested change of `rand` not only fixes the problem with `Dirichlet` described in the OP but also problems with other distributions such as the issue described for MvNormal in https://github.com/JuliaStats/Distributions.jl/issues/1004. Currently,
```julia
julia> d = MvNormal([0, 0], [1, 1]);

julia> rand(d)
ERROR: MethodError: no method matching randn(::Random._GLOBAL_RNG, ::Type{Int64})
Closest candidates are:
  randn(::Random.AbstractRNG, ::Type{T}, ::Tuple{Vararg{Int64,N}} where N) where T at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Random/src/normal.jl:201
  randn(::Random.AbstractRNG, ::Type{T}, ::Integer, ::Integer...) where T at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Random/src/normal.jl:204
  randn(::Random.AbstractRNG) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Random/src/normal.jl:38
  ...
Stacktrace:
 [1] randn!(::Random._GLOBAL_RNG, ::Array{Int64,1}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Random/src/normal.jl:178
 [2] _rand! at /home/david/.julia/packages/Distributions/7Ifyw/src/multivariate/mvnormal.jl:275 [inlined]
 [3] rand at /home/david/.julia/packages/Distributions/7Ifyw/src/multivariates.jl:76 [inlined]
 [4] rand(::MvNormal{Int64,PDMats.PDiagMat{Int64,Array{Int64,1}},Array{Int64,1}}) at /home/david/.julia/packages/Distributions/7Ifyw/src/genericrand.jl:22
 [5] top-level scope at REPL[5]:1
```
whereas with my suggestion
```julia
julia> d = MvNormal([0, 0], [1, 1]);

julia> rand(d)
2-element Array{Float64,1}:
  1.6521247327111197
 -0.2234559539354925
```

andreasnoack:
With these issues, it seems that there isn't a single best solution. There are trade-offs and most of the options have aspects that are annoying. So I'm trying to make sure that we consider some of the implications before applying a fix. A drawback with one solution could be the introduction of allocations and the drawback with another solution could be that it follows a path that pretends that random number generation follows the precision of the parameters. A third solution might make it impossible to create `Dual` random variates but I'd have to admit that I'm not easily convinced that it's of practical relevance to do so.

> My suggestion was to allow to specify the type of exactly (and only) these base variates, i.e., to change such calls to
```julia
x = randn(rng, T)
```

Originally, I thought you were suggesting to use `T=eltype(d)` but I think I now understand your proposal. The default is `T=Float64` and we should then start introducing such `T` arguments in all `rand` methods, right? I think I like that idea.

It wouldn't help with the issue mentioned in my previous post where you could end up with `BigFloat` variates that are based on `Float64` draws which I think is quite unfortunate but maybe it's the smallest of the potential issues. I'd like that the variates have the precision used for the underlying draws but it seems tricky to handle. It might require something like what `Adapt.jl` does for arrays.

Anyway, to avoid that the perfect is the enemy of the good here, I guess we should go with your latest proposal.

devmotion:
I applied the suggestion and added some tests for the uni- and multivariate cases (they revealed that currently `rand(d, ::Dims)` is broken). It seems that at least `MatrixNormal` does not define `eltype` and therefore always falls back to `Float64`.

devmotion:
> I'd like that the variates have the precision used for the underlying draws but it seems tricky to handle.

Yes, I assume ideally internally one would promote the user-specified precision based on the involved parameters.

devmotion:
Thank you for merging the PR! Is it possible to make a new release with the fix? I already bumped the version in this PR :slightly_frowning_face:

PR #1267, opened 2021-01-24
Added Maximum Likelihood Estimation for Beta Distribution (fit_mle)

jg-854:
Included unit tests to check functionality works - ran tests with no errors from my modifications.
Updated documentation too.

### DIFF
--- a/docs/src/fit.md
+++ b/docs/src/fit.md
@@ -38,6 +38,7 @@ The `fit_mle` method has been implemented for the following distributions:
 **Univariate:**
 
 - [`Bernoulli`](@ref)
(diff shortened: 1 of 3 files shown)


andreasnoack:
Thanks for the PR. However, this has been discussed a couple of times in the past and we are not sure if it's generally a good idea that individual distributions implement their own Newton iterations. Is the Hessian guaranteed to be negative definite?

jg-854:
>  Is the Hessian guaranteed to be negative definite?


Yes it is negative definite. I understand your point, a general Newton's method might a be a good idea.

Issue #843, opened 2019-03-17
Truncated should catch truncations outside the support


tpapp:
Eg
```julia
Truncated(Exponential(1.0), -3.0, -1.0)
```
should error, as it leads to an invalid distribution.

Technically, for a continuous distribution `d` with `l, r = extrema(d)`, `Truncated(d, L, R)` should error unless `[l,r]` and `[L,R]` intersect.

(thanks to @mschauer for pointing this out, cf discussion in #842)

matbesancon:
@tpapp I assume this was fixed in #842 ?

tpapp:
No, in general (for `Truncated`) it wasn't, you still can do eg
```julia
julia> Truncated(Gamma(1.0, 2.0), -5.0, -3.0)
Truncated(Gamma{Float64}(α=1.0, θ=2.0), range=(-5.0, -3.0))
```
I think it should be fixed by `Truncated` checking things.

Can you reopen?

matbesancon:
ok sorry for that, I was cleaning things up this morning

Thalassocracy:
This is still an issue now; I'd like to tackle fixing this, but there's a couple layers to this, and I'm curious what the general consensus is on how this issue should be fully resolved.

# Option 1: Allow Truncated Distributions with No Support

Keeping the status quo results in many weird effects like:
```
julia> rand(Distributions.truncated(Distributions.Beta(1, 1), 2, 3))
1.0

julia> rand(Distributions.truncated(Distributions.Beta(1, 1), -5, -1))
0.0

julia> rand(Distributions.truncated(Distributions.Binomial(1, 1), -5, -1))
0

julia> rand(Distributions.truncated(Distributions.Binomial(1, 1), 2, 3))
1
```

There are additional weird effects like:
```
julia> d = Truncated(Gamma(1.0, 2.0), -5.0, -3.0)
Truncated(Gamma{Float64}(α=1.0, θ=2.0), range=(-5.0, -3.0))

julia> maximum(d)
-3.0

julia> minimum(d)
0.0
```

That distribution has a higher reported minimum than a maximum, although both values are ultimately nonsense as the distribution isn't actually supported anywhere; they should theoretically be NaN.

If we continue to allow distributions with no support, we should probably `error` or `warn` on `rand` calls and `NaN` on many other function calls (e.g. `maximum`, `minimum`, `pdf`, `cdf`, `quantile`, etc).

[...]

PR #1273, opened 2021-02-04
Prevent rejection sampling if normal variance is 0

### LINKED ISSUE #1264: rand(truncated(Normal(0, 0), 0.0, Inf)) will crash julia 1.5.3
I found this error on the latest version 0.24.11. Julia will freeze, not Ctrl+C or Ctrl+D will interrupt the execution or close julia itself.

The only solution is to close the terminal or process where julia was running.

Good luck!

Thalassocracy:
Fix #1264

Upon initially investigating this, I discovered that this was not a crash, but instead a process that either does not resolve or a process that takes an extremely long time to resolve. My initial suspicion was that this was an infinite loop. Furthermore, it can be triggered with `rand(truncate(Normal(0,0), x, y))` where `x` and `y` are any numbers that are 0 or of different sign, along with some other cases that I won't immediately delve into. I suspect that the mean of the Normal distribution is also irrelevant, but I did not bother testing this.

We have a method for `rand(d::Truncated{Normal{T}, Continus})` defined at line 140 of `truncated/normal.jl` (https://github.com/JuliaStats/Distributions.jl/blob/863844c88e4153af13996f571fcc612d159de542/src/truncated/normal.jl#L140):

```
function rand(rng::AbstractRNG, d::Truncated{Normal{T},Continuous}) where T <: Real
    d0 = d.untruncated
    μ = mean(d0)
    σ = std(d0)
    if isfinite(μ)
        a = (d.lower - μ) / σ
        b = (d.upper - μ) / σ
        z = randnt(rng, a, b, d.tp)
        return μ + σ * z
    else
        return clamp(μ, d.lower, d.upper)
    end
end
```

`isfinite(mean(truncated(Normal(0,0), 0, 1)))` correctly evaluates to `true`, bringing us to the `randnt` function (although it's worth noting that we're passing exclusively `NaN`, `Inf`, and `-Inf` to the `randnt` function. This is because we are dividing by the standard deviation, which is 0, in order to standardize the bounds.

Within the 2nd part of `randnt`, the first two cases, when triggered, yield acceptable results even for Normal(0,0), e.g. they always yield `0.0` within a reasonable amount of time.

When neither of these cases are triggered, a third case is triggered which, under a wide variety of `x,y` where our underlying distribution is Normal(0,0) and our bounds are `x,y`, never exits.  Under these circumstances, `rho` is always `NaN` and we can only exit the loop when `r < rho` (where `r` is randomly generated). `r < NaN` will never evaluate to `true`, so we never exit the loop.

-----

However, there's no need to actually delve deep into `randnt` to fix this. We should probably just avoid trying to feed `randnt` the "standardized" values from a 0-variance distribution in the first place.

I propose fixing this at the level of `rand(rng::AbstractRNG, d::Truncated{Normal{T},Continuous}) where T <: Real`:

We return the mean when a user attempts to sample from a 0-variance truncated normal distribution.

### DIFF
--- a/src/truncated/normal.jl
+++ b/src/truncated/normal.jl
@@ -142,10 +142,14 @@ function rand(rng::AbstractRNG, d::Truncated{Normal{T},Continuous}) where T <: R
     μ = mean(d0)
     σ = std(d0)
     if isfinite(μ)
-        a = (d.lower - μ) / σ
-        b = (d.upper - μ) / σ
-        z = randnt(rng, a, b, d.tp)
-        return μ + σ * z
+        if σ > 0
+            a = (d.lower - μ) / σ
+            b = (d.upper - μ) / σ
+            z = randnt(rng, a, b, d.tp)
+            return μ + σ * z
+        else
+            return μ
+        end
     else
         return clamp(μ, d.lower, d.upper)
     end


Thalassocracy:
On the other hand, an issue with my "fix":

```
julia> rand(Distributions.truncated(Distributions.Normal(0, 0), 0, Inf))
0.0

julia> rand(Distributions.truncated(Distributions.Normal(-1, 0), 0, Inf))
-1.0
```

It's probably also necessary to prevent (via error) the creation of something like `Distributions.truncated(Distributions.Normal(-1, 0), 0, Inf)` in the first place, since the distribution has no support at all. At the least, there should be a warning.

On the other hand, this behavior is consistent with other behavior currently observed with truncated distributions with no remaining support, such as:
```
julia> rand(Distributions.truncated(Distributions.Beta(1, 1), 2, 3))
1.0

julia> rand(Distributions.truncated(Distributions.Beta(1, 1), -5, -1))
0.0

julia> rand(Distributions.truncated(Distributions.Binomial(1, 1), -5, -1))
0

julia> rand(Distributions.truncated(Distributions.Binomial(1, 1), 2, 3))
1
```

This has been mentioned before back in 2019 in #843, but it does not seem like anything has been done since then.

PR #1277, opened 2021-02-08
Speed up pdf, cdf, mean, and var for FisherNoncentralHypergeometric

andreasnoack:
by implementing the method of Liao and Rosen. For the case in https://github.com/JuliaStats/HypothesisTests.jl/issues/224, I saw a 200k times speed up for `cdf` with this change.

### DIFF
--- a/src/univariate/discrete/noncentralhypergeometric.jl
+++ b/src/univariate/discrete/noncentralhypergeometric.jl
@@ -55,36 +55,164 @@ FisherNoncentralHypergeometric(ns::Integer, nf::Integer, n::Integer, ω::Integer
 convert(::Type{FisherNoncentralHypergeometric{T}}, ns::Real, nf::Real, n::Real, ω::Real) where {T<:Real} = FisherNoncentralHypergeometric(ns, nf, n, T(ω))
 convert(::Type{FisherNoncentralHypergeometric{T}}, d::FisherNoncentralHypergeometric{S}) where {T<:Real, S<:Real} = FisherNoncentralHypergeometric(d.ns, d.nf, d.n, T(d.ω))
 
-# Properties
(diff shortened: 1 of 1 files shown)


src/univariate/discrete/noncentralhypergeometric.jl
```diff
@@ -55,36 +55,164 @@ FisherNoncentralHypergeometric(ns::Integer, nf::Integer, n::Integer, ω::Integer
...
+            fₖ = fᵢ
+        end
+    end
+    fᵢ = one(ω)
+    for i in (η - 1):-1:l
+        rᵢ₊ = (d.ns - i)*ω/((i + 1)*(d.nf - d.n  + i + 1))*(d.n - i)
+        fᵢ /= rᵢ₊
+
+        # break if terms no longer contribute to s
+        sfᵢ = s + fᵢ
```
pdeffebach, line 108: Should we be explicit about the tolerance here? Is there some assumption here about how small `f_i` gets?

pdeffebach (commented):
The calculation seems to match the paper's and tests pass. So looks good to me, but I am not a great expert on this.

src/univariate/discrete/noncentralhypergeometric.jl
andreasnoack: The idea here is to go as far as the precision allows, i.e. until `f_i` is so small that it can't change `s`. In that way, it shouldn't be necessary to introduce an explicit tolerance.

pdeffebach: sounds good!

andreasnoack:
If anybody has further comments please share them soon. Otherwise, I'll go ahead a merge in a couple of days.

PR #1280, opened 2021-02-13
Von mises range fix

Astro-mh:
This pull request reverts part of #965 which added bound checking to the pdf of several distributions.
This shouldn't be done for the von Mises distribution as it is designed to be periodic, repeating every 2π, so values for the pdf beyond the support of the distribution are entirely valid.

To enable the univariate_bounds tests to continue working with as few changes as possible, this also adds an `isperiodic()` method which is set to false in univariates.jl and then set to true in vonmises.jl

### DIFF
--- a/docs/src/univariate.md
+++ b/docs/src/univariate.md
@@ -30,6 +30,7 @@ rate(::UnivariateDistribution)
 ncategories(::UnivariateDistribution)
 ntrials(::UnivariateDistribution)
 dof(::UnivariateDistribution)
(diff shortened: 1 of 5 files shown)


jerlich:
It is frustrating that this PR (and other PR) related to VonMises are not resolved and merged. The current implementation gives unexpected results. 

It seems that the simplest solution to these issues is to define support on [-pi , pi] and then wrap input (to pdf or cdf) to this range and likewise, always give samples in this range.

PR #1281, opened 2021-02-14
fix the rand!() function for the Beta distribution.

sl-solution:
For the Beta distribution, the compiler cannot infer the type of the output, so this causes unnecessary memory allocation. Defining the sampler at the argument level of the function solve the problem.

### DIFF
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -161,8 +161,7 @@ rand(rng::AbstractRNG, s::Sampleable{Univariate,Continuous}, dims::Dims) =
     rand!(rng, sampler(s), Array{float(eltype(s))}(undef, dims))
 
 # multiple univariate with pre-allocated array
-function rand!(rng::AbstractRNG, s::Sampleable{Univariate}, A::AbstractArray)
-    smp = sampler(s)
+function rand!(rng::AbstractRNG, s::Sampleable{Univariate}, A::AbstractArray;smp=sampler(s))
     for i in eachindex(A)
         @inbounds A[i] = rand(rng, smp)
     end


andreasnoack:
I'm not sure why we should expose this as an argument. Could you provide more information about the issue that you are fixing here?

sl-solution:
Consider the following example:

`using Random, Distributions
x = zeros(1000)
@time rand!(Beta(0.1,1.0),x)
@time rand!(Beta(0.1,1.0),x)

0.000091 seconds (1.49 k allocations: 23.328 KiB)`

The 23.328 KiB allocations is undesirable.  `@code_typed rand!(Distributions.GLOBAL_RNG,sampler(Beta(0.1,1.0)),x)` reveals that the compiler cannot infer the type of rand(rng, smp).  By adding smp as an argument of the function, the inference of the type is done before the run of the actual body of the function. 
I thought this should be the minimal way to fix this and avoid any side effect.

andreasnoack:
I see. I think it's better to change the loop to an in-place map like
```julia
map!(t -> rand(rng, smp), A, eachindex(A))
```
That gives a function barrier as well.

sl-solution:
Agree, however, I guess (generally) optimising `for loops` (e.g. exploiting AVX)  is easier than `map!` (?)

devmotion:
I opened a PR that introduces a function barrier without adding any keyword argument to `rand!`: https://github.com/JuliaStats/Distributions.jl/pull/1350

andreasnoack:
Superseded by https://github.com/JuliaStats/Distributions.jl/pull/1350

PR #1285, opened 2021-02-23
More efficient `PoissonBinomial`

devmotion:
For many computations, such as e.g. `mean` and `var`, it is not necessary to evaluate the probability mass function of `PoissonBinomial` for all possible outcomes. However, currently, it is computed when `PoissonBinomial` is constructed. For large number of trials this induces a significant delay and might even make it too inefficient to work with `PoissonBinomial` at all.

In this PR, instead the values of the probability mass function are only computed when a user performs an evaluation that requires it.

Simple benchmark on my computer:

## master

```julia
julia> using Distributions, BenchmarkTools, Random

julia> Random.seed!(1234);

julia> p = rand(100);

julia> @btime PoissonBinomial($p);
  4.993 μs (1 allocation: 896 bytes)

julia> @btime PoissonBinomial($p; check_args=false);
  4.885 μs (1 allocation: 896 bytes)

julia> @btime mean(PoissonBinomial($p));
  5.000 μs (1 allocation: 896 bytes)

julia> @btime var(PoissonBinomial($p));
  5.035 μs (3 allocations: 2.63 KiB)

julia> @btime mode(PoissonBinomial($p));
  5.239 μs (1 allocation: 896 bytes)

julia> @btime median(PoissonBinomial($p));
  5.256 μs (2 allocations: 1.75 KiB)

julia> @btime modes(PoissonBinomial($p));
  6.412 μs (50 allocations: 6.23 KiB)

julia> @btime mgf(PoissonBinomial($p), 0.1);
  5.221 μs (2 allocations: 1.75 KiB)

julia> @btime cf(PoissonBinomial($p), 0.1);
  5.380 μs (2 allocations: 2.64 KiB)

julia> @btime entropy(PoissonBinomial($p));
  5.988 μs (2 allocations: 1.75 KiB)

julia> @btime mgf($(PoissonBinomial(p)), 0.1);
  87.389 ns (1 allocation: 896 bytes)

julia> @btime cf($(PoissonBinomial(p)), 0.1);
  337.177 ns (1 allocation: 1.77 KiB)
```

## This PR

```julia
julia> using Distributions, BenchmarkTools, Random

julia> Random.seed!(1234);

julia> p = rand(100);

julia> @btime PoissonBinomial($p);
  90.943 ns (1 allocation: 32 bytes)

julia> @btime PoissonBinomial($p; check_args=false);
  5.545 ns (1 allocation: 32 bytes)

julia> @btime mean(PoissonBinomial($p));
  104.703 ns (0 allocations: 0 bytes)

julia> @btime var(PoissonBinomial($p));
  191.663 ns (0 allocations: 0 bytes)

julia> @btime mode(PoissonBinomial($p));
  5.118 μs (2 allocations: 928 bytes)

julia> @btime median(PoissonBinomial($p));
  5.004 μs (3 allocations: 1.78 KiB)

julia> @btime modes(PoissonBinomial($p));
  6.132 μs (50 allocations: 6.17 KiB)

julia> @btime mgf(PoissonBinomial($p), 0.1);
  136.762 ns (0 allocations: 0 bytes)

julia> @btime cf(PoissonBinomial($p), 0.1);
  332.973 ns (0 allocations: 0 bytes)

julia> @btime entropy(PoissonBinomial($p));
  5.617 μs (2 allocations: 928 bytes)

julia> @btime mgf($(PoissonBinomial(p)), 0.1);
  43.553 ns (0 allocations: 0 bytes)

julia> @btime cf($(PoissonBinomial(p)), 0.1);
  245.246 ns (0 allocations: 0 bytes)
```

This PR is motivated by https://github.com/TuringLang/MCMCChains.jl/pull/263: often it is sufficient to evaluate the mean and the variance (as an uncertainty measure) of the Poisson binomial distribution of the statistic of interest, and therefor the quadratic complexity of the evaluation of the probability mass function slows down the computations for large number of trials (e.g., MCMC samples in this case) unnecessarily.

### DIFF
--- a/src/univariate/discrete/poissonbinomial.jl
+++ b/src/univariate/discrete/poissonbinomial.jl
@@ -1,4 +1,3 @@
-# TODO: this distribution may need clean-up
 """
     PoissonBinomial(p)
 
@@ -24,35 +23,45 @@ External links:
 * [Poisson-binomial distribution on Wikipedia](http://en.wikipedia.org/wiki/Poisson_binomial_distribution)
 
 """
-struct PoissonBinomial{T<:Real} <: DiscreteUnivariateDistribution
-    p::Vector{T}
-    pmf::Vector{T}
+mutable struct PoissonBinomial{T<:Real,P<:AbstractVector{T}} <: DiscreteUnivariateDistribution
+    p::P
+    pmf::Union{Nothing,Vector{T}} # lazy computation of the probability mass function
 
-    function PoissonBinomial{T}(p::AbstractArray) where {T <: Real}
-        pb = poissonbinomial_pdf(p)
-        @assert isprobvec(pb)
-        new{T}(p, pb)
+    function PoissonBinomial{T}(p::AbstractVector{T}; check_args=true) where {T <: Real}
+        check_args && @check_args(PoissonBinomial, all(x -> zero(x) <= x <= one(x), p))
+        return new{T,typeof(p)}(p, nothing)
     end
 end
 
-function PoissonBinomial(p::AbstractArray{T}; check_args=true) where {T <: Real}
-    if check_args
-        for i in eachindex(p)
-            @check_args(PoissonBinomial, 0 <= p[i] <= 1)
+function PoissonBinomial(p::AbstractVector{T}; check_args=true) where {T<:Real}
+    return PoissonBinomial{T}(p; check_args=check_args)
+end
+
+function Base.getproperty(d::PoissonBinomial, x::Symbol)
+    if x === :pmf
+        z = getfield(d, :pmf)
+        if z === nothing
+            y = poissonbinomial_pdf(d.p)
+            isprobvec(y) || error("probability mass function is not normalized")
+            setfield!(d, :pmf, y)
+            return y
+        else
+            return z
         end
+    else
+        return getfield(d, x)
     end
-    return PoissonBinomial{T}(p)
 end
 
 @distr_support PoissonBinomial 0 length(d.p)
 
 #### Conversions
 
-function PoissonBinomial(::Type{PoissonBinomial{T}}, p::Vector{S}) where {T, S}
-    return PoissonBinomial(Vector{T}(p))
+function PoissonBinomial(::Type{PoissonBinomial{T}}, p::AbstractVector{S}) where {T, S}
+    return PoissonBinomial(AbstractVector{T}(p))
 end
 function PoissonBinomial(::Type{PoissonBinomial{T}}, d::PoissonBinomial{S}) where {T, S}
-    return PoissonBinomial(Vector{T}(d.p), check_args=false)
+    return PoissonBinomial(AbstractVector{T}(d.p), check_args=false)
 end
 
 #### Parameters
@@ -67,7 +76,7 @@ partype(::PoissonBinomial{T}) where {T} = T
 #### Properties
 
 mean(d::PoissonBinomial) = sum(succprob(d))
-var(d::PoissonBinomial) = sum(succprob(d) .* failprob(d))
+var(d::PoissonBinomial) = sum(p * (1 - p) for p in succprob(d))
 
 function skewness(d::PoissonBinomial{T}) where {T}
     v = zero(T)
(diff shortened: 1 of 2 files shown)


src/univariate/discrete/poissonbinomial.jl
```diff
@@ -24,35 +23,45 @@ External links:
...
-    p::Vector{T}
-    pmf::Vector{T}
-
-    function PoissonBinomial{T}(p::AbstractArray) where {T <: Real}
-        pb = poissonbinomial_pdf(p)
-        @assert isprobvec(pb)
-        new{T}(p, pb)
+mutable struct PoissonBinomial{T<:Real,P<:AbstractVector{T}} <: DiscreteUnivariateDistribution
+    p::P
+    pmf::Union{Nothing,Vector{T}} # lazy computation of the probability mass function
```
andreasnoack, line 28: Did you consider using an empty vector here? That would allow the struct to be immutable.

devmotion: Yes, I did. I wanted to try this approach first since arguably it is a bit "clearer" (`nothing` clearly indicates that it was not initialized whereas otherwise one has to check `isempty`) and the alternative would require to change `poissonbinomial_pdf` to an in-place function (which is clearly possible though). But I have no strong opinion here, I can compare it to the other approach.

devmotion:
I compared the implementation in this PR with an implementation of an immutable struct that uses `pmf::Vector{T}` which is empty in the uninitialized state and resized and then filled with an in-place variant of `poissonbinomial_pdf` if needed.

# Without `sizehint!`

Benchmarks without `sizehint!(pmf, length(p) + 1)` in the constructor:

```julia
julia> using Distributions, BenchmarkTools, Random

julia> Random.seed!(1234);

julia> p = rand(100);

julia> @btime PoissonBinomial($p);
  110.461 ns (1 allocation: 80 bytes)

julia> @btime PoissonBinomial($p; check_args=false);
  31.534 ns (1 allocation: 80 bytes)

julia> @btime mean(PoissonBinomial($p));
  121.792 ns (1 allocation: 80 bytes)

julia> @btime var(PoissonBinomial($p));
  217.599 ns (1 allocation: 80 bytes)

julia> @btime mode(PoissonBinomial($p));
  4.675 μs (4 allocations: 960 bytes)

julia> @btime median(PoissonBinomial($p));
  4.706 μs (5 allocations: 1.81 KiB)

julia> @btime modes(PoissonBinomial($p));
  5.808 μs (52 allocations: 6.20 KiB)

julia> @btime mgf(PoissonBinomial($p), 0.1);
  152.058 ns (1 allocation: 80 bytes)

julia> @btime cf(PoissonBinomial($p), 0.1);
  360.134 ns (1 allocation: 80 bytes)

julia> @btime entropy(PoissonBinomial($p));
  5.155 μs (4 allocations: 960 bytes)

julia> @btime mgf($(PoissonBinomial(p)), 0.1);
  42.884 ns (0 allocations: 0 bytes)

julia> @btime cf($(PoissonBinomial(p)), 0.1);
  246.104 ns (0 allocations: 0 bytes)
```

# With `sizehint!`

With `sizehint!(pmf, length(p) + 1)` in the constructor I get:

```julia
julia> using Distributions, BenchmarkTools, Random

julia> Random.seed!(1234);

julia> p = rand(100);

julia> @btime PoissonBinomial($p);
  139.443 ns (2 allocations: 896 bytes)

julia> @btime PoissonBinomial($p; check_args=false);
  57.826 ns (2 allocations: 896 bytes)

julia> @btime mean(PoissonBinomial($p));
  146.683 ns (2 allocations: 896 bytes)

julia> @btime var(PoissonBinomial($p));
  240.535 ns (2 allocations: 896 bytes)

julia> @btime mode(PoissonBinomial($p));
  4.688 μs (4 allocations: 960 bytes)

julia> @btime median(PoissonBinomial($p));
  4.714 μs (5 allocations: 1.81 KiB)

julia> @btime modes(PoissonBinomial($p));
  5.825 μs (52 allocations: 6.20 KiB)

julia> @btime mgf(PoissonBinomial($p), 0.1);
  175.456 ns (2 allocations: 896 bytes)

julia> @btime cf(PoissonBinomial($p), 0.1);
  385.257 ns (2 allocations: 896 bytes)

julia> @btime entropy(PoissonBinomial($p));
  5.162 μs (4 allocations: 960 bytes)

julia> @btime mgf($(PoissonBinomial(p)), 0.1);
  42.805 ns (0 allocations: 0 bytes)

julia> @btime cf($(PoissonBinomial(p)), 0.1);
  244.248 ns (0 allocations: 0 bytes)
```

# Summary

So in both cases the number of allocations is larger than in this PR and the functions that do not require the evaluation of the probability mass function are slower (which I tried to optimize in this PR). I did not expect these results - what's the explanation for them? Are the compiler optimizations for `Union{Nothing,T}` just "too" good?

andreasnoack:
Thanks for benchmarking this. The compiler optimizations for the small union here might be really good. Why would e.g. the `mean` function allocate in the latter case but not in the `Union` case?

devmotion:
> Why would e.g. the mean function allocate in the latter case but not in the Union case?

Yes, IMO this was the most surprising result. I assume that we just see the allocation of the empty vector (80 bytes) in the benchmark without `sizehint!` (with `sizehint!` it increases to 896 bytes) whereas in the `Union` case no vectors have to be allocated. Therefore the number of allocations is reduced and it is faster.

andreasnoack:
This is puzzling to me
```julia
julia> @btime PoissonBinomial($p);
  90.943 ns (1 allocation: 32 bytes)
...
julia> @btime mean(PoissonBinomial($p));
  104.703 ns (0 allocations: 0 bytes)
```
I'd think that the latter would time construction as well as the mean calculation whereas the first would only time the construction. Why is the first allocating when the second one isn't?

devmotion:
Yes, this was surprising to me as well. One can get rid of the allocation by changing the returned value:
```julia
julia> f(p) = (PoissonBinomial(p); return)

julia> @btime f($p);
  88.468 ns (0 allocations: 0 bytes)
```
I just don't fully understand why one gets this additional allocation. I inspected the output of `@code_llvm` for both cases, and it seems the main/only difference apart from the returned value are the additional statements
```
; │││ @ /home/david/.julia/dev/Distributions/src/univariate/discrete/poissonbinomial.jl:32 within `_#64'
     %44 = bitcast %jl_value_t*** %ptls to i8*
     %45 = call noalias nonnull %jl_value_t* @jl_gc_pool_alloc(i8* %44, i32 1424, i32 32) #2
     %46 = bitcast %jl_value_t* %45 to %jl_value_t**
     %47 = getelementptr %jl_value_t*, %jl_value_t** %46, i64 -1
     store %jl_value_t* inttoptr (i64 140610215652240 to %jl_value_t*), %jl_value_t** %47
     %48 = bitcast %jl_value_t* %45 to %jl_value_t**
     store %jl_value_t* %13, %jl_value_t** %48, align 8
     %49 = bitcast %jl_value_t* %45 to i8*
     %50 = getelementptr inbounds i8, i8* %49, i64 8
     %51 = bitcast i8* %50 to %jl_value_t**
     store %jl_value_t* inttoptr (i64 140610869056864 to %jl_value_t*), %jl_value_t** %51, align 8
```
in the version with one allocation. So I assume in the non-allocating case (and also the `mean` etc. examples) the `PoissonBinomial` instance is never instantiated but the call of `new{T,typeof(p)}(p, nothing)` is removed by the compiler?

devmotion:
How should we move on with this PR?

andreasnoack:
I think we should just stick with your current implementation since the alternative one didn't show improvements.

PR #1300, opened 2021-03-31
Add Generalized inverse Gaussian Distribution

samozm:
Style is mostly taken from the Normal-inverse Gaussian distribution and the Contributing Requirements page.

Random sampling is implemented according to [Hörmann & Leydold (2014)](https://doi.org/10.1007/s11222-013-9387-3).

As noted in [this PR](https://github.com/JuliaStats/Distributions.jl/pull/587), the CDF is not in closed form and thus is not implemented here.


azev77:
Thanks, have you seen: 
@LMescheder https://github.com/LMescheder/GenInvGaussian.jl
@dylanfesta https://github.com/dylanfesta/MNIG.jl

dylanfesta:
@azev77  The repo you linked was a partial attempt at fitting a multivariate (not univariate) generalized inverse Gaussian.   I was following   Øigård et al, 2005, doi: 10.1016/j.sigpro.2005.03.005  . If there is interest, I would consider revising the code... although it is a bit off-topic here, because this is about the univariate case.

samozm:
@azev77 I had not seen either of these, thanks. In addition to what @dylanfesta said, it appears as though the @LMescheder implementation of sampling may be less efficient than the Hörmann & Leydold algorithm (which is what I used for this implementation).

PR #1305, opened 2021-04-19
make mean(::Chi) more robust

bicycle1885:
Current:
```
julia> mean(Chi(1000))
NaN
```

PR:
```
julia> mean(Chi(1000))
31.614871896968094
```

### DIFF
--- a/src/univariate/continuous/chi.jl
+++ b/src/univariate/continuous/chi.jl
@@ -48,7 +48,7 @@ params(d::Chi) = (d.ν,)
 
 #### Statistics
 
-mean(d::Chi) = (h = d.ν/2; sqrt2 * gamma(h + 1//2) / gamma(h))
+mean(d::Chi) = (h = d.ν/2; sqrt2 * exp(loggamma(h + 1//2) - loggamma(h)))
 
 var(d::Chi) = d.ν - mean(d)^2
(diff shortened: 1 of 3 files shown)


devmotion (commented):
Thanks, this seems reasonable :+1: Can you add the MWE as a test?

bicycle1885:
I tried to generate the continuous_test.ref.json file on my local environment, but it failed with the following error:
```
...
On Pareto(3.0, 2.0) --> Pareto: alpha=3 beta=2 
On Rayleigh() --> Rayleigh: sigma=1 
On Rayleigh(3.0) --> Rayleigh: sigma=3 
On Rayleigh(8.0) --> Rayleigh: sigma=8 
On StudentizedRange(2.0, 2.0) --> StudentizedRange: nu=2 k=2 
Error in distr$pdf(xs) : attempt to apply non-function
Calls: do.main -> eval.points
3: (function () 
   {
       traceback(2)
       q()
   })()
2: eval.points(distr)
1: do.main("continuous_test")
```

I'm using R 4.0.3. Perhaps newer than expected? Sorry, I'm not so familiar with the R language.

devmotion:
Oh, I didn't realize that it's mainly/only tested with the R/json setup. I thought you could just add a check that the example above does not return `NaN` anymore, possibly in a new Julia test file (although, of course, it could also be added to the R-based checks).

bicycle1885:
Great. That's much easier. I've added a test in a new file chi.jl.

devmotion:
Thank you @bicycle1885!

PR #1306, opened 2021-04-20
Faster sampling from `Gumbel`

devmotion:
On master:
```julia

julia> @btime rand($(Gumbel()));
  34.065 ns (0 allocations: 0 bytes)

julia> @btime rand($(Gumbel()), 100);
  3.433 μs (1 allocation: 896 bytes)
```
With this PR:
```julia

julia> @btime rand($(Gumbel()));
  22.429 ns (0 allocations: 0 bytes)

julia> @btime rand($(Gumbel()), 100);
  1.655 μs (1 allocation: 896 bytes)
```

### DIFF
--- a/test/gumbel.jl
+++ b/test/gumbel.jl
@@ -1,0 +1,24 @@
+@testset "Gumbel" begin
+    @testset "eltype" begin
+        @test eltype(Gumbel()) === Float64
+        @test eltype(Gumbel(1f0)) === Float32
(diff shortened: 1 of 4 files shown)


matbesancon:
thanks!

test/gumbel.jl
```diff
@@ -6,15 +6,15 @@
     end
```
matbesancon, line 7: @JuliaRegistrator register()

Issue #1307, opened 2021-04-20
Type inference for mixture distributions


kaandocal:
I am currently using Distributions.jl with mixture distributions (example: mixture of Gaussians) via `MixtureModel`. I have noticed that the compiler seems unable to perform type inference for basic functions such as `mean` and `logpdf` even when the `MixtureModel` struct is fully instantiated. Checking the source code these functions are implemented on the `UnivariateMixture`. I am currently specialising the functions I need for the specialised types I am working with but I was wondering if there isn't a better way of doing that....

devmotion:
If you only work with mixtures of univariate Gaussians, you might want to use [`UnivariateGMM`](https://github.com/JuliaStats/Distributions.jl/blob/master/src/mixtures/unigmm.jl) which is optimized for this use case.

Generally, if there are type inference problems with `MixtureModel` that are caused by the implementation in Distributions, it would be great to fix them. Do you have a concrete example?

kaandocal:
Thanks for the recommendation! Unfortunately right now I'm using different mixture models including Poisson, NB etc.

Any example would be
```
using Distributions

mixture = MixtureModel([ Poisson(2.0), Poisson(10.0) ], Categorical(2))

@code_warntype mean(mixture)
@code_warntype pdf(mixture, 5)
```

Here several of the intermediate variables (including within `_mixpdf1`) don't have inferred types which leads to a significant performance decrease at the moment. I'd be happy to experiment with solutions but I don't want to break anything and `Distributions` is a very heavily used package...

devmotion:
I made a PR with some fixes: https://github.com/JuliaStats/Distributions.jl/pull/1308

kaandocal:
Upon further inspection the problem seems to be that `MixtureModel` does not fully instantiate `Categorical` in its `prior` field. The compiler is then unable to infer the return value of `probs(d)`; after annotating it all the type problems I've encountered in `mean` and `logpdf` go away.

kaandocal:
This seems to be exactly the fix that's needed! Thanks a lot for submitting this.

PR #1308, opened 2021-04-21
Improve type inference of `MixtureModel`

devmotion:
Fixes https://github.com/JuliaStats/Distributions.jl/issues/1307 (or at least type inference of the functions that are mentioned in the issue, there might be some additional problems not covered by the tests).

The main issue was that the `prior` field is not concretely typed (`Categorical{CT}` is not a concrete type) and some hardcoded Float64 that broke type inference with dual numbers.


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.24.17"
+version = "0.24.18"
```
devmotion, line 4: What is the convention in Distributions - is the different type parameter of `MixtureModel` considered a breaking change?

andreasnoack: Yes

andreasnoack (commented):
These are some nice simplifications. We'd have to bump the minor version to avoid breakage.

Project.toml
devmotion: OK, I updated the version number.

devmotion: But now tests fail since there exists no version of HypothesisTests that is compatible with Distributions 0.25 🤔

devmotion: I opened a PR that removes the test dependency on HypothesisTests to fix these test failures: https://github.com/JuliaStats/Distributions.jl/pull/1311

matbesancon:
@devmotion I think we should merge this, but merge any additional breaking change before tagging a new version

matbesancon:
Following a conversation re-started by @StefanKarpinski, we should probably then tag a 1.0

devmotion:
> Following a conversation re-started by @StefanKarpinski, we should probably then tag a 1.0

I am not opposed to tagging a 1.0, but it wouldn't avoid the problem for downstream packages caused by breaking changes - we just have to bump the major version instead of the minor version.

devmotion:
> any additional breaking change before tagging a new version

What other breaking changes are ready to be merged and released? Maybe https://github.com/JuliaStats/Distributions.jl/pull/1286, or at least the bugfixes therein that could be extracted?

matbesancon:
#1286 is not breaking is it?

devmotion:
It changes the type of `LocationScale`, so it is breaking in the same way as this PR.

devmotion:
I guess it would also be a good opportunity to remove the deprecations in the upcoming breaking releases that have existed for quite some time.

devmotion:
How should we proceed? Merge https://github.com/JuliaStats/Distributions.jl/pull/1286 and this PR and then tag 0.25?

matbesancon:
I would prefer keeping the deprecations and removing them when tagging 1.0, so yes let's go with these two PRs

matbesancon:
@devmotion can you merge master into your branch before we merge?

devmotion:
Done :slightly_smiling_face:

PR #1309, opened 2021-04-21
fix fit for Laplace distribution

salbert83:
The current mle fit for Laplace calculates the mean using the median absolute deviation instead of mean absolute deviation.  The former is not MLE.  Also, I added functionality for a weighted Laplace MLE.

### DIFF
--- a/src/univariate/continuous/laplace.jl
+++ b/src/univariate/continuous/laplace.jl
@@ -114,8 +114,27 @@ rand(rng::AbstractRNG, d::Laplace) =
 
 #### Fitting
 
-function fit_mle(::Type{<:Laplace}, x::Array)
-    xc = copy(x)
-    a = median!(xc)
-    Laplace(a, StatsBase.mad!(xc, center=a))
+function fit_mle(::Type{<:Laplace}, x::AbstractArray{T}) where {T <: Real}
+    μ = median(x)
+    return Laplace(μ, mean(abs.(x .- μ)))
+end
+
+function fit_mle(::Type{<:Laplace}, x::AbstractArray{T}, w::AbstractArray{T}) where {T <: Real}
+    sp = sortperm(x)
+    n = length(x)
+    sw = sum(w)
+    highsum = sw
+    lowsum = zero(T)
+    idx = 0
+    for i = 1:n
+        lowsum += w[sp[i]]
+        highsum -= w[sp[i]]
+        if lowsum >= highsum
+            idx = sp[i]
+            break
+        end
+    end
+    μ = x[idx]
+    θ = sum(w .* abs.(x .- μ)) / sw
+    return Laplace(μ, θ)
 end
--- a/src/univariate/continuous/laplace.jl
+++ b/src/univariate/continuous/laplace.jl
@@ -114,36 +114,10 @@ rand(rng::AbstractRNG, d::Laplace) =
 
 #### Fitting
 
-function fit_mle(::Type{<:Laplace}, x::AbstractArray{T}) where {T<:Real}
-    μ = median(x)
-    θ = zero(T)
-    for v ∈ x
-        θ += abs(v - μ)
-    end
-    θ /= length(x)
-    return Laplace(μ, θ)
-end
-
-function fit_mle(::Type{<:Laplace}, x::AbstractArray{T}, w::AbstractArray{T}) where {T <: Real}
-    sp = sortperm(x)
-    n = length(x)
-    sw = sum(w)
-    highsum = sw
-    lowsum = zero(T)
-    idx = 0
-    for i = 1:n
-        lowsum += w[sp[i]]
-        highsum -= w[sp[i]]
-        if lowsum >= highsum
-            idx = sp[i]
-            break
-        end
-    end
-    μ = x[idx]
-    θ = zero(T)
-    for i = 1:length(x)
-        θ += w[i] * abs(x[i] - μ)
-    end
-    θ /= sw
-    return Laplace(μ, θ)
+function fit_mle(::Type{<:Laplace}, x::AbstractArray{<:Real})
+    xc = similar(x)
+    copyto!(xc, x)
+    m = median!(xc)
+    xc .= abs.(x .- m)
+    return Laplace(m, mean(xc))
 end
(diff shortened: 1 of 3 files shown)


src/univariate/continuous/laplace.jl
```diff
@@ -114,8 +114,27 @@ rand(rng::AbstractRNG, d::Laplace) =
 
 #### Fitting
 
-function fit_mle(::Type{<:Laplace}, x::Array)
-    xc = copy(x)
-    a = median!(xc)
-    Laplace(a, StatsBase.mad!(xc, center=a))
+function fit_mle(::Type{<:Laplace}, x::AbstractArray{T}) where {T <: Real}
+    μ = median(x)
+    return Laplace(μ, mean(abs.(x .- μ)))
```
devmotion, line 119: Probably more efficient would be

```suggestion
    return Laplace(μ, StatsBase.meanad(x, μ))
```

src/univariate/continuous/laplace.jl
```diff
@@ -114,8 +114,27 @@ rand(rng::AbstractRNG, d::Laplace) =
 
 #### Fitting
 
-function fit_mle(::Type{<:Laplace}, x::Array)
-    xc = copy(x)
-    a = median!(xc)
-    Laplace(a, StatsBase.mad!(xc, center=a))
+function fit_mle(::Type{<:Laplace}, x::AbstractArray{T}) where {T <: Real}
```
devmotion, line 117: The type parameter can be omitted:
```suggestion
function fit_mle(::Type{<:Laplace}, x::AbstractArray{<:Real})
```

src/univariate/continuous/laplace.jl
devmotion: But maybe it expects two vectors?

salbert83: From StatsBase.jl,

# MeanAD: mean absolute deviation
"""
    meanad(a, b)
Return the mean absolute deviation between two arrays: `mean(abs(a - b))`.
"""
meanad(a::AbstractArray{T}, b::AbstractArray{T}) where {T<:Number} =
    L1dist(a, b) / length(a)

Seems it wants both arguments to be arrays.

salbert83: But yes, this can be implemented more efficiently.  Would you prefer a simple loop + accumulator?

devmotion: I see. I guess the most efficient implementation would then be more similar to the existing one: one could reuse the array that is used for computing the median. `median` creates an array internally IIRC but one can use `median!` instead with a preallocated array `xc` (as in the current implementation) and then update it with `xc .= abs.(x .- m)` where `m` is the median. I'll add a suggestion below.

devmotion, line 119: I would suggest:
```suggestion
    xc = similar(x)
    copyto!(xc, x)
    m = median!(xc)
    xc .= abs.(x .- m)
    return Laplace(m, mean(xc))
```
`similar` ensures (ideally) that `xc` is a mutable array.

salbert83: I made some changes which reduce memory allocation.

devmotion: As mentioned above, `median` creates an array internally and is not more memory efficient than this suggestion.

salbert83: I removed temporary allocations such as x .- m and abs.(...), which I believe allocate memory

salbert83: Yes, I agree, "median" requires an allocation.

src/univariate/continuous/laplace.jl
salbert83: Note I am using the type parameter for "zero(T)" now.

src/univariate/continuous/laplace.jl
salbert83: Thank you very much for recommending efficiency improvements.  My main issue at the moment is the current implementation is not a MLE.  I figured I would add a weighted MLE so the contribution had some additional marginal value.

devmotion: > I removed temporary allocations such as x .- m and abs.(...), which I believe allocate memory

The line
```julia
xc .= abs.(x .- m)
```
would not allocate memory but update `xc[i]` with `abs(x[i] - m)` for each index `i` (assuming both arrays use the same indices) without any temporary arrays.

>  My main issue at the moment is the current implementation is not a MLE. I figured I would add a weighted MLE so the contribution had some additional marginal value.

The bug should be fixed and it's great that you noted the issue and put together a PR. But actually it would be easier to approve the PR and make a bugfix release if it would not include any new features. I would prefer if bug fixes and PRs that add new features would be separated. I would like to hear the opinion of other contributors before approving and merging new features. In particular, IIRC it was mentioned previously that the general policy is to add not too many or too advanced estimation procedures.

So it would be great if you could minimize the changes in the PR and exclude the new estimation procedure, if possible.

salbert83: Very reasonable policy and I like your suggested implementation.   I will put your implementation in a PR for the bug fix.  Thanks.

salbert83: I created a new PR restricted to your implementation of the bug fix.  Thanks!

devmotion (commented):
Great, thanks for condensing the PR! One final question: could you check if it is possible to reduce the tolerance in https://github.com/JuliaStats/Distributions.jl/blob/ac372f0b3b0ca3931b152d81d9746fbfad363fb3/test/fit.jl#L361? I am curious if it is possible to improve the tests which apparently missed this bug.

salbert83:
Yes, I think reducing the tolerance should catch it.  As a minor note, the MLE of the mean for Laplace need not be unique for an even number of samples.  I know median does select a unique value.  I would use "N+1" in place of "N" for the number of generated values.  Then, since there are closed forms for both the location and scale parameters for this distribution and the random seed is fixed, I think a much smaller tolerance would do.  

I actually came across this when testing my implementation of the weighted MLE code, which I will put in a separate pull request.  Thank you!

devmotion (approved):
Thanks again, I updated the tests as well.

PR #1311, opened 2021-04-22
Remove test dependency on HypothesisTests

devmotion:
This PR removes the test dependency on HypothesisTests.jl and instead uses a custom implementation of the KS test.

The reason for this change is that HypothesisTests.jl introduces a circular dependency: it depends on Distributions.jl and therefore tests fail if a breaking release is prepared (see e.g. https://github.com/JuliaStats/Distributions.jl/pull/1308#discussion_r618720401).


johnczito (approved):

johnczito:
Thanks! I somehow convinced myself at the time that this wouldn't happen if it was just a test dependency. My bad.

PR #1314, opened 2021-04-26
Define `Random.Sampler`

devmotion:
This PR defines `Random.Sampler` for `Sampleable`s.

**Motivation:**
Recently I updated [ConsistencyResampling](https://github.com/devmotion/ConsistencyResampling.jl) and switched to the [Random API](https://docs.julialang.org/en/v1/stdlib/Random/#Hooking-into-the-Random-API). The package now works with all distributions for which the Random interface is implemented. Unfortunately, I noticed that it is not possible to use Distributions without committing type piracy since `Random.Sampler` is not implemented (an example of this "fix" is shown in https://devmotion.github.io/ConsistencyResampling.jl/dev/supported/#Vectors-of-distributions).

**Follow-up:**
There are some questions that could lead to additional PRs:
- Is `sampler` still needed or should one switch to `Random.Sampler` completely and deprecate `sampler`?
- Should one commit to the Random interface more closely? E.g., Random uses `rand!(rng, x, sampler)` whereas Distributions supports `rand!(rng, sampler, x)`. Since it is not relevant for my use case I did not touch it in this PR but, if desired, one could implement `rand!(rng, x, sampler)` as well and possibly deprecate `rand!(rng, sampler, x)`.


matbesancon:
All good here, I haven't looked at the API, but simplifying things in Distributions itself wouldn't hurt so if the Random.Sampler API is redundant we could only keep this one.
If we go in this direction, it needs to be done before 1.0. Could you add an issue to track this?

devmotion:
I opened an issue: https://github.com/JuliaStats/Distributions.jl/issues/1316

matbesancon:
@devmotion we can merge this one too for the release

devmotion:
OK, I fixed the merge conflict. I can merge and tag 0.25 if the tests pass?

matbesancon:
yes absolutely

Issue #1254, opened 2021-01-08
support for discrete distributions with Inf in their support gives InexactError error


mattiasvillani:
Example:
support(Poisson(1)) # gives ERROR: InexactError: trunc(Int64, Inf)

The issue is with the line

support(d::Union{D,Type{D}}) where {D<:DiscreteUnivariateDistribution} = round(Int, minimum(d)):round(Int, maximum(d))

since maximum(Poisson(1)) returns Inf, and round(Int, Inf) gives an InexactError.

One solution is to drop the rounding in the above line, but then the support of a discrete distribution is a RealInterval, which is not great. The root problem is probably that there is no Inf of integer type in Julia.

devmotion:
Related: https://github.com/JuliaLang/julia/issues/15562

I think the correct solution would be to use https://github.com/JuliaArrays/InfiniteArrays.jl (or something similar) to represent the support of `Poisson`.

It seems similarly https://github.com/JuliaMath/IntervalSets.jl and https://github.com/JuliaApproximation/DomainSets.jl could be used to represent the support of continuous distributions (also multivariate ones!) more naturally.

PR #1319, opened 2021-05-05
Add `minimum`, `maximum`, `extrema` for `AbstractMvNormal` and `Product`

agerlach:
This PR adds `minimum`, `maximum`, `extrema` for `AbstractMvNormal` and `Product`. The definition extends that currently used for univariate distributions, i.e. `minimum` and `maximum` returns the min/max of the support for each dimension. 

This extends a common interface between Univariate distributions and as subset of MV distributions for querying the bounds of the support for upstream applications such as DiffEqUncertainty, where quadrature is solved over the support.

### DIFF
--- a/docs/src/multivariate.md
+++ b/docs/src/multivariate.md
@@ -71,6 +71,9 @@ invcov(::Distributions.AbstractMvNormal)
 logdetcov(::Distributions.AbstractMvNormal)
 sqmahal(::Distributions.AbstractMvNormal, ::AbstractArray)
 rand(::AbstractRNG, ::Distributions.AbstractMvNormal)
+minimum(::Distributions.AbstractMvNormal)
+maximum(::Distributions.AbstractMvNormal)
+extrema(::Distributions.AbstractMvNormal)
 ```
 
 ### MvLogNormal
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -80,8 +80,25 @@ abstract type AbstractMvNormal <: ContinuousMultivariateDistribution end
 insupport(d::AbstractMvNormal, x::AbstractVector) =
     length(d) == length(x) && all(isfinite, x)
 
+"""
+    minimum(d::AbstractMvNormal)
+
+Return the minimum of the support of `d`.
+"""
 minimum(d::AbstractMvNormal) = fill(-Inf, length(d))
+
+"""
+    maximum(d::AbstractMvNormal)
+
+Return the maximum of the support of `d`.
+"""
 maximum(d::AbstractMvNormal) = fill(Inf, length(d))
+
+"""
+    extrema(d::AbstractMvNormal)
+
+Return the minimum and maximum of the support of `d` as a 2-tuple.
+"""
 extrema(d::AbstractMvNormal) = minimum(d), maximum(d)
 
 mode(d::AbstractMvNormal) = mean(d)
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -80,6 +80,10 @@ abstract type AbstractMvNormal <: ContinuousMultivariateDistribution end
 insupport(d::AbstractMvNormal, x::AbstractVector) =
     length(d) == length(x) && all(isfinite, x)
 
+minimum(d::AbstractMvNormal) = fill(-Inf, length(d))
+maximum(d::AbstractMvNormal) = fill(Inf, length(d))
+extrema(d::AbstractMvNormal) = minimum(d), maximum(d)
+
 mode(d::AbstractMvNormal) = mean(d)
 modes(d::AbstractMvNormal) = [mean(d)]
 
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -83,21 +83,21 @@ insupport(d::AbstractMvNormal, x::AbstractVector) =
 """
     minimum(d::AbstractMvNormal)
 
-Return the minimum of the support of `d`.
+Return the minimum of the support of each dimension of `d`.
 """
-minimum(d::AbstractMvNormal) = fill(-Inf, length(d))
+minimum(d::AbstractMvNormal) = fill(eltype(d)(-Inf), length(d))
 
 """
     maximum(d::AbstractMvNormal)
 
-Return the maximum of the support of `d`.
+Return the maximum of the support of each dimension of `d`.
 """
-maximum(d::AbstractMvNormal) = fill(Inf, length(d))
+maximum(d::AbstractMvNormal) = fill(eltype(d)(Inf), length(d))
 
 """
     extrema(d::AbstractMvNormal)
 
-Return the minimum and maximum of the support of `d` as a 2-tuple.
+Return the minimum and maximum of the support of each dimenison of `d` as a 2-tuple.
 """
 extrema(d::AbstractMvNormal) = minimum(d), maximum(d)
 
--- a/src/multivariate/product.jl
+++ b/src/multivariate/product.jl
@@ -42,23 +42,23 @@ entropy(d::Product) = sum(entropy, d.v)
 insupport(d::Product, x::AbstractVector) = all(insupport.(d.v, x))
 
 """
-    minimum(d::AbstractMvNormal)
+    minimum(d::Product)
 
-Return the minimum of the support of `d`.
+Return the minimum of the support of each dimension of `d`.
 """
 minimum(d::Product) = minimum.(d.v)
 
 """
-    maximum(d::AbstractMvNormal)
+    maximum(d::Product)
 
-Return the maximum of the support of `d`.
+Return the maximum of the support of each dimension of `d`.
 """
 maximum(d::Product) = maximum.(d.v)
 
 """
-    extrema(d::AbstractMvNormal)
+    extrema(d::Product)
 
-Return the minimum and maximum of the support of `d` as a 2-tuple.
+Return the minimum and maximum of the support of each dimension of `d` as a 2-tuple.
 """
 extrema(d::Product) = minimum(d), maximum(d)
 
--- a/src/multivariate/product.jl
+++ b/src/multivariate/product.jl
@@ -40,6 +40,9 @@ var(d::Product) = var.(d.v)
 cov(d::Product) = Diagonal(var(d))
 entropy(d::Product) = sum(entropy, d.v)
 insupport(d::Product, x::AbstractVector) = all(insupport.(d.v, x))
+minimum(d::Product) = minimum.(d.v)
+maximum(d::Product) = maximum.(d.v)
+extrema(d::Product) = minimum(d), maximum(d)
 
 """
     product_distribution(dists::AbstractVector{<:UnivariateDistribution})
--- a/src/multivariate/product.jl
+++ b/src/multivariate/product.jl
@@ -40,8 +40,26 @@ var(d::Product) = var.(d.v)
 cov(d::Product) = Diagonal(var(d))
 entropy(d::Product) = sum(entropy, d.v)
 insupport(d::Product, x::AbstractVector) = all(insupport.(d.v, x))
+
+"""
+    minimum(d::AbstractMvNormal)
+
+Return the minimum of the support of `d`.
+"""
 minimum(d::Product) = minimum.(d.v)
+
+"""
+    maximum(d::AbstractMvNormal)
+
+Return the maximum of the support of `d`.
+"""
 maximum(d::Product) = maximum.(d.v)
+
+"""
+    extrema(d::AbstractMvNormal)
+
+Return the minimum and maximum of the support of `d` as a 2-tuple.
+"""
 extrema(d::Product) = minimum(d), maximum(d)
 
 """
--- a/test/mvnormal.jl
+++ b/test/mvnormal.jl
@@ -29,6 +29,9 @@ function test_mvnormal(g::AbstractMvNormal, n_tsamples::Int=10^6,
     vs = diag(Σ)
     @test g == typeof(g)(params(g)...)
     @test g == deepcopy(g)
+    @test minimum(g) == fill(-Inf, d)
+    @test maximum(g) == fill(Inf, d)
+    @test extrema(g) == (minimum(g), maximum(g))
 
     # test sampling for AbstractMatrix (here, a SubArray):
     if ismissing(rng)
--- a/test/mvnormal.jl
+++ b/test/mvnormal.jl
@@ -32,6 +32,7 @@ function test_mvnormal(g::AbstractMvNormal, n_tsamples::Int=10^6,
     @test minimum(g) == fill(-Inf, d)
     @test maximum(g) == fill(Inf, d)
     @test extrema(g) == (minimum(g), maximum(g))
+    @test isless(extrema(g)...)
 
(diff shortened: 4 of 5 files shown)


devmotion:
It seems weird to me to define `extrema` etc. for spaces without at least a partial order. Isn't it sufficient if you define custom methods (without overloading `minimum`, `maximum`, and `extrema`), e.g., using `mapreduce`, for this purpose wherever you want to use this information?

agerlach:
I believe the implementation provided satisfies the lexicographic order of `isless` for vector spaces. With that said, I don't disagree that a separate function may be more appropriate. Ultimately, what I desire is a common interface for getting the corners of the bounding box of the support of a distribution.

agerlach:
Would making something like the following be something you are willing to have merged into Distributions.jl?

```
boundingbox(d::UnivariateDistribution) = extrema(d)
boundingbox(d::AbstractMvNormal) = ...
boundingbox(d::Product) = ...
```

devmotion:
The docstrings of `maximum`, `minimum`, and `extrema` in Julia (https://docs.julialang.org/en/v1/base/collections/#Base.minimum etc.) let me think that the definitions here would be a misuse of these functions. The definitions for iterators and arrays always perform reductions of all elements (or along the given dimensions).

mschauer:
I first thought the same, but I convinced myself this is correct:

`extrema(D)` gives the numbers a, b such that `isless(a, X) && isless(X <= b)` for  all `X = rand(D)`, because it Base's `extrema` is for `min` and `max` and those are `isless`-extrema.  
So we still have
```
extrema(D) ≈ extrema([rand(D) for _ in 1:N]) # limit N to Inf
```

After figuring it out I added a clarification to the doc string too: 
I updated https://github.com/JuliaLang/julia/pull/40713

devmotion:
I think a better approach would be to define `support` more generally for multi- and matrixvariate distributions. Instead of the current approach for univariate distributions with ranges and custom types (which is broken e.g. for `Poisson`: https://github.com/JuliaStats/Distributions.jl/issues/1254) probably one should make use of packages such as IntervalSets or DomainSets which would also support multivariate distributions (https://github.com/JuliaStats/Distributions.jl/issues/1254#issuecomment-785218433).

However, this would involve more changes and probably take some time until it would be available, so maybe the easiest short-term solution for you would be to define a custom interface in your package instead of in Distributions.

devmotion:
It would also be more consistent with the current univariate implementation since there `maximum` etc. are coupled with `support`.

mschauer:
But it's the right thing to define, the proposed `extrema` does the  reduction of all possible samples with respect to `isless`, just like our other definitions of `extrema` in `Distributions.jl` do. The 1-d `extrema(::Distribution)` are part of the interface and this PR is just the consequence. That it has a use case outside the package is a nice extra.

devmotion:
I didn't know that `isless` is actually defined for vectors and that it is used by `minimum`, `maximum`, etc. Taking this into account, the PR seems reasonable.

It would be good to add `support` for non-univariate distributions in another PR, the disconnect seems a bit unfortunate (I just checked, `DomainSets` defines `minimum` and `maximum` as well). I guess also it should be implemented for more distributions, it seems to cover only these two special types of multivariate distributions now.

src/multivariate/mvnormal.jl
```diff
@@ -80,6 +80,27 @@ abstract type AbstractMvNormal <: ContinuousMultivariateDistribution end
 insupport(d::AbstractMvNormal, x::AbstractVector) =
     length(d) == length(x) && all(isfinite, x)
 
+"""
+    minimum(d::AbstractMvNormal)
+
+Return the minimum of the support of each dimension of `d`.
+"""
```
devmotion, line 87: The docstrings seem a bit redundant, maybe it would be sufficient to have one docstring for `minimum(d::Distribution)` etc.?

src/multivariate/mvnormal.jl
```diff
@@ -80,6 +80,27 @@ abstract type AbstractMvNormal <: ContinuousMultivariateDistribution end
...
+maximum(d::AbstractMvNormal) = fill(eltype(d)(Inf), length(d))
+
+"""
+    extrema(d::AbstractMvNormal)
+
+Return the minimum and maximum of the support of each dimenison of `d` as a 2-tuple.
+"""
+extrema(d::AbstractMvNormal) = minimum(d), maximum(d)
+
```
devmotion, line 103: It is a bit unfortunate that one has to define this (the default implementation in Julia falls back to iterating over the argument which is not supported by distributions). I guess it would be reasonable to only define
```julia
Base.extrema(d::Distribution) = minimum(d), maximum(d)
```
once in Distributions?

src/multivariate/product.jl
```diff
@@ -41,6 +41,27 @@ cov(d::Product) = Diagonal(var(d))
 entropy(d::Product) = sum(entropy, d.v)
 insupport(d::Product, x::AbstractVector) = all(insupport.(d.v, x))
 
+"""
+    minimum(d::Product)
+
+Return the minimum of the support of each dimension of `d`.
+"""
+minimum(d::Product) = minimum.(d.v)
```
devmotion, line 49: Maybe use `map` instead of broadcasting? It seems broadcasting is not needed here, and so it might be good to avoid the heavy broadcasting machinery.

devmotion:
```suggestion
minimum(d::Product) = map(minimum, d.v)
```

src/multivariate/product.jl
```diff
@@ -41,6 +41,27 @@ cov(d::Product) = Diagonal(var(d))
...
+
+"""
+    maximum(d::Product)
+
+Return the maximum of the support of each dimension of `d`.
+"""
+maximum(d::Product) = maximum.(d.v)
```
devmotion, line 56: Same here:
```suggestion
maximum(d::Product) = map(maximum, d.v)
```

src/multivariate/product.jl
```diff
@@ -41,6 +41,27 @@ cov(d::Product) = Diagonal(var(d))
...
+
+"""
+    extrema(d::Product)
+
+Return the minimum and maximum of the support of each dimension of `d` as a 2-tuple.
+"""
+extrema(d::Product) = minimum(d), maximum(d)
```
devmotion, line 63: With the suggestion above this could be removed.

agerlach:
> I guess also it should be implemented for more distributions, it seems to cover only these two special types of multivariate distributions now.

I am happy to add it for other distributions, however, I am not familiar with many of them and it isn't obvious what this should be in all cases. For example, should Dirichlet be
```
 extrema(d:: Dirichlet) = zeros(eltype(d), length(d)), ones(eltype(d), length(d))
```

devmotion (changes requested):
I made some suggestions.

src/multivariate/mvnormal.jl
mschauer: 👍

agerlach: agreed... I was actually surprised it didn't have a fallback when I implemented it

src/multivariate/mvnormal.jl
agerlach: Where, as in which file, should I define this?

mschauer: Docstring of `minimum(::Distribution)` should be just "Return the `minimum` of the support of `d`."

devmotion:
>  For example, should Dirichlet be

Ah, that's a nice example for why it would be useful to define `support` with something more advanced such as DomainSets. According to the definition of `isless` it should be
```julia
function Base.minimum(d::Dirichlet)
    n = length(d)
	return [i < n ? zero(eltype(d)) : one(eltype(d)) for i in 1:n]
end

function Base.maximum(d::Dirichlet)
	return [i == 1 ? one(eltype(d)) : zero(eltype(d)) for i in 1:length(d)]
end
```
But probably `support` would be much more helpful for your use case.

src/multivariate/mvnormal.jl
mschauer: I.e. just widen the signature of 
```
  minimum(d::UnivariateDistribution)

  Return the minimum of the support of d.
```

src/multivariate/product.jl
mschauer: Doesn't need a docstring then, just an implementation

src/multivariate/mvnormal.jl
agerlach: Do I need an implementation for minimum(::Distribution)? Or can just a docstrring be added?

mschauer:
Yes, on slack I recommended to go for

> Moritz Schauer  18 hours ago: I think you want something which gives a “eps”-bounding box [of the support] similar to https://github.com/JuliaStats/Distributions.jl/blob/35125579a77b938f8d4a8bd2be23cbf1f2ddf225/src/functionals.jl#L1

devmotion:
Since the support is an open set? The nice thing about DomainSets and similar packages is that they provide support (pun intended) for open and closed sets and half-open intervals.

agerlach:
> According to the definition of isless it should be
This doesn't seem correct to me. Maybe I'm trivializing this, but the interpretation I am using for satisfying `isless` is the extrema of the support fo the marginals. Here that is always `0..1 -> 0,1`

agerlach:
I agree that `support` would be useful, this implementation essentially defining the smallest interval set containing the support

devmotion:
> the interpretation I am using for satisfying isless is the extrema of the support of the marginals. Here that is always 0..1 -> 0,1

No, `isless` sorts with respect to lexicographical order, i.e.
```julia
julia> isless([0, 1], [0.5, 0.5])
true

julia> isless([0.5, 0.5], [1, 0])
true

julia> isless([0.1, 0.9], [0.11, 0.89])
true

julia> isless([0.1, 0.9], [0.1, 0.91])
true
```

devmotion:
> I agree that support would be useful, this implementation essentially defining the smallest interval set containing the support

No, we would not want a bounding box, `support` should really return the support of the distribution.

agerlach:
> No, we would not want a bounding box, support should really return the support of the distribution.

Agreed completely. Let me clarify, I meant my proposal for  `extrema` is the corners of the smallest interval set containing the support

devmotion:
E.g., with DomainSets one would define
```julia
support(d::Dirichlet) = VectorUnitSimplex{eltype(d)}(length(d))
```

devmotion:
> Let me clarify, I meant my proposal for extrema is the corners of the smallest interval set containing the support

But that is not how `extrema` (and `minimum` and `maximum`) are supposed to be defined. They should return the extrema etc. with respect to `isless` which uses lexicographical order.

agerlach:
> They should return the extrema etc. with respect to isless which uses lexicographical order.
Thats exactly what the corners of the smallest interval set containing the support is

mschauer:
> No, we would not want a bounding box, support should really return the support of the distribution.

That means that the `epsilon`-bounding box needs a different name than `support`, not that it is a bad idea! I mean we define it already `getEndpoints`, thought the name can be improved

devmotion:
No, in the case of Dirichlet it is not. If we sort all elements in its support, i.e., the probability simplex, in lexicographical order, then `[0, 0, ...., 0, 1]` is the minimum and `[1, 0, ...., 0]` is the maximum and they are unique. We just sort with respect to the first element, and then break ties with the second etc. I tried to illustrate this with the example above.

devmotion:
> That means that the epsilon-bounding box 

DomainSets uses `boundingbox` :slightly_smiling_face:

mschauer:
Except that Lebesgue on the unit simplex and the unit cube are mutually singular, so you cannot just define the density `=0` on the rest of the cube not covered by the simplex and integrate... :-D it's tricky, we know.

agerlach:
@devmotion I follow you now!

So, how would you like to proceed. It seems that this type of definition for `extrema` does not extend in general. Having some general `support` function would be awesome long term, but sounds like will be a significant effort. Adding `boundingbox` would satisfy my immediate needs and seems readily doable. However, I do respect that fact that I won't be the one task w/ maintinging it

I could do something like the following, or use IntervalSets/DomainSets for the output
```
boundingbox(d::UnivariateDistribution) = extrema(d)
boundingbox(d::AbstractMvNormal) = fill(-Inf,length(d)), fill(Inf, length(d))
boundingbox(d::Product) = map(minimum, d.v), map(maximum, d.v)
boundingbox(d::Dirichlet) = zeros(length(d)), ones(length(d))
```

agerlach:
where boundingbox always returns the smallest interval set containing the support

mschauer:
Let's add the more general bounding box in a separate PR.

agerlach:
Ok, do you want me to finalize this PR, or close it in favor of a separate boundingbox PR?

mschauer:
Finalize!

docs/src/multivariate.md
```diff
@@ -71,6 +71,9 @@ invcov(::Distributions.AbstractMvNormal)
 logdetcov(::Distributions.AbstractMvNormal)
 sqmahal(::Distributions.AbstractMvNormal, ::AbstractArray)
 rand(::AbstractRNG, ::Distributions.AbstractMvNormal)
+minimum(::Distributions.AbstractMvNormal)
+maximum(::Distributions.AbstractMvNormal)
+extrema(::Distributions.AbstractMvNormal)
```
agerlach, line 76: For some reason the docs for minimum/maximum are not showing. I widened the docstring in Univariates.jl to `d::Distributions`, so I would think it should. `extrema` is showing

agerlach: ah... I see I have a `Base.extrema(d::Distribution)`, but there is not fallback for `minimum` and `maximum`. Please advise.

src/univariates.jl
```diff
@@ -78,26 +78,19 @@ Get the degrees of freedom.
...
 """
 minimum(d::UnivariateDistribution)
 
 """
-    maximum(d::UnivariateDistribution)
+    maximum(d::Distribution)
 
 Return the maximum of the support of `d`.
 """
 maximum(d::UnivariateDistribution)
```
devmotion, line 92: Can you move these lines to `src/common.jl` and replace `d::UnivariateDistribution` with `d::Distribution`?

docs/src/multivariate.md
devmotion: See https://github.com/JuliaStats/Distributions.jl/pull/1319/files#r626651609, the suggestion there should address it.

agerlach:
Tests pass on my end and docs look clean now.

src/common.jl
```diff
@@ -133,6 +133,26 @@ value_support(::Type{T}) where {T<:Distribution} = value_support(supertype(T))
...
+Return the minimum of the support of `d`.
+"""
+minimum(d::Distribution)
+
+"""
+    maximum(d::Distribution)
+
+Return the maximum of the support of `d`.
+"""
+maximum(d::Distribution)
```
mschauer, line 148: I am not familiar with this idiom. What does it mean?

agerlach: I haven't either, its from https://github.com/JuliaStats/Distributions.jl/pull/1319#discussion_r626645113

agerlach: and https://github.com/JuliaStats/Distributions.jl/pull/1319#discussion_r626651609

devmotion (approved):
Looks good to me!

st--:
@devmotion you approved it & the CI checks pass - want to merge it, too? : )

Issue #806, opened 2019-01-08
Feature request: provide L2 norm of each distribution's pdf


ablaom:
Wishing for a method `l2norm(d::Distribution)` that returns the L2 norm of the probability density function for `d`. Our use case is computing the Brier loss and integrated square loss for machine learning models learning probability distributions: [MLJ issue #34](https://github.com/alan-turing-institute/MLJ.jl/issues/34#issuecomment-451936615)

matbesancon:
@ablaom sorry about the delay.
The best option would be to implement `LinearAlgebra.norm` for Distributions. PR welcome, but you can do that in your own repo (this is technically type piracy, but if it's not implemented in Distributions, it's fine)

ablaom:
@fkiraly Do you want to suggest a shortlist of  distributions for which we want to implement the L2 norm aka `LinearAlgebra.norm( •, 2)`? Here are all the final types of `Distributions.Distribution`:

```
MixtureModel
UnivariateGMM
MvNormal
MvNormalCanon
Arcsine
Bernoulli
Beta
BetaBinomial
BetaPrime
Binomial
Biweight
Categorical
Cauchy
Chi
Chisq
Cosine
Dirichlet
DirichletMultinomial
DiscreteUniform
MvLogNormal
Distributions.GenericMvTDist
EdgeworthMean
EdgeworthSum
EdgeworthZ
EmpiricalUnivariateDistribution
Epanechnikov
Erlang
Exponential
FDist
Frechet
Gamma
GeneralizedExtremeValue
GeneralizedPareto
Geometric
Gumbel
Hypergeometric
InverseGamma
InverseGaussian
InverseWishart
KSDist
KSOneSided
Kolmogorov
Laplace
Levy
LocationScale
LogNormal
Logistic
Multinomial
NegativeBinomial
NoncentralBeta
NoncentralChisq
NoncentralF
FisherNoncentralHypergeometric
WalleniusNoncentralHypergeometric
NoncentralT
Normal
NormalCanon
NormalInverseGaussian
Pareto
Poisson
PoissonBinomial
Rayleigh
Semicircle
Skellam
SymTriangularDist
TDist
TriangularDist
Triweight
Truncated
Uniform
VonMises
VonMisesFisher
Weibull
Wishart
```

matbesancon:
PR welcome.
Also since this is a new feature, a page in the documentation would be great

fkiraly:
[...]

PR #1321, opened 2021-05-05
Add function to compute L2 norm of distributions

giordano:
This implements the L2 norm of some distributions, as requested in #806.  I started with continuous univariate distributions for which I was able to work out the integral.  The default method throws an error explaining that the function isn't implemented for the given distribution, but other methods can be added later.  CC: @ablaom @fkiraly

### DIFF
--- a/src/pdfnorm.jl
+++ b/src/pdfnorm.jl
@@ -4,7 +4,7 @@
 Return the square of the L2 norm of the probability density function ``f(x)`` of the distribution `d`:
 
 ```math
-\\int_{S} f(x)^{2} \\mathrm{d} x
+\\big(\\int_{S} f(x)^{2} \\mathrm{d} x \\big)^{1/2}
 ```
 
 where ``S`` is the support of ``f(x)``.
@@ -18,7 +18,7 @@ function pdfsquaredL2norm(d::Beta)
     return α > 0.5 && β > 0.5 ? z : oftype(z, Inf)
 end
 
-pdfsquaredL2norm(d::Cauchy) = inv(twoπ * d.σ)
+pdfsquaredL2norm(d::Cauchy) = inv2π / d.σ
 
 function pdfsquaredL2norm(d::Chi)
     ν = d.ν
--- a/src/pdfnorm.jl
+++ b/src/pdfnorm.jl
@@ -1,0 +1,58 @@
+export pdfL2norm
+
+"""
+    pdfL2norm(d::Distribution)
+
+Return the L2 norm of the probability distribution function `f(x)` of the distribution `d`:
+
+```math
+\\int_{S} f(x)^{2} \\mathrm{d} x
+```
+
+where `S` is the support of `f(x)`.
+"""
+pdfL2norm
+
+pdfL2norm(d::Distribution) = throw(ArgumentError("L2 norm not implemented for $(d)"))
+
+function pdfL2norm(d::Beta{T}) where T
+    if d.α > 0.5 && d.β > 0.5
+        return beta(2 * d.α - 1, 2 * d.β - 1) / beta(d.α, d.β) ^ 2
+    else
+        return T(Inf)
+    end
+end
+
+pdfL2norm(d::Cauchy) =  1 / (d.σ * 2 * π)
+
+function pdfL2norm(d::Chi{T}) where {T}
+    if d.ν > 0.5
+        return (2 ^ (1 - d.ν) * gamma((2 * d.ν - 1) / 2)) / gamma(d.ν / 2) ^ 2
+    else
+        return T(Inf)
+    end
+end
+
+function pdfL2norm(d::Chisq{T}) where {T}
+    if d.ν > 1
+        return gamma(d.ν - 1) / (gamma(d.ν / 2) ^ 2 * 2 ^ d.ν)
+    else
+        return T(Inf)
+    end
+end
+
+pdfL2norm(d::Exponential) = 1 / (2 * d.θ)
+
+function pdfL2norm(d::Gamma{T}) where {T}
+    if d.α > 0.5
+        return (2^(1 - 2 * d.α) * gamma(2 * d.α - 1)) / (gamma(d.α) ^ 2 * d.θ)
+    else
+        return T(Inf)
+    end
+end
+
+pdfL2norm(d::Logistic) = 1 / (6 * d.θ)
+
+pdfL2norm(d::Normal{T}) where {T} = 1 / (2 * sqrt(T(π)) * d.σ)
+
+pdfL2norm(d::Uniform) = 1 / (d.b - d.a)
--- a/src/pdfnorm.jl
+++ b/src/pdfnorm.jl
@@ -1,58 +1,50 @@
-export pdfL2norm
-
 """
-    pdfL2norm(d::Distribution)
+    pdfsquaredL2norm(d::Distribution)
 
-Return the L2 norm of the probability distribution function `f(x)` of the distribution `d`:
+Return the square of the L2 norm of the probability density function ``f(x)`` of the distribution `d`:
 
 ```math
 \\int_{S} f(x)^{2} \\mathrm{d} x
 ```
 
-where `S` is the support of `f(x)`.
+where ``S`` is the support of ``f(x)``.
 """
-pdfL2norm
+pdfsquaredL2norm
 
-pdfL2norm(d::Distribution) = throw(ArgumentError("L2 norm not implemented for $(d)"))
-
-function pdfL2norm(d::Beta{T}) where T
-    if d.α > 0.5 && d.β > 0.5
-        return beta(2 * d.α - 1, 2 * d.β - 1) / beta(d.α, d.β) ^ 2
-    else
-        return T(Inf)
-    end
+function pdfsquaredL2norm(d::Beta)
+    α, β = params(d)
+    z = beta(2 * α - 1, 2 * β - 1) / beta(α, β) ^ 2
+    # L2 norm of the pdf converges only for α > 0.5 and β > 0.5
+    return α > 0.5 && β > 0.5 ? z : oftype(z, Inf)
 end
 
-pdfL2norm(d::Cauchy) =  1 / (d.σ * 2 * π)
+pdfsquaredL2norm(d::Cauchy) = inv(twoπ * d.σ)
 
-function pdfL2norm(d::Chi{T}) where {T}
-    if d.ν > 0.5
-        return (2 ^ (1 - d.ν) * gamma((2 * d.ν - 1) / 2)) / gamma(d.ν / 2) ^ 2
-    else
-        return T(Inf)
-    end
+function pdfsquaredL2norm(d::Chi)
+    ν = d.ν
+    z = (2 ^ (1 - ν) * gamma((2 * ν - 1) / 2)) / gamma(ν / 2) ^ 2
+    # L2 norm of the pdf converges only for ν > 0.5
+    return d.ν > 0.5 ? z : oftype(z, Inf)
 end
 
-function pdfL2norm(d::Chisq{T}) where {T}
-    if d.ν > 1
-        return gamma(d.ν - 1) / (gamma(d.ν / 2) ^ 2 * 2 ^ d.ν)
-    else
-        return T(Inf)
-    end
+function pdfsquaredL2norm(d::Chisq)
+    ν = d.ν
+    z = gamma(d.ν - 1) / (gamma(d.ν / 2) ^ 2 * 2 ^ d.ν)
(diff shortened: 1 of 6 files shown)


azev77:
Thanks for this PR!
I noticed you take a slightly different approach. 
Usually there is a file for each distribution (such as [Cauchy](https://github.com/JuliaStats/Distributions.jl/blob/master/src/univariate/continuous/cauchy.jl).jl) and there are various properties (mean/var/cdf/fit) within that file.

The approach you took is to instead create one file (property.jl) and implement that property for many distributions in that file. I think I like this.

src/pdfnorm.jl
```diff
@@ -0,0 +1,58 @@
...
+    if d.α > 0.5
+        return (2^(1 - 2 * d.α) * gamma(2 * d.α - 1)) / (gamma(d.α) ^ 2 * d.θ)
+    else
+        return T(Inf)
+    end
+end
+
+pdfL2norm(d::Logistic) = 1 / (6 * d.θ)
+
+pdfL2norm(d::Normal{T}) where {T} = 1 / (2 * sqrt(T(π)) * d.σ)
```
devmotion, line 56: Could make use of the irrational constants in StatsFuns:
```suggestion
pdfL2norm(d::Normal) = inv(sqrt4π * d.σ)
```

src/pdfnorm.jl
```diff
@@ -0,0 +1,58 @@
...
+\\int_{S} f(x)^{2} \\mathrm{d} x
+```
+
+where `S` is the support of `f(x)`.
+"""
+pdfL2norm
+
+pdfL2norm(d::Distribution) = throw(ArgumentError("L2 norm not implemented for $(d)"))
+
+function pdfL2norm(d::Beta{T}) where T
```
devmotion, line 18: IMO it would be good if one could avoid depending on the type parameter `T` as a heuristic, here and in the other functions, and instead compute the result as much as possible and then just return `oftype(result, Inf)` if needed.

src/pdfnorm.jl
```diff
@@ -0,0 +1,58 @@
...
+    if d.α > 0.5 && d.β > 0.5
+        return beta(2 * d.α - 1, 2 * d.β - 1) / beta(d.α, d.β) ^ 2
+    else
+        return T(Inf)
+    end
+end
+
+pdfL2norm(d::Cauchy) =  1 / (d.σ * 2 * π)
```
devmotion, line 26: Here one can also use StatsFuns:
```suggestion
pdfL2norm(d::Cauchy) = inv(twoπ * d.σ)
```

giordano:
> The approach you took is to instead create one file (property.jl) and implement that property for many distributions in that file.

I initially considered defining the methods in the distributions files, but development and testing is much faster:
```
julia> @time include("test/pdfnorm.jl");
Test Summary: | Pass  Total
pdf L2 norm   |   28     28
  0.083126 seconds (12.20 k allocations: 954.906 KiB, 0.00% compilation time)
```
compared to the >10 minutes for running the whole test suite, that's a 7500x improvement in productivity

src/pdfnorm.jl
```diff
@@ -0,0 +1,58 @@
+export pdfL2norm
```
devmotion, line 1: IIRC all exports are in src/Distributions.jl.

src/pdfnorm.jl
```diff
@@ -0,0 +1,58 @@
...
+
+```math
+\\int_{S} f(x)^{2} \\mathrm{d} x
+```
+
+where `S` is the support of `f(x)`.
+"""
+pdfL2norm
+
+pdfL2norm(d::Distribution) = throw(ArgumentError("L2 norm not implemented for $(d)"))
```
devmotion, line 16: Is it needed or would be a MethodError sufficient? It seems the error message does not provide any additional information.

src/pdfnorm.jl
giordano: Can I ask you why?  Because the parameter is `T<:Real` so not necessarily a floating point?  If so, can use `float(T)` as type.  Here and in the other functions the integral diverges after a certain cut-off, and it would have a completely different analytic expression (if any), so there isn't much result to compute at all.  For example, for the `Chi` distribution
```julia
julia> using SpecialFunctions

julia> f(ν) = (2 ^ (1 - ν) * gamma((2 * ν - 1) / 2)) / gamma(ν / 2) ^ 2

julia> f(0.5)
Inf

julia> f(0.25)
-0.14523516581055995
```
Extending the analytic solution to `ν < 0.5` doesn't make sense, you get negative values.

giordano: Ok, I thought you meant "compute as much meaningful result as possible in each branch", instead you're saying "always compute the result for the converging branch, and maybe return `Inf` if necessary disregarding the quantity computed before" :wink:

src/pdfnorm.jl
```diff
@@ -0,0 +1,50 @@
+"""
+    pdfL2norm(d::Distribution)
+
+Return the L2 norm of the probability distribution function `f(x)` of the distribution `d`:
+
+```math
+\\int_{S} f(x)^{2} \\mathrm{d} x
```
devmotion, line 7: I just noticed that this is not the L2 norm but rather the squared L2 norm. So I suggest to either change the name of the function or to change the values to the actual L2 norm:

```suggestion
\\big(\\int_{S} f(x)^{2} \\mathrm{d} x \\big)^{1/2}
```

devmotion (changes requested):

src/pdfnorm.jl
```diff
@@ -0,0 +1,50 @@
+"""
+    pdfL2norm(d::Distribution)
+
+Return the L2 norm of the probability distribution function `f(x)` of the distribution `d`:
```
devmotion, line 4:
```suggestion
Return the L2 norm of the probability density function `f(x)` of the distribution `d`:
```
It would be good to include `pdfL2norm` also in the documentation. I guess, it belongs to https://juliastats.org/Distributions.jl/latest/univariate/#Computation-of-statistics even though `pdf` is listed in https://juliastats.org/Distributions.jl/latest/univariate/#Probability-Evaluation.

src/pdfnorm.jl
giordano: Yes, I wanted to ask about the name.  I believe what I'm computing here is what @ablaom and @fkiraly asked for in #806, so I'd rather change the name, I'm not really attached to it.  What do you suggest?

devmotion: I got the impression they were actually interested in the L2 norm (i.e. the square root of the integral), due to their comments and the use of `l2norm(..)^2` in https://github.com/JuliaStats/Distributions.jl/issues/806#issuecomment-490819682. But maybe they would be more interested in the squared L2 norm? It would also be easy to provide both if there is a need for the squared norm.

ablaom: @devmotion @giordano We are interested in the *square* of the l2 norm. This defines the Brier entropy, as given, for example on p 88 of https://www.ism.ac.jp/editsec/aism/pdf/059_1_0077.pdf

ablaom: Indeed, I think `brier_entropy` is as good name for this as any (need to multiply square of l2 norm by -1/2). Later, we will have the `tsallis_entropy` (so we can implement the Tsallis proper scoring rule), etc. 

In the doc-string we can refer to the David paper. (Of course, the definition depends on a choice of dominating measure, which will either be counting measure or Lebesgue measure.)

devmotion: The name `brier_entropy` might be a bit unfortunate though since there exist different versions of the Brier score and the associated entropy in literature. E.g., the definition by David with the counting measure is neither consistent with the original definition by [Brier](https://journals.ametsoc.org/view/journals/mwre/78/1/1520-0493_1950_078_0001_vofeit_2_0_co_2.xml) nor the definition by [Gneiting](https://viterbi-web.usc.edu/~shaddin/cs699fa17/docs/GR07.pdf). Maybe it would be clearer to just define the squared L2 norm here and then downstream packages that want to implement e.g. Brier entropy can choose whatever definition they want?

giordano: The [`distr6`](https://alan-turing-institute.github.io/distr6/) R package, suggested as reference by Franz, uses the name [`pdfSquared2Norm`](https://www.rdocumentation.org/packages/distr6/versions/1.5.0/topics/pdfSquared2Norm).  A bit mouthful but accurate.  To not use camelCase this could be `pdfsquared2norm`?  (I personally prefer snake_case, but it doesn't seem to the convention here, as far as I can see)

devmotion: It seems there exists a `pdfPNorm` as well though: https://alan-turing-institute.github.io/distr6/reference/ExoticStatistics.html#method-pdfPNorm

I think `pdfsquaredL2norm` or `pdfsquared2norm` are reasonable choices. It is probably better to define the squared L2 norm and avoid `sqrt` and `^2` operations, and only define `pdfL2norm(d) = sqrt(pdfsquaredL2norm(d))` later if needed.

fkiraly:
Hm, @giordano, this takes me back - when was it that we discussed this, 2 years ago? Nice that you were still working on this!

Indeed I like the `property.jl` approach, especially since later you may like to dispatch on *two* distributions, e.g., implementing the L2 inner product $\int f_d(x|\theta) f_{d'}(x|\theta')$ where $d,d'$ are two different parametric distributions.

Just to avoid duplication of effort in working out the integrals: I believe @rsonabend1 and @aintoha have worked out a couple of these (entropy and cross-entropy/interaction terms) for various scoring rules, and there are also a number of other papers which have lists of integrals, e.g., for energy scores and integrated Brier (which @rsonabend1 and @aintoha can point to).

Regarding the naming discussion: there is on occasion a bit of terminology confusion between Brier/Gneiting loss ($L(p,x) = -2p(x) + |p|_2^2$) and integrated Brier loss ($L(F,x) = \int (F(y) - \Theta(y-x))^2 dy$) aka RPS, so I'd avoid using the name "Brier" in distribution properties as opposed to scoring rules or losses.

src/pdfnorm.jl
```diff
@@ -0,0 +1,50 @@
...
+pdfsquaredL2norm
+
+function pdfsquaredL2norm(d::Beta)
+    α, β = params(d)
+    z = beta(2 * α - 1, 2 * β - 1) / beta(α, β) ^ 2
+    # L2 norm of the pdf converges only for α > 0.5 and β > 0.5
+    return α > 0.5 && β > 0.5 ? z : oftype(z, Inf)
+end
+
+pdfsquaredL2norm(d::Cauchy) = inv(twoπ * d.σ)
```
devmotion, line 21: A micro-optimization would be
```suggestion
pdfsquaredL2norm(d::Cauchy) = inv2π / d.σ
```

src/pdfnorm.jl
```diff
@@ -0,0 +1,50 @@
...
+function pdfsquaredL2norm(d::Chisq)
+    ν = d.ν
+    z = gamma(d.ν - 1) / (gamma(d.ν / 2) ^ 2 * 2 ^ d.ν)
+    # L2 norm of the pdf converges only for ν > 1
+    return ν > 1 ? z : oftype(z, Inf)
+end
+
+pdfsquaredL2norm(d::Exponential) = 1 / (2 * d.θ)
+
+function pdfsquaredL2norm(d::Gamma{T}) where {T}
```
devmotion, line 39:
```suggestion
function pdfsquaredL2norm(d::Gamma)
```

devmotion (commented):
Two minor suggestions, otherwise the PR looks good.

src/pdfnorm.jl
```diff
@@ -0,0 +1,50 @@
+"""
+    pdfsquaredL2norm(d::Distribution)
+
+Return the square of the L2 norm of the probability density function ``f(x)`` of the distribution `d`:
+
+```math
+\\big(\\int_{S} f(x)^{2} \\mathrm{d} x \\big)^{1/2}
```
devmotion, line 7: It seems you accidentally merged the wrong outdated suggestion :slightly_smiling_face: This was correct before since we agreed to return the square of the L2 norm:
```suggestion
\\int_{S} f(x)^{2} \\mathrm{d} x
```
The more recent (and I think valid) suggestion was to remove the type parameter `T` in the definition for `Gamma`: https://github.com/JuliaStats/Distributions.jl/pull/1321/files#r632790018

giordano: Whooops

devmotion (approved):

giordano:
Bump :slightly_smiling_face:

devmotion (approved):

devmotion:
Sorry for the delay, I wanted to make sure that people have time to comment on it but then I completely forgot the PR :slightly_smiling_face:

PR #1325, opened 2021-05-11
document Dirac

JeffFessler:
as added in #1231

currently its existence is not evident in the docs.
not sure what else (if anything) might need to be done to make its docstring appear in the docs too.
hopefully Documenter takes care f that?

### DIFF
--- a/docs/src/univariate.md
+++ b/docs/src/univariate.md
@@ -149,6 +149,7 @@ Bernoulli
 BetaBinomial
 Binomial
 Categorical
+Dirac
 DiscreteUniform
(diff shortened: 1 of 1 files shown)


devmotion (approved):
> not sure what else (if anything) might need to be done to make its docstring appear in the docs too.
> hopefully Documenter takes care f that?

Yes, this should be sufficient. Thanks a lot!

devmotion:
It is fixed: https://juliastats.org/Distributions.jl/latest/univariate/#Distributions.Dirac

PR #1327, opened 2021-05-12
Remove unused custom pdf broadcasting

andreasnoack:
I noticed these while working on #1326. It's been a while since broadcasting was extended like this so these should be fine to drop.

### DIFF
--- a/src/univariate/discrete/binomial.jl
+++ b/src/univariate/discrete/binomial.jl
@@ -137,51 +137,6 @@ function rand(rng::AbstractRNG, d::Binomial)
     p <= 0.5 ? y : n-y
 end
 
-struct RecursiveBinomProbEvaluator{T<:Real} <: RecursiveProbabilityEvaluator
-    n::Int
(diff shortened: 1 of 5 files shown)


devmotion:
> It's been a while since broadcasting was extended like this so these should be fine to drop.

Is standard broadcasting without these specializations really equivalent? I always assumed that the main purpose of these recursive evaluations is to avoid redundant evaluation of the pdf for different samples by recursively computing everything in one run instead of multiple independent computations. Is this still the case if these implementations are removed?

andreasnoack:
They are not equivalent. The methods I delete here are cheaper. My point is that these methods haven't been called in a long time because broadcasting no longer dispatches to these functions. Hence I doubt that anybody really relies on them being fast. Alternatively, we could update these methods. I just don't think it's worth the effort and right now it's dead code.

devmotion:
Ah I see, then I'm all up for removing the code.

devmotion (approved):

Issue #1336, opened 2021-06-01
Add LKJ distribution for cholesky factor of correlation matrix


sethaxen:
Distributions already has an `LKJ(d, η)` whose support is `d × d` correlation matrices. More useful for probabilistic programming is the equivalent `LKJ` distribution on whose support is the corresponding cholesky factor `L` (or equivalently `U`), because then the cholesky factor can be directly sampled and used, e.g. in the multivariate normal density, without needing to compute the cholesky decomposition.

There are `d(d-1)/2` unique elements in the strict lower triangle `tril₋(R)` of a correlation matrix `R=L L^T`. Likewise, there are `d(d-1)/2` unique elements in the strict lower triangle `tril₋(L)` of its cholesky factor `L`. The map `ϕ: tril₋(R) → tril₋(L)` is bijective, and the log determinant of its Jacobian is `logdet(J) = \sum_{i=2}^d (i - d) \log(L_{ii})`. The resulting `logkernel` is only adjusted by `-logdet(J)` and works out to `logkernel(L) = \sum_{i=2}^d  (d - i + 2η - 2)) \log(L_{ii})`. `logc0` is unchanged. As a result, we can trivially adapt the code from `LKJ` to define this distribution.

It would also be useful to have a convenient constructor for `MvNormal` that could take a mean vector, the cholesky factor of a correlation matrix, and a vector of standard deviations, or alternatively, a mean vector and the cholesky factor of the covariance matrix.

## Related work

AltDistributions.jl [implements](https://github.com/tpapp/AltDistributions.jl/blob/v0.2.0/src/AltDistributions.jl#L116-L149) this distribution as `LKJL` (however, it pre-dates `LKJ` in this package). Likewise, this distribution is defined in several probabilistic programming languages, notably Stan (see [documentation](https://mc-stan.org/docs/2_26/functions-reference/cholesky-lkj-correlation-distribution.html#probability-density-function-33) and [implementation](https://github.com/stan-dev/math/blob/v4.0.1/stan/math/prim/prob/lkj_corr_cholesky_lpdf.hpp#L18-L53)).

See also discussions in https://github.com/TuringLang/Turing.jl/issues/1629#issuecomment-851959832 and https://github.com/TuringLang/Bijectors.jl/issues/134#issuecomment-679096131.

devmotion:
[...]

PR #1339, opened 2021-06-06
Add LKJCholesky

sethaxen:
This PR adds an `LKJCholesky` distribution whose support are `LinearAlgebra.Cholesky` factorizations of correlation matrices, as discussed in #1336. Essentially all of the functionality is there; I just need to add tests.

### DIFF
--- a/src/cholesky/lkjcholesky.jl
+++ b/src/cholesky/lkjcholesky.jl
@@ -108,6 +108,13 @@ end
 
 logpdf(d::LKJCholesky, R::Cholesky) = logkernel(d, R) + d.logc0
 
+pdf(d::LKJCholesky, R::Cholesky) = exp(logpdf(d, R))
+
+loglikelihood(d::LKJCholesky, R::Cholesky) = logpdf(d, R)
+function loglikelihood(d::LKJCholesky, Rs::AbstractArray{<:Cholesky})
+    return sum(R -> logpdf(d, R), Rs)
+end
+
 #  -----------------------------------------------------------------------------
 #  Sampling
 #  -----------------------------------------------------------------------------
--- a/src/cholesky/lkjcholesky.jl
+++ b/src/cholesky/lkjcholesky.jl
@@ -62,6 +62,12 @@ function insupport(d::LKJCholesky, R::Cholesky)
     return true
 end
 
+function mode(d::LKJCholesky)
+    p = dim(d)
+    factors = Matrix{partype(d)}(I, p, p)
+    return Cholesky(factors, d.uplo, 0)
+end
+
 params(d::LKJCholesky) = (d.d, d.η, d.uplo)
 
 @inline partype(::LKJCholesky{T}) where {T <: Real} = T
--- a/src/cholesky/lkjcholesky.jl
+++ b/src/cholesky/lkjcholesky.jl
@@ -58,3 +58,72 @@ end
 
 logpdf(d::LKJCholesky, x::Cholesky) = logkernel(d, x) + d.logc0
 
+#  -----------------------------------------------------------------------------
+#  Sampling
+#  -----------------------------------------------------------------------------
+
+function rand(rng::AbstractRNG, d::LKJCholesky)
+    p = dim(d)
+    T = eltype(d)
+    factors = Matrix{T}(undef, p, p)
+    R = Cholesky(factors, d.uplo, 0)
+    return rand!(rng, d, R)
+end
+function rand(rng::AbstractRNG, d::LKJCholesky, dims::Dims)
+    p = dim(d)
+    uplo = d.uplo
+    T = eltype(d)
+    TChol = Cholesky{T,Matrix{T}}
+    Rs = Array{TChol}(undef, dims)
+    for i in eachindex(Rs)
+        factors = Matrix{T}(undef, p, p)
+        Rs[i] = R = Cholesky(factors, uplo, 0)
+        rand!(rng, d, R)
+    end
+    return Rs
+end
+
+rand!(rng::AbstractRNG, d::LKJCholesky, R::Cholesky) = _lkj_cholesky_vine_sampler!(rng, d, R)
+
+function _lkj_cholesky_vine_sampler!(rng::AbstractRNG, d::LKJCholesky, R::Cholesky)
+    p, η = params(d)
+    factors = R.factors
+    if R.uplo === 'U'
+        cpcs = _lkj_vine_rand_cpcs!(transpose(factors), p, η, rng)
+        _cpcs_to_cholesky!(factors, cpcs, p)
+    else  # uplo === 'L'
+        cpcs = _lkj_vine_rand_cpcs!(factors, p, η, rng)
+        _cpcs_to_cholesky!(transpose(factors), cpcs, p)
+    end
+    return R
+end
+
+# sample partial canonical correlations using the vine method from Section 2.4 in LKJ (2009 JMA)
+# storage 
+function _lkj_vine_rand_cpcs!(cpcs, d::Integer, η::Real, rng::AbstractRNG)
+    T = eltype(cpcs)
+    β = η + T(d - 1) / 2
+    @inbounds for i in 1:(d - 1)
+        β -= T(0.5)
+        for j in (i + 1):d
+            cpcs[i, j] = 2 * rand(rng, Beta(β, β)) - 1
+        end
+    end
+    return cpcs
+end
+
+# map partial canonical correlations stored in the upper triangle of a matrix
+# to the corresponding upper triangular cholesky factor w
+# adapted from https://github.com/TuringLang/Bijectors.jl/blob/v0.9.4/src/bijectors/corr.jl
+function _cpcs_to_cholesky!(w, z, d::Integer)
+    @inbounds for j in 1:d
+        w[1, j] = 1
+        for i in 1:(j-1)
+            wij = w[i, j]
+            zij = z[i, j]
+            w[i, j] = zij * wij
+            w[i+1, j] = wij * sqrt(1 - zij^2)
+        end
+    end
+    return w
+end
--- a/src/cholesky/lkjcholesky.jl
+++ b/src/cholesky/lkjcholesky.jl
@@ -1,0 +1,6 @@
+struct LKJCholesky{T <: Real, D <: Integer} <: Distribution{CholeskyVariate,Continuous}
+    d::D
+    η::T
+    uplo::Char
+    logc0::T
+end
--- a/src/cholesky/lkjcholesky.jl
+++ b/src/cholesky/lkjcholesky.jl
@@ -39,3 +39,22 @@ params(d::LKJCholesky) = (d.d, d.η, d.uplo)
 
 @inline partype(::LKJCholesky{T}) where {T <: Real} = T
 
+#  -----------------------------------------------------------------------------
+#  Evaluation
+#  -----------------------------------------------------------------------------
+
+function logkernel(d::LKJCholesky, x::Cholesky)
+    factors = x.factors
+    p, η = params(d)
+    c = p + 2(η - 1)
+    T = typeof(one(c) * log(one(eltype(factors))))
+    logp = zero(T)
+    di = diagind(factors)
+    for i in 2:p
+        logp += (c - i) * log(factors[di[i]])
+    end
+    return logp
+end
+
+logpdf(d::LKJCholesky, x::Cholesky) = logkernel(d, x) + d.logc0
+
--- a/src/cholesky/lkjcholesky.jl
+++ b/src/cholesky/lkjcholesky.jl
@@ -25,6 +25,23 @@ _as_char(c::Char) = c
 _as_char(x) = only(string(x))
 
 #  -----------------------------------------------------------------------------
+#  REPL display
+#  -----------------------------------------------------------------------------
+
+show(io::IO, d::LKJCholesky) = show_multline(io, d, [(:d, d.d), (:η, d.η), (:uplo, d.uplo)])
+
+#  -----------------------------------------------------------------------------
+#  Conversion
+#  -----------------------------------------------------------------------------
+
+function convert(::Type{LKJCholesky{T}}, d::LKJCholesky) where T <: Real
+    return LKJCholesky{T, typeof(d.d)}(d.d, T(d.η), d.uplo, T(d.logc0))
+end
+function convert(::Type{LKJCholesky{T}}, d::Integer, η, logc0) where T <: Real
+    return LKJCholesky{T, typeof(d)}(d, T(η), d.uplo, T(logc0))
+end
+
+#  -----------------------------------------------------------------------------
 #  Properties
 #  -----------------------------------------------------------------------------
 
--- a/src/cholesky/lkjcholesky.jl
+++ b/src/cholesky/lkjcholesky.jl
@@ -23,3 +23,19 @@ end
 
 _as_char(c::Char) = c
 _as_char(x) = only(string(x))
+
+#  -----------------------------------------------------------------------------
+#  Properties
+#  -----------------------------------------------------------------------------
+
+dim(d::LKJCholesky) = d.d
+
+function size(d::LKJCholesky)
+    p = dim(d)
+    return (p, p)
+end
+
+params(d::LKJCholesky) = (d.d, d.η, d.uplo)
+
+@inline partype(::LKJCholesky{T}) where {T <: Real} = T
+
--- a/src/cholesky/lkjcholesky.jl
+++ b/src/cholesky/lkjcholesky.jl
@@ -4,3 +4,22 @@ struct LKJCholesky{T <: Real, D <: Integer} <: Distribution{CholeskyVariate,Cont
     uplo::Char
     logc0::T
 end
+
+#  -----------------------------------------------------------------------------
+#  Constructors
+#  -----------------------------------------------------------------------------
+
+function LKJCholesky(d::Integer, η::Real, _uplo::Union{Char,Symbol} = 'L'; check_args = true)
+    if check_args
+        d > 0 || throw(ArgumentError("Matrix dimension must be positive."))
+        η > 0 || throw(ArgumentError("Shape parameter must be positive."))
+    end
+    logc0 = lkj_logc0(d, η)
+    uplo = _as_char(_uplo)
+    uplo ∈ ('U', 'L') || throw(ArgumentError("uplo must be 'U' or 'L'."))
+    T = Base.promote_eltype(η, logc0)
+    LKJCholesky{T, typeof(d)}(d, T(η), uplo, T(logc0))
+end
+
+_as_char(c::Char) = c
+_as_char(x) = only(string(x))
--- a/src/cholesky/lkjcholesky.jl
+++ b/src/cholesky/lkjcholesky.jl
@@ -93,8 +93,8 @@ params(d::LKJCholesky) = (d.d, d.η, d.uplo)
 #  Evaluation
 #  -----------------------------------------------------------------------------
 
-function logkernel(d::LKJCholesky, x::Cholesky)
-    factors = x.factors
+function logkernel(d::LKJCholesky, R::Cholesky)
+    factors = R.factors
     p, η = params(d)
     c = p + 2(η - 1)
     T = typeof(one(c) * log(one(eltype(factors))))
@@ -106,7 +106,7 @@ function logkernel(d::LKJCholesky, x::Cholesky)
     return logp
 end
 
-logpdf(d::LKJCholesky, x::Cholesky) = logkernel(d, x) + d.logc0
+logpdf(d::LKJCholesky, R::Cholesky) = logkernel(d, R) + d.logc0
 
 #  -----------------------------------------------------------------------------
 #  Sampling
--- a/src/cholesky/lkjcholesky.jl
+++ b/src/cholesky/lkjcholesky.jl
@@ -35,6 +35,33 @@ function size(d::LKJCholesky)
     return (p, p)
 end
 
+function insupport(d::LKJCholesky, R::Cholesky)
+    p = dim(d)
+    factors = R.factors
+    (isreal(factors) && size(factors, 1) == p) || return false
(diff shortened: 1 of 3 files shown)


src/cholesky/lkjcholesky.jl
```diff
@@ -0,0 +1,183 @@
...
+            end
+            isapprox(s, 1) || return false
+        end
+    end
+    return true
+end
+
+function mode(d::LKJCholesky)
+    p = dim(d)
+    factors = Matrix{partype(d)}(I, p, p)
```
devmotion, line 84: In general, the type of the parameters is not equivalent with the (element-)type of the samples/support.

src/cholesky/lkjcholesky.jl
```diff
@@ -0,0 +1,183 @@
...
+function logkernel(d::LKJCholesky, R::Cholesky)
+    factors = R.factors
+    p, η = params(d)
+    c = p + 2(η - 1)
+    T = typeof(one(c) * log(one(eltype(factors))))
+    logp = zero(T)
+    di = diagind(factors)
+    for i in 2:p
+        logp += (c - i) * log(factors[di[i]])
+    end
```
devmotion, line 105: Again no need to use type heuristics, one can just compute the sum directly with `sum`.

src/cholesky/lkjcholesky.jl
```diff
@@ -0,0 +1,183 @@
...
+    return logp
+end
+
+logpdf(d::LKJCholesky, R::Cholesky) = logkernel(d, R) + d.logc0
```
devmotion, line 109: Probably one should check `insupport(R)` at some point?

src/cholesky/lkjcholesky.jl
```diff
@@ -0,0 +1,183 @@
...
+
+#  -----------------------------------------------------------------------------
+#  Sampling
+#  -----------------------------------------------------------------------------
+
+function rand(rng::AbstractRNG, d::LKJCholesky)
+    p = dim(d)
+    T = eltype(d)
+    factors = Matrix{T}(undef, p, p)
+    R = Cholesky(factors, d.uplo, 0)
```
devmotion, line 126: Usually it is safer to not fall back to in-place methods here since the type heuristics are sometimes not correct (even though this is the typical approach in Distributions) but I guess this might not be possible in a "nice" way?

src/cholesky/lkjcholesky.jl
```diff
@@ -0,0 +1,183 @@
...
+    end
+    return R
+end
+
+# sample partial canonical correlations z using the vine method from Section 2.4 in LKJ (2009 JMA)
+function _lkj_vine_rand_cpcs!(z, d::Integer, η::Real, rng::AbstractRNG)
+    T = eltype(z)
+    β = η + T(d - 1) / 2
+    @inbounds for i in 1:(d - 1)
+        β -= T(0.5)
```
devmotion, line 163: Maybe
```suggestion
        β -= T(1//2)
```
?

sethaxen:
In the current design, `LKJCholesky` has an `uplo` parameter that is only used to determine the triangle storage of sampled `Cholesky` objects (specified in the `uplo` field of `Cholesky`). If `insupport`, `logkernel`, etc are called on `Cholesky` with a different `uplo`, the distribution's `uplo` is ignored. This is analogous to `logpdf` for a matrix distribution working correctly when passed a structured matrix even if the random sample might be unstructured.

src/cholesky/lkjcholesky.jl
```diff
@@ -0,0 +1,183 @@
...
+        z[i, (i + 1):d] .= 2 .* rand.(rng, Ref(Beta(β, β))) .- 1
```
devmotion, line 164: I would assume that this is less performant than just iterating with a for loop and a precomputed sampler since involves the heavy broadcasting machinery and recomputes the values for the Beta sampler multiple times:
```suggestion
        spl = sampler(Beta(β, β))
        for j in (i + 1):d
            z[i, j] = 2 * rand(rng, spl) - 1
        end
```
I guess if one would iterate in the natural order (rows at the inner level) than one could not use the sampler?

src/matrix/lkj.jl
```diff
@@ -96,10 +96,11 @@ params(d::LKJ) = (d.d, d.η)
 function lkj_logc0(d::Integer, η::Real)
     d > 1 || return zero(η)
     if isone(η)
+        T = float(Base.promote_type(d, η))
```
devmotion: Why is this needed?

src/matrix/lkj.jl
```diff
@@ -182,12 +183,13 @@ end
 
 function lkj_onion_loginvconst(d::Integer, η::Real)
     #  Equation (17) in LKJ (2009 JMA)
-    sumlogs = zero(η)
+    T = typeof(log(one(η + d)))
+    sumlogs = zero(T)
     for k in 2:d - 1
-        sumlogs += 0.5k*logπ + loggamma(η + 0.5(d - 1 - k))
+        sumlogs += T(0.5k)*logπ + loggamma(η + T(0.5(d - 1 - k)))
```
devmotion, line 189: Again it seems type heuristics are not needed and one could just use `sum`?

src/matrix/lkj.jl
```diff
@@ -212,10 +214,11 @@ end
 
 function lkj_vine_loginvconst(d::Integer, η::Real)
     #  Equation (16) in LKJ (2009 JMA)
-    expsum = zero(η)
-    betasum = zero(η)
+    T = typeof(log(one(η + d)))
```
devmotion, line 217: Same here?

src/matrix/lkj.jl
```diff
@@ -238,9 +241,10 @@ end
 
 function lkj_loginvconst_alt(d::Integer, η::Real)
     #  Third line in first proof of Section 3.3 in LKJ (2009 JMA)
-    loginvconst = zero(η)
+    T = typeof(log(one(η + d)))
```
devmotion, line 244: And here.

devmotion (commented):
Great :+1: 

I think it would be good to remove the type computations whenever possible. It seems often one could just use `sum`. If one wants to ensure type stability often a better approach is to perform as many computations as possible and only fall back to e.g. `oftype(result, Inf)` if necessary in the end.

src/cholesky/lkjcholesky.jl
```diff
@@ -0,0 +1,183 @@
...
+#  -----------------------------------------------------------------------------
+
+function LKJCholesky(d::Integer, η::Real, _uplo::Union{Char,Symbol} = 'L'; check_args = true)
+    if check_args
+        d > 0 || throw(ArgumentError("Matrix dimension must be positive."))
+        η > 0 || throw(ArgumentError("Shape parameter must be positive."))
+    end
+    logc0 = lkj_logc0(d, η)
+    uplo = _as_char(_uplo)
+    uplo ∈ ('U', 'L') || throw(ArgumentError("uplo must be 'U' or 'L'."))
```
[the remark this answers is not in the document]
sethaxen: `char_uplo` is not exported by `LinearAlgebra` and so I think is considered an implementation detail that can change. We could basically copy it here for safety.

src/cholesky/lkjcholesky.jl
sethaxen: In general the trade-off we have is that summation via a loop is fastest but with greater code complexity and not as AD-friendly (e.g. with Zygote). e.g. here using `sum` is 10% slower than looping. I think this is acceptable.

src/cholesky/lkjcholesky.jl
sethaxen: Point against that are 1. none of the matrix distributions explicitly check `insupport` during the kernel. 2. for `d=4`, `insupport` has 30% the runtime of `logpdf`. But as `d` increases, `insupport` becomes more expensive (`insupport` is more expensive than `logpdf` around `d=20`).

src/cholesky/lkjcholesky.jl
```diff
@@ -0,0 +1,183 @@
...
+end
+
+_as_char(c::Char) = c
+_as_char(x) = only(string(x))
+
+#  -----------------------------------------------------------------------------
+#  REPL display
+#  -----------------------------------------------------------------------------
+
+show(io::IO, d::LKJCholesky) = show_multline(io, d, [(:d, d.d), (:η, d.η), (:uplo, d.uplo)])
```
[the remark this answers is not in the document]
sethaxen: Perhaps, but this is consistent with how `show` is defined throughout Distributions. There is a simpler implementation I can do using  Distributions' utilities:

```suggestion
show(io::IO, d::LKJCholesky) = show(io, d, (:d, :η, :uplo))
```
which just hides the `logc0` term.

src/cholesky/lkjcholesky.jl
devmotion: `sum` is slightly faster in the following simple benchmark:
```julia
julia> using BenchmarkTools

julia> using LinearAlgebra

julia> function f(c, x)
           T = typeof(one(c) * log(one(eltype(x))))
           di = diagind(x)               
           logp = zero(T)                
           for i in 2:length(di)         
               logp += (c - i) * log(x[di[i]])
           end  
           return logp   
       end             
          
julia> function g(c, x)
           logp = sum((c - i) * log(x[di]) for (i, di) in Iterators.drop(enumerate(diagind(x)), 1))
           return logp
       end      
                    
julia> c = rand();

julia> x = randn(40, 40);

julia> @benchmark f($c, $x)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     311.992 ns (0.00% GC)
  median time:      320.075 ns (0.00% GC)
  mean time:        339.895 ns (0.00% GC)
  maximum time:     740.527 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     241

julia> @benchmark g($c, $x)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     300.394 ns (0.00% GC)
  median time:      305.855 ns (0.00% GC)
  mean time:        321.193 ns (0.00% GC)
  maximum time:     690.703 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     249
```

src/cholesky/lkjcholesky.jl
devmotion: The other matrix distributions at least check the dimensions (using `logpdf` with checks and `_logpdf` without checks) and I'd argue that they should also check the support and return `logpdf(d, x) = -Inf` it `x` is not in the support, as done for uni- and matrixvariate distributions.

johnczito: > none of the matrix distributions explicitly check `insupport` during the kernel.

That's more a historical artifact than a design choice. I agree that `insupport` ought to be checked, but also that you should somehow be able to switch that off for performance purposes.

> I'd argue that they should also check the support and return `logpdf(d, x) = -Inf` it `x` is not in the support, as done for uni- and matrixvariate distributions.

Should it be `-Inf` or an error? Like, how should "isn't posdef" be handled versus "isn't the right dims"?

devmotion: > Should it be -Inf or an error? Like, how should "isn't posdef" be handled versus "isn't the right dims"?

This is a design choice but for multivariate dimensions a `DimensionMismatch` is thrown if the dimension is incorrect (e.g., https://github.com/JuliaStats/Distributions.jl/blob/d285de5e0df4290905c0eab5ed0c94ecd1037567/src/multivariates.jl#L199-L200) and `-Inf` is returned if `x` is of the correct dimension but not in the support (e.g., in the case of `Dirichlet`: https://github.com/JuliaStats/Distributions.jl/blob/d285de5e0df4290905c0eab5ed0c94ecd1037567/src/multivariate/dirichlet.jl#L142-L144). This is slightly inconsistent though since `insupport(d::Dirichlet, x)` also checks the dimension of `x`.

johnczito: Right, so is that how you want the matrix distributions to behave as well?

src/cholesky/lkjcholesky.jl
sethaxen: I'd expect so, but it seems that the broadcasting version is 3x faster than with this change. Perhaps this is because `sampler(Beta(β, β))` is not type-stable.

devmotion: Ah yeah, maybe one has to add a function barrier outside of the for loop (similar to the suggestion in https://github.com/JuliaStats/Distributions.jl/pull/1281#issuecomment-836246397), i.e.,
```julia
function _loop!(rng, z, i, js, sampler)
    @inbounds for j in js
        z[i, j] = 2 * rand(rng, sampler) - 1
    end
end
```
and then use
```julia
    _loop!(rng, z, i, (i + 1):d, sampler(Beta(β, β)))
```
in the for loop.

sethaxen: It's still slightly slower for small matrices but faster for large matrices, so this is worth it.

src/cholesky/lkjcholesky.jl
sethaxen: The speed advantage of `sum` goes away here of you already know the length of `di` (we do). In my `sum` implementations, it's always a little slower than the loops, but I've adopted it anyways.

src/cholesky/lkjcholesky.jl
sethaxen: I've added an `insupport` check to `logpdf` that causes an error to be raised. A user can call `_logpdf` to sidestep this.

src/cholesky/lkjcholesky.jl
sethaxen: I think you mean don't call `rand!` from `rand`. I've refactored so all sampling passes through `_lkj_cholesky_vine_sampler!`.

sethaxen:
Thanks @devmotion for the review already. After incorporating your suggestions, this PR is now ready for review. A few notes:
- I had originally made changes to the `logc0` computation for `LKJ` so it would not force `η` parameters of `Float32` to be promoted to `Float64`, but I decided if this is done it should be part of  larger `LKJ` clean-up and removed it, adding instead a broken test.
-  It's actually quite a bit faster to use this PR's vine method to generate a random `Cholesky` matrix and then call `Matrix` on it than to use `LKJ`'s onion method. A future PR should probably cause LKJ to use this code, though then we need better tests for our sampler here.
- Are there remaining key methods I should implement?

johnczito:
> It's actually quite a bit faster to use this PR's vine method to generate a random Cholesky matrix and then call Matrix on it than to use LKJ's onion method.

Yeah, that's what Stan does. Makes sense to me.

> though then we need better tests for our sampler here.

You could adapt this to check that the elements of `LL'` have the correct marginals:
https://github.com/JuliaStats/Distributions.jl/blob/d285de5e0df4290905c0eab5ed0c94ecd1037567/test/matrixvariates.jl#L424-L441
You could do something similar with a two-sample KS test to check the mutual consistency of draws from the onion sampler and the vine sampler. I'm not sure how to check that all the elements of the matrix are being jointly sampled correctly.

sethaxen:
> You could adapt this to check that the elements of `LL'` have the correct marginals:

Ah, yes currently we just check that mean and variance of the marginals is the same, but we could do something like that: https://github.com/JuliaStats/Distributions.jl/blob/f4cb6e3522ded8993a47b6e0e2ce75adbfd521f9/test/lkjcholesky.jl#L12-L15

johnczito:
Yeah, `_marginal(lkj::LKJ)` returns the entire marginal distribution, and it's the same for each element. So you can check higher moments of the draws and whatever else.

https://github.com/JuliaStats/Distributions.jl/blob/d285de5e0df4290905c0eab5ed0c94ecd1037567/src/matrix/lkj.jl#L150-L159

sethaxen:
> > It's actually quite a bit faster to use this PR's vine method to generate a random Cholesky matrix and then call Matrix on it than to use LKJ's onion method.
> 
> Yeah, that's what Stan does. Makes sense to me.

FWIW, I also implemented an onion sampler that incrementally builds the Cholesky factor, as is mentioned in Section 5 of the paper and as TFP does, but I found that it was 3x slower than the vine sampler for `LKJCholesky(10, 2)` and 6x slower for `LKJCholesky(40, 2)`.

> Yeah, `_marginal(lkj::LKJ)` returns the entire marginal distribution, and it's the same for each element. So you can check higher moments of the draws and whatever else.

I added tests for `mean`, `var`, and the first 5 central moments.

> You could adapt this to check that the elements of `LL'` have the correct marginals:

Adapting that code, I'm getting regular failures of the KS test. This happens even if I replace the `LKJCholesky` sampler with `LKJ`'s sampler followed by `cholesky`. Not certain what's going on here, except maybe the KS test used for testing `LKJ` is not very robust and only works for a specific seed?

johnczito:
> Adapting that code, I'm getting regular failures of the KS test. This happens even if I replace the LKJCholesky sampler with LKJ's sampler followed by cholesky. Not certain what's going on here, except maybe the KS test used for testing LKJ is not very robust and only works for a specific seed?

The test readout looks decent to me. These tests will have some positive failure rate even if the code is error free, so in practice we just want to write down a not-embarrassing (seed, sample size, significance level) for which everything passes and then use this as a regression test ([#143](https://github.com/JuliaStats/Distributions.jl/issues/143)). Do we still have an issue when you bump either the sample size up to `10^5` or bump the significance level up to `0.1`?

sethaxen:
> FWIW, I also implemented an onion sampler that incrementally builds the Cholesky factor, as is mentioned in Section 5 of the paper and as TFP does, but I found that it was 3x slower than the vine sampler for `LKJCholesky(10, 2)` and 6x slower for `LKJCholesky(40, 2)`.

Whoops, had that backwards, the incremental onion sampler is faster, as they found in the paper, so I switched to it. It produces the cholesky factor for the same random matrix sampled from `LKJ` with the same seed, but it does it 8x faster for `LKJCholesky(10, 2)` and 22x faster for `LKJCholesky(100, 2)`.

> Is there still an issue if we bump either the sample size up to `10^5` or bump the significance level up to `0.1`?

It gets better when I bump sample size to `10^5`, but I haven't found a seed for which it passes yet. It looks like increasing the `α` would actually make the test harder to pass, since the test is
```julia
@test pvalue_kolmogorovsmirnoff(zs[i, j, :], marginal) >= α / L
```

johnczito:
> It looks like increasing the α would actually make the test harder to pass...

D'oh, yes of course. Sorry. Might try driving it down to `0.01`.

Anyway, if everything is error free it should be possible to pin down modest conditions under which all of this works, but if you're satisfied that it's right and other reviewers are happy, I don't mean to drag this out. Only the matrix variate tests bother with this KS business, and that's just because I kinda like it. But if that preference isn't widely shared, then don't worry about it.

sethaxen:
> D'oh, yes of course. Sorry. Might try driving it down to `0.01`.
> 
> Anyway, if everything is error free it should be possible to pin down modest conditions under which all of this works, but if you're satisfied that it's right and other reviewers are happy, I don't mean to drag this out. Only the matrix variate tests bother with this KS business, and that's just because I kinda like it. But if that preference isn't widely shared, then don't worry about it.

Okay, I decreased alpha to 0.01 and also added an additional correction factor to account for the number of KS tests being computed. With the chosen seed the test suite passes.

src/cholesky/lkjcholesky.jl
```diff
@@ -0,0 +1,241 @@
...
+    # Section 3.2 in LKJ (2009 JMA)
+    # reformulated to incrementally construct Cholesky factor as mentioned in Section 5
+    # equivalent steps in algorithm in reference are marked.
+    @assert size(A) == (d, d)
+    A[1, 1] = 1
+    d > 1 || return R
+    β = η + (d - 2)//2
+    #  1. Initialization
+    w0 = 2 * rand(rng, Beta(β, β)) - 1
+    @inbounds if TTri <: LowerTriangular
```
devmotion, line 219: Maybe just fall back to `_lkj_cholesky_onion_tri!(rng, A', d, eta, TTri)` in one of both cases?

src/cholesky/lkjcholesky.jl
```diff
@@ -0,0 +1,241 @@
...
+    #  2. Loop, each iteration k adds row/column k+1
+    for k in 2:(d - 1)
+        #  (a)
+        β -= 1//2
+        #  (b)
+        y = rand(rng, Beta(k//2, β))
+        #  (c)-(e)
+        # w is directionally uniform vector of length √y
+        @inbounds w = @views TTri <: LowerTriangular ? A[k + 1, 1:k] : A[1:k, k + 1]
+        w .= randn.(Ref(rng))
```
devmotion, line 234: Maybe just
```suggestion
        randn!(rng, w)
```
?

src/cholesky/lkjcholesky.jl
```diff
@@ -0,0 +1,241 @@
...
+    β = η + (d - 2)//2
+    #  1. Initialization
+    w0 = 2 * rand(rng, Beta(β, β)) - 1
+    @inbounds if TTri <: LowerTriangular
+        A[2, 1] = w0
+    else
+        A[1, 2] = w0
+    end
+    @inbounds A[2, 2] = sqrt(1 - w0^2)
+    #  2. Loop, each iteration k adds row/column k+1
```
devmotion, line 225: Probably it's more efficient to loop through columns?

sethaxen: Perhaps. A quick benchmark for `LKJCholesky(4, 2)` shows that this function is just 2% slower for `uplo='L'`. But the algorithm must proceed row-wise for `uplo='L'` and col-wise for `uplo='U'`.

src/cholesky/lkjcholesky.jl
sethaxen: I introduced the type parameter because operating on the transpose increased the runtime by ~30%. This way we do the fast thing for both triangles with almost no code duplication.

sethaxen:
Is this PR missing anything?

sethaxen:
@devmotion, @johnczito would you like to re-review this PR?

matbesancon:
Apart from a minor addition in the documentation, I'm good with the PR yes

PR #1340, opened 2021-06-07
Add squared L2 norms of some discrete distributions

giordano:
Ref: #806.  CC: @ablaom

### DIFF
--- a/src/pdfnorm.jl
+++ b/src/pdfnorm.jl
@@ -11,12 +11,16 @@ where ``S`` is the support of ``f(x)``.
 """
 pdfsquaredL2norm
 
+pdfsquaredL2norm(d::Bernoulli) = @evalpoly d.p 1 -2 2
+
 function pdfsquaredL2norm(d::Beta)
     α, β = params(d)
     z = beta(2 * α - 1, 2 * β - 1) / beta(α, β) ^ 2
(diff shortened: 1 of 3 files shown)


src/pdfnorm.jl
```diff
@@ -43,8 +49,17 @@ function pdfsquaredL2norm(d::Gamma)
...
+
 pdfsquaredL2norm(d::Logistic) = 1 / (6 * d.θ)
 
 pdfsquaredL2norm(d::Normal) = inv(sqrt4π * d.σ)
 
+# The identity is obvious if you look at the definition of the modified Bessel function of
+# first kind I_0.  Starting from the L2-norm of the Poisson distribution this can be proven
+# by observing this is the Laguerre exponential l−e of x², which is related to said Bessel
+# function, see <https://doi.org/10.1140/epjst/e2018-00073-1> (preprint:
+# <https://arxiv.org/abs/1707.01135>).
```
giordano, line 62: Computing this took me more time than I'd like to admit.  If you have the definition of the [modified Bessel function of the first kind](https://en.wikipedia.org/wiki/Bessel_function#Modified_Bessel_functions), then the identity is easy to see, but I had to dig it up quite a bit to get there.  I don't know whether the reference I dropped here is more useful or confusing, I can remove it if you consider it unnecessary

devmotion (approved):
Looks good!

PR #1342, opened 2021-06-11
Added mle for weibull distribution

### LINKED ISSUE #702: Feature request: fit(Weibull, rand(100))
Hi,
This is a newbie question, thus I might ask in the wrong repo or wrong place in general. 
```julia
julia> fit(Normal, rand(100))
Distributions.Normal{Float64}(μ=0.5084961617475308, σ=0.2908609425853258)

julia> fit(Weibull, rand(100))
ERROR: suffstats is not implemented for (Distributions.Weibull, Array{Float64,1}).
Stacktrace:
 [1] suffstats(::Type{Distributions.Weibull}, ::Array{Float64,1}) at C:\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\Users\tfr004\AppData\Local\JuliaPro-0.6.1.1\pkgs-0.6.1.1\v0.6\Distributions\src\genericfit.jl:5
 [2] fit(::Type{Distributions.Weibull}, ::Array{Float64,1}) at C:\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\Users\tfr004\AppData\Local\JuliaPro-0.6.1.1\pkgs-0.6.1.1\v0.6\Distributions\src\genericfit.jl:33

julia>
```

I am porting a Matlab project to Julia. In Matlab they used `histfit(data,nbins,'weibull')` function. I am actually looking something to replace the histfit function and found Julia fit. Any help is appreciated.

rory-humphries:
I added an implementation that solves for the [Weibull maximum likelihood estimators](https://en.wikipedia.org/wiki/Weibull_distribution#Parameter_estimation) using newtons method. Included some basic tests as sanity checks and updated the docs.

I noticed that there was some discussion in #1267 on whether or not use a generic newtons method and wondering if that's been decided on. 

Merging this can close #702.

### DIFF
--- a/src/univariate/continuous/weibull.jl
+++ b/src/univariate/continuous/weibull.jl
@@ -156,7 +156,7 @@ function fit_mle(::Type{<:Weibull}, x::AbstractArray{<:Real};
 
     α = α0
 
-    lnx = log.(x)
+    lnx = map(log, x)
     n = length(x)
 
     while ϵ > tol && N < maxiter
--- a/src/univariate/continuous/weibull.jl
+++ b/src/univariate/continuous/weibull.jl
@@ -143,14 +143,13 @@ rand(rng::AbstractRNG, d::Weibull) = xv(d, randexp(rng))
 #### Fit model
 
 """
-    fit_mle(::Type{<:Weibull}, x::AbstractArray{T}; 
-    α0::Float64 = 1.0, maxiter::Int = 1000, tol::Float64 = 1e-16)
+    fit_mle(::Type{<:Weibull}, x::AbstractArray{<:Real}; 
+    alpha0::Real = 1, maxiter::Int = 1000, tol::Real = 1e-16)
 
-Maximum Likelihood Estimate of `Weibull` Distribution via Newton's Method
-
+Compute the maximum likelihood estimate of the [`Weibull`](@ref) distribution with Newton's method.
 """
-function fit_mle(::Type{<:Weibull}, x::AbstractArray{T};
-    α0::Float64 = 1.0, maxiter::Int = 1000, tol::Float64 = 1e-16) where T <: Real
+function fit_mle(::Type{<:Weibull}, x::AbstractArray{<:Real};
+    alpha0::Real = 1, maxiter::Int = 1000, tol::Real = 1e-16)
 
     ϵ = Inf
     N = 0
--- a/src/univariate/continuous/weibull.jl
+++ b/src/univariate/continuous/weibull.jl
@@ -151,28 +151,33 @@ Compute the maximum likelihood estimate of the [`Weibull`](@ref) distribution wi
 function fit_mle(::Type{<:Weibull}, x::AbstractArray{<:Real};
     alpha0::Real = 1, maxiter::Int = 1000, tol::Real = 1e-16)
 
-    ϵ = Inf
     N = 0
 
-    α = α0
+    param_type = typeof(AbstractFloat(alpha0))
+    α::param_type = alpha0
 
     lnx = map(log, x)
+    lnxsq = lnx.^2
+    mean_lnx = mean(lnx)
     n = length(x)
 
-    while ϵ > tol && N < maxiter
+    convergence = false
+    while !convergence && N < maxiter
 
         xpow = x.^α
+        sum_xpow = sum(xpow)
+        dot_xpowlnx = dot(xpow, lnx)
 
-        fx = sum(xpow .* lnx) / (sum(xpow)) - (1 / n) * sum(lnx) - 1 / α
-        ∂fx = (-sum(xpow .* lnx)^2 + sum(xpow) * sum((lnx.^2) .* xpow)) / (sum(xpow)^2) + 1 / α^2
+        fx = dot_xpowlnx / sum_xpow - mean_lnx - 1 / α
+        ∂fx = (-dot_xpowlnx^2 + sum_xpow * dot(lnxsq, xpow)) / (sum_xpow^2) + 1 / α^2
 
         αold = α
         α -= fx / ∂fx
 
-        ϵ = abs(α - αold)
+        convergence = abs(α - αold) < tol
         N += 1
     end
 
-    θ = sum((x.^α) ./ n)^(1 / α)
+    θ::param_type = sum((x.^α) ./ n)^(1 / α)
     return Weibull(α, θ)
 end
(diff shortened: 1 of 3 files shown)


src/univariate/continuous/weibull.jl
```diff
@@ -139,3 +139,41 @@ end
...
+
+"""
+    fit_mle(::Type{<:Weibull}, x::AbstractArray{T}; 
+    α0::Float64 = 1.0, maxiter::Int = 1000, tol::Float64 = 1e-16)
+
+Maximum Likelihood Estimate of `Weibull` Distribution via Newton's Method
+
+"""
+function fit_mle(::Type{<:Weibull}, x::AbstractArray{T};
+    α0::Float64 = 1.0, maxiter::Int = 1000, tol::Float64 = 1e-16) where T <: Real
```
devmotion, line 153: The type parameter is not needed it seems and `Float64` might be a bit too restrictive. Additionally, probably it would be good to not use Unicode characters in the official API but at least provide an ASCII alternative.
```suggestion
    fit_mle(::Type{<:Weibull}, x::AbstractArray{<:Real}; 
    alpha0::Real = 1, maxiter::Int = 1000, tol::Real = 1e-16)

Compute the maximum likelihood estimate of the [`Weibull`](@ref) distribution with Newton's method.
"""
function fit_mle(::Type{<:Weibull}, x::AbstractArray{<:Real};
    alpha0::Real = 1, maxiter::Int = 1000, tol::Real = 1e-16)
```

src/univariate/continuous/weibull.jl
```diff
@@ -139,3 +139,41 @@ end
...
+
+    ϵ = Inf
+    N = 0
+
+    α = α0
+
+    lnx = log.(x)
```
devmotion, line 160: No need for broadcasting here:

```suggestion
    lnx = map(log, x)
```

src/univariate/continuous/weibull.jl
```diff
@@ -139,3 +139,41 @@ end
...
+    n = length(x)
+
+    while ϵ > tol && N < maxiter
+
+        xpow = x.^α
+
+        fx = sum(xpow .* lnx) / (sum(xpow)) - (1 / n) * sum(lnx) - 1 / α
+        ∂fx = (-sum(xpow .* lnx)^2 + sum(xpow) * sum((lnx.^2) .* xpow)) / (sum(xpow)^2) + 1 / α^2
```
devmotion, line 168: It would be better to precompute terms (e.g., `sum(lnx) = mean(lnx)` can be computed outside of the loop), to compute terms only once (e.g., `sum(xpow)`), and to use e.g. `dot(xpow, lnx)` instead of `sum(xpow .* lnx)`.

src/univariate/continuous/weibull.jl
```diff
@@ -139,3 +139,41 @@ end
...
+
+        αold = α
+        α -= fx / ∂fx
+
+        ϵ = abs(α - αold)
```
devmotion, line 173: These assignments might create type instabilities since their type might chnage from `Float64` to some other type. A solution would be to not initialize them with the default values `Inf` etc. before the first iteration but just set them to the computed values in the first iteration which would have to be moved outside of the loop.

devmotion (commented):
Thank you! I added some comments and suggestions.

src/univariate/continuous/weibull.jl
rory-humphries: For this I removed `ϵ` and replaced it with the boolean flag `convergence` which is updated inside the loop. For the type instability in `α` I gave it a type 
```
param_type = typeof(AbstractFloat(alpha0))
α::param_type = alpha0
```
so the types of the parameters in the returned `Weibull` distribution are decided by the type passed to `alpha0`. This removes the type instabilities in the loop. If this isn't satisfactory however I can go ahead with your suggestion of writing the first iteration of loop outside the loop body.

devmotion: Unfortunately, this is not sufficient and might even fail in the case of theta: the type of alpha and theta depends also on the type of `x`. E.g., if `x` are dual numbers the parameters should be dual numbers as well even though `alpha0` is only of type `Int` or `Float64`. I think the only reliable way to fix potential type instabilities is to move the initial step outside of the loop and use it to initialize alpha (and don't enforce the type of theta).

rory-humphries: Hopefully the latest commit fixes this, thanks for the help and comments, here's to my first of hopefully many open source contributions :).

src/univariate/continuous/weibull.jl
```diff
@@ -139,3 +139,55 @@ end
...
+    xpow0 = x.^alpha0
+    sum_xpow0 = sum(xpow0)
+    dot_xpowlnx0 = dot(xpow0, lnx)
+
+    fx = dot_xpowlnx0 / sum_xpow0 - mean_lnx - 1 / alpha0
+    ∂fx = (-dot_xpowlnx0^2 + sum_xpow0 * dot(lnxsq, xpow0)) / (sum_xpow0^2) + 1 / alpha0^2
+
+    α = alpha0 - fx / ∂fx
+
+    ϵ = abs(α - alpha0)
```
devmotion, line 172: This is just `abs(fx / ∂fx)`, so it might be easier to compute `Δα = fx / ∂fx` and use it both to update `α` and to compute `ϵ` (also below).

src/univariate/continuous/weibull.jl
```diff
@@ -139,3 +139,55 @@ end
...
+        ∂fx = (-dot_xpowlnx^2 + sum_xpow * dot(lnxsq, xpow)) / (sum_xpow^2) + 1 / α^2
+
+        αold = α
+        α -= fx / ∂fx
+
+        ϵ = abs(α - αold)
+        N += 1
+    end
+
+    θ = sum((x.^α) ./ n)^(1 / α)
```
devmotion, line 191: Maybe a bit simpler
```suggestion
    θ = mean(x.^α)^(1 / α)
```
Then one could even remove `n = length(x)` above it seems.

devmotion (commented):
Looks good now, also the formulas seem to be correct :+1: I added two additional minor suggestions.

src/univariate/continuous/weibull.jl
rory-humphries: Good catch, I'll update that

devmotion:
Thanks a lot!

PR #1348, opened 2021-06-16
Fix `(log)(c)cdf` with `Inf`, `-Inf` and `NaN`

devmotion:
This PR simplifies the `(log)(c)cdf` implementations and fixes the evaluation with `-Inf`, `Inf`, and `NaN`.

There are some problems with the current implementation:
- It does not handle `NaN` consistently: e.g.
  ```julia
  julia> cdf(Normal(), NaN)
  NaN

  julia> cdf(Exponential(), NaN)
  0.0

  julia> cdf(TriangularDist(0, 1), NaN)
  1.0
  ```
- It defines `cdf(d::DiscreteUnivariateDistribution, x::Real) = cdf(d, floor(Int, x))` etc *and* a default implementation for `cdf(::DiscreteUnivariateDistribution, x::Integer)`. This is problematic since
  - one has to define both `cdf(d, x::Real)` and `cdf(d, x::Integer)` for e.g. discrete `LocationScale`, `Truncated` and `MixtureModel` and `DiscreteNonParametric` to avoid method ambiguity errors
  - not all discrete distributions have integer-valued support but still a `cdf` for integers is defined that might not make sense or in the worst case silently produce incorrect results, e.g.
    ```julia
    julia> cdf(Dirac(3.4), 3.5)
    1.0

    julia> logcdf(Dirac(3.4), 3.5)
    -Inf
    ```
    (since `logcdf(Dirac(3.4), 3.5)` is forwarded to `logcdf(Dirac(3.4), 3)` which then calls `log(cdf(Dirac(3.4, 3))) = log(0) = -Inf`). These bugs are difficult to discover and to avoid when implementing a discrete distribution (I fixed some of these for `DiscreteNonParametric` in https://github.com/JuliaStats/Distributions.jl/pull/1286).
  - it is impossible to evaluate `cdf(d, -Inf)`, `cdf(d, Inf)`, or `cdf(d, NaN) with these definitions: e.g.
    ```julia
    julia> cdf(Poisson(), -Inf)
    ERROR: InexactError: trunc(Int64, -Inf)
    ...

    julia> cdf(Poisson(), Inf)
    ERROR: InexactError: trunc(Int64, Inf)
    ...

    julia> cdf(Poisson(), NaN)
    ERROR: InexactError: trunc(Int64, NaN)
    ...
    ```
    For `Poisson` this can be fixed by generalizing the StatsFuns macro to inputs of type `Real` but this is still an issue for other native implementations such as `Categorical` for which the same errors are thrown. The evaluation of `cdf(d, -Inf)` or `cdf(d, Inf)` is useful e.g. in the construction of truncated distributions - it allows to remove the current type heuristics (not included in this PR).

The PR fixes these problems by
- only defining `cdf(d::DiscreteUnivariateDistribution, x::Real)` but not `cdf(::DiscreteUnivariateDistribution, x::Integer)`, similar to https://github.com/JuliaStats/Distributions.jl/pull/1171 
  - the default calls `cdf_int(d, x)` which assumes integer-valued support but not inputs of type `Int`
  - `cdf_int(d, x)` handles `Inf`, `-Inf`, and `NaN` and calls `cdf(d, floor(Int, x))` for other values (not implemented!)
  - for distributions with integer-valued support one has to implement `cdf(d, ::Int)` and gets support for real values including `Inf`, `-Inf`, and `NaN` for free
  - for distributions with non-integer-valued support one has to implement `cdf(d, ::Real)` but not `cdf(d, ::Int)` (and there exists no incorrect default implementation)
  - there are no silent incorrect results since an error is thrown if one does not implement `cdf(d, ::Int)` explicitly
  - one can call `integerunitrange_cdf` etc. instead of implementing `cdf(d, ::Int)` from scratch if the distribution has a unitrange of integers as support (currently the default implementation of `cdf(d, ::Int)`)
  - this also removes method ambiguity issues and the implementation of `Truncated`, `LocationScale` and `MixtureModel` can be simplified.
- generalizing the StatsFuns macro and also defining `cdf(d, x::Real)` etc. instead of `cdf(d, x::Int)` for discrete distributions
  - StatsFuns/R already handles `NaN`, `Inf`, and `-Inf` correctly and can deal with non-integer inputs (hopefully soon there will be many more native implementations: https://github.com/JuliaStats/StatsFuns.jl/pull/113)
- updating implementations of `(log)(c)cdf` of many univariate distributions
  - this includes many continuous distributions
  - removes multiple type heuristics

### DIFF
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -20,12 +20,6 @@ isupperbounded(d::Union{D,Type{D}}) where {D<:UnivariateDistribution} = maximum(
 hasfinitesupport(d::Union{D,Type{D}}) where {D<:DiscreteUnivariateDistribution} = isbounded(d)
 hasfinitesupport(d::Union{D,Type{D}}) where {D<:ContinuousUnivariateDistribution} = false
 
-struct HasNonIntegerSupport end
-struct HasIntegerSupport end
-struct HasIntegerUnitrangeSupport end
-
-supporttype(::UnivariateDistribution) = HasNonIntegerSupport()
-
 """
     params(d::UnivariateDistribution)
 
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -20,6 +20,12 @@ isupperbounded(d::Union{D,Type{D}}) where {D<:UnivariateDistribution} = maximum(
 hasfinitesupport(d::Union{D,Type{D}}) where {D<:DiscreteUnivariateDistribution} = isbounded(d)
 hasfinitesupport(d::Union{D,Type{D}}) where {D<:ContinuousUnivariateDistribution} = false
 
+struct HasNonIntegerSupport end
+struct HasIntegerSupport end
+struct HasIntegerUnitrangeSupport end
+
+supporttype(::UnivariateDistribution) = HasNonIntegerSupport()
+
 """
     params(d::UnivariateDistribution)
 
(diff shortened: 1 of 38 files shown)


devmotion:
**Note:** The simplified implementation for `Truncated` requires https://github.com/JuliaStats/LogExpFunctions.jl/pull/21, thus tests will fail until the LogExpFunctions PR is released.

mschauer:
It's there something to look out for in the review?

devmotion:
Hmm I'm not sure. Hopefully I could convey the main ideas in the comment above and then I guess it depends on which points you want to examine or discuss more carefully. I assume that src/univariates.jl should show the main structure and changes. The changes in src/truncate.jl, src/univariate/locationscale.jl and src/mixtures/... show how it simplifies (and fixes) the existing implementation. And finally the changes in src/univariate/... fix and simplify different (log)(c)cdf implementations. I extended the tests and (log)(c)cdf are tested for all modified distributions (I only changed these distributions since tests failed), so I am quite confident that the changes are correct. But it's always better if someone else checks it as well :slightly_smiling_face:

devmotion:
The LogExpFunctions PR was merged and released, so tests pass now.

mschauer (approved):

src/univariates.jl
```diff
@@ -20,6 +20,12 @@ isupperbounded(d::Union{D,Type{D}}) where {D<:UnivariateDistribution} = maximum(
 hasfinitesupport(d::Union{D,Type{D}}) where {D<:DiscreteUnivariateDistribution} = isbounded(d)
 hasfinitesupport(d::Union{D,Type{D}}) where {D<:ContinuousUnivariateDistribution} = false
 
+struct HasNonIntegerSupport end
+struct HasIntegerSupport end
+struct HasIntegerUnitrangeSupport end
+
+supporttype(::UnivariateDistribution) = HasNonIntegerSupport()
+
```
devmotion, line 28:
```suggestion
```

PR #1349, opened 2021-06-17
add LogUniform

### LINKED ISSUE #1315: A Log Uniform distribution
Hi everyone,

I am using the Distributions.jl package along with [MLJ](https://github.com/alan-turing-institute/MLJ.jl) in order to sample hyper-parameters for hyper-parameter optimization.

The distribution I would like to use is the Log Uniform distribution, also called the Reciprocal Distribution.
I have searched this package but could not find its implementation.

I am currently using this workaround (for the specific case of Discrete Log Uniform Distribution):

```julia
using Random
import Base.rand

struct DiscreteLogUniform
    a
    b    
end

Base.rand(rng::Random.AbstractRNG, d::DiscreteLogUniform) = exp(log(d.a) + (log(d.b) - log(d.a)) * rand(rng)) |> round |> Int
```

that is currently filling my needs (which are only for sampling) but thought it could be nice if this distribution would be added to the repository.

More information concerning the Log Uniform distribution can be find on [Wikipedia](https://en.wikipedia.org/wiki/Reciprocal_distribution) and [Scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.loguniform.html).

Thanks for your time. 

Best regards.

jw3126:
fix #1315

This is bare-bones and not ready for merge. If a maintainer can confirm that we want a LogUniform here, I will add tests, docs, missing methods etc.


devmotion:
Thanks for the PR! I think `LogUniform` would be a good addition, as also suggested in https://github.com/JuliaStats/Distributions.jl/issues/1417. Maybe the implementation in BAT (https://github.com/bat/BAT.jl/blob/master/src/distributions/log_uniform.jl) can be useful for this PR as well.

src/univariate/continuous/loguniform.jl
```diff
@@ -0,0 +1,68 @@
...
+
+
+#### Parameters
+params(d::LogUniform) = (d.a, d.b)
+partype(::LogUniform{T}) where {T<:Real} = T
+
+#### Statistics
+
+function mean(d::LogUniform)
+    a = d.a; b = d.b
```
devmotion, line 29: Let's use `params`:
```suggestion
    a, b = params(d)
```

src/univariate/continuous/loguniform.jl
```diff
@@ -0,0 +1,68 @@
...
+    (b - a) / log(b/a)
+end
+function var(d::LogUniform)
+    a = d.a; b = d.b
```
devmotion, line 33:
```suggestion
    a, b = params(d)
```

src/univariate/continuous/loguniform.jl
```diff
@@ -0,0 +1,68 @@
...
+
+#### Evaluation
+function pdf(d::LogUniform{T}, x) where {T}
+    a = d.a; b = d.b
+    if insupport(d, x)
+        1/(x*log(b/a))
+    else
+        zero(T)
+    end
+end
```
devmotion, line 46: This implementation (and the methods below) are not type stable since they include a branch that returns a value based solely on `T` but not on the type of `x`. You can have a look at other distributions (such as the recently updated `Uniform`) how this can be avoided. The general guideline is:
- Don't use the type parameter
- Perform all computations, even if `x` is not in the support (sometimes it is necessary to use special values for values that are not in the support; e.g., for a distribution with support on the positive real line it might be necessary to perform computations for negative `x` with `_x = zero(x)` to avoid calling `log` with negative values)
- Only return special values such as `zero(result)` or `oftype(result, -Inf)` for values outside the support at the end of the computations.

E.g., for `pdf` here the preferred implementation would be
```suggestion
function pdf(d::LogUniform, x::Real)
    a, b = params(d)
    res = inv(x * log(b / a))
    return insupport(d, x) ? res : zero(res)
end
```

src/univariate/continuous/loguniform.jl
```diff
@@ -0,0 +1,68 @@
...
+function cdf(d::LogUniform{T}, x) where {T}
+    a = d.a; b = d.b
+    if x <= a
+        zero(T)
+    elseif x <= b
+        log(x/a) / log(x/b)
+    else
+        one(T)
+    end
+end
```
devmotion, line 56: Similarly, here something like (BTW it seems there is an error in your formula here - it should be divided by `log(b / a)`)
```suggestion
function cdf(d::LogUniform, x::Real)
    a, b = params(d)
    _x = clamp(x, a, b)
    return log(_x / a) / log(b / a)
end
```
should be used.

src/univariate/continuous/loguniform.jl
```diff
@@ -0,0 +1,68 @@
...
+function rand(rng::AbstractRNG, d::LogUniform)
+    u = Uniform(log(d.a), log(d.b))
+    exp(rand(rng, u))
+end
+function rand!(rng::AbstractRNG, d::LogUniform, out::AbstractArray)
+    u = Uniform(log(d.a), log(d.b))
+    rand!(rng, u, out)
+    out .= exp.(out)
+    out
+end
```
devmotion, line 68: If you define `quantile` these will be defined automatically.

devmotion (changes requested):

src/univariate/continuous/loguniform.jl
jw3126: Thanks a lot @devmotion for all the feedback. Does this mean this PR will be merged, once it is polished?

devmotion: Yes, this is my plan 🙂

jw3126: I ask because I did not receive any feedback for months, so I assumed there was no interest in having LogUniform here. Before I put work into this, I want to make sure it is not already doomed.

jw3126: Ok thanks a lot for confirming @devmotion

test/loguniform.jl
```diff
@@ -0,0 +1,39 @@
+module TestLogUniform
+using Test
+using Distributions
+import Random
+
+@testset "LogUniform" begin
```
jw3126, line 6: @devmotion how should I test LogUniform?

jw3126: Is it acceptable to calculate some values using `scipy.stats.loguniform` and hardcode these here as reference?

src/univariate/continuous/loguniform.jl
```diff
@@ -0,0 +1,82 @@
+"""
+    LogUniform(a,b)
+
+The *log uniform* distribution also known as *reciprocal distribution* is the exponential
+of the [`Uniform`](@ref) distribution `Uniform(log(a), log(b))`.
```
mschauer, line 5: "exponential of the distribution", I think it is not clear what that means. Perhaps follow Wikipedia and define `A positive random variable X is log-uniformly with parameters `a` and `b` if the logarithm of X is Uniform(log(a), log(b)).`

src/univariate/continuous/loguniform.jl
```diff
@@ -0,0 +1,81 @@
...
+External links
+
+* [Log uniform distribution on Wikipedia](https://en.wikipedia.org/wiki/Reciprocal_distribution)
+"""
+struct LogUniform{T<:Real} <: ContinuousUnivariateDistribution
+    a::T
+    b::T
+    LogUniform{T}(a::T, b::T) where {T <: Real} = new{T}(a, b)
+end
+Base.eltype(::Type{LogUniform{T}}) where {T} = float(T)
```
devmotion, line 18: This is something that is discussed from time to time but the current policy is not let the type of the parameters determine the type of the samples:
```suggestion
```
I am currently working on a PR that would still allow to generate samples of different types (and not only `Float64`), and there I'll also list more reasons for why IMO the type of the parameters and samples should be decoupled (simple example: `Poisson` - the type of the rate and samples have to be decoupled since the first one is usually a floating point number and the latter one an integer).

src/univariate/continuous/loguniform.jl
```diff
@@ -0,0 +1,81 @@
...
+
+function LogUniform(a::T, b::T; check_args=true) where {T <: Real}
+    check_args && @check_args(LogUniform, 0 < a < b)
+    LogUniform{T}(a, b)
+end
+
+function LogUniform(a::Real, b::Real)
+    a1,b1 = promote(float(a), float(b))
+    LogUniform(a1,b2)
+end
```
devmotion, line 28: It should not be necessary to convert the parameters to floating point numbers. In fact, for most computations it is easier if they are kept as integers since then no undesired type promotions occur (e.g., `Int` would be changed to `Float64` and thus computations with samples of type `Float32` will be promoted to `Float64`).
```suggestion
LogUniform(a::Real, b::Real) = LogUniform(promote(a, b)...)
```

src/univariate/continuous/loguniform.jl
```diff
@@ -0,0 +1,81 @@
...
+Base.maximum(d::LogUniform) = d.b
+
+#### Parameters
+params(d::LogUniform) = (d.a, d.b)
+partype(::LogUniform{T}) where {T<:Real} = T
+
+#### Statistics
+
+function mean(d::LogUniform)
+    a = d.a; b = d.b
```
devmotion, line 43:
```suggestion
    a, b = params(d)
```

src/univariate/continuous/loguniform.jl
```diff
@@ -0,0 +1,81 @@
...
+    (b - a) / log(b/a)
+end
+function var(d::LogUniform)
+    a, b = params(d)
+    log_ba = log(b/a)
+    (b^2 - a^2) / (2*log_ba) - ((b-a)/ log_ba)^2
+end
+StatsBase.median(d::LogUniform) = quantile(d, 1//2)
+StatsBase.mode(d::LogUniform)   = d.a
+StatsBase.modes(d::LogUniform)  = FillArrays.Fill(mode(d), 0)
```
devmotion, line 53: This is already the default fallback of `median`, so it can be removed. I haven't seen the use of `Fill` for `modes`, I suggest we just create an empty vector as for other distributions:
```suggestion
mode(d::LogUniform)   = d.a
modes(d::LogUniform)  = partype(d)[]
```

src/univariate/continuous/loguniform.jl
```diff
@@ -0,0 +1,81 @@
...
+    return insupport(d, x) ? res : zero(res)
+end
+function cdf(d::LogUniform, x::Real)
+    a, b = params(d)
+    _x = clamp(x, a, b)
+    return log(_x / a) / log(b / a)
+end
+logpdf(d::LogUniform, x::Real) = log(pdf(d,x))
+
+function quantile(d::LogUniform{T}, p::U) where {T,U<:Real}
```
devmotion, line 68: The type parameters are not needed here:
```suggestion
function quantile(d::LogUniform, p::Real)
```

src/univariate/continuous/loguniform.jl
```diff
@@ -0,0 +1,81 @@
...
+
+truncated(d::LogUniform, lo, hi) = truncated_LogUniform(d, promote(lo, hi)...)
+function truncated(d::LogUniform, lo::T, hi::T) where {T<:Integer}
+    # this method is needed to fix ambiguities
+    truncated_LogUniform(d, promote(lo, hi)...)
+end
+function truncated_LogUniform(d::LogUniform, lo::T, hi::T) where {T}
+    a,b = params(d)
+    LogUniform(max(a, lo), min(b, hi))
+end
```
devmotion, line 81: Can you move the `truncated` implementation to `src/truncated/loguniform.jl`, similar to `src/truncated/uniform.jl`? The implementation should also be simplified: one only has to implement
```julia
function truncated(d::LogUniform, lo::T, hi::T) where {T<:Real}
    a, b = params(d)
    return LogUniform(max(a, lo), min(b, hi))
end
```
The promotion is handled already in https://github.com/JuliaStats/Distributions.jl/blob/5d879bacf208ab7531ff9d438a8d6c75e30aa8d1/src/truncate.jl#L16-L18, which explains the method ambiguities in the current implementation here.

test/loguniform.jl
devmotion: We should test all definitions, i.e., methods and types defined in loguniform.jl. All methods should be type-stable, i.e., checked with `@inferred`. The correctness of e.g. `mean`, `var`, `pdf`, `logpdf`, `cdf`, and `quantile` can be checked with hard-coded values, e.g., obtained from `scipy`. Consistency checks are also simple and useful (e.g., `cdf(quantile(d, p)) \isapprox p` or similar).

test/loguniform.jl
```diff
@@ -0,0 +1,50 @@
...
+    rng = Random.MersenneTwister(0)
+    d = LogUniform(1,10)
+    @test 1 <= rand(rng, d) <= 10
+    @test typeof(rand(rng, d) ) == eltype(d)
```
devmotion, line 10: A bit simpler:
```suggestion
    @test rand(rng, d) isa eltype(d)
```

test/loguniform.jl
```diff
@@ -0,0 +1,50 @@
...
+    @test quantile(d, 0) ≈ 1
+    @test quantile(d, 0.5) ≈ sqrt(10) # geomean
+    @test quantile(d, 1) ≈ 10
+    @test mode(d) ≈ 1
+    @test insupport(d, 0) == false
```
devmotion, line 15:
```suggestion
    @test !insupport(d, 0)
```

test/loguniform.jl
```diff
@@ -0,0 +1,50 @@
...
+    @inferred pdf(d, 1)
+    @inferred cdf(d, 1)
+    @inferred ccdf(d, 1)
+    @inferred logpdf(d, 1)
+    @inferred logcdf(d, 1)
+    @inferred quantile(d, 1)
+    @inferred cquantile(d, 1)
+    @inferred rand(rng, d)
+    @inferred params(d)
+    @inferred partype(d)
```
devmotion, line 44: Can you merge the `@inferred` checks with the tests above? E.g.
```julia
@test @inferred(pdf(d, 5)) ≈ 0.08685889638065035
```

test/loguniform.jl
mschauer: In particular, you can test `median`, `quantile` and `cdf` against the corresponding transformed median, quantile, cdf of the uniform distribution.

test/loguniform.jl
jw3126: Mhh `eltype(d) = Any` would pass the test you suggest. This is undesirable I think?

devmotion: Ah, I see your point. If your interested in `eltype(d)` and not that `rand(rng, d)` is of `eltype(d)`, then I would suggest checking it directly:
```suggestion
    @test eltype(d) === Float64
    @test rand(rng, d) isa eltype(d)
```
Generally, it is preferable to check equality of types and singletons with `===`.

jw3126: I am a bit confused now. What information does the eltype of a distribution encode? Is `eltype(d) === typeof(rand(rng, d))` expected to hold in general? How about `typeof(pdf(d, 0)) === eltype(d)`?

devmotion: > What information does the eltype of a distribution encode? Is eltype(d) === typeof(rand(rng, d)) expected to hold in general?

According to the docstring of `eltype` it should return the element type of a sample. But there's a lot of issues and inconsistencies with how it is implemented currently. I suggest we don't focus on this too much here (I hope to improve the situation soon) but just use the same approach as in most other distributions currently.

> How about typeof(pdf(d, 0)) === eltype(d)?

No, the return type of `pdf`, `cdf` etc. is not enforced and is just whatever the function returns. One main motivation is that it also depends on the type of `x` (the point where these functions are evaluated).

src/truncated/loguniform.jl
```diff
@@ -0,0 +1,7 @@
+truncated(d::LogUniform, lo::T, hi::T) where {T<:Real} = truncated_LogUniform(d,lo,hi)
+truncated(d::LogUniform, lo::Integer, hi::Integer) = truncated_LogUniform(d,lo,hi)
+
+function truncated_LogUniform(d::LogUniform, lo, hi)
+    a,b = params(d)
+    LogUniform(max(a, lo), min(b, hi))
+end
```
devmotion, line 7: We only need the following:
```suggestion
function truncated(d::LogUniform, lo::T, hi::T) where {T<:Real}
    a, b = params(d)
    return LogUniform(max(a, lo), min(b, hi))
end
```

jw3126: Mhh I got ambiguities with your suggestion. The problem is 
https://github.com/JuliaStats/Distributions.jl/blob/master/src/truncate.jl#L42

src/univariate/continuous/loguniform.jl
```diff
@@ -0,0 +1,66 @@
...
+    a, b = params(d)
+    log_ba = log(b/a)
+    (b^2 - a^2) / (2*log_ba) - ((b-a)/ log_ba)^2
+end
+mode(d::LogUniform)   = d.a
+modes(d::LogUniform)  = partype(d)[]
+
+#### Evaluation
+function pdf(d::LogUniform, x::Real)
+    x, a, b = promote(x, params(d)...)
```
devmotion, line 52: These promotions should not be needed, what is the motivation for them?

jw3126: So that `Float32` is preserved. E.g. `pdf(LogUniform(1f0,2f0), 3f0)::Float32`

src/truncated/loguniform.jl
devmotion: Ah well spotted, I missed the definition for integers (I am not sure if it's a good default, maybe it should be removed in a separate PR). In any case, can you check if the following works?
```suggestion
function truncated(d::LogUniform, lo::T, hi::T) where {T<:Real}
    a, b = params(d)
    return LogUniform(max(a, lo), min(b, hi))
end

# fix method ambiguity
truncated(d::LogUniform, lo::Integer, hi::Integer) = truncated(d, float(lo), float(hi))
```
This would basically be the same as the definition for `Uniform`.

src/univariate/continuous/loguniform.jl
devmotion: Yes, this is seems desirable. Can you add a comment that explains it? And use e.g.
```suggestion
    _x, a, b = promote(x, params(d)...)
```
(or some other variable name) to ensure that there are no type stability issues (sometimes the compiler chokes if assign a new value of a different type to the same variable name).

src/truncated/loguniform.jl
jw3126: Mhhh it has the following unexpected (to me) consequence:
```
truncated(LogUniform(2, 10), 2, 14) === LogUniform(2.0, 10.0)
```
We started with integers and ended up with floats for no good reason.

devmotion: Ah, this is annoying but expected... Ideally we would just remove the default method for `Integer`s in src/truncated.jl but it's a bit unrelated to this PR and might break things. I'll check and see if I can fix it in a separate PR.

devmotion: It seems to work, I made a PR: https://github.com/JuliaStats/Distributions.jl/pull/1419

devmotion: The PR was merged and released. If you merge the latest changes from the master branch, `truncated(d::LogUniform, lo::T, hi::T)` should be sufficient.

src/truncated/loguniform.jl
```diff
@@ -0,0 +1,9 @@
+truncated(d::LogUniform, lo::T, hi::T) where {T<:Real} = truncated_LogUniform(d,lo,hi)
+
+function truncated_LogUniform(d::LogUniform, lo,hi)
+    a, b = params(d)
+    return LogUniform(max(a, lo), min(b, hi))
+end
+
+# fix method ambiguity
+truncated(d::LogUniform, lo::Integer, hi::Integer) = truncated_LogUniform(d,lo,hi)
```
devmotion, line 9: If you merge the latest changes from the master branch, you should be able to change this to:
```suggestion
truncated(d::LogUniform, lo::T, hi::T) where {T<:Real} = LogUniform(max(d.a, lo), min(d.b, hi))
```

src/univariate/continuous/loguniform.jl
```diff
@@ -0,0 +1,66 @@
...
+function LogUniform(a::T, b::T; check_args=true) where {T <: Real}
+    check_args && @check_args(LogUniform, 0 < a < b)
+    LogUniform{T}(a, b)
+end
+
+LogUniform(a::Real, b::Real) = LogUniform(promote(a, b)...)
+
+function convert(::Type{LogUniform{T}}, d::LogUniform) where {T}
+    LogUniform(T(d.a), T(d.b))::LogUniform{T}
+end
```
devmotion, line 28: I would assume that the type annotation is not needed?
```suggestion
convert(::Type{LogUniform{T}}, d::LogUniform) where {T<:Real} = LogUniform(T(d.a), T(d.b))
```

src/univariate/continuous/loguniform.jl
```diff
@@ -0,0 +1,66 @@
...
+#### Evaluation
+function pdf(d::LogUniform, x::Real)
+    x1, a, b = promote(x, params(d)...) # ensure e.g. pfd(LogUniform(1,2), 1f0)::Float32
+    res = inv(x1 * log(b / a))
+    return insupport(d, x1) ? res : zero(res)
+end
+function cdf(d::LogUniform, x::Real)
+    x1, a, b = promote(x, params(d)...)
+    x2 = clamp(x, a, b)
+    return log(x2 / a) / log(b / a)
```
devmotion, line 59:
```suggestion
    x2 = clamp(x1, a, b)
    return log(x2 / a) / log(b / a)
```
Alternatively, since `x1`, `a`, and `b` are already of the same type you should be able to use
```suggestion
    x1 = clamp(x1, a, b)
    return log(x1 / a) / log(b / a)
```

devmotion (commented):
Looks quite good, I think it's almost ready!

Can you please also add it to the list of continuous distributions in https://github.com/JuliaStats/Distributions.jl/blob/master/docs/src/univariate.md, together with a plot (similar to the other distributions)?

src/univariate/continuous/loguniform.jl
jw3126: @devmotion that one is tricky. Tha `clamp` takes care of promoting `x` but does not promote `a,b,`. Consider `a::Int, b::Int, x::Float32`. With the variant, you suggest the `log(b/a)` term will poison everything into `Float64`.

src/truncated/loguniform.jl
jw3126: Works!

src/univariate/continuous/loguniform.jl
devmotion: Ah no, I didn't suggest removing the `promote` line. Just fixing and changing the two lines below.

src/univariate/continuous/loguniform.jl
```diff
@@ -53,8 +53,8 @@ function pdf(d::LogUniform, x::Real)
 end
 function cdf(d::LogUniform, x::Real)
     x1, a, b = promote(x, params(d)...)
```
mschauer, line 55: Is there an advantage to use explicit `promote` instead of just doing the computation?

devmotion: Yes, no unintended peomotions for `Int` + `Float32`. It's the same reason as explained in `pdf`.

devmotion: @jw3126 Can you add a similar comment as for `pdf` to all promotions, please?

mschauer: Ah, I see, `1/Int` pollutes with `Float64`

mschauer: "pollute" no offence to the `Int`s 😁

jw3126: done

docs/src/univariate.md
```diff
@@ -327,6 +327,13 @@ LogNormal
 plotdensity((0, 5), LogNormal, (0, 1)) # hide
 ```
 
+```@docs
+LogUniform
+```
+```@example plotdensity
+plotdensity((0, 11), LogUniform, (1, 10)) # hide
```
mschauer, line 334: Did you see that the plot looks ok? I think we are done.

src/univariate/continuous/loguniform.jl
```diff
@@ -0,0 +1,64 @@
...
+    b::T
+    LogUniform{T}(a::T, b::T) where {T <: Real} = new{T}(a, b)
+end
+
+function LogUniform(a::T, b::T; check_args=true) where {T <: Real}
+    check_args && @check_args(LogUniform, 0 < a < b)
+    LogUniform{T}(a, b)
+end
+
+LogUniform(a::Real, b::Real) = LogUniform(promote(a, b)...)
```
devmotion, line 24: I know this is missing in many distributions but I think we should allow to skip the checks with this constructor as well:
```suggestion
LogUniform(a::Real, b::Real; kwargs...) = LogUniform(promote(a, b)...; kwargs...)
```

devmotion (commented):
Looks good, just one final suggestion :slightly_smiling_face:

docs/src/univariate.md
devmotion: image

devmotion: image

devmotion: It's copied from the docs that I built locally.

mschauer (approved):
I went ahead and committed the suggestion.

devmotion (approved):
Looks good to me, thanks a lot @jw3126 !

jw3126:
Btw is there an expression for the entropy of `LogUniform`?

devmotion:
Mathematica says it is `ln(a * b) / 2 + ln(ln(b / a))` (well, if I typed in everything correctly :smile:).

devmotion:
Seems to be correct:
```julia
julia> using Distributions

julia> function f(a, b)
           d = LogUniform(a, b)
           return - Distributions.expectation(d, Base.Fix1(logpdf, d))
       end

julia> g(a, b) = log(a * b) / 2 + log(log(b / a))

julia> a = rand()
0.7419428118736766

julia> b = a + 5 * rand()
1.5350807295138775

julia> f(a, b)
-0.2536879759021778

julia> g(a, b)
-0.2536879759021778
```

devmotion:
@jw3126 do you want to make a PR?

jw3126:
sure

mschauer:
Kullback-Leiber is invariant under parameter transformation, you could perhaps reduce it to a call of the `Uniform` version of the function and use @devmotion for the test. Or the other way around.

oschulz:
Nice, I'll get rid of `BAT.LogUniform` - had always meant to contribute it to Distributions eventually, but never got around to do it ... :-) Thanks @jw3126 !

PR #1351, opened 2021-06-18
Update CITATION.bib

davibarreira:
The `july` in the month is causing an error with the new `PkgCite.jl`.

### DIFF
--- a/CITATION.bib
+++ b/CITATION.bib
@@ -32,7 +32,7 @@
                   Moritz Schauer and
                   other contributors},
   title        = {{JuliaStats/Distributions.jl: a Julia package for probability distributions and associated functions}},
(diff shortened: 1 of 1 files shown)


CITATION.bib
```diff
@@ -32,7 +32,7 @@ @misc{Distributions.jl-2019
                   Moritz Schauer and
                   other contributors},
   title        = {{JuliaStats/Distributions.jl: a Julia package for probability distributions and associated functions}},
-  month        = july,
+  month        = "JUL",
```
andreasnoack, line 35:
```suggestion
  month        = jul,
```
according to the the BibTeX documentation, but I don't know if it matters.

devmotion (approved):

PR #1352, opened 2021-06-24
Implement std(::MultivariateDistribution)

bicycle1885:
This implements `std` for multivariate distributions as an element-wise square root of variance: `std(d) = sqrt.(var(d))`.


src/multivariates.jl
```diff
@@ -121,6 +121,13 @@ Compute the vector of element-wise variances for distribution `d`.
 """
 var(d::MultivariateDistribution)
 
+"""
+    std(d::MultivariateDistribution)
+
+Compute the vector of element-wise standard deviations for distribution `d`.
+"""
+std(d::MultivariateDistribution) = sqrt.(var(d))
```
devmotion, line 129: Maybe we could use
```suggestion
function std(d::MultivariateDistribution)
    v = var(d)
    map!(sqrt, v, v)
    return v
end
```
to avoid the broadcast machinery and reduce allocations?

devmotion (commented):
Seems reasonable! Can you also add `std` to the documentation of multivariate distributions?

devmotion (approved):

PR #1355, opened 2021-07-01
Fix `rand(Uniform{Float32})`

oscardssmith:
generate random numbers using the type of the distribution.

### DIFF
--- a/src/univariate/continuous/uniform.jl
+++ b/src/univariate/continuous/uniform.jl
@@ -109,7 +109,7 @@ end
 
 #### Sampling
 
-rand(rng::AbstractRNG, d::Uniform) = d.a + (d.b - d.a) * rand(rng)
+rand(rng::AbstractRNG, d::Uniform{T}) where {T <: Real} = d.a + (d.b - d.a) * rand(rng, T)
 
 rand!(rng::AbstractRNG, d::Uniform, A::AbstractArray) =
     A .= quantile.(d, rand!(rng, A))


devmotion:
This seems to be a subset of the changes in https://github.com/JuliaStats/Distributions.jl/pull/1045. This issue has been raised before but no decision has been made how to deal with the inconsistencies of `rand`: https://github.com/JuliaStats/Distributions.jl/pull/1045#issuecomment-571559677 The output type of rand is not always equivalent to the type of the parameters (e.g., continuous distributions with integer valued parameters) and sometimes is difficult to determine in advance (if `rand` involves some more advanced computations and/or parameters are dual numbers etc.).

oscardssmith:
Yeah. I understand that the issue has been raised before, but this is a 1 line change that brings it more in line with `Normal` and what seemed to be the majority opinion from that issue. I've made this as a separate PR in hopes that it being a 1 line change would make it easier to merge.

devmotion:
IMO this is not the right approach since it is not a general solution - as I said, sometimes you don't know the output type of `rand` in advance since the output type can't be deduced in an obvious way from the parameters. Currently, also `Normal` is based on heuristics (it has to call `float(T)` since otherwise it would already break for integer-valued parameters). I think it would be better to add an optional type parameter `T` to `rand(rng, dist)` that determines the type that is used for underlying `rand` or `randn` calls: https://github.com/JuliaStats/Distributions.jl/pull/1262#issuecomment-764995042

PR #1357, opened 2021-07-03
Fix #1356

### LINKED ISSUE #1356: Warning on cdf in src/univariate/discrete/betabinomial.jl in v0.25.7
With v0.25.7, I ran the test, which was passed without errors, yet generated the following warning message.
The contents of this warning message concerns me, so I want to post it here to ask if this should be OK or not.
Thanks!
```
 Testing Running tests...
WARNING: Method definition cdf(Distributions.BetaBinomial{T} where T<:Real, Int64) in module Distributions at C:\Users\xxx\.julia\packages\Distributions\nSnDu\src\univariate\discrete\betabinomial.jl:96 overwritten 
at C:\Users\xxx\.julia\packages\Distributions\nSnDu\src\univariate\discrete\betabinomial.jl:101.
  ** incremental compilation may be fatally broken for this module **
```

devmotion:
Fixes https://github.com/JuliaStats/Distributions.jl/issues/1356.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.25.7"
+version = "0.25.8"
 
 [deps]
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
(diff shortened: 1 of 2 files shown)


mschauer:
What about cdf(d::BetaBinomial, k::Real)?

devmotion:
There's no need to implement it, the default fallback in https://github.com/JuliaStats/Distributions.jl/blob/a6b99600a4d3182e533bb3bb4e0358f6aff50c8d/src/univariates.jl#L351 is sufficient. In general, for discrete distributions whose support is of type `Int` one just has to define `cdf(d, ::Int)` (otherwise one has to implement `cdf(d, ::Real)`). One does not have to define `ccdf`, `logcdf`, `logccdf` (they are computed from `cdf`) but it is numerically more stable to compute them by direct summation (with `logsumexp` for the `log` variants).

Issue #1333, opened 2021-05-24
MvNormal constructor is inconsistent between dim=1 and dim=2+


eperim:
The `MvNormal` makes different assumptions depending on `dim` being 1 or more. For `dim=1`, it takes the argument as the std, while for `dim=2+` it takes the argument as the cov:
```julia
julia> MvNormal(2.0 .* [1 0; 0 1])
ZeroMeanFullNormal(
dim: 2
μ: 2-element Zeros{Float64}
Σ: [2.0 0.0; 0.0 2.0]
)

julia> MvNormal(2.0 .* [1])
ZeroMeanDiagNormal(
dim: 1
μ: 1-element Zeros{Float64}
Σ: [4.0]
)
```
I understand that it is calling different constructors. In the second case, it assumes it's receiving a vector of stds:
```julia
julia> MvNormal(2.0 .* [1, 1])
ZeroMeanDiagNormal(
dim: 2
μ: 2-element Zeros{Float64}
Σ: [4.0 0.0; 0.0 4.0]
)
```
but this leads to a situation like this:
```julia
julia> A
1×1 Matrix{Any}:
 1
 
julia> MvNormal(2.0 .* A)
ZeroMeanFullNormal(
dim: 1
μ: 1-element Zeros{Float64}
Σ: [2.0]
)

julia> MvNormal(2.0 .* [1])
ZeroMeanDiagNormal(
dim: 1
μ: 1-element Zeros{Float64}
Σ: [4.0]
)
```
which seems dangerous. Probably having the constructor for the diagonal normal receive a vector of variances would make things more consistent and safer.

st--:
I agree, having the vector case behave qualitatively differently from the matrix case is highly counter-intuitive, and an easy source of bugs.

devmotion:
IMO the constructors for distributions are generally a bit confusing since they only use positional arguments. I think constructors with keyword arguments would be a lot clearer.

However, I disagree with the OP here, there is really no inconsistency here - vectors and matrices are fundamentally different types so it is fine to treat them differently. Moreover, vectors are treated in the same way as scalars and the second positional argument in the `Normal` constructor, they all indicate standard deviations. If you want to provide a vector of variances you can just provide a `Diagonal` matrix:
```julia
julia> MvNormal(Diagonal([2.0, 3.0]))
ZeroMeanDiagNormal(
dim: 2
μ: 2-element Zeros{Float64}
Σ: [2.0 0.0; 0.0 3.0]
[...]

Issue #1203, opened 2020-10-21
Inconsistency between single-arg MvNormal and Normal


sethaxen:
On Distributions v0.24.1, the mean-only constructor for `Normal` defaults to a variance of 1:
```julia
julia> using Distributions

julia> Normal(0)
Normal{Float64}(μ=0.0, σ=1.0)
```
while the mean-only constructor for `MvNormal` defaults to a covariance of 0:
```julia
julia> MvNormal(zeros(2))
ZeroMeanDiagNormal(
dim: 2
μ: 2-element Zeros{Float64}
Σ: [0.0 0.0; 0.0 0.0]
)
```

It seems to me that by providing the mean only, the user would usually expect an identity covariance matrix, analogous to `Normal`.

johnczito:
Apparently it's not a mean-only constructor. It's a "covariance"-only constructor:

https://github.com/JuliaStats/Distributions.jl/blob/6a22a99a6915ee42922d10bc8b89e88b337cab18/src/multivariate/mvnormal.jl#L223

Given a vector it always sets the mean to zero and fills in the diagonal of the covariance, which is why you got a zero matrix in your case. Alternatively:
```julia
julia> MvNormal([1., 2.])
ZeroMeanDiagNormal(
dim: 2
μ: 2-element Zeros{Float64}
Σ: [1.0 0.0; 0.0 4.0]
)
```
I prefer `MvNormal(x::AbstractVector) = MvNormal(x, I)` to `MvNormal(x::AbstractVector) = MvNormal(0, diagm(x))`, and that would be consistent with the univariate behavior, but I guess the current version isn't "wrong" if people like it.

sethaxen:
Wow, that's even more surprising. I can see how that could make sense, but the inconsistency with `Normal` seems like it could introduce bugs in users' code.

e.g. in Turing when you have `n` variables iid drawn from the same `Normal(0, 1)` distribution, you get an efficiency boost by using `MvNormal(zeros(n), 1)`. I encountered this issue by converting `Normal(0)` to `MvNormal(zeros(n))` and watching my model go insane.

devmotion:
> e.g. in Turing when you have `n` variables iid drawn from the same `Normal(0, 1)` distribution, you get an efficiency boost by using `MvNormal(zeros(n), 1)`.

Actually, if possible you should just specify `MvNormal(1)` since internally `Zeros` (from FillArrays) is used in such cases.

[...]

Issue #657, opened 2017-08-21
Bad default for multivariate normal


andreasnoack:
```julia
julia> D = MvNormal(zeros(2))
ZeroMeanDiagNormal(
dim: 2
μ: [0.0, 0.0]
Σ: [0.0 0.0; 0.0 0.0]
)
```

simonbyrne:
Yes, let's deprecate `MvNormal(::AbstractVector)` and replace it with `MvNormal(::Diagonal)`

andreasnoack:
Oh. It didn't even cross my mind that it would be the variance. I assumed that a vector would be the mean. So yes, let's definitely deprecate that for a `Diagonal` version. Eventually, I think we could consider reintroducing the constructor where the vector specifies the mean.

simonbyrne:
Actually, even more confusingly, it's standard deviation.

PR #1362, opened 2021-07-09
Deprecate use of vectors + scalars of standard deviations in constructors of multivariate normal distributions

devmotion:
There are multiple and regular complaints about the current constructors of `MvNormal` (and hence also of `MvNormalCanon` and `MvLogNormal`) being suprising and unintuitive (see, e.g., https://github.com/JuliaStats/Distributions.jl/issues/1203, https://github.com/JuliaStats/Distributions.jl/issues/1333, https://github.com/JuliaStats/Distributions.jl/issues/584, https://github.com/JuliaStats/Distributions.jl/issues/826, https://github.com/JuliaStats/Distributions.jl/issues/657).

This PR implements the suggestion in https://github.com/JuliaStats/Distributions.jl/issues/1203#issuecomment-716165782 and https://github.com/JuliaStats/Distributions.jl/issues/1203#issuecomment-716205791, and deprecates the constructors of `MvNormal` that accept a scalar standard deviation or a vector of standard deviations. Instead one should use
```julia
MvNormal(mu::AbstractVector{<:Real}, Sigma::AbstractMatrix{<:Real})
```
and
```julia
MvNormal(Sigma::AbstractMatrix{<:Real})
```
for distributions with zero mean. If the deprecations are removed, this will allow us to define
```julia
MvNormal(mu::AbstractVector{<:Real}) = MvNormal(mu, I)
```
which would fix the inconsistencies with the constructor of `Normal`. More concretely, the deprecations are:
```julia
Base.@deprecate MvNormal(μ::AbstractVector{<:Real}, σ::AbstractVector{<:Real}) MvNormal(μ, Diagonal(map(abs2, σ)))
Base.@deprecate MvNormal(μ::AbstractVector{<:Real}, σ::Real) MvNormal(μ, σ^2 * I)
Base.@deprecate MvNormal(σ::AbstractVector{<:Real}) MvNormal(Diagonal(map(abs2, σ)))
Base.@deprecate MvNormal(d::Int, σ::Real) MvNormal(Diagonal(Fill(σ^2, d)))
```

In the same way, the constructors of `MvNormalCanon` and `MvLogNormal` are updated and deprecated.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.25.10"
+version = "0.25.11"
 
 [deps]
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
--- a/src/multivariate/mvlognormal.jl
+++ b/src/multivariate/mvlognormal.jl
@@ -165,16 +165,15 @@ struct MvLogNormal{T<:Real,Cov<:AbstractPDMat,Mean<:AbstractVector} <: AbstractM
     normal::MvNormal{T,Cov,Mean}
 end
 
-#Constructors mirror the ones for MvNormmal
-MvLogNormal(μ::AbstractVector,Σ::AbstractPDMat) = MvLogNormal(MvNormal(μ,Σ))
-MvLogNormal(Σ::AbstractPDMat) = MvLogNormal(MvNormal(Zeros{eltype(Σ)}(dim(Σ)),Σ))
-MvLogNormal(μ::AbstractVector,Σ::Matrix) = MvLogNormal(MvNormal(μ,Σ))
+# Constructors mirror the ones for MvNormmal
+MvLogNormal(μ::AbstractVector{<:Real}, Σ::AbstractMatrix{<:Real}) = MvLogNormal(MvNormal(μ, Σ))
+MvLogNormal(Σ::AbstractMatrix{<:Real}) = MvLogNormal(MvNormal(Σ))
+
+# Deprecated constructors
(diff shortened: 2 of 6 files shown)


andreasnoack (approved):

mschauer (approved):

torfjelde:
I don't know _why_ this causes issues, but it seems like this PR is causing Turing.jl tests to grind to a halt https://github.com/TuringLang/Turing.jl/pull/1660#issuecomment-881907433. It _might_ be an issue on the Turing.jl side that interacts badly with this change, but I figured I'd mention it here in case it also affect other packages.

torfjelde:
I figured out what the "bug" is. With `--depwarn=yes`:
```julia

julia> @benchmark $(MvNormal)($(zeros(1)), $(ones(1)))
┌ Warning: `MvNormal(μ::AbstractVector{<:Real}, σ::AbstractVector{<:Real})` is deprecated, use `MvNormal(μ, Diagonal(map(abs2, σ)))` instead.
│   caller = ip:0x0
└ @ Core :-1
BenchmarkTools.Trial: 
  memory estimate:  27.44 KiB
  allocs estimate:  363
  --------------
  minimum time:     64.617 ms (0.00% GC)
  median time:      67.703 ms (0.00% GC)
  mean time:        68.065 ms (0.00% GC)
  maximum time:     85.660 ms (0.00% GC)
  --------------
  samples:          74
  evals/sample:     1
```
without `--depwarn=yes`:
```julia
julia> @benchmark $(MvNormal)($(zeros(1)), $(ones(1)))
BenchmarkTools.Trial: 
  memory estimate:  192 bytes
  allocs estimate:  2
  --------------
  minimum time:     81.207 ns (0.00% GC)
  median time:      84.698 ns (0.00% GC)
  mean time:        96.957 ns (10.22% GC)
  maximum time:     3.617 μs (96.73% GC)
  --------------
  samples:          10000
  evals/sample:     962
```

PR #1363, opened 2021-07-10
Use `show` method for all derived types

wildart:


src/mixtures/mixturemodel.jl
```diff
@@ -254,7 +254,7 @@ end
 
 #### show
 
-function show(io::IO, d::MixtureModel)
+function show(io::IO, d::AbstractMixtureModel)
     K = ncomponents(d)
     pr = probs(d)
     println(io, "MixtureModel{$(component_type(d))}(K = $K)")
```
devmotion, line 260: Probably this should be modified since in general the distribution is not a `MixtureModel`.

wildart: Maybe by putting space between two words.

andreasnoack: Not directly related to the change here but the two argument show methods shouldn't have line breaks so I think this should be changed to a three argument show methods.

src/mixtures/mixturemodel.jl
```diff
@@ -254,17 +254,17 @@ end
 
 #### show
 
-function show(io::IO, d::MixtureModel)
+function show(io::IO, d::AbstractMixtureModel)
```
andreasnoack, line 257:
```suggestion
function show(io::IO, ::MIME"text/plain", d::AbstractMixtureModel)
```

src/mixtures/mixturemodel.jl
```diff
@@ -254,17 +254,17 @@ end
...
     K = ncomponents(d)
     pr = probs(d)
-    println(io, "MixtureModel{$(component_type(d))}(K = $K)")
+    println(io, "Mixture Model{$(component_type(d))}(K = $K)")
     Ks = min(K, 8)
```
andreasnoack, line 261:
```suggestion
    Ks = get(io, :limit, true) ? min(K, 8) : K
```

src/mixtures/mixturemodel.jl
andreasnoack: @devmotion Are you okay with the updated version?

devmotion (commented):
Looks good :+1: Could we add a test that maybe just ensures that the method can be called without errors?

src/mixtures/mixturemodel.jl
```diff
@@ -254,17 +254,17 @@ end
...
+    println(io, "Mixture Model{$(component_type(d))}(K = $K)")
+    Ks = get(io, :limit, true) ? min(K, 8) : K
     for i = 1:Ks
         @printf(io, "components[%d] (prior = %.4f): ", i, pr[i])
-        println(io, component(d, i))
+        print(io, component(d, i), i < Ks || Ks < K ? "\n" : "")
     end
     if Ks < K
-        println(io, "The rest are omitted ...")
+        print(io, "The rest are omitted ...")
```
nalimilan, line 267: Isn't it the convention to end the output with a newline?

Issue #518, opened 2016-08-02
piecewise uniform distribution?


tawheeler:
Is there any chance for implementing a piecewise uniform distribution? (univariate or multivariate).
This is what a histogram generates.

Most stats functions on it should be well-defined; just not sure whether a `fit` function would be.

I can take a stab at it and submit a PR if desired.

andreasnoack:
Would it make sense for the applications you have in mind to overload the `Histogram` type in `StatsBase` and thereby making it a distribution?

tawheeler:
Do you mean `AbstractHistogram{T<:Real,N,E} <: Distribution`?
Thanks, I did not know about the Histogram type.

andreasnoack:
Maybe. I haven't really thought enough about it but it be worth trying out.

tawheeler:
Worst case you could do encapsulation:

````
type PiecewiseUniform <: ContinuousUnivariate
   hist::Histogram
end
```

or similar

andreasnoack:
Reopening this since it doesn't appear to be resolved

BenCurran98:
Hi, just wanted to see if this was still being considered? It would be very handy to some work I'm doing :)

RoyiAvital:
Couldn't it be considered as a mixture model of Uniform (In case that is supported)?

PR #1367, opened 2021-07-16
Add PieceweiseLinearDist

davidanthoff:
The idea is that one manually specifies points on the CDF, and anything in-between these points is then linearly interpolated. At the moment I require the first and last point to have a y value of 0 and 1 respectively, I _think_ anything else doesn't really make sense?

I'll add tests and docs later, but wanted to get some feedback first whether a) the maintainers think this belongs in this package and b) get some additional eyes on the implementation and check whether I made some stupid mistake somewhere.

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -137,6 +137,7 @@ export
     NormalInverseGaussian,
     Pareto,
     PGeneralizedGaussian,
+    PiecewiseLinearDist,
     Product,
     Poisson,
     PoissonBinomial,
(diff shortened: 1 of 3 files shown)


andreasnoack:
Thanks for opining the PR. A couple of immediate high-level comment: I believe this distributions is the same as the one proposed in https://github.com/JuliaStats/Distributions.jl/issues/518 and https://github.com/JuliaStats/Distributions.jl/pull/877. I think it's more common to use the shape of the density to inspire the name (e.g. semicircle and triangular) so "piecewise uniform" as proposed in #518 seems more appropriate to me than "piecewise linear".

ParadaCarleton:
I think this definitely looks like a good idea, but it should probably work in more than 1 dimension. `uniform` is currently specified with upper and lower endpoints, so I think that could naturally be extended to higher dimensions by having users provide a set of vertices.

ParadaCarleton:
(Worth noting that because of this convention that random variables should be described in terms of their PDF, `PiecewiseLinear` as a name suggests to me a _PDF_ that is piecewise linear, i.e. it should look like a bunch of trapezoids.)

PR #1368, opened 2021-07-17
Re-export from LinearAlgebra and FillArrays

torfjelde:
Deprecations introduced in #1362 suggest alternatives that use the following:
- `LinearAlgebra.Diagonal`
- `LinearAlgebra.I`
- `FillArrays.Fill`

It seems like these should then also be re-exported, in particular since Distributions.jl is a package that is often used by Julia beginners who might not be aware of where the above can be found.

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -29,6 +29,13 @@ export
     # re-export Statistics
     mean, median, quantile, std, var, cov, cor,
 
+    # re-export from LinearAlgebra
+    Diagonal,
+    I,
+
+    # re-export from FillArrays
+    Fill,
+
     # generic types
     VariateForm,
     ArrayLikeVariate,


ParadaCarleton:
I will add a slight comment (which you can feel free to ignore) -- I personally had exactly this problem when working with Distributions.jl a couple months ago as a beginner in Julia (I didn't realize `I` is only part of LinearAlgebra).

devmotion:
I think we should qualify all functions, structs etc. from other packages in the docs, e.g., we should write `LinearAlgebra.I` or `LinearAlgebra.Diagonal` instead of `I` or `Diagonal` to make it clear where these objects are defined. IMO, though, we should not reexport `Fill`, `Diagonal`, `I`, `PDMat` etc.

torfjelde:
> I think we should qualify all functions, structs etc. from other packages in the docs, e.g., we should write `LinearAlgebra.I` or `LinearAlgebra.Diagonal` instead of `I` or `Diagonal` to make it clear where these objects are defined. IMO, though, we should not reexport `Fill`, `Diagonal`, `I`, `PDMat` etc.

Sounds good :+1:

Closed-Limelike-Curves:
I like that idea.

devmotion:
It seems this was fixed by #1428.

PR #1369, opened 2021-07-21
Add `quantile` for `UnivariateMixture`

ForceBru:
Added the `quantile` function for `UnivariateMixture`s using `quantile_bisect`. Inspired by [this blog post](http://www.awebb.info/probability/2017/05/12/quantiles-of-mixture-distributions.html).

### DIFF
--- a/src/mixtures/mixturemodel.jl
+++ b/src/mixtures/mixturemodel.jl
@@ -219,6 +219,26 @@ function var(d::UnivariateMixture)
     return v
 end
 
+"""
+    quantile(d::UnivariateMixture, p::Real; tol=1e-12)
+
+Compute `p`th quantile of a `UnivariateMixture`.
+"""
+function quantile(d::UnivariateMixture, p::Real; tol=1e-12)
+    # The quantile of the mixture is
+    # greater than the smallest quantile of its components (`x_left`)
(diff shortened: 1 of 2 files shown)


devmotion:
Related: https://github.com/JuliaStats/Distributions.jl/pull/1195

ForceBru:
@devmotion, IMO, one of these should be merged, because:

1. Presumably, it would make plotting univariate  mixture models much easier. Currently, StatsPlots.jl [has to special-case them](https://github.com/JuliaPlots/StatsPlots.jl/issues/448#issuecomment-876003083). With support for quantiles one can just plot from some left quantile to some right one.
2. It's inconvenient that it's not possible to calculate quantiles of mixture models

My main reason for implementing this was plotting: I wanted to create a custom plot with mixtures, but suddenly found that there's no way to easily calculate quantiles of a mixture

devmotion:
Sure, I definitely agree, it would be great since it allows to simplify StatsPlots and is useful in general (I used the algorithm myself in one of my projects recently). However, it would be good to use an improved implementaton, either by adding parts of the other PR here or finishing the other PR.

devmotion:
Closed in favour of https://github.com/JuliaStats/Distributions.jl/pull/1389.

PR #1373, opened 2021-07-28
**BREAKING** Fix loss of structure in cov() and invcov() of AbstractMvNormal subtypes

### LINKED ISSUE #572: why do cov and invcov call full?
Internally many multivariate distributions represent the scale as a subtype of `PDMats.AbstractPDMat`, yet `cov` and `invcov` call `full` before returning it. This seems premature, eg suppose the user wants to do an eigendecomposition before some transformation, then it has to be converted back, or reconstructed from the parameters.

st--:
Resolves #572. Distributions.jl has a lower bound of PDMats=0.10, in which PDMat and PDiagMat are subtypes of AbstractMatrix, so no need to call Matrix() anymore.

What, if anything, do I still need to add to this PR to get it merged ?


st--:
Hi, is there anything that would hold back merging this PR ? @devmotion ?

devmotion:
I wonder if it is possible to check if/how breaking this would be for downstream packages. Is there actually any package that relies on `cov` returning a Matrix? I would like to merge this PR but it's blocked since it's technically breaking. In my experience though (also with e.g. SpecialFunctions 2) for such central packages such a tiny breaking change means putting a lot of work on downstream developers and requiring manual compat updates in many packages, but if the breaking change only affects very few packages it might not be worth it and easier to make a non-breaking release and fix a handful of packages instead of > 200.

devmotion:
I think we should do this. We already avoid the `Matrix` conversion for `mean`, so possibly we should just treat it as a bug. I also doubt it will have any implications for downstream packages apart from reduced number of allocations.

PR #1374, opened 2021-07-29
Fix the numerical instability in `Truncated`.

MrVPlusOne:
The current constructor for Truncated is not numerically stable, for example, 
```julia
original = Normal(-5.0, 0.2)
trunc = truncated(original, 0.0, 5.0)
test_point = 0.5
logpdf(trunc, test_point)  # returns Inf!
```
This is because the constructor directly performs the computation using probabilities instead of log probabilities:
```julia
function truncated(d::UnivariateDistribution, l::T, u::T) where {T <: Real}
    l < u || error("lower bound should be less than upper bound.")
    T2 = promote_type(T, eltype(d))
    lcdf = isinf(l) ? zero(T2) : T2(cdf(d, l))
    ucdf = isinf(u) ? one(T2) : T2(cdf(d, u))
    tp = ucdf - lcdf
    Truncated(d, promote(l, u, lcdf, ucdf, tp, log(tp))...)
end
```

This pull request fixes this issue by using `logcdf` and the numerically stable operation `logsubexp`.

### DIFF
--- a/src/truncate.jl
+++ b/src/truncate.jl
@@ -25,8 +25,11 @@ function truncated(d::UnivariateDistribution, l::T, u::T) where {T <: Real}
     T2 = promote_type(T, eltype(d))
     lcdf = isinf(l) ? zero(T2) : T2(cdf(d, l))
     ucdf = isinf(u) ? one(T2) : T2(cdf(d, u))
-    tp = ucdf - lcdf
-    Truncated(d, promote(l, u, lcdf, ucdf, tp, log(tp))...)
+    logcdf_l = isinf(l) ? T2(-Inf) : T2(logcdf(d, l))
+    logcdf_u = isinf(u) ? zero(T2) : T2(logcdf(d, u))
+    log_tp = logsubexp(logcdf_l, logcdf_u)
+    tp = exp(log_tp)
+    Truncated(d, promote(l, u, lcdf, ucdf, tp, log_tp)...)
 end
 
 truncated(d::UnivariateDistribution, l::Integer, u::Integer) = truncated(d, float(l), float(u))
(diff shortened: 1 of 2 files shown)


src/truncate.jl
```diff
@@ -25,8 +25,11 @@ function truncated(d::UnivariateDistribution, l::T, u::T) where {T <: Real}
     T2 = promote_type(T, eltype(d))
     lcdf = isinf(l) ? zero(T2) : T2(cdf(d, l))
     ucdf = isinf(u) ? one(T2) : T2(cdf(d, u))
-    tp = ucdf - lcdf
-    Truncated(d, promote(l, u, lcdf, ucdf, tp, log(tp))...)
+    logcdf_l = isinf(l) ? T2(-Inf) : T2(logcdf(d, l))
+    logcdf_u = isinf(u) ? zero(T2) : T2(logcdf(d, u))
```
devmotion, line 29: The changes in https://github.com/JuliaStats/Distributions.jl/pull/1348 should allow us to just use
```suggestion
    logcdf_l = logcdf(d, l)
    logcdf_u = logcdf(d, u)
```
Note, however, that this does not handle discrete distributions correctly - for these one would have to subtract the pdf of the lower bound from the cdf. But since it is already incorrect in the current release, this could be addressed in a separate PR.

devmotion: Similarly, one could drop `isinf` in the computation of `lcdf` and `ucdf`, and remove `T2` completely.

src/truncate.jl
```diff
@@ -59,12 +62,7 @@ end
...
-    l < u || error("lower bound should be less than upper bound.")
-    T2 = promote_type(T, eltype(d))
-    lcdf = isinf(l) ? zero(T2) : T2(cdf(d, l))
-    ucdf = isinf(u) ? one(T2) : T2(cdf(d, u))
-    tp = ucdf - lcdf
-    Truncated(d, promote(l, u, lcdf, ucdf, tp, log(tp))...)
+    truncated(d, l, u)
 end
 
 Truncated(d::UnivariateDistribution, l::Integer, u::Integer) = Truncated(d, float(l), float(u))
```
devmotion, line 68: These should be removed completely - `Truncated` is deprecated below.

src/truncate.jl
MrVPlusOne: Thanks for the comment! I have pushed a new commit to incorporate the suggested changes.

src/truncate.jl
MrVPlusOne: I tried to remove `Truncated` but it causes a bunch of tests to fail (and I guess potentially some code from downstream users), so I left it here.

PR #1376, opened 2021-08-04
Fix AbstractMvNormal docstring

st--:
The docstring for `AbstractMvNormal` has outdated definitions of the `MvNormal` struct and its aliases, which is quite confusing when trying to write own methods dispatching on the covariance type. This PR also fixes an outdated link and a minor inconsistency.


devmotion:
Thanks!

devmotion (approved):

Issue #1290, opened 2021-03-03
Rician distribution


mchitre:
I've implemented most of the functions in `Distributions.jl` for a Rician distribution. I note that this is a distribution in the _wanted_ list in #124. Would you be open to a PR to bake this into `Distributions.jl`, or should I make a standalone `RicianDistributions.jl` package for it?

**Consideration**:
I am currently using `Roots.jl` and `Optim.jl` in the implementations of `quantile()`, `mode()` and `fit()`. These are not currently dependencies of `Distributions.jl`, so is this a concern?

devmotion:
Maybe you can use one of the algorithms in https://github.com/JuliaStats/Distributions.jl/blob/master/src/quantilealgs.jl instead of depending on Roots and Optim?

mchitre:
yes, that should be doable for `quantile()`, and I can do something similar for `mode()` quite easily. The `fit()` still might need `Optim.jl`, but if that is undesirable, I can manually implement a minimization algorithm (it is only a 1D search, so not hard to do).

hdocmsu:
Hi @mchitre, I am very interested in using the Rician distribution. Looking forward to using it when you have it incorporated into Distributions.jl. Is there a timeline for when this PR will be completed?

mchitre:
I'll try and prepare an initial version over the next week or so, and then we review and take it from there, I guess.

hdocmsu:
> I'll try and prepare an initial version over the next week or so, and then we review and take it from there, I guess.

Great, thanks @mchitre for sharing the tentative timeline. Appreciate that!

mchitre:
I haven't dropped this, but have been delayed due to some emergencies to handle. I should be getting back to this in a couple of weeks time.

hdocmsu:
> I haven't dropped this, but have been delayed due to some emergencies to handle. I should be getting back to this in a couple of weeks time.

Thanks a lot for the heads up, @mchitre! I look forward to using the feature when it is available.

MattEttus:
If this is stalled, is the code that has already been written available anywhere?  I'd love to finish it up and get it in the package.

mchitre:
Thanks for the nudge @MattEttus. I'm on it today, and will submit a PR with the code I have shortly.

mchitre:
PR #1387 raised.

MattEttus:
Thanks, @mchitre !  Looks like you've made a ton of progress.

devmotion:
[...]

PR #1387, opened 2021-08-29
Add Rician distribution

mchitre:
Rician distribution added as per #1290


src/univariate/continuous/rician.jl
```diff
@@ -0,0 +1,152 @@
...
+
+#### PDF/CDF
+
+function logpdf(d::Rician, x::Real)
+    isnan(x) && return x
+    x ≤ 0.0 && return -convert(eltype(x), Inf)
+    isinf(x) && return -convert(eltype(x), Inf)
+    σ² = d.σ^2
+    log.(x) - log(σ²) - (x.^2 + d.ν^2)/(2σ²) + log.(besseli.(zero(eltype(x)), x .* (d.ν / σ²)))
+end
```
devmotion, line 101: The function is not type stable, e.g., if `x isa Int` or `x isa Rational{Int}`. We should use the same approach as in other distributions and perform as many calculations as possible and only return e.g. `oftype(result, NaN)` etc. in special cases if necessary. Since `NaN` and `Inf` usually are propagated just fine it might not be necessary to handle them at all.

The broadcasting operations should be removed since we only operate with scalars here.

src/univariate/continuous/rician.jl
```diff
@@ -0,0 +1,152 @@
...
+
+pdf(d::Rician, x::Real) = exp.(logpdf(d, x))
+
```
devmotion, line 104: This is already the default implementation and should not be needed:
```suggestion
```

src/univariate/continuous/rician.jl
```diff
@@ -0,0 +1,152 @@
...
+function cdf(d::Rician, x::Real)
+    isnan(x) && return x
+    x ≤ 0.0 && return zero(eltype(x))
+    isinf(x) && return one(eltype(x))
+    first(quadgk(x -> pdf(d, x), zero(eltype(x)), x; rtol=1e-8))
+end
```
devmotion, line 110: Also this function is type unstable and should be updated according to the suggestions above. `zero(eltype(x))` etc. can be replaced with `zero(x)`.

src/univariate/continuous/rician.jl
```diff
@@ -0,0 +1,152 @@
...
+function rand(rng::AbstractRNG, d::Rician)
+    x = randn(rng) * d.σ + d.ν
+    y = randn(rng) * d.σ
+    hypot(x, y)
+end
+
+#### Fitting
+
+# implementation based on the Koay inversion technique
+function fit(::Type{<:Rician}, x::AbstractArray{T}; tol=1e-12, maxiters=500) where T
```
devmotion, line 123: It would be good to base the computations not on `T` as it will currently lead to type instability problems e.g. when `theta` is redefined in the loop in the second branch.

devmotion (commented):
Thanks for the PR! I added some comments and suggestions :slightly_smiling_face:

src/univariate/continuous/rician.jl
```diff
@@ -0,0 +1,152 @@
...
+
+External links:
+
+* [Rician distribution on Wikipedia](https://en.wikipedia.org/wiki/Rice_distribution)
+
+"""
+struct Rician{T<:Real} <: ContinuousUnivariateDistribution
+    ν::T
+    σ::T
+    Rician{T}(ν, σ) where {T} = new{T}(ν, σ)
```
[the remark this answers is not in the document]
mchitre: I agree. It came from implementation of `Rayleigh()` in the `Distributions.jl` codebase, and so I assumed there was a good reason for it. If not, happy to remove it.

src/univariate/continuous/rician.jl
```diff
@@ -0,0 +1,152 @@
...
+end
+
+function Rician(ν::T, σ::T; check_args=true) where {T<:Real}
+    check_args && @check_args(Rician, ν ≥ zero(ν) && σ ≥ zero(σ))
+    return Rician{T}(ν, σ)
+end
+
+Rician(ν::Real, σ::Real) = Rician(promote(ν, σ)...)
+Rician(ν::Integer, σ::Integer) = Rician(float(ν), float(σ))
```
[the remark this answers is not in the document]
mchitre: I agree. It came from implementation of `Rayleigh()` in the `Distributions.jl` codebase, and so I assumed there was a good reason for it. If not, happy to remove it.

src/univariate/continuous/rician.jl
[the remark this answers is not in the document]
devmotion: If it's not needed and nothing breaks without it, I'd suggest removing it.

src/univariate/continuous/rician.jl
```diff
@@ -0,0 +1,152 @@
...
+
+function Rician(; K=0.0, Ω=2.0)
+    σ = sqrt(Ω / (K + 1) / 2)
+    ν = √2 * σ * sqrt(K)
+    Rician(ν, σ)
+end
```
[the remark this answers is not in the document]
mchitre: I thought long about it before suggesting this option.

The `ν` and `σ` parameters are more _standard_, and so only having the `K` and `Ω` parameters feels somewhat _wrong_. `K` and `Ω` parameters are somewhat specific to communications applications, although they have a nice interpretation as shape and scale.

I personally would still prefer to have both parametrization available to the user. But if I was forced to use only one, I'd prefer `ν` and `σ`. The only problem I see with that is that the `scale()` and `shape()` functions would return something that can't be used to create `Rician()`.

Thoughts?

src/univariate/continuous/rician.jl
```diff
@@ -0,0 +1,152 @@
...
+    Rician(T(d.ν), T(d.σ); check_args=false)
+end
+
+#### Parameters
+
+shape(d::Rician) = d.ν^2 / (2 * d.σ^2)
+scale(d::Rician) = d.ν^2 + 2 * d.σ^2
+
+params(d::Rician) = (d.ν, d.σ)
+@inline partype(d::Rician{T}) where {T<:Real} = T
```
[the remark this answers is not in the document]
mchitre: Again, it came from implementation of `Rayleigh()` in the `Distributions.jl` codebase, and I am happy to drop it if it wasn't there for a good reason.

src/univariate/continuous/rician.jl
```diff
@@ -0,0 +1,152 @@
...
+
+#### Statistics
+
+# helper
+L½(x) = exp(x/2) * ((1-x) * besseli(zero(x), -x/2) - x * besseli(oneunit(x), -x/2))
```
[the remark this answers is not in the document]
mchitre: OK, will rename.

src/univariate/continuous/rician.jl
```diff
@@ -0,0 +1,152 @@
...
+
+mean(d::Rician{T}) where T = d.σ * √T(π/2) * L½(-d.ν^2/(2 * d.σ^2))
+var(d::Rician) = 2 * d.σ^2 + d.ν^2 - (π * d.σ^2 / 2) * L½(-d.ν^2/(2 * d.σ^2))^2
+
+quantile(d::Rician, q::Real) = quantile_newton(d, q, mean(d))
+mode(d::Rician{T}) where T = _minimize_gss(x -> -pdf(d, x), zero(T), mean(d))
```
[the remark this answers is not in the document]
mchitre: Will do

mchitre:
@devmotion thanks for the comments. Will fix them after the parametrization approach is discussed and agreed on.

src/univariate/continuous/rician.jl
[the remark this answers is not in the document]
devmotion: I know it is used in other implementations but as far as I know it is just because of historical reasons. The compiler was improved a lot and actually it can be harmful to use too many `@inline` statements. So my guideline is to use it only if one actually notices performance problems but not in advance.

src/univariate/continuous/rician.jl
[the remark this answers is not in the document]
devmotion: Given these considerations, I think we should at least initially only use the standard parameters `nu` and `sigma` and not include the keyword argument constructor with `K` and `Omega`. We don't generally guarantee that one can reconstruct a distribution with the scale, shape, or location (e.g., `NormalCanon` does not satisfy this property), so it is fine if `scale` and `shape` return something that can't be used to create a `Rician` distribution.

src/univariate/continuous/rician.jl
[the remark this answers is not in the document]
mchitre: Done in e7f6b63f45cf182a3e219dd1ce973d3770bc5857

src/univariate/continuous/rician.jl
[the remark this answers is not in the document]
mchitre: Done in e7f6b63f45cf182a3e219dd1ce973d3770bc5857

src/univariate/continuous/rician.jl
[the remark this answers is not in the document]
mchitre: Actually inner constructor is needed to avoid a method definition overridden error, so reverting this change.

src/univariate/continuous/rician.jl
mchitre: Good point. Computation now uses `Float64`, with a type conversion at the end for stability.

4344ea13cd21c2597ec520d1a7bed9a32044e380

src/univariate/continuous/rician.jl
mchitre: Done. Needs special handling for `NaN` and `Inf` as `quadgk` doesn't like to deal with them.

4344ea13cd21c2597ec520d1a7bed9a32044e380

src/univariate/continuous/rician.jl
mchitre: Broadcasting removed. Improved type stability. `Inf` propagation does not work for this case, as it results in `NaN` instead of `-Inf`, and so needs to be dealt with as a special case.

4344ea13cd21c2597ec520d1a7bed9a32044e380

mchitre:
@devmotion All comments addressed. Accepted changes are marked resolved. Other comments are left open for you to take a look at my changes and then to mark resolved. Thanks.

src/univariate/continuous/rician.jl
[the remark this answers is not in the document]
devmotion: I see. At some point we should remove the test and change the other distributions as well but let's keep it for now.

src/univariate/continuous/rician.jl
```diff
@@ -0,0 +1,153 @@
...
+end
+
+function cdf(d::Rician, x::Real)
+    let x = float(x)
+        isnan(x) && return x
+        x ≤ 0.0 && return zero(x)
+        isinf(x) && return one(x)
+        oftype(x, first(quadgk(x -> pdf(d, x), zero(x), x; rtol=1e-8)))
+    end
+end
```
devmotion, line 111: These functions are still not type stable, in general the result is not guaranteed to be of type `float(x)`. I am also not completely satisfied with the use of `quadgk` and `rtol = 1e-8` since it is unclear if the result will be precise enough in general.

Maybe the easiest would be to just forward evaluations to `NoncentralChisq`: if `X ~ Riccian(nu, sigma)`, then `Y := X^2 / sigma^2` is distributed according to `Y ~ NoncentralChisq(2, nu^2 / sigma^2)`. So we can define e.g.
```julia
function pdf(d::Rician, x::Real)
    ν, σ = params(d)
    return 2 * x / σ^2 * pdf(NoncentralChisq(2, (ν / σ)^2), (x / σ)^2)
end
function logpdf(d::Rician, x::Real)
    ν, σ = params(d)
    return log(2 * x / σ^2) + logpdf(NoncentralChisq(2, (ν / σ)^2), (x / σ)^2)
end
for f in (:cdf, :logcdf, :ccdf, :logccdf, :quantile, :cquantile)
    @eval begin
        function $f(d::Rician, x::Real)
            ν, σ = params(d)
            return $f(NoncentralChisq(2, (ν / σ)^2), (x / σ)^2)
        end
    end
end
```

mchitre: Also, the `NoncentralChisq` implementation is broken on many typical data types.

Example:
```julia
julia> logpdf(NoncentralChisq(2, 1.0), 1.0f0)
ERROR: MethodError: no method matching nchisqlogpdf(::Float64, ::Float64, ::Float32)
Closest candidates are:
  nchisqlogpdf(::Union{Float64, Int64}, ::Union{Float64, Int64}, ::Union{Float64, Int64}) at /Users/mandar/.julia/packages/StatsFuns/aZlbb/src/rmath.jl:73
Stacktrace:
 [1] logpdf(d::NoncentralChisq{Float64}, x::Float32)
   @ Distributions ~/Projects/Distributions.jl/src/univariates.jl:653
```

devmotion: Yes, IMO it would be better to use the existing functionality for `NoncentralChisq` and in StatsFuns. Hopefully the functions in StatsFuns will be moved to Julia implementations in the near future (as started e.g. in https://github.com/JuliaStats/StatsFuns.jl/pull/113), and then the limitations will be fixed automatically. Otherwise we can always reevaluate at a later point if we should implement some functions here, in a possibly numerically less optimal way.

mchitre: OK, will shortly update with code based on `NoncentralChisq`, but with some fixes for dealing with `Float32` and edge cases.

devmotion: > Also, the `NoncentralChisq` implementation is broken on many typical data types.
> 
> Example:

I guess the example was not tested/considered since it mixes different types. Nevertheless, of course it should work. It seems StatsFuns is missing support for `Float32` in https://github.com/JuliaStats/StatsFuns.jl/blob/master/src/rmath.jl. This should be fixed in StatsFuns, not in Distributions.

mchitre: Implementation updated to use `NoncentralChisq` in c22230174f83e233923b9bccceadf20d808ac0f7

src/univariate/continuous/rician.jl
```diff
@@ -0,0 +1,175 @@
...
+        σ = scale(fit(Rayleigh, float.(x)))
+    else
+        ξ(θ) = 2 + θ^2 - π/8 * exp(-θ^2 / 2) * ((2 + θ^2) * besseli(0, θ^2 / 4) + θ^2 * besseli(1, θ^2 / 4))^2
+        g(θ) = sqrt(ξ(θ) * (1+r^2) - 2)
+        θ = 1.0
+        for j ∈ 1:maxiters
+            θ⁻ = θ
+            θ = g(θ)
+            abs(θ - θ⁻) < tol && break
+        end
```
devmotion, line 161: The type of `θ` might change in the first step and hence create a type instability. I'd suggest
```suggestion
        θ = g(1)
        for j in 1:maxiters
            θ⁻ = θ
            θ = g(θ)
            abs(θ - θ⁻) < tol && break
        end
```

src/univariate/continuous/rician.jl
```diff
@@ -0,0 +1,175 @@
...
+        d = a + (b - a) / ϕ
+    end
+    (b + a) / 2
+end
+
+#### PDF/CDF/quantile delegated to NoncentralChisq
+
+# NOTE:
+# 1.0ν and 1.0x used to support x::Float32, since the underlying NoncentralChisq delegates
+# to StatsFuns(rmath.jl) which doesn't have support for Float32
```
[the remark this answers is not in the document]
mchitre: I initially tried to keep everything to the precision it was, but realized that many of the other distributions in this package propagate to `Float64`, and that `NoncentralChisq` doesn't even work with `Float32`. So while I agree with you that this should be fixed when `StatsFuns` in fixed, having a broken implementation in this PR seems extreme. My preference was to come back and change the implementation once `StatsFuns` is fixed, but you seem to prefer otherwise. I'll defer to your wisdom on this.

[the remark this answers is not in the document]
mchitre: Fixed in f8a7847 and other accepted changes.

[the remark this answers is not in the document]
devmotion: I opened a PR to StatsFuns that fixes the `Float32` issue: https://github.com/JuliaStats/StatsFuns.jl/pull/124

src/univariate/continuous/rician.jl
```diff
@@ -0,0 +1,171 @@
...
+
+params(d::Rician) = (d.ν, d.σ)
+partype(d::Rician{T}) where {T<:Real} = T
+
+#### Statistics
+
+# helper
+_Lhalf(x) = exp(x/2) * ((1-x) * besseli(zero(x), -x/2) - x * besseli(oneunit(x), -x/2))
+
+mean(d::Rician{T}) where T = d.σ * √T(π/2) * _Lhalf(-d.ν^2/(2 * d.σ^2))
```
devmotion, line 70:
```suggestion
mean(d::Rician) = d.σ * sqrthalfπ * _Lhalf(-d.ν^2/(2 * d.σ^2))
```

src/univariate/continuous/rician.jl
```diff
@@ -0,0 +1,171 @@
...
+var(d::Rician) = 2 * d.σ^2 + d.ν^2 - (π * d.σ^2 / 2) * _Lhalf(-d.ν^2/(2 * d.σ^2))^2
```
devmotion, line 71:
```suggestion
var(d::Rician) = 2 * d.σ^2 + d.ν^2 - halfπ * d.σ^2 * _Lhalf(-d.ν^2/(2 * d.σ^2))^2
```

src/univariate/continuous/rician.jl
```diff
@@ -0,0 +1,171 @@
...
+        d = a + (b - a) / ϕ
+    end
+    (b + a) / 2
+end
+
+#### PDF/CDF/quantile delegated to NoncentralChisq
+
+function quantile(d::Rician, x::Real)
+    ν, σ = params(d)
+    return sqrt(quantile(NoncentralChisq(2, (1.0ν / σ)^2), 1.0x)) * σ
```
devmotion, line 99:
```suggestion
    return sqrt(quantile(NoncentralChisq(2, (ν / σ)^2), x)) * σ
```

src/univariate/continuous/rician.jl
```diff
@@ -0,0 +1,171 @@
...
+end
+
+function cquantile(d::Rician, x::Real)
+    ν, σ = params(d)
+    return sqrt(cquantile(NoncentralChisq(2, (1.0ν / σ)^2), 1.0x)) * σ
```
devmotion, line 104:
```suggestion
    return sqrt(cquantile(NoncentralChisq(2, (ν / σ)^2), x)) * σ
```

test/truncate.jl
```diff
@@ -81,7 +81,8 @@ function verify_and_test(d::UnivariateDistribution, dct::Dict, n_tsamples::Int)
         if !(typeof(d) in [Distributions.Truncated{Distributions.NoncentralChisq{Float64},Distributions.Continuous, Float64},
                            Distributions.Truncated{Distributions.NoncentralF{Float64},Distributions.Continuous, Float64},
                            Distributions.Truncated{Distributions.NoncentralT{Float64},Distributions.Continuous, Float64},
-                           Distributions.Truncated{Distributions.StudentizedRange{Float64},Distributions.Continuous, Float64}])
+                           Distributions.Truncated{Distributions.StudentizedRange{Float64},Distributions.Continuous, Float64},
+                           Distributions.Truncated{Distributions.Rician{Float64},Distributions.Continuous, Float64}])
```
devmotion, line 85: Do you know what the problem is here?

devmotion (commented):
The PR looks very good, thanks a lot! I just had some minor suggestions, if tests pass I think we can merge it :slightly_smiling_face: 

I opened a PR to StatsFuns that fixes the `Float32` issues (https://github.com/JuliaStats/StatsFuns.jl/pull/124), so we shouldn't worry about them in this PR.

test/rician.jl
```diff
@@ -0,0 +1,97 @@
...
+
+    d1 = Rician(10.0, 10.0)
+    @test median(d1) == quantile(d1, 0.5)
+    x = quantile.(d1, [0.25, 0.45, 0.60, 0.80, 0.90])
+    @test all(cdf.(d1, x) .≈ [0.25, 0.45, 0.60, 0.80, 0.90])
+
+    x = rand(Rician(5.0, 5.0), 100000)
+    d1 = fit(Rician, x)
+    @test d1 isa Rician{Float64}
+    @test params(d1)[1] ≈ 5.0 atol=1e-1
```
devmotion, line 29: It seems this fails with Julia nightly since the error is 0.12 (https://github.com/JuliaStats/Distributions.jl/pull/1387/checks?check_run_id=3503702946#step:6:647). I assume it's just due to a different `x`. Can you relax the test a bit?

test/truncate.jl
mchitre: Yes, needs Dual numbers support for this test.

devmotion: Ah makes sense, then this will be fixed if there is a native Julia implementation of `nchisq` in StatsFuns :+1:

test/rician.jl
mchitre: That's strange. `runtests.jl` sets a random number seed to make these tests repeatable:
https://github.com/JuliaStats/Distributions.jl/blob/ed52948fb47e7e0aecaf81e99ebb090c2738cf0e/test/runtests.jl#L76

And on my machine, the tests pass. Anyway I've now added a seed while generating random numbers to see if it fixes the problem:
`x = rand(MersenneTwister(3456789), Rician(5.0, 5.0), 100000)`

Can you  check if the tests pass now?

devmotion: Random numbers are not guaranteed to be reproducible with different Julia versions and operating systems, even if a seed is fixed.

devmotion: So specifying the RNG explicitly here does not guarantee that the tests pass. StableRNG.jl fixes this issue but often the simpler solution is just to use a slightly larger tolerance.

mchitre: I see, I didn't realize that. Increased tolerance now -- hopefully that fixes this?

mchitre: These tests pass now

mchitre:
But two still fail...
```
 MethodError: no method matching nchisqinvcdf(::Float64, ::Float64, ::Rational{Int64})
  Closest candidates are:
    nchisqinvcdf(::Union{Float64, Int64}, ::Union{Float64, Int64}, ::Union{Float64, Int64}) at /home/runner/.julia/packages/StatsFuns/vyFzQ/src/rmath.jl:89
```
because `nchisqinvcdf` doesn't support rationals. To fix in `StatsFuns`, or should we convert to float here?

devmotion:
I extended the StatsFuns PR, it covers `Rational`s as well now.

mchitre:
@devmotion is there anything else needed from my end before we're ready to merge this?

devmotion:
No, the PR looks great :+1: I just wanted to wait for https://github.com/JuliaStats/StatsFuns.jl/pull/125, I still hope that it can be merged within the next days/week. If it takes longer, I think we can just mark the `median` tests as broken for now.

devmotion:
Thanks for the PR, with StatsFuns 0.9.12 all tests pass now :slightly_smiling_face:

PR #1198, opened 2020-10-16
Faster, more accurate Poisson Binomial PMF

bdeonovic:
I have done a lot of research in evaluating poisson binomial pmf in last 3 years. The simplest and fastest continues to be the recursive formula. I've cited Thomas & Taub (1982) but it predates this article as well because it is related to evaluating the elementary symmetric function. The Hong (2013) article does a very bad job at evaluating and comparing their method 

```julia
julia> x = [3.5118, .6219, .2905, .8450, 1.8648]
julia> p = big.(x) ./ (1 .+ big.(x))
julia> @benchmark poisbin_sum_taub(p)
BenchmarkTools.Trial: 
  memory estimate:  10.20 KiB
  allocs estimate:  230
  --------------
  minimum time:     7.075 μs (0.00% GC)
  median time:      7.525 μs (0.00% GC)
  mean time:        9.559 μs (12.27% GC)
  maximum time:     2.087 ms (66.47% GC)
  --------------
  samples:          10000
  evals/sample:     4

julia> @benchmark Distributions.poissonbinomial_pdf_fft(p)
BenchmarkTools.Trial: 
  memory estimate:  188.90 KiB
  allocs estimate:  4155
  --------------
  minimum time:     515.800 μs (0.00% GC)
  median time:      523.201 μs (0.00% GC)
  mean time:        558.614 μs (3.83% GC)
  maximum time:     9.248 ms (62.80% GC)
  --------------
  samples:          8935
  evals/sample:     1

julia> sum(abs.(poisbin_sum_taub(p) .- naive_sol))
1.106509096121475717627626332421301195049237546693398552456162007115350824780186e-77
julia> sum(abs.(Distributions.poissonbinomial_pdf_fft(p) .- naive_sol))
3.197437330491599664330118659436447055371737441489833207203143282015769955768403e-17
```
One method which does truly scale better than the recursive algorithm is Biscarri (2018), but it is a lot more involved but I do have it implemented in julia in a separate package right now. In simulation study you can see that it has comparable error rate, but is faster than the recursive algorithm for long inputs (see figure below where DFT-CF is Hong (2013), Taub is recursive algorithm implemented in this PR, and DC-FFT is Biscarri (2018)).

pmf_sim

PR #1390, opened 2021-08-30
Add ChainRules definitions for pdf of `PoissonBinomial`

devmotion:
I am in the process of updating DistributionsAD to ChainRules 1. I thought it would be a good opportunity to reduce the amount of type piracy and move rules for functions in Distributions here. Currently, there's only one such rule in DistributionsAD. It defines the pullback for the pdf of the `PoissonBinomial` distribution, so the PR adds and tests only this rule.

ChainRulesCore 1 is a very lightweight package (see [@oxinabox's comment](https://github.com/JuliaStats/StatsFuns.jl/pull/106#issuecomment-905063473) for more details) and Distributions already depends on ChainRulesCore indirectly via SpecialFunctions and StatsFuns (support for ChainRules was added in version 0.9.10 which was released today). DistributionsAD does not support ChainRulesCore 1 yet, so the definitions in Distributions won't lead to any conflicts with DistributionsAD if we remove the `rrule`s in DistributionsAD before we tag a CR1-compatible release.

The implementation is based on the same dynamic programming "trick" as the implementation of `poissonbinomial_pdf`. I didn't find any reference for it, I only derived it a while ago and added it to DistributionsAD. However, I am pretty confident that the ChainRules tests utilities would have revealed any errors.

/ cc: @bdeonovic who added the primal definition in https://github.com/JuliaStats/Distributions.jl/pull/1198

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ authors = ["JuliaStats"]
 version = "0.25.14"
 
 [deps]
+ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 PDMats = "90014a1f-27ba-587c-ab20-58faa44d9150"
@@ -17,6 +18,7 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 
 [compat]
+ChainRulesCore = "1"
 FillArrays = "0.9, 0.10, 0.11, 0.12"
 PDMats = "0.10, 0.11"
 QuadGK = "2"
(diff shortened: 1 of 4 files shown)


matbesancon (approved):

matbesancon:
Looks good, I had tried to start something in https://github.com/matbesancon/DistributionsDiff.jl/ but never got around sitting to continue this

matbesancon:
do you want to implement frule in this PR or a following one?

devmotion:
I'll add the `frule`s in this PR :+1:  I did not include them in DistributionsAD since the only AD backend in DistributionsAD with CR support is Zygote, so it didn't seem necessary. However, I guess we should add `frule`s in Distributions as well.

devmotion:
I added them :+1:

bdeonovic:
Not sure if it helps at all (or simplifies some code) but if p(x_1,...,x_k) is the poisson binomial PMF evaluated at (x_1,...,x_k) then the partial derivative of p with respect to x_i is equal to p(x_1,...,x_{i-1},0,x_{i+1},...x_k)

storopoli:
Great! Is this is the first AD in `Distributions.jl`?

devmotion:
AFAIK yes. `gradlogpdf` has existed for a long time but there were no AD backends used in the package. There are many tests with ForwardDiff though.

bdeonovic:
> 
> 
> Great! Is this is the first AD in `Distributions.jl`?

Yes; but only because AD generally works off the shelf without having to implement anything extra.

oschulz:
Awesome!

Issue #1320, opened 2021-05-05
`Product` methods not type stable for mixed distribution types


agerlach:
```
p1 = product_distribution([Uniform(1,2),Uniform(3,4)])
@code_warntype pdf(p1, [1.5, 3.5]) #Float64

p2 = product_distribution([Normal(0.0,1.0),Normal(10.0,1.0)])
@code_warntype pdf(p2, [0.0, 10.0] ) #Float64

p3 = product_distribution([Uniform(1,2),Normal(0.0,1.0)])
@code_warntype pdf(p3 [1.5, 0.0) ) #Any
```
I suspect this stems from  `Product.v<:AbstractVector{T}`.  However `eltype(p3)` is stable and returns a concrete type, so I would expect it possible to make these `methods` typestable without changing `Product`, pos

devmotion:
In general, `eltype` is not sufficient to determine the return type of `pdf`, so I don't think there's much hope with a vector of distributions that is not concretely typed. I think for small collections of distributions (in particular of different types) it would be beneficial if the restriction to `AbstractVector{T}` would be removed completely and one allow other containers such as `Tuple`s as well.

agerlach:
Something like this, or typed looser yet?
```
struct Product{
    S<:ValueSupport,
    T<:UnivariateDistribution{S},
    V<:Union{AbstractVector{T},Tuple{Varag{T}}}
} <: MultivariateDistribution{S}
    v::V
    function Product(v::V) where
        V<:Union{AbstractVector{T},Tuple{Varag{T}}} where
        T<:UnivariateDistribution{S} where
        S<:ValueSupport
        return new{S, T, V}(v)
    end
end

Product(ds...) = Product(d)
```

devmotion:
I'd remove the super type completely, similar to e.g. KernelSum in KernelFunctions.

devmotion:
On a second thought, it's maybe a bit unclear what the type of the samples should be if it is a tuple of distributions. Still a `Vector`, I assume?

agerlach:
I was debating that too. `Vector` was my initial reaction.

devmotion:
I think it makes sense - I view it as an inherent property of the Product distribution type that is not affected by how the distributions are provided.

agerlach:
hmmm... I don't know how to remove the super type for `V` and still set `S` and `T`

devmotion:
Do we need `T` really?

agerlach:
[...]

PR #1391, opened 2021-08-31
Generalize `Product`

devmotion:
DistributionsAD contains custom distribution types for vectors of multivariate distributions and matrices of univariate distributions. I think it would be good to add support for such more general product distributions to Distributions and remove them from DistributionsAD.

In this PR, I propose to generalize the existing `Product` such that it allows to construct `M + N`-dimensional product distributions from `AbstractArray{<:Distribution{ArrayLikeVariate{M}},N}` by stacking the `M`-dimensional distributions. The `ArrayLikeVariate` variate form (thanks to @oschulz!) enables us nicely to define these product distributions in an even more general way than what currently is implemented in DistributionsAD.

I tried to perform this generalization in a non-breaking way by adding a new type `ProductDistribution`~~, making `Product` a special alias of it, and deprecating the `Product` constructor. I did not find a way to deprecate `Product` with `Base.@deprecate_binding`, it seems it can't deal with type parameters: https://github.com/JuliaLang/julia/issues/9830~~. `Product` still exists and `product_distribution(::AbstractVector{<:UnivariateDistribution})` still creates a `Product` to avoid breaking changes (otherwise e.g. the downstream tests fail).

This PR requires:
- [x] implementation of interface for `MatrixOfUnivariateDistribution`
- [x] implementation of interface for `VectorOfMultivariateDistribution`
- [x] basic methods for more general cases
- [x] tests
- [x] implementations of more efficient dispatches for product distributions of `Fill{<:Distribution{<:ArrayLikeVariate}}` (if necessary; not all implementations in DistributionsAD might be needed)

cc: @yebai


oschulz:
I like the ability to create `M + N` products!

Question: Shouldn't we also allow for products of multivaritate dists (of same length) that result in matrixvariate dists, and so on for higher orders?

devmotion:
> Question: Shouldn't we also allow for products of multivaritate dists (of same length) that result in matrixvariate dists, and so on for higher orders?

That's exactly what the PR proposes to do :slightly_smiling_face:

matbesancon:
like the idea as well

oschulz:
Any news on this one?

devmotion:
I have quite many changes in a local branch but haven't had time to clean and polish them. There were some test errors caused by bugs in upstream packages such as StatsBase, so I had to fix these first. I'll try to finish the PR this week.

mschauer:
How do you treat the value support? Force all factors to have the same value support?

oschulz:
Thanks for doing this, @devmotion , this will come in *so* handy!

devmotion:
A small update: This PR works basically (unsurprisingly type inference in Julia 1.6 is better than in Julia 1.0 which explains the test - I'll just mark it as broken/not test it on Julia 1.0), I think mainly the docstrings have to be completed and coverage be improved:
- It adds a `ProductDistribution` (generalization of to-be-deprecated (? - see below) `Product`):
  - composed of a N-dimensional array of distributions **or** a tuple of distributions (fixes https://github.com/JuliaStats/Distributions.jl/issues/1320 and allows type stable code if the involved distributions are of different types)
  - both `eltype` and `value_support` are promoted (support of discrete distributions in some dimension is embedded in the continuous space of the other dimensions)
- It adds a `ReshapedDistribution` (generalization of to-be-deprecated `MatrixReshaped`):
   - defines the distribution of `reshape(X, size)` where `X` is a RV with N-dimensional arrays as variates

The reason to add both distributions in this PR are:
- both require and benefit from the generalizations of `logpdf` and `rand`/`rand!`
- `ReshapedDistribution` allows to specify also a N-dimensional product distribution of distributions of different types in a type stable way by specifying them as a tuple and reshaping the resulting product distribution

The main questions/problems are:
- Should one restrict `ProductDistribution` to vectors (but not N-dim. arrays) and tuples of distributions? And support product distributions of different shapes only via `ReshapedDistribution`?
- The promotion of `eltype` can only work if the `eltype` is a type parameter - however, this makes it impossible to deprecate `Product` in a non-breaking way as this type parameter does not exist in `Product` and can't be obtained statically from the array of distributions in general (if they are of different types)
  - Should we just keep `Product` around for now and only change `product_distribution(::AbstractVector)` to produce a `ProductDistribution` (I think this can be viewed as non-breaking)?
  - Should we just make a breaking release and remove `Product`?

oschulz:
> Should one restrict ProductDistribution to vectors (but not N-dim. arrays) and tuples of distributions? And support product distributions of different shapes only via ReshapedDistribution

Just a suggestion in case `ProductDistribution` restricted to vectors is easier/cleaner to implement.: `product_distribution(d, size1, size2, ...)` could automatically produce a `ProductDistribution` wrapped in a `ReshapedDistribution`.

mschauer:
``` 
# promotions: embed discrete set in continuous space
Base.promote_rule(::Type{Continuous}, ::Type{Discrete}) = Continuous
```
 
This would break with our current understanding of a `Continuous` distribution being  [a continuous distribution](https://en.wikipedia.org/wiki/Probability_distribution#Continuous_probability_distribution).

devmotion:
> Just a suggestion in case ProductDistribution restricted to vectors is easier/cleaner to implement.: product_distribution(d, size1, size2, ...) could automatically produce a ProductDistribution wrapped in a ReshapedDistribution.

My suggestion/proposal would be to only restrict the internal `ProductDistribution` but not the user-facing `product_distribution`. `product_distribution(x::AbstractArray)` would just return `reshape(ProductDistribution(vec(x)), size(x))` if `ndims(x) > 1`.

> This would break with our current understanding of a Continuous distribution being a continuous distribution

I don't think it would - at least not in the context of product distributions. But in the end it really comes to how we define `Discrete` and `Continuous` :shrug:

My understanding is: The value support is `Discrete` if the support is a discrete set (possibly countably infinite), and `Continuous` if the support is an uncountable set. This seems to be in line with the definition on Wikipedia that you linked to:
*A continuous probability distribution is a probability distribution whose support is an uncountable set*.

Using these definitions, the support of a product distribution with a mix of continuous and discrete distributions is not a discrete set but an uncountable set. Thus the value support of such a distribution in our type system should not be `Discrete` but `Continuous`.

That being said, I was wondering if it would be useful or misleading to use `promote` to achieve this behaviour instead of some custom promotion or trait system. I guess this is what you are concerned about? My motivation was that it seemed easier to use the standard promotion system instead of rolling our own. Additionally, I assumed it could be useful for e.g. mixture models as well. However, I see that it is also problematic: if one does not consider product (or mixture) distributions, it does not make sense to "promote" the support - if the support of a distribution is a discrete set, it can't be made uncountable by promotion. This seems to indicate that we should just use some custom promotion system for `ProductDistribution` and not define `promote_rule`, doesn't it?

mschauer:
> A continuous probability distribution is a probability distribution whose support is an uncountable set, such as an interval in the real line.

that’s a property of continuous distributions, not a definition. As mathematician you’ll find examples where this property holds without the distribution being continuous as defined in the subsection

devmotion:
Hmm I'm a bit confused - I didn't intend to make any statements about continuous distributions but mainly point out that my understanding of `Continuous` (the `ValueSupport`) seems to be nicely aligned with the first sentence in the link that you mentioned. Do you have a different understanding or interpretation of `Discrete` and `Continuous` (as types in the `ValueSupport` hierarchy, not in the sense of discrete and continuous distributions)?

oschulz:
>>>> `Base.promote_rule(::Type{Continuous}, ::Type{Discrete}) = Continuous`

>>> This would break with our current understanding of a Continuous distribution being a continuous distribution.

>> [...] in the end it really comes to how we define Discrete and Continuous [...] *A continuous probability distribution is a probability distribution whose support is an uncountable set*

> that’s a property of continuous distributions, not a definition.

Hm, but maybe we could make this work, still? I'm not sure about the exact definition, but I expect we want a continuous distribution to

* Have a support that and uncountable set

* Be defined on variates that have continous values

* hmmm, anything else?

The first condition is obviously satisfied for a product of continuous and discrete. Now the second condition - maybe there's room for interpretation there? The discrete distributions are already defined for continuous variate - for example, `logpdf(Poisson(), 0.4)` is allowed. And we already have continuous Distributions like `Dirichlet` whose support is a non-dense subset of R^n. I would say a `Dirichlet` and a product of `Normal` and `Poisson` aren't that different in that sense.

I guess it would come down to clearly define the implied base measure for the PDF of a product of continuous and discrete (the implementation of the PDF is clear, of course). But if `Dirichlet` can be `Continuous`, why can't `Normal()` times `Poisson()`?

devmotion:
As I said before, it really comes down to whatever we mean with a `ValueSupport` that is `Continuous` (or `Discrete`) :man_shrugging:

I tried to explain my understanding above, but it seems it differs from your interpretation, @oschulz (and I'm not sure what @mschauer's view on these types is but I got the impression from your comments that you also had something else in mind?). In my opinion, `Discrete` and `Continuous` only encode the countability of the support but in particular these types don't contain any information about
- for which values `logpdf` can be evaluated
- what the base measures of the densities are
(as the name `ValueSupport` indicates they only focus on the support :smile:). So in my view the second and third point in your list, @oschulz, are not relevant here.

oschulz:
> my understanding above, but it seems it differs from your interpretation, @oschulz

Maybe I misunderstood you - I actually meant to support your approach of promoting `Continuous` and `Discrete` to `Continuous`.

> in particular these types don't contain any information about [...] for which values logpdf can be evaluated

Hm, I think users would be surprised if a `Continuous` distribution would only accept integer variate values, though. ;-)  But like I said, it's not a concern since both our current `Continuous` and `Discrete` already accept floating point variate values.

> what the base measures of the densities are

Sure, the nature of the distribution itself shouldn't depend on the choice of default measure for `[log]pdf`. I just meant that we might define what the default base measure is (in words, and at least for ourselves) in a way that's compatible with what `[log]pdf` will *do* (quite naturally) for a product of `Continuous` and `Discrete`. It doesn't have to be in the docs of `Distributions`,  I was just trying to build a bridge to `MeasureTheory`.

devmotion:
Ah OK, then I misunderstood your comment, I thought you are worried about the handling and interpretation of `Continuous` and `Discrete` in `ProductDistribution` :slightly_smiling_face: 

I agree that it should be possible to evaluate `logpdf` of distributions with `Discrete` support also with floating point numbers - which is the case already and also the same behaviour as e.g. in Rmath.

> I just meant that we might define what the default base measure is (in words, and at least for ourselves) in a way that's compatible with what [log]pdf will do (quite naturally) for a product of Continuous and Discrete. It doesn't have to be in the docs of Distributions, I was just trying to build a bridge to MeasureTheory.

The base measure of `ProductDistribution` is the (uniquely defined) [product measure](https://en.wikipedia.org/wiki/Product_measure) of the individual base measures. I am not completely familiar with the interface in MeasureBase (maybe it'll be changed a bit in the refactoring as well) but it seems there is already some support for product measures (https://github.com/cscherrer/MeasureBase.jl/blob/master/src/combinators/product.jl and https://github.com/cscherrer/MeasureBase.jl/blob/master/src/combinators/pointwise.jl).

oschulz:
>   I thought you are worried about the handling and interpretation of `Continuous` and `Discrete` in `ProductDistribution`

No, quite the contrary! :-) I had never thought before about what the result of such a product would be, regarding "continuousness", but I think your approach makes perfect sense.

oschulz:
> The base measure of ProductDistribution is the (uniquely defined) product measure of the individual base measures.

Oh yes, you're right, it's simple, actually. :-)

mschauer:
Yeah,
```
Continuous <: ValueSupport
```
is the original sin, as `ValueSupport` is speaking about the support and we started small so that the distinction between

* the property that the support is not a countable set 

and

* the property that the distribution function is continuous or absolutely continuous

didn't matter yet, so at the beginning `Continuous` meant *both* (it meant even additionally that the samples where `Float64`!) and we didn't have to make up our mind.

But they can be different (e.g. a mixture between a Dirac and a Uniform has uncountable support but not a continuous distribution function.) So at the point we introduce distributions for which both notions diverge we need to make up our mind. We had to make a similar step when we introduced a `DiscreteNonParametric` which is discrete but not necessarily  `Int` or `Float64` so we had to fix the type system to accommodate it and needed to make up our minds.

So I just notice now we have to become clear "what we mean" and we should give it a thought.

devmotion:
I thought about these questions - and arrived at the interptetation of `Discrete` and `Continuous` as countable and uncountable support, respectively 🙂

I think this is the cleanest and most useful definition of these types. It is in line with the name of the supertype `ValueSupport` and makes the distinction between both types clear and, even more importantly, ensures that the `ValueSupport` of every distribution is either `Discrete` or `Continuous`. This allows us to cover eg also the mixture model (uniform + Dirac) in our `ValueSupport` system. The interpretation also matches a standard definition of discrete and continuous random variables: https://en.wikipedia.org/wiki/Random_variable#Standard_case

mschauer:
Oh, that's a bold definition, you really don't want "continuous random variables" having discontinuous distributions - that word "continuous" comes from somewhere!

But that's about the name, could still be that you are right and defining our `Continuous` as "not discrete" could be the right choice to make the type system work, and then your promotion rule works. Then a dedicated base measure function could make up for that that `Continuous` doesn't tell the full story anymore.

devmotion:
> Then a dedicated base measure function could make up for that that Continuous doesn't tell the full story anymore.

Exactly, that was my idea. I think base measures etc should be defined via some MeasureBase interface, the type system is just not expressive enough. Generally it seems easier to define such things as properties of individual distributions that return instances (of eg base measures) instead of working with type parameters for such more concrete properties. Another advantage is that the type system would be both simple and complete.

mschauer:
Okay!

oschulz:
I've been discussing notations for product measures with @cscherrer. We already use `*` for something else in Distributions (#1438), but we could use a Cartesian product notation (so `×`). I think being able to write `Normal() × Exponential()` and the like would be neat. If that results in a `ProductDistribution` that uses a `Tuple` instead of an `Array` internally, it should be type-stable.

devmotion:
Sure, this could be done (and I've thought of it - with a different notation) but I thought such convenience functions could be discussed and possibly added in a separate PR as it is not needed for this PR. Hence I don't really want to start a discussion here but just mention that I thought `\otimes` (ie tensor product) would be the natural choice.

cscherrer:
Ideally, yes, but in LinearAlgebra, `× === cross`. See e.g. 
https://julialang.zulipchat.com/#narrow/stream/225540-gripes/topic/.5Ctimes
and
https://github.com/JuliaLang/julia/issues/37109

It's unfortunate, because the symbol is very common in other contexts. But at least for now, IMO this is reason enough to avoid it when possible.

oschulz:
> Sure, this could be done (and I've thought of it - with a different notation) but I thought such convenience functions could be discussed and possibly added in a separate PR as it is not needed for this PR.

Sure, of course.

> Hence I don't really want to start a discussion here but just mention that I thought `\otimes` (ie tensor product) would be the natural choice.

Oh yes, that would be neat too, and avoid the `× === cross` awkwardness.

oschulz:
@cscherrer, maybe you can start playing with `\otimes` in MeasureBase to see how that "feels"?

cscherrer:
I think we'll need to step back and look at the notation as a whole, hopefully get something self-consistent that doesn't conflict too badly with Distributions. But let's discuss this elsewhere 🙂

devmotion:
Principally, this PR is ready for review - but I really would like to revert the last commit which fixed the downstream tests. It introduces an annoying inconsistency, namely that `product_distribution(::AbstractVector{<:UnivariateDistribution})` falls back to creating a `Product` whereas for all other inputs the default definition of `product_distribution` creates a `ProductDistribution`. The `ProductDistribution` is clearly superior, it supports everything that `Product` does and fixes some of its bugs (such as support of `FillArray`s). However, it seems that `Broadcast.broadcased` expressions (used instead of `zip` to benefit from pairwise summation) are problematic for AD.

src/product.jl
```diff
@@ -0,0 +1,246 @@
...
+end
+
+"""
+    product_distribution(dists::AbstractArray{<:Distribution{<:ArrayLikeVariate{M}},N})
+
+Create a distribution of `M + N`-dimensional arrays as a product distribution of
+independent `M`-dimensional distributions by stacking them.
+
+The function falls back to constructing a [`ProductDistribution`](@ref) distribution but
+specialized methods can be defined.
```
mschauer, line 223: Should we say here what `pdf(d::ProductDistribution, x)` does?

oschulz:
This is going to be awesome, thanks so much for this @devmotion !

devmotion:
> Principally, this PR is ready for review - but I really would like to revert the last commit which fixed the downstream tests. It introduces an annoying inconsistency, namely that `product_distribution(::AbstractVector{<:UnivariateDistribution})` falls back to creating a `Product` whereas for all other inputs the default definition of `product_distribution` creates a `ProductDistribution`. The `ProductDistribution` is clearly superior, it supports everything that `Product` does and fixes some of its bugs (such as support of `FillArray`s). However, it seems that `Broadcast.broadcased` expressions (used instead of `zip` to benefit from pairwise summation) are problematic for AD.

Any comments? In principle the PR seems ready.

oschulz:
> Any comments? In principle the PR seems ready.

Just that I can't wait to use it! :-)

mschauer (approved):

src/multivariate/product.jl
```diff
@@ -20,6 +21,10 @@ struct Product{
         V<:AbstractVector{T} where
         T<:UnivariateDistribution{S} where
         S<:ValueSupport
+        Base.depwarn(
+            "`Product(v)` is deprecated, please use `product_distribution(v)`",
+            :Product,
+        )
```
torfjelde, line 27: Maybe I'm just brainfarting, but it seems like this combined with the `@deprecate` below means that no matter which constructor I use for a vector of univariate distributions, I'm going to get a deprecation-warning?

oschulz: Yes, at the moment one always get's a depwarn ... #1589

devmotion: Yeah it's unfortunate and fixed by https://github.com/JuliaStats/Distributions.jl/pull/1590. I thought someone would approve the PR and I would be able to make a bugfix release shortly after the issue was discovered but it seems nobody has approved it within almost two weeks. I'm going to merge it and tag a release now since the issue is quite annoying for downstream packages (and AFAICT even causes time outs in Turing).

oschulz: Thanks @devmotion !

torfjelde: Wonderful stuff @devmotion !

PR #1393, opened 2021-09-01
Add ZeroInflatedPoisson distribution

emfeltham:
In the wake of a [discussion over on StatisticalRethinkingTuring](https://github.com/StatisticalRethinkingJulia/StatisticalRethinkingTuring.jl/issues/1), I figured that it would be a good idea to put this together. It is used fairly extensively in the social sciences, and is probably not something that researchers should have to construct themselves. Thanks again.


emfeltham:
Hi again, I ran and added tests, and changed the code according to the test criteria. It should pass now, at least it does locally. Apologies, and thanks again.

src/univariate/discrete/zeroinflatedpoisson.jl
```diff
@@ -0,0 +1,178 @@
...
+* [Zero-inflated Poisson Regression on UCLA IDRE Statistical Consulting](https://stats.idre.ucla.edu/stata/dae/zero-inflated-poisson-regression/)
+* [Zero-inflated model on Wikipedia](https://en.wikipedia.org/wiki/Zero-inflated_model)
+* McElreath, R. (2020). Statistical Rethinking: A Bayesian Course with Examples in R and Stan (2nd ed.). Chapman and Hall/CRC. https://doi.org/10.1201/9780429029608
+
+"""
+struct ZeroInflatedPoisson{T<:Real} <: DiscreteUnivariateDistribution
+  λ::T
+  p::T
+
+  function ZeroInflatedPoisson{T}(λ::T, p::T) where {T <: Real}
```
mschauer, line 29: Can you adapt your indentation to 4 spaces?

mschauer:
Can’t  we have the entire thing as actual MixtureDistribution of Dirac at 0 and Poisson, or make that work?

Project.toml
```diff
@@ -6,7 +6,9 @@ version = "0.25.15"
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
+LambertW = "984bce1d-4616-540c-a9ee-88d1112d94c9"
```
nalimilan, line 9: Maybe it would make sense to define this distribution in StatsFuns rather than using a separate package?

@jlapeyre Would you be OK with that?

devmotion: IIRC there was a discussion about moving it to SpecialFunctions and there even exists a PR.

nalimilan: Ah yes that's https://github.com/JuliaMath/SpecialFunctions.jl/pull/84. Though it's quite outdated now and there have been new commits in LambertW since then.

@emfeltham Do you feel like reviving this PR (or opening a new one)? There seems to be lots of interest in it.

src/univariate/discrete/zeroinflatedpoisson.jl
emfeltham: Sure, will do.

src/univariate/discrete/zeroinflatedpoisson.jl
```diff
@@ -0,0 +1,178 @@
...
+end
+
+function convert(::Type{ZeroInflatedPoisson{T}}, d::ZeroInflatedPoisson{S}) where {T <: Real, S <: Real}
+  return ZeroInflatedPoisson(T(d.λ), T(d.p), check_args = false)
+end
+
+#### Parameters
+
+import StatsBase:params
+
```
devmotion, line 68: This is not needed:
```suggestion
```

src/Distributions.jl
```diff
@@ -239,6 +240,7 @@ export
     quantile,           # inverse of cdf (defined for p in (0,1))
     qqbuild,            # build a paired quantiles data structure for qqplots
     rate,               # get the rate parameter
+    excessprob,         # get the exess probability of zeros parameter (ZeroInflatedPoison)
```
devmotion, line 243: I am not sure if we should introduce and in particular export a new function here just for `ZeroInflatedPoisson`.

src/univariate/discrete/zeroinflatedpoisson.jl
```diff
@@ -0,0 +1,178 @@
...
+  if y == 0
+    LLs = zeros(typeof(log(d.λ)), 2)
+    LLs[1] = log(d.p)
+    LLs[2] = log1p(-d.p) - d.λ
+    LL = logsumexp(LLs)
+  else
+    LL = log1p(-d.p) + logpdf(Poisson(d.λ), y)
+  end
+  return LL
+end
```
devmotion, line 88: This function changes the type of `LL` which should be avoided and can be simplified:
```suggestion
function logpdf(d::ZeroInflatedPoisson, y::Real)
    lp = if iszero(y)
        logaddexp(log(d.p), log1p(-d.p) - d.λ)
    else
        log1p(-d.p) + logpdf(Poisson(d.λ), y)
    end
    return lp
end
```

src/univariate/discrete/zeroinflatedpoisson.jl
```diff
@@ -0,0 +1,178 @@
...
+  return out
+end
+
+## Sampling
+
+function rand(rng::AbstractRNG, d::ZeroInflatedPoisson)
+  zo = rand(rng, Uniform(0, 1))
+  return quantile(d, zo)
+end
+
```
devmotion, line 128: This is the default fallback and can be removed:
```suggestion
```

src/univariate/discrete/zeroinflatedpoisson.jl
```diff
@@ -0,0 +1,178 @@
...
+struct ZeroInflatedPoissonStats <: SufficientStats
+  sx::Float64   # (weighted) sum of x
+  p0::Float64   # observed proportion of zeros
+  tw::Float64   # total sample weight
+end
+
+suffstats(::Type{<:ZeroInflatedPoisson}, x::AbstractArray{T}) where {T<:Integer} = ZeroInflatedPoissonStats(
+    sum(x),
+    sum(x .== 0) / length(x),
```
devmotion, line 139:
```suggestion
    mean(iszero, x),
```

src/univariate/discrete/zeroinflatedpoisson.jl
```diff
@@ -0,0 +1,178 @@
...
+  zo = rand(rng, Uniform(0, 1))
+  return quantile(d, zo)
+end
+
+#### Fitting
+
+struct ZeroInflatedPoissonStats <: SufficientStats
+  sx::Float64   # (weighted) sum of x
+  p0::Float64   # observed proportion of zeros
+  tw::Float64   # total sample weight
```
devmotion, line 134: The types seem a bit restrictive.

src/univariate/discrete/zeroinflatedpoisson.jl
```diff
@@ -0,0 +1,178 @@
...
+end
+
+suffstats(::Type{<:ZeroInflatedPoisson}, x::AbstractArray{T}) where {T<:Integer} = ZeroInflatedPoissonStats(
+    sum(x),
+    sum(x .== 0) / length(x),
+    length(x)
+  )
+
+# weighted
+function suffstats(::Type{<:ZeroInflatedPoisson}, x::AbstractArray{T}, w::AbstractArray{Float64}) where T<:Integer
```
devmotion, line 144: The `Float64` type parameter seems a bit restrictive.

src/univariate/discrete/zeroinflatedpoisson.jl
```diff
@@ -0,0 +1,178 @@
...
+    LLs[1] = log(d.p)
+    LLs[2] = log1p(-d.p) - d.λ
+    LL = logsumexp(LLs)
+  else
+    LL = log1p(-d.p) + logpdf(Poisson(d.λ), y)
+  end
+  return LL
+end
+
+function cdf(d::ZeroInflatedPoisson, x::Real)
```
devmotion, line 90: This function is type unstable (eg. if the parameters and `x` are of type `Float32`). The best approach is to perform all calculations and just return `zero(result)` or `oftype(result, NaN)` if necessary for some values of `x` in the end.

src/univariate/discrete/zeroinflatedpoisson.jl
```diff
@@ -0,0 +1,178 @@
...
+  pd = Poisson(d.λ)
+
+  deflat_limit = -1.0 / expm1(d.λ)
```
devmotion, line 93: We should avoid harcoded `Float64` literals since they might cause unwanted promotions.

src/univariate/discrete/zeroinflatedpoisson.jl
```diff
@@ -0,0 +1,178 @@
...
+  elseif d.p < deflat_limit
+    out = NaN
+  else
+    out = d.p + (1 - d.p) * cdf(pd, x)
+  end
+  return out
+end
+
+# quantile
+function quantile(d::ZeroInflatedPoisson, q::Real)
```
devmotion, line 106: This function has the same problems as `cdf`.

src/univariate/discrete/zeroinflatedpoisson.jl
```diff
@@ -0,0 +1,178 @@
...
+
+  deflat_limit = -1.0 / expm1(d.λ)
+
+  if (q <= d.p)
+    out = 0
+  elseif (d.p < deflat_limit)
+    out = convert(Int64, NaN)
```
devmotion, line 113: This will throw an `InexactError`. In general, one should also avoid hardcoding `Int64` and use `Int` (defaults to `Int64` on 64bit) if possible since `Int64` can lead to a mixup of `Int32` and `Int64` on 32bit machines.

src/univariate/discrete/zeroinflatedpoisson.jl
```diff
@@ -0,0 +1,178 @@
...
+
+rate(d::ZeroInflatedPoisson) = d.λ
+excessprob(d::ZeroInflatedPoisson) = d.p
+
+#### Evaluation
+
+function logpdf(d::ZeroInflatedPoisson, y::Real)
+  LL = 0.0
+  if y == 0
+    LLs = zeros(typeof(log(d.λ)), 2)
```
mschauer, line 80: Is this really the way to compute it? Perhaps  a comment why we need this little dance

mschauer: I think this would be the right way for two numbers https://github.com/numpy/numpy/blob/e02af0b2991171c39bb3013fa83bdbe727cd3cda/numpy/core/src/npymath/npy_math_internal.h.src#L585

devmotion: Yes, this should be simplified. I suggested an alternative in https://github.com/JuliaStats/Distributions.jl/pull/1393#discussion_r702345981 :slightly_smiling_face:

mschauer: Ah, exactly!

mschauer:
So let’s make it ZeroInflated{Poisson}? For the same amount of code we get a couple of related zero inflated distributions, e.g ZeroInflated{NegativeBinomial}.

src/univariate/discrete/zeroinflatedpoisson.jl
emfeltham: Great -- thank you.

azev77:
Building on @mschauer we can generically allow creating all kinds of transformations (Zero-inflated/truncated/modified): 
(from my [discourse post](https://discourse.julialang.org/t/truncated-poisson-distribution-probs-not-adding-up-to-one/61053))
```julia
using Distributions
λ = 2.0; #Poisson parameter
zp = 0.4; #zero prob. 
dpoi = Poisson(λ)

dzip = MixtureModel([Dirac(0.0), dpoi], [zp, 1.0-zp])  #ZeroInflated ZIP
dztp = Truncated(dpoi, 1, Inf)                         #ZeroTruncated ZTP 
dzmp = MixtureModel([Dirac(0.0), dztp], [zp,1.0-zp])   #ZeroModified ZMP
```

sparktseung:
@azev77 

* Currently in [LRMoE](https://github.com/sparktseung/LRMoE.jl).jl v0.2.0, zero-inflated distributions are implemented as a separate object. e.g. there is `PoissonExpert(λ)` and there is `ZIPoissonExpert(p0, λ)`. I don't this is a good way and I may need to change it up in the future.

* I think your example using the `MixtureModel` in Distributions.jl is much better and more maintainable.

* For discrete distributions with zero inflation & modification, I'd suggest adding some notes in the documentation, to explicitly specify what is the actual zero probability. For zero-inflated Poisson, it should be `p0+(1-p0)*exp(-λ)`. A lot of people tend forget about the second term.

* For continuous distributions, some may have infinite density at zero, e.g. Gamma with shape<1. One should be careful about writing and interpreting `pdf`/`logpdf` in such case.

PR #1403, opened 2021-10-10
Plot pdf of continuous univariate distributions in documentation

devmotion:
This PR is an update of https://github.com/JuliaStats/Distributions.jl/pull/1115. I took the old PR and
- merged the master branch,
- replaced Gadfly with GR and unified and standardized the plots (e.g., one plotting function and same number of points),
- added missing plots for e.g. `Rician` which was added recently,
- enabled preview of documentation (you can check how the plots will look here: https://juliastats.org/Distributions.jl/previews/PR1403/univariate/#Continuous-Distributions),
- reverted an unrelated change in the docstring of `Arcsine`.

Since the plotting implementation is in a separate function `plotdensity` now and used for the pdf plots of all distributions, it is easier to change the style or backend of the plots. The main reasons for switching from Gadfly to GR are
- Gadfly depends on Distributions and hence building the documentation of breaking releases of Distributions will fail (there will always be some delay until the compat bounds in Gadfly are updated) whereas GR does not depend on Distributions,
- in my experience Gadfly seemed to be less actively maintained and releases, e.g., for updating compat bounds, were tagged less frequently (e.g., it took multiple months until a version compatible with Distributions 0.25 was released https://github.com/GiovineItalia/Gadfly.jl/issues/1551) than for GR or other plotting packages.

If more plotting themes or features are desired, one could easily replace GR with Plots (it does not depend on Distributions, only StatsPlots does). I chose GR for this initial PR since it is more lightweight than Plots and pulls in fewer dependencies.

### DIFF
--- a/docs/src/univariate.md
+++ b/docs/src/univariate.md
@@ -446,7 +446,7 @@ plotdensity((0, 2), Triweight, (1, 1)) # hide
 Uniform
 ```
 ```@example plotdensity
-plotdensity((-0.5, 1.5), Uniform, (0, 1)) # hide
+plotdensity((-0.5, 1.5), Uniform, (0, 1); ylim=(0, 1.5)) # hide
 ```
 
 ```@docs


devmotion:
Any additional suggestions? :slightly_smiling_face:

PR #1408, opened 2021-10-21
allow pmf call for discrete distributions

EvoArt:


devmotion:
You can always define
```julia
pmf(d::DiscreteDistribution, x) = pdf(d, x)
```
and then use `pmf` in your scripts, even if it is not defined in Distributions (I guess maybe you already do this right now?). I think it might be better to open an issue and gather feedback from other contributors and users if you would like `pmf` to be added to Distributions. I know that it is a common terminology and widely used, so maybe there are different opinions on this matter. It is still a probability density function as well, with respect to the counting measure, so `pdf` is not incorrect.

Due to the reasons mentioned above and since it was removed some time ago, I'll close this PR for now until there is more feedback and support for adding a `pmf` function.

PR #1409, opened 2021-10-22
Use pairwise summation in `loglikelihood`

devmotion:
This PR changes the implementation of `loglikelihood` such that it uses pairwise summation. This approach is based on https://github.com/JuliaLang/julia/pull/31020 and uses a fast path in recent Julia versions. It is already used in e.g. StatsBase successfully (xref: https://github.com/JuliaStats/StatsBase.jl/issues/518).


devmotion:
@andreasnoack would you like to wait for a potential change in base? While the implementation is more verbose, I still think it has an advantage even if `sum(f, x)` is redefined in base since no anonymous functions or `Base.Fix1` are needed.

src/matrixvariates.jl
```diff
@@ -226,11 +226,15 @@ rows and `size(d, 2)` columns, or an array of matrices of size `size(d)`.
...
-    return sum(i -> _logpdf(d, view(X, :, :, i)), axes(X, 3))
+    (size(X, 1), size(X, 2)) == size(d) || throw(DimensionMismatch("inconsistent array dimensions"))
+    # we use pairwise summation (https://github.com/JuliaLang/julia/pull/31020)
+    broadcasted = Broadcast.broadcasted(_logpdf, (d,), (view(X, :, :, i) for i in axes(X, 3)))
+    return sum(Broadcast.instantiate(broadcasted))
 end
 function loglikelihood(d::MatrixDistribution, X::AbstractArray{<:AbstractMatrix{<:Real}})
-    return sum(x -> logpdf(d, x), X)
+    # we use pairwise summation (https://github.com/JuliaLang/julia/pull/31020)
+    broadcasted = Broadcast.broadcasted(logpdf, (d,), X)
```
mschauer, line 236:
```suggestion
    broadcasted = Broadcast.broadcasted(logpdf, (d,), X) # computes sum(x -> logpdf(d, x), X)
```

src/matrixvariates.jl
```diff
@@ -226,11 +226,15 @@ rows and `size(d, 2)` columns, or an array of matrices of size `size(d)`.
 """
 loglikelihood(d::MatrixDistribution, X::AbstractMatrix{<:Real}) = logpdf(d, X)
 function loglikelihood(d::MatrixDistribution, X::AbstractArray{<:Real,3})
-    (size(X, 1), size(X, 2)) == size(d) || throw(DimensionMismatch("Inconsistent array dimensions."))
-    return sum(i -> _logpdf(d, view(X, :, :, i)), axes(X, 3))
+    (size(X, 1), size(X, 2)) == size(d) || throw(DimensionMismatch("inconsistent array dimensions"))
+    # we use pairwise summation (https://github.com/JuliaLang/julia/pull/31020)
```
mschauer, line 230:
```suggestion
    # we use pairwise summation (https://github.com/JuliaLang/julia/pull/31020)
    # to compute sum(i -> _logpdf(d, view(X, :, :, i)), axes(X, 3))
```

src/matrixvariates.jl
devmotion: The comment is not at the right place I think, nothing is computed in this line. And typically one would just evaluate it to an array. I'll add some comments above.

devmotion: I added a comment above (similar to your suggestion below I noticed afterwards).

src/matrixvariates.jl
devmotion: I added a similar comment, somehow I missed this suggestion.

src/matrixvariates.jl
```diff
@@ -226,11 +226,17 @@ rows and `size(d, 2)` columns, or an array of matrices of size `size(d)`.
...
+    # to compute `sum(Base.Fix1(_logpdf, d), eachslice(X; dims=3))`
+    broadcasted = Broadcast.broadcasted(_logpdf, (d,), (view(X, :, :, i) for i in axes(X, 3)))
+    return sum(Broadcast.instantiate(broadcasted))
 end
 function loglikelihood(d::MatrixDistribution, X::AbstractArray{<:AbstractMatrix{<:Real}})
-    return sum(x -> logpdf(d, x), X)
```
nalimilan, line 233: Doesn't this call already use pairwise summation?

devmotion: Hmm, I thought it doesn't but seems you're right (was it always the case or did it change at some point?):
```julia
julia> A = vcat([Float32(1E0)], fill(Float32(1E-8), 10^8));

julia> sum(identity, A)
1.9999989f0

julia> sum(x -> x, A)
1.9999989f0

julia> sum(Broadcast.instantiate(Broadcast.broadcasted(x -> x, A)))
1.9999989f0
```

In any case, I'll close this issue in favour of https://github.com/JuliaStats/Distributions.jl/pull/1391 which will generalize `logpdf` and `loglikelihood`.

devmotion:
In any case, I'll close this issue in favour of https://github.com/JuliaStats/Distributions.jl/pull/1391 which will generalize `logpdf` and `loglikelihood`.

src/matrixvariates.jl
nalimilan: No idea. That said, I get slightly different results in some cases:
```julia
julia> x = vcat([Float32(1E0)], fill(Float32(1E-8), 10^8));

julia> sum(sin, x) - sum(sin.(x))
-1.41859055f-5

```

PR #1414, opened 2021-11-01
Analytic KL divergences

theogf:
This PR proposes to add some analytic formulations of the Kullback-Leibler divergence between two distributions.
This basically imports all the code from https://github.com/theogf/KLDivergences.jl as suggested in https://github.com/theogf/KLDivergences.jl/issues/12
This introduces it for `Beta`, `Exponential` `Gamma` `InverseGamma` `Normal` `Poisson` `MvNormal` and `AbstractMvNormal`.
A general approach for `MultivariateDistribution` is also added by using **Monte Carlo Integration** with `mcexpectation`.
Tests are also added to compare analytical and numerical evaluations

`expectation` and `kldivergence` for univariate distributions have also been reworked.

~~There are docs with references to the existing Latex formula of each of the KL but that could go if needed~~


src/univariate/continuous/normal.jl
```diff
@@ -75,6 +75,17 @@ kurtosis(d::Normal{T}) where {T<:Real} = zero(T)
...
+"""
+    kldivergence(p::Normal, q::Normal)
+
+See [KL Gaussian](https://en.wikipedia.org/wiki/Normal_distribution#Other_properties)
+"""
+function kldivergence(p::Normal, q::Normal)
+    μp, σp = params(p)
+    μq, σq = params(q)
+    return 0.5 * (abs2(σp / σq) + abs2((μp - μq) / σq) - 1 + 2 * (log(σq) - log(σp)))
+end
```
devmotion, line 87: Eliminating some computations:
```suggestion
    μp = mean(p)
    σ2p = var(p)
    μq = mean(q)
    σ2q = var(q)
    return ((μp - μq)^2 + σ2p) / σ2q - 1 + log(σ2q / σ2p)) / 2
```

src/univariate/discrete/poisson.jl
```diff
@@ -85,6 +85,19 @@ function entropy(d::Poisson{T}) where T<:Real
 end
 
 
+"""
+    kldivergence(p::Poisson, q::Poisson)
+
+See [KL Poisson](https://en.wikipedia.org/wiki/Poisson_distribution#Other_properties)
+"""
```
devmotion, line 92:
```suggestion
```

src/univariate/discrete/poisson.jl
```diff
@@ -85,6 +85,19 @@ function entropy(d::Poisson{T}) where T<:Real
...
+function kldivergence(p::Poisson, q::Poisson)
+    λp, λq = rate.((p, q))
+    return λq - λp + λp * (log(λp) - log(λq))
+end
+
+
+
+
```
devmotion, line 100: Again we should remove broadcasting and reduce some computations. Additionally, `λp` can be zero and hence it is better to use `xlogy` here:
```suggestion
    λp = rate(p)
    λq = rate(q)
    λp_over_λq = λp / λq
    return xlogy(λp, λp_over_λq) - λp_over_λq
end

```

test/kldivergences.jl
```diff
@@ -0,0 +1,69 @@
+@testset "KL divergences" begin
+    function test_kl(p, q)
+        @test kldivergence(p, q) > 0
```
devmotion, line 3: Maybe also check that `kldivergence(p, p)` and `kldivergence(q, q)` is approximately zero?

test/kldivergences.jl
```diff
@@ -0,0 +1,69 @@
...
+        @test kldivergence(p, q) ≈ Distributions.mcexpectation(p, x -> let p = pdf(P,x); (p > 0) * log(p / pdf(Q, x)) end; n_samples=100_000)
```
devmotion, line 4: This should be updated. It might also be better to just call the fallback method with `invoke` instead of reimplementing it here.

test/kldivergences.jl
```diff
@@ -0,0 +1,69 @@
...
+        end
+        @testset "MvNormal" begin
+            n_dim = 2
+            p = MvNormal(zeros(n_dim), Matrix(0.5 * I(n_dim)))
+            q = MvNormal(ones(n_dim), Matrix(0.3 * I(n_dim)))
+            test_kl(p, q)
+        end
+    end
+
+end
```
devmotion, line 69:
```suggestion
end
```

devmotion (changes requested):
Thanks for opening this PR!

I made some suggestions for improvements :slightly_smiling_face:

theogf:
Thanks a lot for the review @devmotion I will address the comments tomorrow

src/functionals.jl
```diff
@@ -32,3 +32,12 @@ end
...
 end
+
+function kldivergence(P::MultivariateDistribution, Q::MultivariateDistribution; n_samples=100)
+    mcexpectation(P, x -> let p = pdf(P,x); (p > 0) * log(p) - logpdf(Q, x) end; n_samples)
+end
+
+function mcexpectation(distr::Distribution, f; n_samples::Int=100)
+    n_samples > 0 || throw(ArgumentError("number of samples should be > 0"))
+    sum(f, rand(distr) for _ in 1:n_samples) / n_samples
+end
```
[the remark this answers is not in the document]
theogf: I think `mean(f, itr)` only exists since 1.3

src/multivariate/mvnormal.jl
```diff
@@ -100,6 +100,13 @@ end
 
 mvnormal_c0(g::AbstractMvNormal) = -(length(g) * convert(eltype(g), log2π) + logdetcov(g))/2
 
+function kldivergence(p::AbstractMvNormal, q::AbstractMvNormal)
+    # This is the generic implementation for AbstractMvNormal, you might need to specialize for your type
+    length(p) == length(q) || 
+        throw(DimensionMismatch("Distributions p and q have different dimensions $(length(p)) and $(length(q))"))
+    return 0.5 * (tr(cov(q) \ cov(p)) + sqmahal(q, mean(p)) - length(p) + logdetcov(q) - logdetcov(p))
+end
```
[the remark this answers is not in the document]
theogf: Not necessarily. For example, in the test with the Cholesky representation it does not make sense to compute the `cov` to get `logdetcov`

src/functionals.jl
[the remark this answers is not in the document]
theogf: Also what you wrote (`xlogx((p > 0) * log(p))`) is wrong, since we sample from `p`, `p(x)` should not be part of the function

src/univariate/continuous/exponential.jl
```diff
@@ -60,6 +60,17 @@ kurtosis(::Exponential{T}) where {T} = T(6)
...
 
+"""
+    kldivergence(p::Exponential, q::Exponential)
+
+See [KL Exponential](https://en.wikipedia.org/wiki/Exponential_distribution#Kullback%E2%80%93Leibler_divergence)
+"""
+function kldivergence(p::Exponential, q::Exponential)
+    λp = scale(p)
+    λq = scale(q)
+    return log(λp) - log(λq) + λq / λp - 1
```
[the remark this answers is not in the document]
theogf: `logmxp1` return `log(x) - x +1` so one would still need to compute the `log` additionally to correct for it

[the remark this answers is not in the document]
theogf: Also : isn't `log(x) - log(y)` more stable than `log(x / y)` ?

src/univariate/continuous/normal.jl
theogf: How about doing similarly to your other propositions and having a `sigma2p_over_sigma2q` ?

src/univariate/discrete/poisson.jl
theogf: Your new formula was incorrect but I still used the `xlogy`

src/functionals.jl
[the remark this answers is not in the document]
devmotion: > I think mean(f, itr) only exists since 1.3

Ah yes, I missed this note in the docstring :+1:

> Also what you wrote (xlogx((p > 0) * log(p))) is wrong, since we sample from p, p(x) should not be part of the function

Ah, yes, I didn't mean to, I guess I copied something incorrectly.

The better suggestion:
```suggestion
function kldivergence(P::MultivariateDistribution, Q::MultivariateDistribution; kwargs...)
    expectation(P, x -> let p = pdf(P,x); (p > 0) * (log(p) - logpdf(Q, x)) end; kwargs..)
end

# Monte Carlo fallback
function expectation(distr::MultivariateDistribution, f; nsamples::Int=100, rng::AbstractRNG=GLOBAL_RNG)
    nsamples > 0 || throw(ArgumentError("number of samples should be > 0"))
    return _expectation(rng, f, sampler(distr), nsamples) 
end

# Function barrier for `sampler` type instabilities
_expectation(rng, f, sampler, n) =  sum(f, rand(rng, sampler) for _ in 1:n) / n
```

src/multivariate/mvnormal.jl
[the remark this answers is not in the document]
devmotion: I think it's OK if it's not the most efficient implementation in all cases, it's a fallback :shrug: But let's keep it for now and only get rid of `0.5`.

src/univariate/discrete/poisson.jl
devmotion: Ah true :smile:

src/univariate/continuous/exponential.jl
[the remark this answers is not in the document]
devmotion: No, not necessarily since due to cancellation the result of `log(x) - log(y)` can be much less precise.

> logmxp1 return log(x) - x +1 so one would still need to compute the log additionally to correct for it

No, if one would compute `logmxp1(lp_over_lq) = log(lp_over_lq) - lp_over_lq + 1` one would just have to subtract 2. But the precision wouldn't be as optimal since `logmxp1` is optimized for the `+1` addition.

src/univariate/continuous/normal.jl
devmotion: There's no efficiency gain since you still have to define the squared difference of the means by sigma2q separately.

test/kldivergences.jl
```diff
@@ -0,0 +1,71 @@
+@testset "KL divergences" begin
+    function test_kl(p, q)
+        @test kldivergence(p, q) > 0
+        @test kldivergence(p, p) ≈ 0 atol=1e-1
+        @test kldivergence(q, q) ≈ 0 atol=1e-1
+        @test kldivergence(p, q) ≈ invoke(kldivergence, Tuple{MultivariateDistribution,MultivariateDistribution}, p, q; nsamples=10000) atol=1e-1
```
theogf, line 6: Somehow I never got to use invoke,
I am trying
`invoke(kldivergence, Tuple{MultivariateDistribution,MultivariateDistribution}, p, q; nsamples=10000)` but I get an error `invoke: argument type error`

src/functionals.jl
```diff
@@ -22,6 +22,13 @@ function expectation(distr::UnivariateDistribution, g::Function)
     expectation(distr, g, 1e-10)
 end
 
+function expectation(distr::MultivariateDistribution, g::Function; nsamples::Int=100, rng::AbstractRNG=GLOBAL_RNG)
+    nsamples > 0 || throw(ArgumentError("number of samples should be > 0"))
```
devmotion, line 26: I think it's helpful to add a comment here that explains why a separate function is used:
```suggestion
    nsamples > 0 || throw(ArgumentError("number of samples should be > 0"))
    # We use a function barrier to work around type instability of `sampler(dist)`
```

src/functionals.jl
```diff
@@ -30,5 +37,9 @@ end
 # end
 
 function kldivergence(P::UnivariateDistribution, Q::UnivariateDistribution)
```
devmotion, line 39: IMO it is easiest if you just change this line to
```suggestion
function kldivergence(P::Distribution{V}, Q::Distribution{V}; kwargs...) where {V<:VariateForm}
```

src/functionals.jl
```diff
@@ -30,5 +37,9 @@ end
...
-    expectation(P, x -> let p = pdf(P,x); (p > 0)*log(p/pdf(Q,x)) end)
+    expectation(P, x -> let p = pdf(P, x); (p > 0) * log(p / pdf(Q, x)) end)
+end
+
+function kldivergence(P::MultivariateDistribution, Q::MultivariateDistribution; kwargs...)
+    expectation(P, x -> let p = pdf(P, x); (p > 0) * log(p / pdf(Q, x)) end; kwargs...)
```
devmotion, line 44: Can we change both to use `logpdf` instead? The samples have to have non-zero probability but `pdf` could still be zero (e.g. far away from the mean of a normal distribution); I think this is more important in general than potential cancellations due to the subtraction:
```suggestion
    function kldivergence_f(x)
        logpdf_P = logpdf(P, x)
        return (logpdf_P > oftype(logpdf_P, -Inf)) * (logpdf_P - logpdf(Q, x))
    end
    return expectation(P, kldivergence_f; kwargs...)
end
```

test/kldivergences.jl
```diff
@@ -0,0 +1,76 @@
+@testset "KL divergences" begin
+    function logdiff(P, Q, x)
+        return let p = pdf(P, x)
+            (p > 0) * log(p / pdf(Q, x))
+        end
```
devmotion, line 5: The `let` block is not needed/useful here, I think. You could also use the `logpdf` version to handle small probabilities better.
```suggestion
        p = pdf(P, x)
        return (p > 0) * log(p / pdf(Q, x))
```
or
```suggestion
        logp = logpdf(P, x)
        return (logp > -Inf) * (logp - logpdf(Q, x))
```

src/functionals.jl
theogf: Somehow that does not sound very AD-friendly. Why not using `isinf` ?

devmotion: `isinf` would catch `+Inf` as well but we only want to check for `-Inf` here. It's not clear to me though why it would be less AD-friendly than `p > 0`?

theogf: > isinf would catch +Inf 

That's true!

>why it would be less AD-friendly than p > 0?

I am just thinking that `oftype` could mess up some of the AD mechanisms, `Dual`s for example?

devmotion: Type promotion will happen implicitly also for `p > 0`, it's just more explicit. It also seems to work fine for `Dual`s:
```julia
julia> using ForwardDiff

julia> oftype(ForwardDiff.Dual(300.0), -Inf)
Dual{Nothing}(-Inf)

julia> typeof(oftype(ForwardDiff.Dual(300.0), -Inf))
ForwardDiff.Dual{Nothing, Float64, 0}

julia> typeof(oftype(ForwardDiff.Dual(300f0), -Inf))
ForwardDiff.Dual{Nothing, Float32, 0}

julia> oftype(ForwardDiff.Dual(300f0), -Inf)
Dual{Nothing}(-Inf)
```

test/kldivergences.jl
devmotion: You have to use `UnivariateDistribution` for univariate distributions, maybe this is the problem?

theogf: But the whole point is to use the `MultivariateDistribution` one right? Anyway since I externalized the KL computation in an external function, we don't have to use this now

theogf: Oh I see what you mean now

theogf:
So the only thing stopping this PR to pass is #1415

devmotion:
I had a look and I think one should change `getEndPoints` and the `expectation`. They seem to be implemented a bit suboptimally. E.g. `quadgk` accepts infinite interval endpoints so the `+- Inf` workaround should not be needed. Moreover, it would be better to check if the minimum or maximum of the distribution is unbounded instead of computing the quantile.

devmotion:
More concretely, I suggest to use
```julia
function expectation(distr::ContinuousUnivariateDistribution, g::Function; kwargs...)
    return first(quadgk(x -> pdf(distr, x) * g(x), extrema(distr)...; kwargs...))
end
```
and
```julia
function getEndPoints(dist::DiscreteUnivariateDistribution, eps)
    mindist, maxdist = extrema(dist)
    minval = isfinite(mindist) ? mindist : quantile(dist, eps)
    maxval = isfinite(maxdist) ? maxdist : quantile(dist, 1 - eps)
    return minval, maxval
end
```
(maybe we should use `ifelse` for type stability). This is still not correct for discrete distributions with non-integer valued support (we really need a better system there) but at least the error should be fixed.

theogf:
> 
> 
> More concretely, I suggest to use
> 
> ```julia
> function expectation(distr::ContinuousUnivariateDistribution, g::Function; kwargs...)
>     return first(quadgk(x -> pdf(distr, x) * g(x), extrema(distr)...; kwargs...))
> end
> ```
> 
> and
> 
> ```julia
> function getEndPoints(dist::DiscreteUnivariateDistribution, eps)
>     mindist, maxdist = extrema(dist)
>     minval = isfinite(mindist) ? mindist : quantile(dist, eps)
>     maxval = isfinite(maxdist) ? maxdist : quantile(dist, 1 - eps)
>     return minval, maxval
> end
> ```
> 
> (maybe we should use `ifelse` for type stability). This is still not correct for discrete distributions with non-integer valued support (we really need a better system there) but at least the error should be fixed.

I just started to do this and it's definitely nicer but it does not solve the `Poisson` problem, because right now `expectation` for `Discrete` distributions is defined as
```
sum(x -> f(x) * g(x), leftEnd:rightEnd)
```
and `0:Inf` obviously doesn't work

src/functionals.jl
```diff
@@ -22,13 +22,32 @@ function expectation(distr::UnivariateDistribution, g::Function)
...
+        return (logp > oftype(logp, -Inf)) * (logp - logpdf(Q, x))
+    end
+end
+
+function kldivergence(P::Distribution{V}, Q::Distribution{V}; kwargs...) where {V<:VariateForm}
+    expectation(P, safe_logdiff(P, Q))
+end
+
+function kldivergence(P::MultivariateDistribution, Q::MultivariateDistribution; kwargs...)
+    expectation(P, safe_logdiff(P, Q); kwargs...)
```
devmotion, line 52: What I wanted to suggest is to have only a single fallback definition. This should cover all cases:
```suggestion
function kldivergence(P::Distribution{V}, Q::Distribution{V}; kwargs...) where {V<:VariateForm}
    function logdiff(x)
        logp = logpdf(P, x)
        return (logp > oftype(logp, -Inf)) * (logp - logpdf(Q, x))
    end
    expectation(P, logdiff; kwargs..)
end
```

test/kldivergences.jl
```diff
@@ -0,0 +1,75 @@
...
+        @testset "Normal" begin
+            p = Normal(0, 1)
+            q = Normal(0.5, 0.5)
+            test_kl(p, q)
+        end
+        @testset "Poisson" begin
+            p = Poisson(4.0)
+            q = Poisson(3.0)
+            test_kl(p, q)
```
devmotion, line 41: Can we add a test with `lambdap = 0`?

src/functionals.jl
```diff
@@ -1,34 +1,39 @@
...
+    mindist, maxdist = extrema(distr)
+    # We want to avoid taking values up to infinity
+    minval = isfinite(mindist) ? mindist : quantile(distr, epsilon)
+    maxval = isfinite(maxdist) ? maxdist : quantile(distr, 1 - epsilon)
+    return sum(x -> pdf(distr, x) * g(x), minval:maxval)
+end
+
+function expectation(distr::DiscreteUnivariateDistribution, g::Function; epsilon::Real=1e-10, kwargs...)
+    return expectation(distr, g, ϵ)
 end
```
devmotion, line 16: Maybe just change the `expectation` definition above to
```julia
function expectation(distr::DiscreteUnivariateDistribution, g::Function; epsilon::Real=1e-10)
...
```
? In any case, the unicode epsilon is not defined.

src/univariate/continuous/exponential.jl
```diff
@@ -60,6 +60,11 @@ kurtosis(::Exponential{T}) where {T} = T(6)
 
 entropy(d::Exponential{T}) where {T} = one(T) + log(d.θ)
 
+function kldivergence(p::Exponential, q::Exponential)
+    λq_over_λp = scale(q) / scale(p)
+    return -log(λq_over_λp) + λq_over_λp - 1
```
devmotion, line 65: It seems here we can actually use `logmxp1`:
```suggestion
    return -logmxp1(λq_over_λp)
```

src/univariate/discrete/poisson.jl
```diff
@@ -84,6 +84,11 @@ function entropy(d::Poisson{T}) where T<:Real
     end
 end
 
+function kldivergence(p::Poisson, q::Poisson)
+    λp = rate(p)
+    λq = rate(q)
+    return λq - λp + xlogx(λp) - xlogy(λp, λq)
```
devmotion, line 90: What's the reason for not using
```suggestion
    return λq - λp + xlogx(λp, λp / λq)
```
? This would probably be faster.

src/univariate/continuous/normal.jl
```diff
@@ -75,6 +75,14 @@ kurtosis(d::Normal{T}) where {T<:Real} = zero(T)
 
 entropy(d::Normal) = (log2π + 1)/2 + log(d.σ)
 
+function kldivergence(p::Normal, q::Normal)
+    μp = mean(p)
+    σ²p = var(p)
+    μq = mean(q)
+    σ²q = var(q)
+    return ((abs2(μp - μq) + σ²p) / σ²q - 1 + log(σ²q / σ²p)) / 2
```
devmotion, line 83: I just realized we can use `logmxp1` here as well:
```suggestion
    σ²p_over_σ²q = σ²p / σ²q
    return (abs2(μp - μq) / σ²q - logmxp1(σ²p_over_σ²q)) / 2
```

src/functionals.jl
```diff
@@ -1,34 +1,39 @@
...
-    leftEnd = left!=-Inf ? left : quantile(distr, epsilon)
-    rightEnd = right!=-Inf ? right : quantile(distr, 1-epsilon)
-    (leftEnd, rightEnd)
-end
-
-function expectation(distr::ContinuousUnivariateDistribution, g::Function, epsilon::Real)
-    f = x->pdf(distr,x)
-    (leftEnd, rightEnd) = getEndpoints(distr, epsilon)
-    quadgk(x -> f(x)*g(x), leftEnd, rightEnd)[1]
+function expectation(distr::ContinuousUnivariateDistribution, g::Function; kwargs...)
```
devmotion, line 1: Can you deprecate `expectation(dist::ContinuosUnivariateDistribution, g::Function, epsilon::Real)` to ensure that the PR is not breaking?

src/functionals.jl
```diff
@@ -1,34 +1,39 @@
...
+    return first(quadgk(x -> pdf(distr, x) * g(x), extrema(distr)...; kwargs...))
 end
 
 ## Assuming that discrete distributions only take integer values.
 function expectation(distr::DiscreteUnivariateDistribution, g::Function, epsilon::Real)
```
devmotion, line 6: Can you (for consistency) also deprecate `expectation(distr::DiscreteUnivariateDistribution, g::Function, epsilon::Real)` and use
```suggestion
function expectation(distr::DiscreteUnivariateDistribution, g::Function; epsilon::Real=1e-10)
```
instead?

src/functionals.jl
devmotion: Just add it to src/deprecates.jl :slightly_smiling_face:

devmotion (commented):
Sorry that it takes a while, here are some additional suggestions :slightly_smiling_face: Overall it looks good, I think we can merge it once the remaining comments are addressed (or we figured out that the suggestions are not helpful :wink:)

Can you also update the version number?

src/functionals.jl
```diff
@@ -1,34 +1,39 @@
...
+    mindist, maxdist = extrema(distr)
+    # We want to avoid taking values up to infinity
+    minval = isfinite(mindist) ? mindist : quantile(distr, epsilon)
+    maxval = isfinite(maxdist) ? maxdist : quantile(distr, 1 - epsilon)
+    return sum(x -> pdf(distr, x) * g(x), minval:maxval)
+end
+
+function expectation(distr::DiscreteUnivariateDistribution, g::Function; epsilon::Real=1e-10)
+    return expectation(distr, g, epsilon)
 end
```
devmotion, line 16: If you apply the suggestions above, these lines can be removed:
```suggestion
```

test/kldivergences.jl
```diff
@@ -0,0 +1,78 @@
+@testset "KL divergences" begin
```
devmotion, line 1: I just noticed that there are already some tests for `kldivergence` in `test/functionals.jl`. Can you move the tests here to this file?

src/univariate/discrete/poisson.jl
theogf: That handle the case where lambdaq is 0

devmotion: :+1: Ah yes I missed that `xlogy(0, 0 / 0) === NaN`. The following would handle the case `λp = λq = 0` as well with only one `log` (`false` is a strong zero, i.e., `false * NaN = 0.0`):
```suggestion
    # `false` is a strong zero and ensures that `λp = 0` is handled correctly
    # we don't use `xlogy` since it returns `NaN` for `λp = λq = 0`
    return λq - λp + (λp > 0) * (λp * log(λp / λq))
```
In any case, we should add tests for these special cases to avoid that regressions are introduced accidentally.

src/deprecates.jl
```diff
@@ -46,3 +46,7 @@ end
 @deprecate Wishart(df::Real, S::AbstractPDMat, warn::Bool) Wishart(df, S)
 @deprecate Wishart(df::Real, S::Matrix, warn::Bool) Wishart(df, S)
 @deprecate Wishart(df::Real, S::Cholesky, warn::Bool) Wishart(df, S)
+
+# Deprecate 3 arguments expectation
+@deprecate expectation(distr::DiscreteUnivariateDistribution, g, epsilon) expectation(distr, g; epsilon=epsilon)
+@deprecate expectation(distr::ContinuousUnivariateDistribution, g, epsilon) expectation(distr, g)
```
devmotion, line 52: Maybe good to limit it to the types that it is defined for currently:
```suggestion
@deprecate expectation(distr::DiscreteUnivariateDistribution, g::Function, epsilon::Real) expectation(distr, g; epsilon=epsilon)
@deprecate expectation(distr::ContinuousUnivariateDistribution, g::Function, epsilon::Real) expectation(distr, g)
```

test/kldivergences.jl
```diff
@@ -0,0 +1,86 @@
...
+        elseif p isa MultivariateDistribution
+            @test kldivergence(p, q) ≈ invoke(kldivergence, Tuple{MultivariateDistribution,MultivariateDistribution}, p, q; nsamples=10000) atol=1e-1
+        end
+    end
+    
+    @testset "Categorical" begin
+        @test kldivergence(Categorical([0.0, 0.1, 0.9]), Categorical([0.1, 0.1, 0.8])) ≥ 0
+        @test kldivergence(Categorical([0.0, 0.1, 0.9]), Categorical([0.1, 0.1, 0.8])) ≈
+            kldivergence([0.0, 0.1, 0.9], [0.1, 0.1, 0.8])
+        @test expectation(Normal(0.0, 1.0), identity, 1e-10) ≤ 1e-9
```
devmotion, line 17: Doesn't really fit in this test set :smile: It's also one of the deprecated calls (which might be good to test as well).

What about:
- moving the `kldivergence` of ´`Categorical` to the univariate section below?
- have a separate section before the KL divergence test sets for expectations?

I think we should compute expectations of e.g. a normal distribution, a Poisson distribution (unbounded discrete) and a Binomial distribution (bounded discrete), with the new and the deprecated calls (the deprecation could also be checked with `@test_deprecated`).

The file name is not completely correct then anymore, so maybe we could just rename it to `functionals.jl` again (which is also consistent with the source file src/functionals.jl).

test/kldivergences.jl
```diff
@@ -0,0 +1,86 @@
...
+            p_0 = Poisson(0.0)
+            test_kl(p_0, p)
+        end
+    end
+    @testset "multivariate" begin
+        # We try on a newly created AbstractMvNormal
+        struct CholeskyMvNormal{TL,Tm} <: Distributions.AbstractMvNormal
+            m::Tm
+            L::TL
+        end
```
devmotion, line 60: Julia 1.0 doesn't allow type definitions in a local scope. Can you move this outside of the testsets?

devmotion (approved):
LGTM, thanks a lot @theogf! I just fixed the test errors.

PR #1416, opened 2021-11-05
Support DensityInterface API

oschulz:
Implements `DensityInterface.hasdensity(::Distribution)` and `DensityInterface.logdensityof(d::Distribution, x)`.

This will form a basis for using distributions in generic code that needs to handle distributions as well as (non-normalized) densities.

CC @cscherrer, @devmotion, @mschauer, @phipsgabler, @tpapp,


oschulz:
Failures on nightly seem unrelated.

mschauer:
Can you write more about what this would be good for, maybe a specific use case, also compared to the alternative of just creating an anonymous function `x -> logpdf(MyObject, x)` and passing that around?

oschulz:
> Can you write more about what this would be good for, maybe a specific use case

Sure. I think @phipsgabler put it very nicely in TuringLang/AbstractPPL.jl#38: "

> * to avoid the IMHO artificial distiction between "PDFs" and "PMFs",
> * to use an appropriate term for densities which are not probability densities, like (unnormalized) posteriors and likelihoods, and
> * to have a generic hypernym covering loglikelihood, logprior, etc., which are all expressible in terms of logdensity of differently conditioned models.

All of this works much better, of course, if `Distributions` takes part in it. `DensityInterface` is about things that *are* a density or can be said to *have* a density. Distributions definitely do have a (probability) density of course. Even discrete distributions can be said to have a density in respect to a natural base measure (the counting measure) - which corresponds exactly to what `logpdf(d, x)` calculates at the moment.


> compared to the alternative of just creating an anonymous function x -> logpdf(MyObject, x) and passing that around

Using plain log-density functions can be very awkward at higher levels of code. To illustrate, in Julia examples for Bayesian analysis I used to write `log_likelihood = some_func; prior = SomeDistribution()` - only to be asked by people "why one with log_ and not the other?". It just doesn't read well.

On the other hand, `likelihood = some_func; prior = SomeDistribution()` is misleading, dangerous and wrong, if `some_func` returns the *log* of the likelihood. And using `log_likelihood = some_func; log_prior = x -> logpdf(SomeDistribution(), x)` also won't work in many cases, since important information is lost: An algorithm can't sample log-prior function to get starting values, for example.

But `likelihood = logfuncdensity(some_func); prior = SomeDistribution()` is clear to the reader and safe: the code using likelihood and prior (and the posterior, and so on) can use logdensityof and be sure to get the log. And it more complex use cases it will often be `likelihood = SomeLikelihood(data)`, an object that can be passed around without loss of information (can still acces the data at any time as well as calculate it's (log)-density value.

DensityInterface is not limited to Bayesian applications of course - in fact, it's intended to be very general. The current API originated through (I think very productive) discussions (oschulz/DensityInterface.jl#1, oschulz/DensityInterface.jl#3, oschulz/DensityInterface.jl#4) among several people who have a an interest in (log-)densities in the Julia community and are involved in packages that would profit from this. Maintainers of Distributions were also involved. Having Distributions support DensityInterface was an important aspect of the design.

DensityInterface is an extremely lightweight package with no dependencies beyond `Test` and `InverseFunctions` (which is extremely lightweight itself and introduces no further dependencies), so there should be no impact on the load time of `Distributions` (a `@time using Distributions` test resulted in aload time of 1.05s before and 0.99s after this PR, so no significant effect).

oschulz:
Added `DensityInterface.densityof(d::Distribution, x) = pdf(d, x)`.

I think this is complete now.

src/density_interface.jl
```diff
@@ -0,0 +1,5 @@
+@inline DensityInterface.hasdensity(::Distribution) = true
+
+DensityInterface.logdensityof(d::Distribution, x) = logpdf(d, x)
```
devmotion, line 3: The log likelihood example makes me wonder if this should be

```suggestion
DensityInterface.logdensityof(d::Distribution, x) = loglikelihood(d, x)
```

The difference would be:
- `loglikelihood` is guaranteed to return a `Real` whereas `logpdf` can return an array of `Real`s if `x` consists of multiple samples (in the univariate case though it has to be broadcasted)
- `loglikelihood` evaluates the log likelihood of multiple samples without constructing intermediate arrays
- they coincide for single samples

Due to this more consistent interface and since it does not allocate eg Turing mainly uses `loglikelihood` but not `logpdf` internally.

oschulz:
@devmotion, now that we have `DensityInterface.logdensityof(d::Distribution, x) = loglikelihood(d, x)` should `DensityInterface.densityof(d::Distribution, x)` call `exp(loglikelihood(d, x))` or `pdf(d, x)`?

phipsgabler:
I'd say specialize to `Distributions.pdf`. Many distributions probably will then do the `exp` thing internally anyway, but maybe some can have simpler implementations. It also gives Distributions.jl the full control and gets rid of possible numerical differences between `pdf` and `densityof`. (There's no extra `pmf`, right?)

Something related: I remember one bug originating from a distribution which did it the other way round and implemented `logpdf` in terms of `pdf`, which caused a stack overflow at some place. Could that be a problem?

oschulz:
> I'd say specialize to Distributions.pdf

Seems more natural to me as well (it's what this PR currently does).

devmotion:
> should DensityInterface.densityof(d::Distribution, x) call exp(loglikelihood(d, x)) or pdf(d, x)

It should call `likelihood(d, x)` ideally which has the same properties as `loglikelihood` but returns `exp(loglikelihood(d, x))`. Unfortunately, such a method does not exist yet :smile: So I think it is better to use the default fallback for now (i.e., `exp(loglikelihood(d, x))` but not defining it) since otherwise `densityof` and `logdensityof` would be very inconsistent. `pdf` and `loglikelihood` return different types for some inputs (`pdf` may also return arrays) and support different inputs.

> Something related: I remember one bug originating from a distribution which did it the other way round and implemented logpdf in terms of pdf, which caused a stack overflow at some place. Could that be a problem?

I think I fixed all of these potential StackOverflowErrors, the dispatching should be much more sane nowadays.

oschulz:
> So I think it is better to use the default fallback for now

Ok, I took the specialization of `DensityInterface.densityof` out and left a comment about this.

oschulz:
@mschauer ,@devmotion, is this fine like this from your side?

test/density_interface.jl
```diff
@@ -0,0 +1,16 @@
+@testset "DensityInterface" begin
+    using DensityInterface
+
+    d_uv_continous = Normal(-1.5, 2.3)
+    d_uv_discrete = Poisson(4.7)
+    d_mv = MvNormal([2.3 0.4; 0.4 1.2])
+
+    for d in (d_uv_continous, d_uv_discrete, d_mv)
+        x = rand(d)
```
devmotion, line 9: Maybe add a test with multiple samples as well?

oschulz: You mean multiple samples in one call? I don't think DensityInterface should support this, really. It predates Julia having broadcasting in the design of Distributions - but nowadays, I don't think it's a good idea. And it's not cleanly done in Distributions currently anyway: Multiple samples get passed as a matrix to a multivariate dist, but as a vector of matrices to a matrix-variate dist ...

oschulz: Ok, I've blocked this now, and added some tests for it. Ok with you?

oschulz: Note: I did it this way (blocking instead of restricting to specific array dimensionality in the primary definition of `logdensityof`) to support Distributions that have non-array-like variates (e.g. `ValueShapes.NamedTupleDist`).

devmotion: > You mean multiple samples in one call? I don't think DensityInterface should support this, really. It predates Julia having broadcasting in the design of Distributions - but nowadays, I don't think it's a good idea.

I think we talk about different things here. I don't want to support broadcasting, ie. `logdensityof.((dist,), samples)`. Multiple samples here means `sum(logpdf.((dist,), samples))`, i.e., it just returns the log-likelihood of a data set of `samples`. I think it would be useful to support this with `logdensityof` as it would simplify working with log-likelihoods. It would also be much more difficult to use `logdensityof` in Turing if it is not supported as it is based on calls such as `loglikelihood(Normal(), data)` etc.

> And it's not cleanly done in Distributions currently anyway: Multiple samples get passed as a matrix to a multivariate dist, but as a vector of matrices to a matrix-variate dist ...

There is some inconsistency but only because multivariate data/samples as matrices is so popular. `loglikelihood` also supports arrays (not only vectors) of vectors, so one does not have to use matrices in the evaluation. Additionally, all these implementation details and optimizations of `loglikelihood` are handled by Distributions and hence nothing that DensityInterface has to be concerned about.

Also due to these inconsistencies, it would be very annoying to replace calls of `loglikelihood(dist, samples)` in Turing with something like `sum(logdensityof.((dist,), samples))` - it would just not work for all combinations of distributions and samples, and hence we would have to write our own custom function that dispatches on the types of distributions. And we would have to implement optimizations for specific distributions on our own. It seems much more sane to just rely on Distributions implementing `loglikelihood` in a reasonable and optimized way.

src/density_interface.jl
```diff
@@ -0,0 +1,19 @@
...
+
+function DensityInterface.logdensityof(d::Distribution{ArrayLikeVariate{N}}, x::AbstractArray{<:Real}) where N
+    throw(ArgumentError("logdensityof doesn't support multiple samples in a single function call"))
+end
+
+function DensityInterface.logdensityof(d::Distribution{ArrayLikeVariate{N}}, x::AbstractVector{<:AbstractArray{<:Real}}) where N
+    throw(ArgumentError("logdensityof doesn't support multiple samples in a single function call"))
+end
+
+
```
devmotion, line 16: As mentioned in my comment (https://github.com/JuliaStats/Distributions.jl/pull/1416/files#r744310342), I think these errors and definitions should be removed. Instead we should just rely on the implementation of `loglikelihood` in Distributions - it will throw an error if there is no implementation for `dist` and `x`. An important point of `loglikelihood` exploited in Turing is that it supports the evaluation for a set of samples (here for us it also does not matter how they look). Since DensityInterface wants to support working with (log-)likelihoods, I think we should only forward `logdensityof` to `loglikelihood` (it is no broadcasting, it is the sum of the individual log probabilities):
```suggestion
```

test/density_interface.jl
cscherrer: > I think we talk about different things here. I don't want to support broadcasting, ie. `logdensityof.((dist,), samples)`. Multiple samples here means `sum(logpdf.((dist,), samples))`, i.e., it just returns the log-likelihood of a data set of `samples`.

Say `dist` is a distribution over some complex type, so "is it an Array or not" isn't enough. This could make it very hard to tell whether what's being passed is intended as a set of samples, vs one sample from a distribution over arrays.

devmotion: I'm not sure what you mean and it's not clear to me how it is related to the discussion. `sum(logpdf.((dist,), samples))` was just a simple illustration for univariate samples to make it clear what I meant with "multiple samples" and that it is not just broadcasting. This is not how `loglikelihood` is implemented for general distributions. And in the implementation of `loglikelihood` for a specific distribution there is no confusion about whether a single sample or a set of samples is passed. However, all these discussions about the implementation of `loglikelihood` are completely irrelevant for the definition of `logdensityof`.

oschulz: Yes, from a dispatch viewpoint this can be awkward, supporting both a single and multiple samples in the same function call. I'm also not sure that it matches the semantics we defined for `logdensityof` well: We're just using `loglikelihood` instead of `logpdf` here because of implementation reasons. But the density we speak of in regard to `logdensityof` is the PDF, and `logdensityof` is supposed to return that density at a single point.

It's of course debatable what "a single point" is: it can of course be a concrete set of samples drawn from the set of all possible sets of samples. Still, I don't think we should say that the likelihood is the natural density of a distribution.

I do however see you points regarding the practical aspects, @devmotion. How about we do this, in addition to defining `logdensityof` for distributions:

```julia
struct Likelihood{D<:Distribution}
    d::D
end

@inline DensityInterface.hasdensity(l::Likelihood) = true

@inline DensityInterface.logdensityof(l::Likelihood, x) = likelihood(l.d, x)
```

I would argue that the PDF and the likelihood are very different things, semantically and mathematically, so we should not use `logdensityof(d, x)` to get the one or the other depending on the type of `x`.

oschulz: Ah, no, that's also not so nice, since the argument of a likelihood density are the parameters of the distribution, not the data.

What does Turing need here, exactly?

devmotion: We need and use `loglikelihood`. There's not much use for a `logdensityof` that only gives us the `logpdf`, due to all the inconsistencies and dispatches that we would have to solve on our own - even though `loglikelihood` already solves them.

cscherrer: From
> I think it would be useful to support this with `logdensityof` as it would simplify working with log-likelihoods.

it seemed to me you were suggesting `logdensity` adopt this behavior.

> However, all these discussions about the implementation of `loglikelihood` are completely irrelevant for the definition of `logdensityof`.

I was commenting on something you brought up. Maybe I misunderstood something. I don't assume you have deep knowledge of how MeasureTheory works, and I'm happy to help with sincere misunderstandings. I'd appreciate the same in return.

devmotion: If the intention of DensityInterface is to not support this use case (which is OK but should be stated and communicated more clearly probably, in particular since discussions involved likelihoods), I think it will be unlikely that `logdensityof` is adopted in the near future in Turing.

devmotion: > I would argue that the PDF and the likelihood are very different things, semantically and mathematically, so we should not use logdensityof(d, x) to get the one or the other depending on the type of x.
> Ah, no, that's also not so nice, since the argument of a likelihood density are the parameters of the distribution, not the data.

I guess there might be another difference in our interpretation of `loglikelihood` here: in Turing and in my view, `loglikelihood(dist, x)` gives us just the log density of the conditional distribution `law(x | theta)` where `theta` are the parameters of the distribution `x`. Whereas it seems you think about the log-likelihood function `L(theta) = p(x | theta)` of the parameters `theta` for a fixed set of observations `x`. I agree that in the latter case the use of `logdensityof(dist, x)` seems weird but I would argue if it is viewed as a distribution of `x` it seems quite natural.

oschulz: I do think we should use a wrapper, but `Likelihood` is the wrong name: What we mean here (for Turing, etc.) is the Distribution/PDF over a set of possible outcomes. So it's like the PDF of an implicit product distribution, actually. So we could write

```julia
struct DatasetDensity{D<:Distribution}
    distribution::D
end

@inline DensityInterface.hasdensity(d::DatasetDensity) = true

@inline DensityInterface.logdensityof(d::DatasetDensity, X) = likelihood(d.distribution, X)
```

And Turing could just use `logdensity(DatasetDensity(d), X)`. This would be clean.

And `logdensityof(d::Distribution, x)` should actually call `logpdf(d, x)`, *not* `loglikelihood`.

I'm sorry, I think I contributed to the confusion: When @devmotion proposed changing to `loglikelihood` I misunderstood/misread things a bit, I think.

oschulz: > another difference in our interpretation of loglikelihood

Well, in the Bayesian inference world we typically think of the data as fixed, and when people say log-likelihood they do indeed mean `L(theta) = p(x | theta)`. My main use case for supporting `logdensityof` for Distributions is to get the log-density of a prior distribution, and the log-density of a likelihood, both as a function of theta.

But I get that there are other use cases, and of course we should strive to serve those as well.

However, I think we shouldn't muddle the semantics of `logdensityof(d, x)` too much. The definition is that it gives us the log-value of the density of `d` (if `d` is not a density itself) in respect to it's base measure at the point x. So for a distribution, that would be it's log-PDF at x.

> log density of the conditional distribution law(x | theta)

But this distribution is not the original distribution `d(theta)` - instead, it's a product distribution over `d` (`d` repeated n-times for n elements in `x`), correct?

For a single point I think it's clear that we want `logdensityof(d, x) == logpdf(d, x) == loglikelihood(d, x)` (in the sense of `Distributions.loglikelihood`), correct?

The question is, given multiple points, should we define the "natural" density of `d` to be the density of it's product distribution? It seems a bit too magical too me. `logdensityof(ProductDensity(d), X)` would be a way to state that we want that explicitly. Would there be any problems for Turing with this approach?

devmotion: Thanks @oschulz, I think this would be a clean solution! I was debating with myself back and forth whether one should use `loglikelihood` in the definition of `logdensityof`. I was not completey happy about it but a) it seemed a more general and, e.g. in Turing, more useful definition than `logpdf` and b) its implementation in Distributions satisfies properties that we expect from `logdensityof` but do not hold for `logpdf` for all inputs (e.g., always returns a real number). However, I agree that probably it is better to separate the "base case" and the implicit product distribution of iid samples more clearly.

I'm not completely sure if `DatasetDensity` is the best name. The name `DatsetDensity` seems to suggest that `X` is a set of multiple samples (or at least a set of samples) but it is fine to call `loglikelihood` with a single sample (it just falls back to `logpdf`), and this property is used in Turing as well. Maybe something like `IIDDensity` could be an alternative that does not imply necessarily that `X` are multiple samples?

Overall it seems we agree that we should implement the base case as
```julia
hasdensity(::Distribution) = true
logdensityof(d::Distribution, x) = logpdf(d, x)
densityof(d::Distribution, x) = pdf(d, x)
```
and the iid case with a dedicated density such as suggested above (preferably without the `@inline`, at least in `logdensityof` :stuck_out_tongue:). I think we should not be worried that we can't determine a priori which types of `x` (or `X`) are valid for the distribution at hand - the implementations (or rather non-existing implementations) of `logpdf` and `loglikelihood` for invalid sample(s) will take care of it.

oschulz: > I'm not completely sure if DatasetDensity is the best name

Me neither, was just the best I could come up with quickly for a prototype. :-) I like `IIDDensity`.

I'll implement it, can you just review JuliaMath/DensityInterface.jl#6 quickly? We need approximate comparison in `test_density_interface` for this, since `exp(logpdf(d, x))` is not always exactly `pdf(d, x)`.

oschulz: One problem with the name `IIDDensity`: We don't really require the sample to be independent, just to be identically distributed, right? But on the other hand, `IDDensity` sound like a density of IDs ...

devmotion: >  We don't really require the sample to be independent, just to be identically distributed, right?

We do require independence, don't we? Otherwise it is possible that the density of the samples can't be factorized and we don't end up with `loglikelihood(d, samples) = sum(logpdf(d, x) for x in samples)`.

oschulz: Ah - no, I mean the sample could, for example, result from an MCMC sampling. Even it it was a faithful sampling of `d`, the samples wouldn't be independent, since they would stem from a Markov chain.

oschulz:
CI will fail, requires JuliaMath/DensityInterface.jl#6

test/density_interface.jl
devmotion: Ah OK, you're concerned with the fact that `X` for which we perform the evaluation is not actually a set of iid samples but of correlated samples? I would say this is irrelevant here, in the same way as it is e.g. irrelevant if a single sample `x` is actually a sample from the distribution for which we evaluate the density. IMO the name `IIDDensity` would still be correct since we perform the evaluation under the assumption that samples are iid.

oschulz: Ok, let's go with `IIDDensity` then, I don't have a better proposal anyway. :-)

oschulz:
Have to wait for JuliaRegistries/General#48369

oschulz:
Oh, it did pass on Julia v1.0 and v1.6, actually. It failed for me locally on v1.7 without JuliaMath/DensityInterface.jl#6. But you can review then changes already. :-)

docs/src/density_interface.md
```diff
@@ -0,0 +1,9 @@
+# Support for DensityInterface
+
+`Distributions` supports [`DensityInterface`](https://github.com/JuliaMath/DensityInterface.jl) for distributions: `DensityInterface.logdensityof(d, x)` is equivalent to `logpdf(d, x)` and `DensityInterface.densityof(d::Distribution, x)` is equivalent to `pdf(d, x)` for *single* variate values `x`.
```
devmotion, line 3:
```suggestion
`Distributions` supports [`DensityInterface`](https://github.com/JuliaMath/DensityInterface.jl) for distributions.

For *single* variate values `x`, `DensityInterface.logdensityof(d::Distribution, x)` is equivalent to `logpdf(d, x)` and `DensityInterface.densityof(d::Distribution, x)` is equivalent to `pdf(d, x)`.
```

oschulz:
Test failures on nightly seem unrelated. Ready for a fresh review round. :-)

docs/src/density_interface.md
```diff
@@ -0,0 +1,9 @@
...
+
+To get the log-density value at either a single variate value or at a set of variate values (implying a product distribution over `d`), use `logdensityof(IIDDensity(d), x)`, this is equivalent to `loglikelihood(d, x)`.
+
+```@docs
+IIDDensity
+```
```
devmotion, line 9: I don't think there's a real use case for `IIDDensity` yet apart from the potential use in Turing, so I think we should maybe not export it currently.

````suggestion
You can use
```julia
DensityInterface.logdensityof(Distributions.IIDDensity(d), x)
```
and
```julia
DensityInterface.densityof(Distributions.IIDDensity(d), x)
```
to evaluate the log-density and density of variate value(s) `x` that are assumed to be identically and independently
distributed according to distribution `d`. Here `x` can be either a single variate value or at a set of variate values.
`DensityInferface.logdensityof(Distributions.IIDDensity(d), x)` falls back to `loglikelihood(d, x)`.

```@docs
Distributions.IIDDensity
```
````

File: src/density_interface.jl [hunk 6]
```diff
@@ -0,0 +1,47 @@
...
+        end
+    end
+end
+
+
+"""
+    IIDDensity(distribution::Distribution)
+
+Represents the probability density of `d` and implicit product distributions
+over d.
```
devmotion, line 24:
```suggestion
Represents the probability density of an implicit product distribution of variates that are identically and
independently distributed according to `distribution`.
```

src/density_interface.jl
```diff
@@ -0,0 +1,47 @@
...
+of `d`.
+
+If `x` is a single variate of `d`, the density is the PDF of `d`.
+
+If `x` is a set of variates (given as a higher-dimensional array or
+and array of arrays), the density is the PDF of an implicit product
+distribution over `d`, the size of the product is implied by the size of
+the set.
+
+`DensityInterface.logdensityof(d, x)` is equivalent to `loglikelihood(d, x)`.
```
devmotion, line 37: Can you move this to the definition of `logdensityof(::IIDDensity, x)`? This would it also make clearer what `d` is, I assume.

src/density_interface.jl
```diff
@@ -0,0 +1,47 @@
...
+"""
+struct IIDDensity{D<:Distribution}
+    distribution::D
+end
+
+export IIDDensity
```
devmotion, line 43: As mentioned above, I think we should not export `IIDDensity` currently:
```suggestion
```

src/density_interface.jl
oschulz: Do we really want to append all of that to the function documentation of `logdensityof`?

oschulz: I have improved the docstring, to make it clearer what `d` is.

I don't think every type that implements DensityInterface should add to the docstring of `logdensityof` though, otherwise it could be come excessively long.

devmotion: Yes, one should only add a docstring for `logdensityof` if it differs from the default behaviour, i.e., is special in some way. I think this is the case for `IIDDensity`, in contrast to a standard `Distribution`, but it's a matter of taste.

One can filter docstrings in the REPL if there are multiple ones for e.g. a function such as `any` by the call syntax and the types of the arguments:
```julia
julia> d = Uniform(); x = 0.3;

julia>? logdensityof(d, x) # should only display docstring(s) that match `logdensityof(::Uniform, ::Float64)`

julia>? logdensityof(d) # should only display docstring(s) that match `logdensityof(::Uniform)`
```

oschulz: Sure, but I usually type `?somefunction` first. :-)

As you noted, use of `IIDDensity` may be limited to a few applications for now, which is why we're not exporting it. So I would prefer not to add IIDDensity to the docstring of `logdensityof`. Ok with you?

oschulz: The behavior of `IIDDensity` is indeed a bit "special", and I don't want to "advertise" it too much as something people should imitate when implementing `logdensityof`. ;-)

devmotion: I assume other packages will also want to add docstrings to `logdensityof` in some cases, to explain specific assumptions or behaviour (e.g., AbstractPPL will also add a docstring). IMO this is reasonable and fine if it additional explanations are required, e.g., to clarify which density is evaluated.

I agree though that it is a bit weird to extend the documentation of `logdensityof` with the `IIDDensity` case as long as `IIDDensity` is not exported. I suggest keeping it in the documentation of the type for now. However, I think this should be changed when/if `IIDDensity` is exported, so maybe it would be good to add a `TODO:` note as a comment.

mschauer:
I find `Distribution`'s `loglikelihood` accepting both a single sample and multiple samples a code smell. Is the single-sample version used in Turing? We might not want to support the single sample form for "IIDDensity" it if not.

oschulz:
> Is the single-sample version used in Turing

I has similar concerns, but from what I understood Turing actually depends on having densities that will work for both single and multiple samples - correct, @devmotion ?

docs/src/density_interface.md
```diff
@@ -0,0 +1,21 @@
+# Support for DensityInterface
+
+`Distributions` supports [`DensityInterface`](https://github.com/JuliaMath/DensityInterface.jl) for distributions.
+
+For *single* variate values `x`, `DensityInterface.logdensityof(d::Distribution, x)` is equivalent to `logpdf(d, x)` and `DensityInterface.densityof(d::Distribution, x)` is equivalent to `pdf(d, x)`.
+
+You can use
+```julia
+DensityInterface.logdensityof(Distributions.IIDDensity(d), x)
```
mschauer, line 9: `Distributions`'s `Product` can only represent products of univariate distributions, that is annoying here.  So a `IIDDensity` seems to be necessary.

src/density_interface.jl
```diff
@@ -0,0 +1,48 @@
...
+
+for (di_func, d_func) in ((:logdensityof, :logpdf), (:densityof, :pdf))
+    @eval begin
+        DensityInterface.$di_func(d::Distribution, x) = $d_func(d, x)
+
+        function DensityInterface.$di_func(d::UnivariateDistribution, x::AbstractArray)
+            throw(ArgumentError("$(DensityInterface.$di_func) doesn't support multiple samples as an argument"))
+        end
+
+        function DensityInterface.$di_func(d::MultivariateDistribution, x::AbstractMatrix)
```
mschauer, line 11: What would and what should happen for `AbstractVector{<:AbstractVector}` arguments?

src/density_interface.jl
oschulz: I agree.

> maybe it would be good to add a TODO: note as a comment.

Done.

src/density_interface.jl
```diff
@@ -0,0 +1,50 @@
...
+of `d`.
+
+If `x` is a single variate of `d`, the density is the PDF of `d`.
+
+If `x` is a set of variates (given as a higher-dimensional array or
+and array of arrays), the density is the PDF of an implicit product
+distribution over `d`, the size of the product is implied by the size of
+the set.
+
+`DensityInterface.logdensityof(d, x)` is equivalent to `loglikelihood(d, x)`.
```
mschauer, line 40: Is `d::IIDDensity` in this line?

src/density_interface.jl
oschulz: > What would and what should happen for AbstractVector{<:AbstractVector} arguments?

`logpdf` (called by `logdensityof`) fails with `DimensionMismatch("Inconsistent array dimensions.")`. Which is what we want to happen, since `logdensityof` should only support evaluation at single points.

For the parameter type combinations where Distributions provides legacy `[log]pdf` multi-point evaluation (implicit broadcasting-equivalent of `[log]pdf`) we've blocked this for `logdensityof` explicitly. Otherwise we'll just let `[log]pdf` itself fail.

test/density_interface.jl
```diff
@@ -0,0 +1,37 @@
...
+    d_av = Distributions.MatrixReshaped(MvNormal(rand(10)), 2, 5)
+
+    @testset "Distribution" begin
+        for d in (d_uv_continous, d_uv_discrete, d_mv, d_av)
+            x = rand(d)
+            ref_logd_at_x = logpdf(d, x)
+            DensityInterface.test_density_interface(d, x, ref_logd_at_x)
+
+            # Stricter than required by test_density_interface:
+            @test logfuncdensity(logdensityof(d)) === d
```
mschauer, line 16: How does this work?

src/density_interface.jl
oschulz: Whoops, yes, that was wrong. I just changed it to `DensityInterface.logdensityof(IIDDensity(d), x)`.

I'll close this to make it easier to keep track of things, but please feel free to reopen if it's still not Ok.

test/density_interface.jl
oschulz: It's the default way `DensityInterface.logfuncdensity` behaves, but it's not required for specializations of it. So strictly speaking, we don't need to test it here - I just thought it would be nice to formally require it for distributions.

docs/src/density_interface.md
oschulz: Yes - also, for a product we'd need to specify the size of it. From what I understand, Turing needs the "auto-scaling" of the product size (down to one), bases on the shape of `x`. Correct, @devmotion ?

test/density_interface.jl
mschauer: Yes, but isn't `logdensityof(d)` a function? How do you get the distribution back?

devmotion:
> I has similar concerns, but from what I understood Turing actually depends on having densities that will work for both single and multiple samples - correct, @devmotion ?

Yes, in expressions of the form `x ~ dist` where `x` is some data/observation(s)/... that is assumed to be i.i.d. according to distribution `dist` we evaluate just `loglikelihood(dist, x)` (of course, the computation is skipped if we only compute the prior probabilities). IMO it is a feature that `loglikelihood` accepts both a single and multiple samples - the only assumption of `loglikelihood` is that the data is iid distributed but the number of samples in the data set is arbitrary. And why should one have to artificially wrap a single sample in a container with one element or have to define custom dispatches to `logpdf`?

Also implementation-wise there are no ambiguities as it is clear from the type of the distribution if `x` is a single sample (then one can just fall back to `logpdf`), an array of samples, or another supported collection of samples. It only becomes confusing and difficult to implement if it is left for other packages to determine if `x` is suitable for `dist`, if it is a single sample or if it is a suitable collection of samples: then they have to start dispatching on the distribution type which requires knowledge about and synchronization with implementation details in Distributions and other packages that define distributions. Moreover, it means that only distributions and collections that are known when the dispatches are defined are supported. Therefore in my opinion the current `loglikelihood` interface is very valuable and removing support for single samples would be a major regression.

test/density_interface.jl
oschulz: That's because we have (the equivalent of)

```julia
logdensityof(d) = Base.Fix1(logdensityof, d)

logfuncdensity(log_f::Base.Fix1{typeof(logdensityof)}) = log_f.x
```

in DensityInterface. Makes for a nice symmetry.

oschulz: We can take the `logfuncdensity(logdensityof(d)) === d` for distributions out though, if you prefer. I don't think it's really important.

mschauer: No, it's fine, just wanted to check how it works.

docs/src/density_interface.md
devmotion: Yes, we need the iid assumption of `loglikelihood` and do not fix a size of the implicit iid product distribution a priori. `Product` or a generalization of it are not sufficient here. I tried to explain it in more detail in this comment: https://github.com/JuliaStats/Distributions.jl/pull/1416#issuecomment-963707803

test/density_interface.jl
oschulz: Ok, thanks!

docs/src/density_interface.md
oschulz: I guess for clean semantics, one could just think of `IIDDensity` as a non-normalized probability density over the set of all possible sets of samples, right?

docs/src/density_interface.md
```diff
@@ -0,0 +1,21 @@
+# Support for DensityInterface
+
+`Distributions` supports [`DensityInterface`](https://github.com/JuliaMath/DensityInterface.jl) for distributions.
+
+For *single* variate values `x`, `DensityInterface.logdensityof(d::Distribution, x)` is equivalent to `logpdf(d, x)` and `DensityInterface.densityof(d::Distribution, x)` is equivalent to `pdf(d, x)`.
+
+You can use
```
mschauer, line 7:
```suggestion
Additionally—though not supported by other packages using this API—you can use
```

devmotion: I don't think this should be added. Or at least I think it should be phrased differently. Currently it seems to suggest no other package supports and makes use of `IIDDensity` even though it is supposed to be used by DynamicPPL/Turing. It also only accepts a `Distribution`, so it doesn't seem to suggest that it can be used with arbitrary measures or densities. In the same way as eg a `RadonNikodymDerivative` probably would only support specific measures.

docs/src/density_interface.md
mschauer: Yes

docs/src/density_interface.md
mschauer: You can export an `IIDDensity` from your package which uses Distributions. But you cannot call `IIDDensity` on an object outside the Distributions context. So if you are using `IIDDensity` on the importing side, you do not participate in `DensityInterface`, because you will restrict yourself to `Distribution`s and then you can just call `logpdf` instead and forget about `logdensityof`.

This is in contrast to the *semantics* of `IIDDensity` which (except the conflation of `value` and `[value]`) would work with any object that already supports `logdensityof` interface.

devmotion: > This is in contrast to the semantics of IIDDensity which (except the conflation of value and [value]) would work with any object that already supports logdensityof interface.

A default implementation wouldn't cover the implementation of `loglikelihood` since it does not assume that multiple samples are not necessarily arrays of individual samples. It's not relevant here if this is a good choice (I don't think so) but the only purpose of the (unexported!) `IIDDensity` is to be able to start the transition to DensityInterface if one relies on the behaviour of `loglikelihood`. It is still possible to support other densities by e.g. dispatching on `Distribution` and only using `IIDDensity` for them. Maybe in the long run `IIDDensity` won't be needed or useful, or maybe it will be upstreamed :shrug: But it allows to adopt DensityInterface without any breaking changes.

oschulz: > though not supported by other packages using this API

I also don't like this - it sound like `IIDDensity` is violating the contract of the `LogDensity`-API. But I think with the right interpretation/definition it doesn't: If we say it's the density over all possible outcomes of all possible draws from `d`, over all possible numbers of samples in each draw, then the behavior of `logdensityof` is just as defined, right?

oschulz: I would rather argue that the name of `Distributions.loglikelihood` isn't ideal ... but that's a different topic. :-)

oschulz: > I also don't like this - it sound like IIDDensity is violating the contract of the LogDensity-API. 

Ah, sorry @mschauer , I think I misunderstood - with "though not supported by other packages using this API" you were referring not to what `logdensityof` means for `IIDDensity` but you were referring to the relationship to `loglikelihood`?

mschauer: That's what I meant: the semantics if `IIDDensity` is as valid as the rest, it's just impossible to support `IIDDensity` without inheriting from `Distributions` because it is living here. So it's something a producer of densities using Distributions can use but not a consumer of density-like objects. It's perhaps ok like it is for now, I am just reviewing the PR because I care

docs/src/density_interface.md
oschulz: Ok, I'll mark this resolved, then? Please feel free to reopen if not.

docs/src/density_interface.md
oschulz: Ah right, sorry for the misunderstanding!

> just impossible to support IIDDensity without inheriting from Distributions because it is living here.

Yes, it is a bit of a weird thing, being in Distributions but being just a density. I guess we could, in principle turn IIDDensity into an actual `Distribution`: We can define `rand` for it (won't be very useful, but we can define it), and it's `logpdf` would be implemented by calling `loglikelihood` on the underlying Distribution. We wouldn't even need a special method for `logdensity` then. It wouldn't be possible to normalize the PDF though. Would that be acceptable?

mschauer: I think it would need to live outside `<:Distribution` because otherwise it breaks assumptions on how `Distribution`s behave, I am not even sure if defining a new variate form would do the trick. Compare that `Product` is only defined for univariate densities, not because products of multivariate densities don't make sense, just because `Product` was forced to take a place in the type hierarchy. Perhaps now it would even work with making product `ArrayLikeVariate{N} where N` without loosing too much, but that's a bit outside of our scope right now.

oschulz: Hm, if it was restricted to a number of draws/samples, we *could* normalize such a distribution. @devmotion , could you do something like this in Turing?

```julia
logdensityof(ProductOver(d, nsamples(x)), x)
```

There would need to be a function `nsamples(x)` that determines the number of "entries" in `x`, But that we should be able to code, right?

oschulz: I think a `ProductOver(d::Distribution, n::Integer)` could be very useful in general, even for a univariate `d`, because `Distributions.Product(FillArrays.Fill(d, n))` doesn't work (`rand` on it produces vectors filled with the same value in each entries).

devmotion: This will be fixed in my refactoring of `Product` (hopefully done within the next days), so it should be covered by the regular product distribution.

devmotion:
Maybe we could do the following, as I'm also not particularly satisfied with the slightly unusual design of `IIDDensity` and it's limitations but we need something that behaves like `loglikelihood` (I still think it's a feature that it supports both single and multiple samples - why shouldn't it be possible to evaluate the log-likelihood of a single observation or alternative why should one have to wrap it in a container? I think the `broadcast` argument/view is quite fitting):
- Only define the standard `logdensityof` and `densityof` for `Distribution` in this PR, with the errors for cases with multiple samples for which `logpdf` and `pdf` are defined in Distributions (in all other cases it will error automatically)
- Define an internal `IIDDensity` in DynamicPPL to get the `loglikelihood` behaviour in Turing for `Distribution`s (this PR shows this can be done in a very simple way) until there is a better/different solution (e.g. based on the refactored `Product`)

oschulz:
> Define an internal IIDDensity in DynamicPPL

So we just move IIDDensity to DynamicPPL? Sounds good.

oschulz:
Ok, here's our current IIDDensity implementation, to make it easy to find for DynamicPPL later on:

```julia


"""
    Distributions.IIDDensity(d::Distribution)

Represents the probability density of an implicit product distribution of
variates that are identically and independently distributed according to
the distribution `d`.

Use `DensityInterface.logdensityof(d, x)` to compute the logarithmic density
value at `x`. `x` may be a single variate of `d` or a whole set of variates
of `d`.

If `x` is a single variate of `d`, the density is the PDF of `d`.

If `x` is a set of variates (given as a higher-dimensional array or
and array of arrays), the density is the PDF of an implicit product
distribution over `d`, the size of the product is implied by the size of
the set.

`DensityInterface.logdensityof(IIDDensity(d::Distribution), x)` is equivalent
to `loglikelihood(d, x)`.
"""
struct IIDDensity{D<:Distribution}
    distribution::D
end

@inline DensityInterface.hasdensity(d::IIDDensity) = true

# ToDo: Move documentation of behavior of `logdensityof(d::IIDDensity, x)` to
# a method docstring of logdensityof here if/when IIDDensity becomes exported.
DensityInterface.logdensityof(d::IIDDensity, x) = loglikelihood(d.distribution, x)
```

oschulz:
Ok, `IIDDensity` is gone ... bye for now, `IIDDensity`, see you again in DynamicPPL. :-)

docs/src/density_interface.md
mschauer: That's great!

oschulz: > This will be fixed in my refactoring of Product

Awesome!

src/density_interface.jl
```diff
@@ -0,0 +1,19 @@
+@inline DensityInterface.hasdensity(::Distribution) = true
```
mschauer, line 1: Last bit, let's check that this should indeed return a Boolean value and not a type, so we are relying on constant propagation instead of Holy-type trait like `Base.IteratorEltype`
```
julia> Base.IteratorEltype([1,1])
Base.HasEltype()
```

oschulz:
@devmotion and @mschauer, Ok with you in the new state?

src/density_interface.jl
devmotion: IMO it should return a `Boolean` and only possibly reevaluated if there are any major issues in some implementation that can't be fixed otherwise. In all cases such as
```julia
f(d) = hasdensity(d) ? g(d) : h(d)
```
the compiler should be able to elide one branch. Generally, I think it is much nicer to work with `Boolean`s since they can be used in if statements etc. whereas otherwise one always has to add a helper function and dispatch.

devmotion (approved):
Looks good to me, can you update the version number?

src/density_interface.jl
oschulz: > Last bit, let's check that this should indeed return a Boolean value and not a type

You mean test that returning a Bool is correct (`test_density_interface` checks it)? Or changing DensityInterface itself, and have `hasdensity` return a type?

Hm, constant propagation is supposed to be pretty good nowadays, but ... @devmotion , what do you think?

oschulz: My guess would be that `Val(hasdensity(density_to_test))` should work well nowadays.

mschauer: Ok

devmotion: I trust the compiler here. If we notice any problems, we can reevaluate.

oschulz: Ok, quick test:

```julia
using DensityInterface, Test

struct MyDensity end
@inline DensityInterface.hasdensity(::MyDensity) = true
DensityInterface.logdensityof(::MyDensity, x) = - x^2


get_log_density_value_impl(::Val{true}, d, x) = logdensityof(d, x)
get_log_density_value_impl(::Val{false}, d, x) = "blah, blah"
get_log_density_value(d, x) = get_log_density_value_impl(Val(hasdensity(d)), d, x)

d = MyDensity()
get_log_density_value(d, 4.2)
```

```julia
julia> @inferred get_log_density_value(d, 4.2)
-17.64
```

`@code_warntype get_log_density_value(d, 4.2)` looks fine, too.

oschulz: Also

```julia
julia> hasdensity_as_val(d) = Val(hasdensity(d))
hasdensity_as_val (generic function with 1 method)

julia> @code_llvm hasdensity_as_val(d)
;  @ REPL[28]:1 within `hasdensity_as_val`
define void @julia_hasdensity_as_val_1265() #0 {
top:
  ret void
}
```

oschulz:
> Looks good to me, can you update the version number?

Done. Let's wait until we've resolved the last open issue (question regarding return type of `hasdensity`) though.

src/density_interface.jl
oschulz: @mschauer will this work for you?

oschulz: > @mschauer  Ok

Ah, sorry, I overlooked that one. Ok, I'll mark as resolved.

oschulz:
Ok, I think we're through, all good from my side.

mschauer:
I merged this, trusting that we revisit how to allow the DensityInterface to preserve and make available within the interface some of the information and context which defines what kind of object `pdf` actually returns and which is available through our type parameters.

oschulz:
> that we revisit how to allow the DensityInterface to preserve and make available within the interface some of the information and context which defines what kind of object pdf actually returns

Absolutely. From my side, DensityInterface is a beginning, not the end of this story.

PR #1418, opened 2021-11-09
MvNormal Test Suite inside src

willtebbutt:
This PR moves the tests for the MvNormal type inside a new module inside Distributions entitled `TestUtils`. The benefit of locating this code here is that it makes it easy for other packages which implement subtypes of `AbstractMvNormal` to run the same tests as Distributions to make them confident that they have implemented Distributions' `AbstractMvNormal` API properly.

This is non-breaking.

I've also sorted a couple of style-related things, but no additional tests have been added / removed.

cc @devmotion

### DIFF
--- a/test/mvnormal.jl
+++ b/test/mvnormal.jl
@@ -10,92 +10,6 @@ using LinearAlgebra, Random, Test
 using SparseArrays
 using FillArrays
 
-import Distributions: distrname
-
-
-
-####### Core testing procedure
-
-function test_mvnormal(g::AbstractMvNormal, n_tsamples::Int=10^6,
-                       rng::Union{AbstractRNG, Missing} = missing)
-    d = length(g)
-    μ = mean(g)
-    Σ = cov(g)
-    @test length(μ) == d
-    @test size(Σ) == (d, d)
-    @test var(g)     ≈ diag(Σ)
-    @test entropy(g) ≈ 0.5 * logdet(2π * ℯ * Σ)
-    ldcov = logdetcov(g)
-    @test ldcov ≈ logdet(Σ)
-    vs = diag(Σ)
-    @test g == typeof(g)(params(g)...)
-    @test g == deepcopy(g)
-    @test minimum(g) == fill(-Inf, d)
-    @test maximum(g) == fill(Inf, d)
-    @test extrema(g) == (minimum(g), maximum(g))
-    @test isless(extrema(g)...)
-
-    # test sampling for AbstractMatrix (here, a SubArray):
-    if ismissing(rng)
-        subX = view(rand(d, 2d), :, 1:d)
-        @test isa(rand!(g, subX), SubArray)
-    else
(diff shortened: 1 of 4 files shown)


test/mvnormal.jl
```diff
@@ -10,92 +10,6 @@ using LinearAlgebra, Random, Test
...
-    @test g == deepcopy(g)
-    @test minimum(g) == fill(-Inf, d)
-    @test maximum(g) == fill(Inf, d)
-    @test extrema(g) == (minimum(g), maximum(g))
-    @test isless(extrema(g)...)
-
-    # test sampling for AbstractMatrix (here, a SubArray):
-    if ismissing(rng)
-        subX = view(rand(d, 2d), :, 1:d)
-        @test isa(rand!(g, subX), SubArray)
```
devmotion, line 41: It seems you removed the tests without RNG argument? I don't think the `missing` implementation here is good style but I think maybe it can be useful to still check it. We could e.g. just perform both tests, i.e., sample with and without the RNG provided explicitly, and loop over the checks. What do you think?

devmotion (commented):
Thanks, it's great to clean up the test infrastructure a bit :slightly_smiling_face:

test/mvnormal.jl
willtebbutt: Ah, yes. I had been assuming the purpose of the `missing` was to make it possible to run the tests when the caller couldn't be bothered to provide a random number generator. Was the point actually to ensure that the various `rand` methods (with and without an `AbstractRNG`) were covered? If so, yes, running both make sense to me.

devmotion: Hmm seems `missing` was added in  https://github.com/JuliaStats/Distributions.jl/pull/830 when support for custom RNGs was added. I guess the motivation was to only extend but not really modify the tests.

After having thought a bit more about it I think it's not a good idea to test both variants: then the global RNG would be used and tested even if you specify a different one. Ie., the state of the global RNG is affected and you would always also have to seed the global RNG if you want to set seeds.

Maybe the easiest is to use something similar to the existing version. IMO, we could switch from `missing` to `nothing` (it's not really missing, we don't specify an RNG on purpose) and we could use helper functions such as eg
```julia
_rand(::Nothing, args...) = rand(args...)
_rand(rng::AbstractRNG, args...) = rand(rng, args...)

_rand!(::Nothing, args...) = rand!(args...)
_rand!(rng::AbstractRNG, args...) = rand!(rng, args...)
```
to remove duplicate code.

willtebbutt: Interesting. I'll adopt your suggestion.

willtebbutt: Suggestion adopted. Please resolve if you're happy @devmotion

devmotion (approved):
LGTM, thanks!

willtebbutt:
Excellent.

I don't have merge rights, so please merge and tag when you get a minute :)

PR #1419, opened 2021-11-10
Remove `truncated(d, ::Integer, ::Integer)`

devmotion:
This PR removes a default fallback for `truncated(d::UnivariateDistribution, lo::Integer, hi::Integer)` which is not needed anymore (possibly due to recent changes in the default implementation of `truncated`).

The main advantage is that it makes it easier to overload `truncated`: currently one has to resolve method ambiguities that arise from the `Integer` fallback (see e.g. https://github.com/JuliaStats/Distributions.jl/pull/1349).

### DIFF
--- a/src/truncated/uniform.jl
+++ b/src/truncated/uniform.jl
@@ -2,6 +2,4 @@
 ##### Shortcut for truncating uniform distributions.
 #####
 
-truncated(d::Uniform, l::T, u::T) where {T <: Real} = Uniform(promote(max(l, d.a), min(u, d.b))...)
-
-truncated(d::Uniform, l::Integer, u::Integer) = truncated(d, float(l), float(u))
+truncated(d::Uniform, l::T, u::T) where {T <: Real} = Uniform(max(l, d.a), min(u, d.b))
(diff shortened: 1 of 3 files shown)


src/truncated/uniform.jl
```diff
@@ -2,6 +2,4 @@
 ##### Shortcut for truncating uniform distributions.
 #####
 
-truncated(d::Uniform, l::T, u::T) where {T <: Real} = Uniform(promote(max(l, d.a), min(u, d.b))...)
-
-truncated(d::Uniform, l::Integer, u::Integer) = truncated(d, float(l), float(u))
+truncated(d::Uniform, l::T, u::T) where {T <: Real} = Uniform(max(l, d.a), min(u, d.b))
```
devmotion, line 5: Promotion is handled by the `Uniform` constructor.

matbesancon (approved):

matbesancon:
looks good thanks! Although I think codecov would have been happier if we had increased coverage by 0% instead of decreasing it by the same amount :smile:

Issue #968, opened 2019-09-09
Why can't this package use numerical methods for mean of Truncated Distributions?


Deduction42:
I noticed that Truncated distributions generally don't support mean, var, etc. But they could easily be calculated using numerical methods. In fact, I wrote my own custom module for them:

function numeric_mean( Dist::ContinuousUnivariateDistribution )
    (LB, UB) = extrema(Dist)
    F(x) = x*pdf(Dist,x)
    return quadgk(F, LB, UB)[1]
end

function numeric_mean( Dist::DiscreteUnivariateDistribution )
    (LB, UB) = extrema(Dist)
    F(x) = x*pdf(Dist,x)
    ϵ = 1E-9

    #Ensure loop is finite
    if !isfinite(LB)
        LB = quantile(Dist, ϵ)
    end
    if !isfinite(UB)
        UB = quantile(Dist, 1-ϵ)
    end

    #Loop through and numerically calculate the mean
    y = 0.0
    for x in LB:UB
        y += F(x)
    end
    return y
end

I'm wondering why this sort of method (numeric_mean) isn't used as a fallback for the mean truncated distributions? I tried to extend "mean" myself, but it doesn't seem to work smoothly; I get errors and there's possible issues around type piracy. If performance is an issue, would it not be enough to give a warning that a numerical method is being used as an approximation?

matbesancon:
Numerical methods can be complex to implement right and will have **a lot** of tricky corner cases.

cossio:
A generic routine doing numerical integration like that is likely to have a lot of issues, for instance if you truncate on the tail of the distribution where the pdf is decaying to zero.

I think we should think about each case separately and handle them when possible. For instance Truncated Uniform is trivial. 
And `Truncated(Normal(), -1, 1)` is also well handled by Distributions.jl.

PR #1420, opened 2021-11-11
Change order of arguments in `expectation`

devmotion:
I noticed in https://github.com/JuliaStats/Distributions.jl/pull/1414 that the order of arguments in `expectation` does not follow the general Julia recommendations and style. IMO the function should be the first argument since this would also allow to use the `do` syntax. The PR fixes this issue.

Additionally, I wonder if `expectation` should be exported.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.25.28"
+version = "0.25.29"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/deprecates.jl
+++ b/src/deprecates.jl
@@ -47,6 +47,7 @@ end
 @deprecate Wishart(df::Real, S::Matrix, warn::Bool) Wishart(df, S)
 @deprecate Wishart(df::Real, S::Cholesky, warn::Bool) Wishart(df, S)
 
-# Deprecate 3 arguments expectation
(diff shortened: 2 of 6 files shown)


mschauer:
Not really a fan of having numerical integration in `Distributions.jl`, not sure if worth fixing it and even exporting it. Similar to want @matbesancon says here: https://github.com/JuliaStats/Distributions.jl/issues/968#issuecomment-535432617

devmotion:
I understand these concerns and I don't think we *have* to export it. However, I think it is just a fallback (and the natural one), so no guarantees on accuracy and/or ineffeciency. Additionally, `expectation` has existed for more than 9 years: https://github.com/JuliaStats/Distributions.jl/commit/87eca10bbcbab484260f1b96c7d3a3fb60a41aef

QuadGK is also used in other places, e.g., for computations with the Chernoff distribution and the non-central hypergeometric distribution.

mschauer:
Yeah, we don't need to remove it, but I am fine with it to continue living under the stairs with the wrong argument order (I agree that the current order bad for `do`-operations)

devmotion:
So you think we shouldn't fix the order internally and be able to use `do` syntax internally? This PR does not export `expectation`, it was only a question that occurred to me.

mschauer:
In any case it seems that people don't rely much on it outside: https://juliahub.com/ui/Search?q=Distributions.%2aexpectation&type=code&r=true

devmotion:
The PR is not intended for othee packages or users.

mschauer:
No harm is done by checking if people are using unexpected function and if changing them will break that code.

mschauer (approved):
Ok, let's do it, as we are able to provide update path via the deprecation of the old order.

PR #1421, opened 2021-11-11
add entropy(::LogUniform)

jw3126:
cc @devmotion

### DIFF
--- a/src/univariate/continuous/loguniform.jl
+++ b/src/univariate/continuous/loguniform.jl
@@ -45,6 +45,10 @@ end
 mode(d::LogUniform)   = d.a
 modes(d::LogUniform)  = partype(d)[]
 
+function entropy(d::LogUniform)
+    a,b = params(d)
+    log(a * b) / 2 + log(log(b / a))
+end
 #### Evaluation
 function pdf(d::LogUniform, x::Real)
     x1, a, b = promote(x, params(d)...) # ensure e.g. pdf(LogUniform(1,2), 1f0)::Float32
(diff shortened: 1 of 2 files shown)


src/univariate/continuous/loguniform.jl
```diff
@@ -62,3 +66,14 @@ function quantile(d::LogUniform, p::Real)
     p1,a,b = promote(p, params(d)...) # ensure e.g. quantile(LogUniform(1,2), 1f0)::Float32
     exp(p1 * log(b/a)) * a
 end
+
+function kldivergence(p::LogUniform, q::LogUniform)
+    ap, bp, aq, bq = promote(params(p)..., params(q)...)
+    finite = (0 < aq <= ap < bp <= bq)
```
devmotion, line 72: `aq > 0` (and also `ap > 0`) should always hold, so I guess we don't have to check it.
```suggestion
    finite = aq <= ap < bp <= bq
```

src/univariate/continuous/loguniform.jl
```diff
@@ -62,3 +66,14 @@ function quantile(d::LogUniform, p::Real)
...
+    res = log(log(bq / aq) / log(bp / ap))
+    if finite
+        return res
+    else
+        return typeof(res)(Inf)
+    end
```
devmotion, line 78: I guess it's a bit more idiomatic to use `oftype(res, Inf)`. Only a matter of taste but I think the following would be a bit nicer:
```suggestion
    return finite ? res : oftype(res, Inf)
```

devmotion (commented):
I just have two minor suggestions.

devmotion (approved):
Looks good, thanks for the PR! I'll merge if tests pass.

devmotion:
Thanks!

PR #1424, opened 2021-11-16
New Distribution: MvDiscreteNonParametric

davibarreira:
I've created a new distribution which implements the `DiscreteNonParametric` distribution to the multivariate case.
The distribution followed the guidelines for new distributions (one can sample, calculate mean, var, cov, etc).
If this PR is accepted, I'll submit another PR with documentation.

Highlights:
* `mvdiscretenonparametric.jl` - Has the implementation of the new `MvDiscreteNonParametric` distribution;
* Dependency on ArraysOfArrays - `MvDiscreteNonParametric` uses ArraysOfArarys in order to properly work with vector of vectors, where each vector is a "sample" (point-mass).

The new distribution does not have a `cdf` yet. I'm working on a more efficient way to perform the computations instead of naively sorting and comparing everything.


src/multivariate/mvdiscretenonparametric.jl
```diff
@@ -0,0 +1,153 @@
+struct MvDiscreteNonParametric{T <: Real,P <: Real,Ts <:  ArrayOfSimilarArrays{T},Ps <: AbstractVector{P}} <: DiscreteMultivariateDistribution
```
devmotion, line 1: Can we make this more general and change it to something like
```suggestion
struct GeneralDiscreteNonParametric{VF,T,P<:Real,Ts<:AbstractVector{T},Ps<:AbstractVector{P}} <: Distribution{VF,Discrete}
```
? Then it could also be used instead of `DiscreteNonParametric` for the univariate case and we could define the alias
```julia
const DiscreteNonParametric{T<:Real,P<:Real,Ts<:AbstractVector{T},Ps<:AbstractVector{P}} = GeneralDiscreteNonParametric{Univariate,T,P,Ts,Ps}
```
to make it a non-breaking change.

src/multivariate/mvdiscretenonparametric.jl
```diff
@@ -0,0 +1,153 @@
...
+end
+
+function cov(d::MvDiscreteNonParametric)
+    x = support(d)
+    p = probs(d)
+    return cov(x, Weights(p, one(eltype(p))), corrected=false)
+end
+
+entropy(d::MvDiscreteNonParametric) = entropy(probs(d))
+entropy(d::MvDiscreteNonParametric, b::Real) = entropy(probs(d), b)
```
devmotion, line 153: This is defined in StatsBase (https://github.com/JuliaStats/StatsBase.jl/blob/0a179531f14d0140356fcf1f1d55bd0240030cf2/src/scalarstats.jl#L537):
```suggestion
```

src/multivariates.jl
```diff
@@ -29,8 +29,8 @@ function _rand!(rng::AbstractRNG, s::Sampleable{Multivariate}, m::AbstractMatrix
     @boundscheck size(m, 1) == length(s) ||
         throw(DimensionMismatch("Output size inconsistent with sample length."))
     smp = sampler(s)
-    for i in Base.OneTo(size(m,2))
-        _rand!(rng, smp, view(m,:,i))
+    for i in Base.OneTo(size(m, 2))
+        _rand!(rng, smp, view(m, :, i))
```
devmotion, line 33: Can you revert the unrelated changes?

Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["JuliaStats"]
 version = "0.25.29"
 
 [deps]
+ArraysOfArrays = "65a8f2f4-9b39-5baf-92e2-a9cc46fdf018"
```
devmotion, line 7: I don't think the implementation has to or should use `ArraysOfArrays` explicitly - I guess you can just work with generic `AbstractVector{<:AbstractVector}`, and then one can e.g. provide such Array wrappers from ArraysOfArrays if desired.

src/multivariate/mvdiscretenonparametric.jl
devmotion: And one could also define the alias
```suggestion
const MvDiscreteNonParametric{T<:AbstractVector{<:Real},P<:Real,Ts<:AbstractVector{T},Ps<:AbstractVector{P}} = GeneralDiscreteNonParametric{Multivariate,T,P,Ts,Ps}
```

src/multivariate/mvdiscretenonparametric.jl
```diff
@@ -0,0 +1,153 @@
...
+"""
+    size(d::MvDiscreteNonParametric)
+Returns the size of the support as a tuple where
+the first value is the number of points (samples)
+and the second is the dimension of the samples (e.g ℝⁿ).
+It corresponds to `size(flatview(d.support)')`
+where `flatview` is a function from from ArraysOfArrays.jl
+that turns an Array of Arrays into a matrix.
+"""
+Base.size(d::MvDiscreteNonParametric) = size(flatview(d.support)')
```
devmotion, line 103: For instance, here one could just use
```suggestion
Base.length(d::MvDiscreteNonParametric) = length(first(d.support))
Base.size(d::MvDiscreteNonParametric) = (length(d), length(d.support))
```

src/multivariate/mvdiscretenonparametric.jl
```diff
@@ -0,0 +1,153 @@
...
+
+function _rand!(rng::AbstractRNG, d::MvDiscreteNonParametric, x::AbstractVector{T}) where T <: Real
```
devmotion, line 105: Since this is basically the same as the univariate implementation, it would be good to merge them.

src/multivariate/mvdiscretenonparametric.jl
```diff
@@ -0,0 +1,153 @@
...
+    draw = Base.rand(rng, float(eltype(p)))
+    cp = p[1]
+    i = 1
+    while cp <= draw && i < n
+        @inbounds cp += p[i += 1]
+    end
+    for (j, v) in enumerate(s[i])
+        x[j] = v
+    end
+    return x
```
devmotion, line 121: Probably easier and better (currently it does not work with non-standard indices):
```suggestion
    copyto!(x, s[i])
    return x
end
```

Project.toml
davibarreira: Ok. But then there are some significant adjustments I have to make, cause I've relied on ArraysOfArrrays to do stuff like convert matrices to vector of vectors. What do you suggest in order to dispatch them?

devmotion: IMO we should remove the matrix dispatch (even regardless of ArraysOfArrays) as data can be organized in rows or columns - but this is not clear if just a matrix is provided. I think users should just use `eachrow`/`eachcol` (in a future Julia version or with Compat: https://github.com/JuliaLang/julia/pull/32310), ArraysOfArrays or ColVecs/RowVecs in KernelFunctions (I wanted to wait with extracting them a bit since I was hoping for the Julia PR).

src/multivariate/mvdiscretenonparametric.jl
davibarreira: Can you expand on how the `GeneralDiscreteNonParametric` shoudl work? It's not clear to me how this `VF` work, for example, I tried doing the following:
```julia

struct GeneralDiscreteNonParametric{VF,T,P<:Real,Ts<:AbstractVector{T},Ps<:AbstractVector{P}} <: Distribution{VF,Discrete}

    support::Ts
    p::Ps

    function GeneralDiscreteNonParametric{VF,T,P,Ts,Ps}(support::Ts,
        p::Ps) where {T,P <: Real,Ts <: AbstractVector{<:AbstractVector{T}},Ps <: AbstractVector{P}}

        length(support) == length(p) || error("length of `support` and `p` must be equal")
        isprobvec(p) || error("`p` must be a probability vector")
        allunique(support) || error("`support` must contain only unique value")
        new{VF,T,P,Ts,Ps}(support, p)
    end
end
```
But of course I get an error, since `VF` is not defined.

Project.toml
davibarreira: Ok, I'll drop the matrix dispatch.

src/multivariate/mvdiscretenonparametric.jl
davibarreira: Also, should the `GeneralDiscreteNonParametric` dispatch to `MvDiscreteNonParametric` and `DiscreteNonParametric`? Would something like CategoricalDistributions.jl be inside `GeneralDiscreteNonParametric`?

davibarreira: @devmotion , should I leave the functions for `MvDiscreteNonParametric` or should I drop it and just do everything inside the `GeneralMvDiscreteNonParametric`?

Project.toml
davibarreira: I just realized that there is an issue with not using `ArraysOfArrays`. I've heavily used it for computing `mean`, `var` and `cov`. I doesn't seem to work with a regular vector of vectors. Do you have a suggestion for it?

davibarreira: I just realized that computations of weighted `mean`, `var` and `cov` depend on `ArraysOfArrays`. Do you know a workaround?

src/multivariate/mvdiscretenonparametric.jl
devmotion: > But of course I get an error, since VF is not defined.

You have to provide it, like
```julia
struct GeneralDiscreteNonParametric{VF,T,P<:Real,Ts<:AbstractVector{T},Ps<:AbstractVector{P}} <: Distribution{VF,Discrete}
    support::Ts
    p::Ps

    function GeneralDiscreteNonParametric{VF,T,P,Ts,Ps}(support::Ts,
        p::Ps; check_args=true) where {VF,T,P<: Real,Ts<:AbstractVector{T},Ps<:AbstractVector{P}}
        if check_args
            length(support) == length(p) || error("length of `support` and `p` must be equal")
            isprobvec(p) || error("`p` must be a probability vector")
            allunique(support) || error("`support` must contain only unique values")
        end
        new{VF,T,P,Ts,Ps}(support, p)
    end
end
```
Then you could define a univariate discrete non-parametric distribution with `GeneralDiscreteNonParametric{Univariate,Float64,Float64,Vector{Float64},Vector{Float64}}([-10.0, -3.4, 2.1, 3.5], [0.4, 0.2, 0.1, 0.1])`. But, of course, users shouldn't have to do this. Therefore we could use the aliases (also for the multivariate case) and define outer constructors such as
```julia
function DiscreteNonParametric(support::AbstractVector{<:Real}, p::AbstractVector{<:Real}; issorted=false)
    _p = issorted ? p : sort(p)
    return DiscreteNonParametric{eltype(support),eltype(_p),typeof(support),typeof(_p)}(support, _p)
end
```

davibarreira: Besides Multivariate and Univariate, what kind of distribution you intend to keep under GeneralDiscreteNonParametric? Categorical Distributions?

devmotion: They are already a type alias of `DiscreteNonParametric`, so if `DiscreteNonParametric` works they should just work automatically.

davibarreira:
@devmotion, can I run the `format` from the terminal here in the same way we do in OptimalTransport.jl? My editor sometimes "auto-formats" stuff -.- , so I end up messing the previous formatting.

devmotion:
No, we don't use a specific style here (yet). (BTW in general running the JuliaFormatter won't necessarily revert all of your changes.)

davibarreira:
> No, we don't use a specific style here (yet). (BTW in general running the JuliaFormatter won't necessarily revert all of your changes.)

Thanks. I ended up restoring the originals and doing the changes again, avoiding the auto-formatting.

Project.toml
devmotion: They all work for vectors of vectors (of course, `ArraysOfArrays`, `ColVecs` or `EachCol` may implement more optimized versions of `mean` etc., but the default fallback works always):
```julia
julia> using Statistics

julia> x = [rand(5) for _ in 1:5];

julia> mean(x)
5-element Vector{Float64}:
 0.2609279080845831
 0.5184741034528969
 0.523898105188404
 0.6408166253497452
 0.33381236831836236

julia> var(x)
5-element Vector{Float64}:
 0.019751971071567172
 0.07655721742732527
 0.06221826760951234
 0.022761714823869204
 0.10639711437483416

julia> cov(x)
5×5 Matrix{Float64}:
  0.019752    -0.0182541  -0.00331084   0.0142255  -0.0231928
 -0.0182541    0.0765572  -0.0500437   -0.0402498   0.0677088
 -0.00331084  -0.0500437   0.0622183    0.0214676  -0.0425375
  0.0142255   -0.0402498   0.0214676    0.0227617  -0.0405306
 -0.0231928    0.0677088  -0.0425375   -0.0405306   0.106397
```

davibarreira: Not the weighted version. I ended up writing a for-loop. Don't know if it's the most efficient way.

devmotion: Ah I didn't realize that you need the weighted ones. They are defined in StatsBase, there are some issues about exactly the vector of vectors case: https://github.com/JuliaStats/StatsBase.jl/issues/534 and https://github.com/JuliaStats/StatsBase.jl/issues/518

davibarreira: Here is what I came up with:
```julia
function cov(d::MvDiscreteNonParametric)
    x = hcat(support(d)...)
    p = probs(d)
    return cov(x, Weights(p, one(eltype(p))), 2,corrected = false)
end
```

davibarreira: @devmotion , are you fine with this implementation?

davibarreira:
@devmotion , I've implemented some of your suggestions (I still have to adjust DiscreteNonParametric to be used with GeneralNonParametric), but I'm having a hard time to get the sampler to work for `MvDiscreteNonParametric`. The `rand!()` function simply does not work. I mean, when I do:
```julia
x = collect(eachrow(rand(10,3)))
p = normalize!(rand(10),1)
μ = MvDiscreteNonParametric(x,p)
rand(μ)
```
it's working. But if I do
```
rand(μ,1)
```
I get instead:
```
3×2 Matrix{SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}}:
 [0.0969373, 0.623776, 0.447577]  [0.824966, 0.174585, 0.805307]
 [0.824966, 0.174585, 0.805307]   [0.0969373, 0.623776, 0.447577]
 [0.562016, 0.439165, 0.146882]   [0.562016, 0.439165, 0.146882]
```

devmotion:
The `eltype` definition is incorrect. It should return the element type of the samples for arrayvariate distributions (ie., e.g., `Float64` if the samples are `AbstractVector{Float64}`) but currently in the multivariate case you define it as a vector.

davibarreira:
I've modified the code to
```
Base.eltype(::Type{<:MvDiscreteNonParametric{T}}) where T = Base.eltype(T)
```
Now, this returns for example `Float64`, but the `rand(mu, 1)` now does not work. I get an errror:
```DimensionMismatch("inconsistent array dimensions")```

File: src/multivariate/mvdiscretenonparametric.jl [hunk 8]
```diff
@@ -0,0 +1,104 @@
...
+
+"""
+    probs(d::MvDiscreteNonParametric)
+Get the vector of probabilities associated with the support of `d`.
+"""
+probs(d::MvDiscreteNonParametric) = d.p
+
+
+Base.length(d::MvDiscreteNonParametric) = length(first(d.support))
+Base.size(d::MvDiscreteNonParametric) = (length(d), length(d.support))
```
devmotion, line 55: This is wrong and defined automatically based on `length`:

```suggestion
```

devmotion (commented):
The definition of `size` is not correct, it can just be removed.

src/multivariate/mvdiscretenonparametric.jl
davibarreira: Ok, done. But I'm still getting the erro :/, but instead it's:
```
StackOverflowError:

Stacktrace:
     [1] rand!(rng::Random._GLOBAL_RNG, s::MvDiscreteNonParametric{SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}, Float64, Vector{SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}}, Vector{Float64}}, x::SubArray{Float64, 1, Matrix{Float64}, Tuple{Base.Slice{Base.OneTo{Int64}}, Int64}, true})
       @ Distributions ~/MEGA/EMAp/Distributions.jl/src/genericrand.jl:83
     [2] _rand!(rng::Random._GLOBAL_RNG, s::MvDiscreteNonParametric{SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}, Float64, Vector{SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}}, Vector{Float64}}, x::SubArray{Float64, 1, Matrix{Float64}, Tuple{Base.Slice{Base.OneTo{Int64}}, Int64}, true})
       @ Distributions ~/MEGA/EMAp/Distributions.jl/src/genericrand.jl:117
     [3] rand!(rng::Random._GLOBAL_RNG, s::MvDiscreteNonParametric{SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}, Float64, Vector{SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}}, Vector{Float64}}, x::SubArray{Float64, 1, Matrix{Float64}, Tuple{Base.Slice{Base.OneTo{Int64}}, Int64}, true})
       @ Distributions ~/MEGA/EMAp/Distributions.jl/src/genericrand.jl:91
--- the last 2 lines are repeated 14426 more times ---
 [28856] _rand!(rng::Random._GLOBAL_RNG, s::MvDiscreteNonParametric{SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}, Float64, Vector{SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}}, Vector{Float64}}, x::SubArray{Float64, 1, Matrix{Float64}, Tuple{Base.Slice{Base.OneTo{Int64}}, Int64}, true})
       @ Distributions ~/MEGA/EMAp/Distributions.jl/src/genericrand.jl:117
 [28857] rand!
       @ ~/MEGA/EMAp/Distributions.jl/src/genericrand.jl:91 [inlined]
 [28858] _rand!(rng::Random._GLOBAL_RNG, s::MvDiscreteNonParametric{SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}, Float64, Vector{SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}}, Vector{Float64}}, x::Matrix{Float64})
       @ Distributions ~/MEGA/EMAp/Distributions.jl/src/genericrand.jl:117
 [28859] rand!
       @ ~/MEGA/EMAp/Distributions.jl/src/genericrand.jl:108 [inlined]
 [28860] rand
       @ ~/MEGA/EMAp/Distributions.jl/src/multivariates.jl:21 [inlined]
 [28861] rand(s::MvDiscreteNonParametric{SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}, Float64, Vector{SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}}, Vector{Float64}}, dims::Int64)
       @ Distributions ~/MEGA/EMAp/Distributions.jl/src/genericrand.jl:22
 [28862] top-level scope
       @ In[31]:1
 [28863] eval
       @ ./boot.jl:373 [inlined]
```

davibarreira: I had added some changes. The `_rand!` was missing. I added it back, and it worked.

davibarreira:
@devmotion , this PR has possibly all your corrections with the exception of incorporating the `DiscreteNonParametric` inside the `GeneralNonParametric`. But I `MvDiscreteNonParametric` should be working. Should we do the modifications of `DiscreteNonParametric` in this PR?

src/Distributions.jl
```diff
@@ -98,6 +98,7 @@ export
     FullNormalCanon,
     Gamma,
     DiscreteNonParametric,
+    GeneralDiscreteNonParametric,
```
devmotion, line 101: Maybe let's keep it unexported until everything is figured out?
```suggestion
```

src/multivariate/generaldiscretenonparametric.jl
```diff
@@ -0,0 +1,79 @@
...
+    p = probs(d)
+    n = length(p)
+    draw = rand(rng, float(eltype(p)))
+    cp = p[1]
+    i = 1
+    while cp <= draw && i < n
+        @inbounds cp += p[i +=1]
+    end
+    return x[i]
+end
```
devmotion, line 31: While this method is correct, it messes with the `rand` dispatches. Maybe either change it to the univariate case only (i.e., use `::GeneralDiscreteNonParametric{Univariate}`)
```suggestion
function rand(rng::AbstractRNG, d::GeneralDiscreteNonParametric{Univariate})
    x = support(d)
    p = probs(d)
    n = length(p)
    draw = rand(rng, float(eltype(p)))
    cp = p[1]
    i = 1
    while cp <= draw && i < n
        @inbounds cp += p[i +=1]
    end
    return x[i]
end
```
 or let's remove it for now and add it in a follow-up PR that updates `DiscreteNonParametric`:
```suggestion
```

src/multivariate/generaldiscretenonparametric.jl
```diff
@@ -0,0 +1,79 @@
...
+
+"""
+    support(d::MvDiscreteNonParametric)
+Get a sorted AbstractVector defining the support of `d`.
+"""
```
devmotion, line 36: No need for a docstring, `support` is already documented (and I don't think it is sorted here actually?)
```suggestion
```

src/multivariate/generaldiscretenonparametric.jl
```diff
@@ -0,0 +1,79 @@
...
+support(d::GeneralDiscreteNonParametric) = d.support
+
+"""
+    probs(d::MvDiscreteNonParametric)
+Get the vector of probabilities associated with the support of `d`.
```
devmotion, line 41:
```suggestion
    probs(d::GeneralDiscreteNonParametric)

Return the vector of probabilities associated with the support of `d`.
```

src/multivariate/generaldiscretenonparametric.jl
```diff
@@ -0,0 +1,79 @@
...
+"""
+probs(d::GeneralDiscreteNonParametric) = d.p
+
+
+Base.length(d::GeneralDiscreteNonParametric) = length(first(d.support))
```
devmotion, line 46: `length` should be defined for the multivariate case only, for other cases we need `size`: 
```suggestion
Base.length(d::GeneralDiscreteNonParametric{Multivariate}) = length(first(d.support))
Base.size(d::GeneralDiscreteNonParametric{<:ArrayLikeVariate}) = size(first(d.support))
```
IMO it is better to be a bit conservative here and only define it for `ArrayLikeVariate`s.

src/multivariate/generaldiscretenonparametric.jl
```diff
@@ -0,0 +1,79 @@
...
+
+function _rand!(
+    rng::AbstractRNG,
+    d::GeneralDiscreteNonParametric,
+    x::AbstractVector{T},
+    ) where {T<:Real}
```
devmotion, line 52: `_rand!` is only used by `ArrayLikeVariate`s, hence better to restrict the method a bit - and generalize it to arbitrary sizes:
```suggestion
function _rand!(
    rng::AbstractRNG,
    d::GeneralDiscreteNonParametric{ArrayLikeVariate{N}},
    x::AbstractArray{<:Real,N},
) where {N}
```

src/multivariate/generaldiscretenonparametric.jl
```diff
@@ -0,0 +1,79 @@
...
+
+    length(x) == length(d) || throw(DimensionMismatch("Invalid argument dimension."))
```
devmotion, line 54: The dimensions are already checked before `_rand!` is called, no need to do it again here:
```suggestion
```

src/multivariate/generaldiscretenonparametric.jl
```diff
@@ -0,0 +1,79 @@
...
+    s = d.support
+    p = d.p
+
+    n = length(p)
+    draw = Base.rand(rng, float(eltype(p)))
```
devmotion, line 59: No need to qualify `rand`
```suggestion
    draw = rand(rng, float(eltype(p)))
```

src/multivariate/generaldiscretenonparametric.jl
```diff
@@ -0,0 +1,79 @@
...
+    i = 1
+    while cp <= draw && i < n
+        @inbounds cp += p[i+=1]
+    end
+    copyto!(x, s[i])
+    return x
+end
+
+
+function _logpdf(d::GeneralDiscreteNonParametric, x::AbstractVector{T}) where {T<:Real}
```
devmotion, line 70: Again, the dispatches are only used by `ArrayLikeVariate` distributions, so let's just use
```suggestion
function _logpdf(d::GeneralDiscreteNonParametric{ArrayLikeVariate{N}}, x::AbstractArray{<:Real,N}) where {N}
```

src/multivariate/generaldiscretenonparametric.jl
```diff
@@ -0,0 +1,79 @@
...
+    s = support(d)
+    p = probs(d)
+    for i = 1:length(p)
+        if s[i] == x
+            return log(p[i])
+        end
+    end
```
devmotion, line 77: In general, this could be optimized with `searchsortedfirst` if the support would be supported. At least for `ArrayLikeVariate` distributions sorting is clearly defined and works well, so maybe sort the support at least in these cases?
```julia
julia> sort([[0.0, 1.0, 0.0], [1.0, 0.0, 0.0], [0.5, 0.0, 0.5]])
3-element Vector{Vector{Float64}}:
 [0.0, 1.0, 0.0]
 [0.5, 0.0, 0.5]
 [1.0, 0.0, 0.0]

julia> searchsortedfirst(ans, [0.5, 0.0, 0.5])
```

src/multivariate/mvdiscretenonparametric.jl
```diff
@@ -0,0 +1,55 @@
+const MvDiscreteNonParametric{
+    T<:AbstractVector{<:Real},
+    P<:Real,
+    Ts<:AbstractVector{T},
+    Ps<:AbstractVector{P},
+} = GeneralDiscreteNonParametric{Multivariate,T,P,Ts,Ps}
```
devmotion, line 6: What's the motivation for introducing this type-alias? It seems we could work with `GeneralDiscreteNonParametric` directly? Also it would only cover the multivariate case even though other `ArrayLikeVariate` cases could be handled in the same way.

I suggest removing this file completely and moving the definitions to `src/generaldiscretenonparametric.jl` instead (it should not be in the `multivariate` subfolder since it is not necessarily a multivariate distribution).

src/multivariate/mvdiscretenonparametric.jl
```diff
@@ -0,0 +1,55 @@
...
+x = collect(eachrow(rand(10,2)))
+μ = MvDiscreteNonParametric(x)
+
+# columns correspond to samples
+y = collect(eachcol(rand(7,12)))
+ν = MvDiscreteNonParametric(y)
+```
+"""
+function MvDiscreteNonParametric(
+    support::AbstractArray{<:AbstractVector{<:Real}},
```
devmotion, line 30: This should be
```suggestion
    support::AbstractVector{<:AbstractVector{<:Real}},
```
in the multivariate case, and
```suggestion
    support::AbstractVector{<:AbstractArray{<:Real,N}},
```
only in the more general case of `ArrayLikeVariate{N}`.

src/multivariate/mvdiscretenonparametric.jl
```diff
@@ -0,0 +1,55 @@
...
+    p::AbstractVector{<:Real} = fill(inv(length(support)), length(support)),
```
devmotion, line 31: I'm not sure if we should add a default definition for `p`. In any case, we should use the non-allocating `Fill` instead.
```suggestion
    p::AbstractVector{<:Real} = Fill(inv(length(support)), length(support)),
```

src/multivariate/mvdiscretenonparametric.jl
```diff
@@ -0,0 +1,55 @@
...
+)
+    return MvDiscreteNonParametric{eltype(support),eltype(p),typeof(support),typeof(p)}(
+        support,
+        p,
+    )
+end
+
+Base.eltype(::Type{<:MvDiscreteNonParametric{T}}) where T = Base.eltype(T)
```
devmotion, line 39: This should be defined more generally:
```suggestion
Base.eltype(::Type{<:GeneralDiscreteNonParametric{<:ArrayLikeVariate,T}}) where T = eltype(T)
```

src/multivariate/mvdiscretenonparametric.jl
```diff
@@ -0,0 +1,55 @@
...
+    x = hcat(support(d)...)
+    p = probs(d)
+    return StatsBase.var(x, Weights(p, one(eltype(p))), 2,corrected = false)
+end
+
+function cov(d::MvDiscreteNonParametric)
+    x = hcat(support(d)...)
+    p = probs(d)
+    return cov(x, Weights(p, one(eltype(p))), 2,corrected = false)
+end
```
devmotion, line 55: I requested to add support for vectors of vectors to StatsBase but it might take a bit until it is available. IMO it is fine to have possibly less optimized implementations until then. However, we should avoid splatting and use
```suggestion
function mean(d::MvDiscreteNonParametric)
    return mean(reduce(hcat, d.support), Weights(d.p, one(eltype(d.p))); dims=2)
end

function var(d::MvDiscreteNonParametric)
    x = reduce(hcat, support(d))
    p = probs(d)
    return var(x, Weights(p, one(eltype(p))), 2; corrected = false)
end

function cov(d::MvDiscreteNonParametric)
    x = reduce(hcat, support(d))
    p = probs(d)
    return cov(x, Weights(p, one(eltype(p))), 2; corrected = false)
end
```

src/univariate/discrete/discretenonparametric.jl
```diff
@@ -69,19 +69,6 @@ Base.isapprox(c1::D, c2::D) where D<:DiscreteNonParametric =
...
-    n = length(p)
-    draw = rand(rng, float(eltype(p)))
-    cp = p[1]
-    i = 1
-    while cp <= draw && i < n
-        @inbounds cp += p[i +=1]
-    end
-    return x[i]
-end
-
```
devmotion, line 84: I guess this was removed accidentally? I guess it would be cleaner to not make any changes to `DiscreteNonParametric` in this PR.

src/multivariate/generaldiscretenonparametric.jl
davibarreira: I actually have been implementing an algorithm that does this more efficiently. But I only got the 2D case working. My idea is to submit a future PR improving this.

src/univariate/discrete/discretenonparametric.jl
davibarreira: yeah. my bad

src/multivariate/generaldiscretenonparametric.jl
devmotion: Why do we need a custom algorithm? Can't we just use `searchsortedfirst`? `sort` etc works with arrays just fine in Julia.

davibarreira: I meant for computing the empirical cdf. Not exactly the sorting of the support.

src/multivariate/mvdiscretenonparametric.jl
davibarreira: Do we just drop the `MvDiscreteNonParametric` then? Should I change all for `GeneralNonParametric`?

src/multivariate/mvdiscretenonparametric.jl
davibarreira: Should I just use the second option and change the function to `GeneralDiscreteNonParametric`?

davibarreira:
@juliohm , this might interest you.

PR #1425, opened 2021-11-16
Remove some utilities and use external functions

devmotion:
There are some utility functions such as `exp!`, `multiply!`, and `add!` in Distributions. This PR removes them and replaces them with standard broadcasting or mapping (already the default definition currently). The main motivation is that by using standard functionality it is likely that the code works for custom array types and also in an efficient way (e.g., FillArrays defines optimizations for `map` and `broadcast`) - it seems very unlikely that custom array types would implement internal functions in Distributions and hence Distributions would have to maintain support for them. Arguably it also increases readability if the amount of internal functions is reduced.

Additionally, the PR removes some utility functions that are not used in the package anymore.


src/multivariate/mvnormalcanon.jl
```diff
@@ -174,7 +174,8 @@ if isdefined(PDMats, :PDSparseMat)
     unwhiten_winv!(J::PDSparseMat, x::AbstractVecOrMat) = x[:] = J.chol.PtL' \ x
 end
 
-_rand!(rng::AbstractRNG, d::MvNormalCanon, x::AbstractVector) =
-    add!(unwhiten_winv!(d.J, randn!(rng,x)), d.μ)
-_rand!(rng::AbstractRNG, d::MvNormalCanon, x::AbstractMatrix) =
-    add!(unwhiten_winv!(d.J, randn!(rng,x)), d.μ)
+function _rand!(rng::AbstractRNG, d::MvNormalCanon, x::AbstractVecOrMat)
```
devmotion, line 177:
```suggestion
function _rand!(rng::AbstractRNG, d::MvNormalCanon, x::AbstractVector)
    unwhiten_winv!(d.J, randn!(rng, x))
    x .+= d.μ
    return x
end
function _rand!(rng::AbstractRNG, d::MvNormalCanon, x::AbstractMatrix)
```

andreasnoack (approved):

mschauer:
The integration tests have great return on investment... https://github.com/TuringLang/DistributionsAD.jl/blob/21641cf20ce75ed31e780f8419120f90955489fe/src/multivariate.jl#L249

devmotion:
Yep I almost expected some errors and hence ran the downstream tests. I'll fix DistributionsAD before merging this PR. Even though technically it's not a problem with this PR but DistributionsAD's excessive use of internal functions (did I mention already that I don't like this package and look forward to when it's not needed anymore? 😄).

PR #1426, opened 2021-11-16
Fix precondition for dirichlet_mode

siddharthab:
I believe the current code was a typo and not intentional.

### DIFF
--- a/src/multivariate/dirichlet.jl
+++ b/src/multivariate/dirichlet.jl
@@ -111,6 +111,7 @@ function entropy(d::Dirichlet)
 end
 
 function dirichlet_mode!(r::AbstractVector{<:Real}, α::AbstractVector{<:Real}, α0::Real)
+    all(x -> x > 1, α) || error("Dirichlet has a mode only when alpha[i] > 1 for all i")
     k = length(α)
     inv_s = inv(α0 - k)
     @. r = inv_s * (α - 1)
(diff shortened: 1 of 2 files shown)


src/multivariate/dirichlet.jl
```diff
@@ -118,7 +118,7 @@ function dirichlet_mode!(r::AbstractVector{<:Real}, α::AbstractVector{<:Real},
 end
 
 function dirichlet_mode(α::AbstractVector{<:Real}, α0::Real)
-    all(αi < 1 for αi in α) || error("Dirichlet has a mode only when alpha[i] > 1 for all i")
+    all(αi > 1 for αi in α) || error("Dirichlet has a mode only when alpha[i] > 1 for all i")
```
devmotion, line 121: Thanks, yes this seems to be a typo! A bit faster for `FillArray`s would be
```suggestion
    all(x -> x > 1, α) || error("Dirichlet has a mode only when alpha[i] > 1 for all i")
```
So maybe let's change it?

devmotion (commented):
Thank you for the PR, yes this should be changed!

Could you add the same check to the in-place method `dirichlet_mode!` as well? And add tests for both methods, both for alphas > 1 and alphas < 1?

devmotion (approved):
Thanks!

PR #1427, opened 2021-11-16
Support DensityInterface v0.4

oschulz:
This is non-breaking and keeps support for DensityInterface v0.3.2.

### DIFF
--- a/src/density_interface.jl
+++ b/src/density_interface.jl
@@ -1,9 +1,4 @@
-@static if isdefined(DensityInterface, :DensityKind)
-    @inline DensityInterface.DensityKind(::Distribution) = HasDensity()
-else
-    # DensityInterface v0.3
-    @inline DensityInterface.hasdensity(::Distribution) = true
-end
+@inline DensityInterface.DensityKind(::Distribution) = HasDensity()
 
 for (di_func, d_func) in ((:logdensityof, :logpdf), (:densityof, :pdf))
     @eval begin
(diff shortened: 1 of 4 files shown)


devmotion (commented):
Thanks for the PR! Is there a good reason to keep support for < 0.4? I would assume it is easier to just switch to 0.4.

Can you also update the version number of Distributions?

oschulz:
> I would assume it is easier to just switch to 0.4.

I guess so, not many packages depend on DensityInterface so far.

Ok, dropped DensityInterface v0.3 and updated Distributions version number.

src/density_interface.jl
```diff
@@ -1,4 +1,4 @@
-@inline DensityInterface.hasdensity(::Distribution) = true
+@inline DensityInterface.DensityKind(::Distribution) = HasDensity()
```
devmotion, line 1: I think it's not exported, is it?
```suggestion
@inline DensityInterface.DensityKind(::Distribution) = DensityInterface.HasDensity()
```

oschulz: Hm, it actually *is* exported:

https://github.com/JuliaMath/DensityInterface.jl/blob/e71eb2ded6f3320d66a6d7efc615e3e60dddb58f/src/interface.jl#L69

oschulz: Ah, but we "import DensityInterface" :-)

devmotion (approved):
Thank you!

oschulz:
Thanks!

PR #1428, opened 2021-11-18
Make symbols in deprecation warnings more explicit

rikhuijzer:
The deprecation warning read:
```
┌ Warning: `MvNormal(d::Int, σ::Real)` is deprecated, use `MvNormal(Diagonal(Fill(σ ^ 2, d)))` instead.
```
However, `Fill` is not exported by Distributions and is likely to cause much searching for many readers.

This PR fixes it by using `fill` instead of `Fill`. Another solution would be to `export Fill` in `src/Distributions.jl`.

cc: @devmotion

### DIFF
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -217,10 +217,10 @@ Construct a multivariate normal distribution with zero mean and covariance matri
 MvNormal(Σ::AbstractMatrix{<:Real}) = MvNormal(Zeros{eltype(Σ)}(size(Σ, 1)), Σ)
 
 # deprecated constructors with standard deviations
-Base.@deprecate MvNormal(μ::AbstractVector{<:Real}, σ::AbstractVector{<:Real}) MvNormal(μ, Diagonal(map(abs2, σ)))
+Base.@deprecate MvNormal(μ::AbstractVector{<:Real}, σ::AbstractVector{<:Real}) MvNormal(μ, LinearAlgebraDiagonal(map(abs2, σ)))
 Base.@deprecate MvNormal(μ::AbstractVector{<:Real}, σ::Real) MvNormal(μ, σ^2 * I)
-Base.@deprecate MvNormal(σ::AbstractVector{<:Real}) MvNormal(Diagonal(map(abs2, σ)))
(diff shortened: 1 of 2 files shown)


devmotion:
No, I definitely think users should use `Fill` - otherwise the behaviour is different than before and their code will become allocating.

Also a no on the second point, we agreed to not re-export FillArrays (and other dependencies) but instead document them clearly: https://github.com/JuliaStats/Distributions.jl/pull/1368

devmotion:
I wonder if the deprecation message becomes clearer if we use `LinearAlgebra.Diagonal` and `FillArrays.Fill` in the `@deprecated` definition. Ideally users should see
```julia
┌ Warning: `MvNormal(d::Int, σ::Real)` is deprecated, use `MvNormal(LinearAlgebra.Diagonal(FillArrays.Fill(σ ^ 2, d)))` instead.
```
which would match our (ideal) documentation (all external structs and functions are qualified).

devmotion:
It seems to work locally (don't pay attention to the stupid example):
```julia
julia> using LinearAlgebra

julia> g(x) = sum(abs2, x)

julia> Base.@deprecate f(x) g(LinearAlgebra.Diagonal(x))

julia> f(rand(3, 3))
┌ Warning: f(x) is deprecated, use g(LinearAlgebra.Diagonal(x)) instead.
│   caller = top-level scope at REPL[4]:1
└ @ Core REPL[4]:1
0.2616974132900723
```

rikhuijzer:
> I wonder if the deprecation message becomes clearer if we use `LinearAlgebra.Diagonal` and `FillArrays.Fill` in the `@deprecated` definition. Ideally users should see
> 
> ```julia
> ┌ Warning: `MvNormal(d::Int, σ::Real)` is deprecated, use `MvNormal(LinearAlgebra.Diagonal(FillArrays.Fill(σ ^ 2, d)))` instead.
> ```
> 
> which would match our (ideal) documentation (all external structs and functions are qualified).

I love it!

devmotion (commented):
Thanks, looks good! While you're at it, maybe update the deprecations in `mvnormalcanon.jl` as well?

devmotion (approved):
LGTM, thanks! I'll merge when tests pass.

devmotion:
Test errors are unrelated (recent release of PDMats uncovered a bug in the implementation of `Wishart`).

Issue #1371, opened 2021-07-26
cdf returns negative zero


aplavin:
In a recent `Distributions.jl` version, `cdf(Rayleigh(1), 0)` started returning the negative zero:
```
cdf(Rayleigh(1), 0) === 0.0  # false
cdf(Rayleigh(1), 0) === -0.0  # true
``` 
Negative zero is compared equal to positive zero of course, but mathematical functions treat them differently sometimes. Including distributions: e.g., `rand(Binomial(100, -0.))` throws an exception.
So, I believe that `-0.0` should never be returned from `cdf` methods.

PR #1430, opened 2021-11-21
Rayleigh cdf always nonnegative

aplavin:
fix https://github.com/JuliaStats/Distributions.jl/issues/1371

### DIFF
--- a/src/univariate/continuous/rayleigh.jl
+++ b/src/univariate/continuous/rayleigh.jl
@@ -79,7 +79,8 @@ end
 
 function logccdf(d::Rayleigh, x::Real)
     z = - x^2 / (2 * d.σ^2)
-    return x > 0 ? z : isnan(x) ? oftype(z, NaN) : zero(x)
+    # return negative zero so that cdf is +0, not -0
+    return x > 0 ? z : isnan(x) ? oftype(z, NaN) : -zero(z)
 end
 ccdf(d::Rayleigh, x::Real) = exp(logccdf(d, x))
 
--- a/test/continuous.jl
+++ b/test/continuous.jl
@@ -98,3 +98,9 @@ end
     @test var(d) ≈ δ * α^2 / g^3
     @test skewness(d) ≈ 3β/(α*sqrt(δ*g))
 end
+
+@testset "edge cases" begin
+    # issue #1371: cdf should not return -0.0
+    @test cdf(Rayleigh(1), 0) === 0.0
+    @test cdf(Rayleigh(1), -10) === 0.0
+end


src/univariate/continuous/rayleigh.jl
```diff
@@ -83,7 +83,7 @@ function logccdf(d::Rayleigh, x::Real)
 end
 ccdf(d::Rayleigh, x::Real) = exp(logccdf(d, x))
 
-cdf(d::Rayleigh, x::Real) = -expm1(logccdf(d, x))
+cdf(d::Rayleigh, x::Real) = false - expm1(logccdf(d, x))  # zero (false) ensures that cdf cannot be -0
```
devmotion, line 86: I would prefer something that fixes the issue more directly - this seems like a quite weird workaround. The underlying problem here is that the implementation of `expm1` in base returns
```julia
julia> expm1(0.0)
0.0

julia> expm1(-0.0)
-0.0
```
Since
```julia
julia> iszero(-0.0)
true

julia> iszero(0.0)
true
```
both positive and negative zero are proper zeros, I'm not completely sure why a negative zero would be a problem here. Do you have a MWE where it causes problems?

If we want to change it maybe we could just use
```suggestion
# we ensure that the cdf cannot be -0
cdf(d::Rayleigh, x::Real) = max(- expm1(logccdf(d, x)), 0)
```

aplavin: See the linked issue for motivation. A simple example:
```
julia> using Distributions
julia> d = Rayleigh(1)
# nonzero works fine:
julia> rand(Binomial(100, cdf(d, 0.1)))
2
# zero doesn't:
julia> rand(Binomial(100, cdf(d, 0)))
ERROR: InexactError: trunc(Int64, -Inf)
<...>
# but it would, if cdf didn't return negative zero:
julia> rand(Binomial(100, 0.0))
0
```

test/continuous.jl
```diff
@@ -98,3 +98,7 @@ end
     @test var(d) ≈ δ * α^2 / g^3
     @test skewness(d) ≈ 3β/(α*sqrt(δ*g))
 end
+
+@testset "edge cases" begin
+    @test cdf(Rayleigh(1), 0) === 0.0  # not -0
```
devmotion, line 103: Maybe it's helpful to mention the issue here?
```suggestion
    # issue #1371: cdf should not return -0.0
    @test cdf(Rayleigh(1), 0) === 0.0
```

src/univariate/continuous/rayleigh.jl
aplavin: Actually, a simpler solution is possible. Changed.

src/univariate/continuous/rayleigh.jl
```diff
@@ -79,7 +79,7 @@ end
 
 function logccdf(d::Rayleigh, x::Real)
     z = - x^2 / (2 * d.σ^2)
-    return x > 0 ? z : isnan(x) ? oftype(z, NaN) : zero(x)
+    return x >= 0 ? z : isnan(x) ? oftype(z, NaN) : zero(x)
```
devmotion, line 82: Oh, actually we should use
```suggestion
    return x >= 0 ? z : isnan(x) ? oftype(z, NaN) : zero(z)
```
here to ensure type stability. It shouldn't matter though if we use `>` or `>=`, should it?

aplavin: Almost - it should be negative zero(z), see updated commit.

src/univariate/continuous/rayleigh.jl
```diff
@@ -79,7 +79,7 @@ end
 
 function logccdf(d::Rayleigh, x::Real)
     z = - x^2 / (2 * d.σ^2)
-    return x > 0 ? z : isnan(x) ? oftype(z, NaN) : zero(x)
+    return x >= 0 ? z : isnan(x) ? oftype(z, NaN) : -zero(z)
```
devmotion, line 82: Yeah, this is simpler (thought about it actually but assumed that it doesn't matter if we return negative or positive zeros for `logccdf`). Could you add the comment about positive cdf here? Otherwise I assume it can be a bit unclear for a reader why `-zero(z)` is used here.

devmotion:
It seems one of the tests fails? https://github.com/JuliaStats/Distributions.jl/runs/4279041031?check_suite_focus=true#step:6:469

aplavin:
fixed

devmotion (approved):
Thanks, looks good to me! The test error is unrelated (and there's an open PR that fixes it).

PR #1435, opened 2021-11-23
Add skewed exponential power distribution

lukketotte:
Implements the skewed exponential power distribution (SEPD) as given by Eq. 6 in [Zhy & Zinde-Walsh (2009)](https://www.sciencedirect.com/science/article/pii/S0304407608001668)


src/univariate/continuous/skewexponentialpower.jl
```diff
@@ -0,0 +1,115 @@
+"""
```
devmotion, line 1: If you use
```suggestion
@doc raw"""
```
you can use single backslashes, e.g. ``\mu`` instead of ``\\mu`` (you don't have to change it).

src/univariate/continuous/skewexponentialpower.jl
```diff
@@ -0,0 +1,115 @@
...
+The *Skewed exponential power distribution*, with location `μ`, scale `σ`, shape `p`, and skewness `α`
+has the probability density function [1]
+```math
+f(x; \\mu, \\sigma, p, \\alpha) =
+\\begin{cases}
+\\frac{1}{\\sigma 2p^{1/p}\\Gamma(1+1/p)} \\exp \\left\\{ - \\frac{1}{2p}\\Big| \\frac{x-\\mu}{\\alpha \\sigma} \\Big|^p \\right\\}, & \\text{if } x \\leq \\mu \\\\
+\\frac{1}{\\sigma 2p^{1/p}\\Gamma(1+1/p)} \\exp \\left\\{ - \\frac{1}{2p}\\Big| \\frac{x-\\mu}{(1-\\alpha) \\sigma} \\Big|^p \\right\\}, & \\text{if } x > \\mu
+\\end{cases}.
+```
+The Skewed exponential power distribution (SEPD) incorporates the laplace (``p=1, \\alpha=0.5``),
```
devmotion, line 13:
```suggestion
The Skewed exponential power distribution (SEPD) incorporates the Laplace (``p=1, \\alpha=0.5``),
```

src/univariate/continuous/skewexponentialpower.jl
```diff
@@ -0,0 +1,115 @@
...
+normal (``p=2, \\alpha=0.5``), uniform (``p\\rightarrow \\infty, \\alpha=0.5``), asymmetric laplace (``p=1``), skew normal (``p=2``),
```
devmotion, line 14:
```suggestion
normal (``p=2, \\alpha=0.5``), uniform (``p\\rightarrow \\infty, \\alpha=0.5``), asymmetric Laplace (``p=1``), skew normal (``p=2``),
```

src/univariate/continuous/skewexponentialpower.jl
```diff
@@ -0,0 +1,115 @@
...
+    σ::T
+    p::T
+    α::T
+    SkewExponentialPower{T}(μ::T, σ::T, p::T, α::T) where {T} = new{T}(μ, σ, p, α)
+end
+
+function SkewExponentialPower(µ::T, σ::T, p::T, α::T; check_args=true) where {T <: Real}
+    check_args && @check_args(SkewExponentialPower, σ > zero(σ))
+    check_args && @check_args(SkewExponentialPower, p > zero(p))
+    check_args && @check_args(SkewExponentialPower, α > zero(α) && α < one(α))
```
devmotion, line 43: A bit simpler:
```suggestion
    if check_args
        @check_args(SkewExponentialPower, σ > zero(σ))
        @check_args(SkewExponentialPower, p > zero(p))
        @check_args(SkewExponentialPower, zero(α) < α < one(α))
    end
```

src/univariate/continuous/skewexponentialpower.jl
```diff
@@ -0,0 +1,115 @@
...
+    return SkewExponentialPower{T}(µ, σ, p, α)
+end
+
+SkewExponentialPower(μ::Real, σ::Real, p::Real, α::Real) = SkewExponentialPower(promote(μ, σ, p, α)...)
+SkewExponentialPower(μ::Real, σ::Real, p::Real) = SkewExponentialPower(promote(μ, σ, p, 0.5)...)
+SkewExponentialPower(μ::Real, σ::Real) = SkewExponentialPower(promote(μ, σ, 2., 0.5)...)
+SkewExponentialPower(μ::Real) = SkewExponentialPower(promote(μ, 1., 2., 0.5)...)
+SkewExponentialPower() = SkewExponentialPower(0., 1., 2., 0.5)
```
devmotion, line 51: A bit simpler, with less invasive promotions (useful e.g. when working with `Float32`), and the possibility to specify `check_args`:
```suggestion
function SkewExponentialPower(μ::Real=0, σ::Real=1, p::Real=2, α::Real=1//2; kwargs...)
    return SkewExponentialPower(promote(μ, σ, p, α)...; kwargs...)
end
```

src/univariate/continuous/skewexponentialpower.jl
```diff
@@ -0,0 +1,115 @@
...
+
+@distr_support SkewExponentialPower -Inf Inf
+
+### Conversions
+convert(::Type{SkewExponentialPower{T}}, μ::S, σ::S, p::S, α::S) where {T <: Real, S <: Real} = SkewExponentialPower(T(μ), T(σ), T(p), T(α))
+convert(::Type{SkewExponentialPower{T}}, d::SkewExponentialPower{S}) where {T <: Real, S <: Real} = SkewExponentialPower(T(d.μ), T(d.σ), T(d.p), T(d.α), check_args=false)
```
devmotion, line 57: We could add a specialization if the type argument does not change:
```suggestion
convert(::Type{SkewExponentialPower{T}}, d::SkewExponentialPower{S}) where {T <: Real, S <: Real} = SkewExponentialPower(T(d.μ), T(d.σ), T(d.p), T(d.α), check_args=false)
convert(::Type{SkewExponentialPower{T}}, d::SkewExponentialPower{T}) where {T<:Real} = d
```

src/univariate/continuous/skewexponentialpower.jl
```diff
@@ -0,0 +1,115 @@
...
+shape(d::SkewExponentialPower) = d.p
+scale(d::SkewExponentialPower) = d.σ
+
+### Statistics
+
+#Calculates the kth central moment of the SEPD, see Equation 18 in [1].
+function m_k(d::SkewExponentialPower, k::Integer)
+    _, σ, p, α = params(d)
+    (2*p^(1/p))^k*((-1)^k*α^(1+k) + (1-α)^(1+k)) *
+        σ^k * gamma((1+k)/p) / gamma(1/p)
```
devmotion, line 73: Maybe the calculation should be performed in log space (then you could also use e.g. `loggamma`)? I assume that this might be better (i.e., more accurate and no over- or underflow) for extreme parameter values. Did you try e.g. very small values of `p`?

src/univariate/continuous/skewexponentialpower.jl
```diff
@@ -0,0 +1,115 @@
...
+end
+
+mean(d::SkewExponentialPower) = m_k(d, 1) + d.μ
+mode(d::SkewExponentialPower) =  m_k(d, 1) + d.μ
+var(d::SkewExponentialPower) = m_k(d, 2) - m_k(d, 1)^2
+std(d::SkewExponentialPower) = √var(d)
```
devmotion, line 79: Isn't this the default definition? I would assume this can be removed:
```suggestion
```

src/univariate/continuous/skewexponentialpower.jl
```diff
@@ -0,0 +1,115 @@
...
+
+#Calculates the kth central moment of the SEPD, see Equation 18 in [1].
+function m_k(d::SkewExponentialPower, k::Integer)
+    _, σ, p, α = params(d)
+    (2*p^(1/p))^k*((-1)^k*α^(1+k) + (1-α)^(1+k)) *
+        σ^k * gamma((1+k)/p) / gamma(1/p)
+end
+
+mean(d::SkewExponentialPower) = m_k(d, 1) + d.μ
+mode(d::SkewExponentialPower) =  m_k(d, 1) + d.μ
```
devmotion, line 77: Maybe simpler:
```suggestion
mode(d::SkewExponentialPower) = mean(d)
```

src/univariate/continuous/skewexponentialpower.jl
```diff
@@ -0,0 +1,115 @@
...
+var(d::SkewExponentialPower) = m_k(d, 2) - m_k(d, 1)^2
+std(d::SkewExponentialPower) = √var(d)
+skewness(d::SkewExponentialPower) = m_k(d, 3)/(std(d))^3
+kurtosis(d::SkewExponentialPower) = m_k(d, 4)/var(d)^2 - 3.
```
devmotion, line 81: We should avoid hard-coded `Float64` literals since they can cause undesired promotions:
```suggestion
kurtosis(d::SkewExponentialPower) = m_k(d, 4)/var(d)^2 - 3
```

src/univariate/continuous/skewexponentialpower.jl
```diff
@@ -0,0 +1,115 @@
...
+
+function logpdf(d::SkewExponentialPower, x::Real)
+    μ, σ, p, α = params(d)
+    -log(σ) - log(2*p^(1/p)*gamma(1+1/p)) - 1/p*(x < μ ? ((μ-x)/(2*σ*α))^p : ((x-μ)/(2*σ*(1-α)))^p)
```
devmotion, line 85: Again, I think it would be better to perform more calculations in log space and use `loggamma`:
```suggestion
    a = x < μ ? α : 1 - α
    inv_p = inv(p)
    return -(logtwo + log(σ) + inv_p * log(p) + loggamma(1 + inv_p) + inv_p * (abs(μ - x) / (2 * σ * a))^p)
```

src/univariate/continuous/skewexponentialpower.jl
```diff
@@ -0,0 +1,115 @@
...
+end
+
+pdf(d::SkewExponentialPower, x::Real) = exp(logpdf(d, x))
+
```
devmotion, line 89: This is the default definition and should not be needed:
```suggestion
```

src/univariate/continuous/skewexponentialpower.jl
```diff
@@ -0,0 +1,115 @@
+"""
+    SkewExponentialPower(μ, σ, p, α)
```
devmotion, line 2: Is there a reason why you called it `SkewExponentialPower` instead of e.g. `SkewedExponentialPower`?

devmotion (commented):
Thank you for the PR! It looks very good, I added some minor comments and suggestions.

(BTW great to see other people at UU who use Julia :slightly_smiling_face:)

src/univariate/continuous/skewexponentialpower.jl
lukketotte: Thanks for all the comments! I will adjust asap. Named it SkewedExponentialPower first but noticed the skewnormal.jl naming and changed skewed to skew. But I'll revert to skewed as that is the term used in the paper :)

src/univariate/continuous/skewexponentialpower.jl
lukketotte: Fixed lowercases

src/univariate/continuous/skewexponentialpower.jl
lukketotte: Indeed, fixed :)

src/univariate/continuous/skewexponentialpower.jl
lukketotte: Good point, did some comparisons and the log-scale allows computations of moments for lower p. This might be worth looking into for the PGenerealizedGaussian also.

Changed to log-scale, also using `loggamma`

src/univariate/continuous/skewexponentialpower.jl
lukketotte: Removed

src/univariate/continuous/skewexponentialpower.jl
lukketotte: fixed

src/univariate/continuous/skewexponentialpower.jl
lukketotte: my bad! fixed

src/univariate/continuous/skewexponentialpower.jl
lukketotte: Changed to use `loggamma`

src/univariate/continuous/skewexponentialpower.jl
lukketotte: fixed

lukketotte:
Made changes per your comments :)

src/univariate/continuous/skewedexponentialpower.jl
```diff
@@ -0,0 +1,118 @@
...
+shape(d::SkewedExponentialPower) = d.p
+scale(d::SkewedExponentialPower) = d.σ
+
+### Statistics
+
+#Calculates the kth central moment of the SEPD on the log-scale
+function m_k(d::SkewedExponentialPower, k::Integer)
+    _, σ, p, α = params(d)
+    return (k*log(2) + k*inv(p) * log(p) + k*log(σ) + loggamma((1+k)*inv(p)) -
+        loggamma(inv(p))) + log((-1)^k*α^(1+k) + (1-α)^(1+k))
```
devmotion, line 74: The `log(2)` will promote everything to `Float64`. Maybe we could use something like
```suggestion
    inv_p = inv(p)
    return k * (logtwo + inv_p * log(p) + log(σ)) + loggamma((1 + k) * inv_p) -
        loggamma(inv_p) + log((-1)^k * α^(1 + k) + (1 - α)^(1 + k))
```
to avoid such promotions.

src/univariate/continuous/skewedexponentialpower.jl
```diff
@@ -0,0 +1,118 @@
...
+end
+
+mean(d::SkewedExponentialPower) = d.α == 0.5 ? d.μ : exp(m_k(d, 1)) + d.μ
```
devmotion, line 77: This can be type unstable, e.g., if `d.μ` is a `Rational` or `Int`. The safest approach is to compute the other branch and convert `d.μ` to its type. A good heuristic here would be also
```suggestion
mean(d::SkewedExponentialPower) = d.α == 1//2 ? float(d.μ) : exp(m_k(d, 1)) + d.μ
```

devmotion: Could you also add some type inference tests (you can just add `@inferred` around the computations) and make sure that the second branch is tested as well? It seems currently all examples use `alpha = 0.5`.

src/univariate/continuous/skewedexponentialpower.jl
```diff
@@ -0,0 +1,118 @@
...
+mode(d::SkewedExponentialPower) =  mean(d)
+var(d::SkewedExponentialPower) = exp(m_k(d, 2)) - exp(2*m_k(d, 1))
+skewness(d::SkewedExponentialPower) = d.α == 0.5 ? 0 : exp(m_k(d, 3)) / (√(v(d)))^3
```
devmotion, line 80: Similarly, for certain types of parameters `skewness` is type unstable. Either compute the second branch and return a zero of its type or use a heuristic such as
```suggestion
skewness(d::SkewedExponentialPower) = d.α == 1//2 ? float(zero(partype(d))) : exp(m_k(d, 3)) / (√(v(d)))^3
```
I assume that `v` is not defined? It is not covered by the tests currently though since you always use `alpha = 0.5`. Can you fix this?

src/univariate/continuous/skewedexponentialpower.jl
```diff
@@ -0,0 +1,118 @@
...
+    a = x < μ ? α : 1 - α
+    inv_p = inv(p)
+    return -(logtwo + log(σ) + inv_p * log(p) + loggamma(1 + inv_p) + inv_p * (abs(μ - x) / (2 * σ * a))^p)
+end
+
+function cdf(d::SkewedExponentialPower, x::Real)
+    μ, σ, p, α = params(d)
+    inv_p = inv(p)
+    if x <= μ
+        α * (1 - cdf(Gamma(inv_p), inv_p * (abs((x-μ)/σ) / (2*α))^p))
```
devmotion, line 94: Here you could use
```suggestion
        α * ccdf(Gamma(inv_p), inv_p * (abs((x-μ)/σ) / (2*α))^p)
```

lukketotte:
Updated based on your comments. Sorry about the `v()` function, error on my part. Have extended the tests to the case where alpha != 0.5 also.

docs/src/univariate.md
```diff
@@ -418,6 +418,13 @@ Semicircle
 plotdensity((-1, 1), Semicircle, (1,)) # hide
 ```
 
+```@docs
+SkewedExponentialPower
+```
+```@example plotdensity
+plotdensity((-8, 5), SkewedExponentialPower, (0, 1, 0.7, 0.7)) # hide
+```
```
devmotion, line 426: For anyone who watches this PR: I built the documentation locally and the plot seems fine.
image

docs/Project.toml
```diff
@@ -1,4 +1,5 @@
 [deps]
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
```
devmotion, line 2:
```suggestion
```

src/univariate/continuous/skewedexponentialpower.jl
```diff
@@ -0,0 +1,122 @@
...
+scale(d::SkewedExponentialPower) = d.σ
+
+### Statistics
+
+#Calculates the kth central moment of the SEPD
+function m_k(d::SkewedExponentialPower, k::Integer)
+    _, σ, p, α = params(d)
+    inv_p = inv(p)
+    return  ((k*logtwo + k*inv(p) * log(p) + k*log(σ) + loggamma((1+k)*inv(p)) -
+        loggamma(inv(p))) + log(abs((-1)^k*α^(1+k) + (1-α)^(1+k))))
```
devmotion, line 75: It's better to move the multiplication with `k` since otherwise the first term will always be a `Float64`. Also should use `inv_p` :slightly_smiling_face: 
```suggestion
    return  k * (logtwo + inv_p * log(p) + log(σ)) + loggamma((1 + k) * inv_p) -
        loggamma(inv_p) + log(abs((-1)^k * α^(1 + k) + (1 - α)^(1 + k)))
```

devmotion (approved):
Looks good, thanks a lot for the PR! I also checked the plot.

I made two small changes and will merge if tests pass.

PR #1437, opened 2021-11-24
Generalize (log)pdf, rand(!), and MatrixReshaped

devmotion:
This PR is extracted from https://github.com/JuliaStats/Distributions.jl/pull/1391 and contains only
- the generalization of `pdf`, `logpdf` etc.,
- the generalization of `rand` and `rand!`,
- and the generalization of `MatrixReshaped` based on `reshape` and `vec`.

The reason for splitting up https://github.com/JuliaStats/Distributions.jl/pull/1391 is that the changes in this PR are non-breaking whereas, depending on how we deal with `Product`, the other changes in the original PR are breaking. I did not divide it into even smaller chunks since the generalizations of pdf and rand are needed for `reshape`, and without `reshape` they are not really needed and not clearly motivated.


mschauer:
What was the rationale for the implicit broadcast of `logpdf` over its argument here?

devmotion:
It is the currently implemented behaviour and has to be kept if it is supposed to be non-breakimg 🤷‍♂️

`logpdf` is implemented not only for single samples but also for arrays of samples (which is a bit weird since one could just use `map` the curried `logpdf` function - IIRC `broadcast` is not used in the implementation in this PR) and higher-dimensional arrays where the samples are stacked along the last dimensions (probably due to Julias memory layout). The latter is more annoying to implement and can't be achieved with a simple `map` or `broadccast`. So while I think the standard `map` could be removed, the other case becomes more inconvenient for users.

Or rather became, since the `eachvariate` function introduced in this PR makes it more straightforward (and maybe could/should be extended to arrays of samples as well for a unified interface).

Apart from (in)convenience, however, I guess another motivation for `logpdf` with multiple samples is that mapping is not the most performant implementation: Eg, with `MvNormal` one can perform evaluation with matrices instead of each column separately. (And even in the univariate case, an improvement should be possible by separating constant terms and computing them only once.) This is used eg in AbstractGPs.

I'm not sure if such performance improvements would be possible in a reliable way without `logpdf` for higher-dimensional arrays. I assume one approach could be to add specializations for `map(::Base.Fix1{typeof(logpdf),D}, ::EachVariate{N})` where `D <: Distribution{ArrayLikeVariate{N}}` that do not iterate but work with the underlying array directly. But then users would have to use `Base.Fix1` and `eachvariate` instead of a simple `logpdf` call. And eg. it would not work with the proposed `eachcol` types in base.

src/multivariate/mvnormal.jl
```diff
@@ -142,7 +142,7 @@ sqmahal(d::AbstractMvNormal, x::AbstractMatrix) = sqmahal!(Vector{promote_type(p
 
 _logpdf(d::AbstractMvNormal, x::AbstractVector) = mvnormal_c0(d) - sqmahal(d, x)/2
 
-function _logpdf!(r::AbstractArray, d::AbstractMvNormal, x::AbstractMatrix)
+function _logpdf!(r::AbstractVector{<:Real}, d::AbstractMvNormal, x::AbstractMatrix{<:Real})
```
mschauer, line 145: Does this introduce a new restriction?

devmotion: Ah it seems we currently allow outputs of any shape and only check if the length is correct. I'll change this, also in the `logpdf!` definition.

BTW maybe a not completely obvious consequence of this PR is that it would be possible to remove the non-exported `_logpdf` and `_rand!` functions from the interface. Currently we call them if we want to avoid bounds checking but this leads to problems if users or developers implement `logpdf` or `rand!` instead (sometimes one wants to use more efficient implementations for the matrix case and then it is easier to just implement a single `logpdf` function for matrices and vectors; this is done e.g. in AbstractGPs). With this PR, however, we only call `logpdf` and `rand!` internally but use `@inbounds` to indicate that bounds checks should be elided if we are certain that the sizes are correct.

mschauer:
Would it be a good idea to test this with some downstream packages to check it's not breaking anything? Do we do something like that?

devmotion:
I think regardless of this PR integration tests are a good idea. Some months ago I thought about adding some, in particular for DistributionsAD which is broken by upstream changes all the time (not only Distributions). However, it didn't seem feasible at that time since tests took more than two hours, and I didn't know which other dependencies shoild be tested. I made the DistributionsAD tests more modular some weeks ago (so the AD backends can be tested separately in a more standard way that works with the common integration test setup) but eg ForwardDiff tests alone still take more than 1 hour.

Do you have any specific packages in mind that would be useful to test?

mschauer:
Nothing specific, just like throwing a net out.

src/multivariate/mvnormal.jl
devmotion: I fixed the inplace methods.

devmotion:
The DistributionsAD tests pass, I only had to fix conversions of `Multinomial` (showed up in the inner constructor of `Base.Fix1` with ReverseDiff): https://github.com/JuliaStats/Distributions.jl/pull/1437/commits/d0e1a9e167060e680e5a85973fb3d65bfa0f5424

mschauer:
Looks good and tests are green?

devmotion:
Seems good to me, I also added a short section about `reshape` and `vec` in the documentation: https://juliastats.org/Distributions.jl/previews/PR1437/reshape/

Can you approve officially if you are satisfied with the PR?

mschauer (approved):

PR #1439, opened 2021-11-25
Add integration tests

devmotion:
@mschauer  This PR adds some integration tests. To me it seemed DistributionsAD would be the most important downstream package, and the one which is easiest to break accidentally.

Requires the following fix in DistributionsAD (which indicates we just broke it again without noticing): https://github.com/TuringLang/DistributionsAD.jl/pull/207.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,4 +1,5 @@
 name: CI
+
 on:
   pull_request:
     branches:
@@ -7,6 +8,13 @@ on:
     branches:
       - master
     tags: '*'
+
+concurrency:
+  # Skip intermediate builds: always.
+  # Cancel intermediate builds: only if it is a pull request build.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
+
 jobs:
   test:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
--- a/.github/workflows/IntegrationTest.yml
+++ b/.github/workflows/IntegrationTest.yml
@@ -1,0 +1,61 @@
+name: IntegrationTest
+
+on:
+  pull_request:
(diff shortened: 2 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -7,6 +8,13 @@ on:
     branches:
       - master
     tags: '*'
+
+concurrency:
+  # Skip intermediate builds: always.
+  # Cancel intermediate builds: only if it is a pull request build.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
```
devmotion, line 16: This will cancel running tests if a new commit is pushed to a PR. Particularly useful for the longer running integration tests but also helpful to reduce the load on Github actions for our regular tests.

.github/workflows/IntegrationTest.yml
```diff
@@ -0,0 +1,61 @@
...
+
+concurrency:
+  # Skip intermediate builds: always.
+  # Cancel intermediate builds: only if it is a pull request build.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
+
+jobs:
+  test:
+    name: ${{ matrix.package.repo }}
```
devmotion, line 19: Show group as well:
```suggestion
    name: ${{ matrix.package.repo }}/${{ matrix.package.group }}
```

devmotion:
The DistributionsAD/Others issue is fixed in the master branch and the upcoming release of DistributionsAD.

mschauer:
Why does it pass, are we testing against the master branch of DistributionsAD?

devmotion:
Yes, the action checks out the downstream repo in a subfolder. And then runs the specific subset of the tests. The action is the standard for integration tests in the Julia ecosystem and used eg in StatsFuns and ChainRules, SciML, and Turing packages.

mschauer (approved):

PR #1442, opened 2021-11-28
Add Affine distributions

ParadaCarleton:


src/univariate/affine.jl
```diff
@@ -0,0 +1,182 @@
...
+    T = promote_type(eltype(base_dist), typeof(shift), typeof(scale))
+    D = typeof(base_dist)
+    S = value_support(D)
+    return Affine{T,S,D}(T(shift), T(scale), base_dist)
+end
+
+# Composing affine transformations
+function affine(shift::Real, scale::Real, d::Affine)
+    return affine(shift + scale * d.shift, scale * d.scale, d.base_dist)
+end
```
devmotion, line 61: I suggest removing this:

```suggestion
```

We already have an API and should just use `*`, `+`, etc.

src/univariate/affine.jl
```diff
@@ -0,0 +1,182 @@
...
+
+reflected(d::Affine) = d.scale < 0
+
+# Aliases
+const ContinuousAffine{T<:Real,D<:ContinuousUnivariateDistribution} = Affine{T,Continuous,D}
+const DiscreteAffine{T<:Real,D<:DiscreteUnivariateDistribution} = Affine{T,Discrete,D}
+
+Base.eltype(::Type{<:Affine{T}}) where {T} = T
```
devmotion, line 69: This is wrong (see the `eltype` PR for a discussion).

src/univariate/affine.jl
```diff
@@ -0,0 +1,182 @@
...
+    if name == :μ
+        return x.shift
+    elseif name == :σ
+        return x.scale
+    elseif name == :ρ
+        return x.base_dist
+    else
+        return getfield(x, name)
+    end
+end
```
devmotion, line 82: Fields are internal, users should use `location` and `scale` to access them.

```suggestion
```

src/univariate/affine.jl
```diff
@@ -0,0 +1,182 @@
...
+
+# Support
+minimum(d::Affine) = d.shift + d.scale * minimum(d.base_dist)
+maximum(d::Affine) = d.shift + d.scale * maximum(d.base_dist)
```
devmotion, line 86: These are wrong for negative scales.

src/univariate/affine.jl
```diff
@@ -0,0 +1,182 @@
...
+support(d::Affine) = affine_support(d.shift, d.scale, support(d.base_dist))
+
+affine_support(shift::Real, scale::Real, support) = shift .+ scale .* support
+function affine_support(shift::Real, scale::Real, support::RealInterval)
+    _transform = x -> shift + scale * x
+    return RealInterval(_transform(support.lb), _transform(support.ub))
+end
```
devmotion, line 93: Also wrong for negative scales. I suggest removing `_transform`.

src/univariate/affine.jl
```diff
@@ -0,0 +1,182 @@
...
+    S = value_support(D)
+    return Affine{T,S,D}(T(shift), T(scale), base_dist)
+end
+
+# Composing affine transformations
+function affine(shift::Real, scale::Real, d::Affine)
+    return affine(shift + scale * d.shift, scale * d.scale, d.base_dist)
+end
+
+reflected(d::Affine) = d.scale < 0
```
devmotion, line 63: Is this used anywhere?

```suggestion
```

src/Distributions.jl
```diff
@@ -115,7 +116,6 @@ export
     Levy,
     LKJ,
     LKJCholesky,
-    LocationScale,
```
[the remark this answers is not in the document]
ParadaCarleton: @devmotion Isn't it still being exported through the `@deprecate` macro?

src/univariate/affine.jl
ParadaCarleton: `*`, `+`, etc. are implemented as `affine` using args 0/1; is there any reason not to keep it at least internally?

src/Distributions.jl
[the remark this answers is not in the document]
devmotion: Yep, I only saw later that you deprecated it.

src/univariate/affine.jl
ParadaCarleton: It might actually be worth exporting to let people do an affine transformation all at once, instead of having to separately create 2 structs.

And keeping it makes it easier to overload the function, since someone can just overload `affine` instead of overloading each of `+`.

src/univariate/affine.jl
ParadaCarleton: @devmotion Hmm... would `Base.eltype(::Type{<:AffineDistribution{T, S, D}}) where {T, S, D} = promote(T, eltype(D))` be what we'd want?

src/Distributions.jl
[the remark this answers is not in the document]
ParadaCarleton: Ahh, only the constructor is, got it.

src/univariate/affine.jl
devmotion: Only for discrete distributions and only when `eltype(D)` *has* to be promoted - e.g. if it is `Int` but the location or scale are of type `Rational` or `Float64`. In general, the `eltype` should not depend on the type of the parameters.

devmotion:
Some more high-level comments:
- I think we should not introduce any additional API. and hence in particular not add `affine`: We should continue to use `*` and `+` (and `/` and `-`) to construct affine transformations of random variables/distributions.
- Maybe we should use the opportunity and not just rename `LocationScale` and remove the check for the scale but actually change the design and allow general array-like variates, i.e., make it a subtype of `Distribution{ArrayLikeVariate{N},<:ValueSupport}` and allow arrays as location and scale.
- I wonder if it would be useful to split `AffineDistribution` into a `ShiftedDistribution` and a `ScaledDistribution`

src/univariate/affine.jl
```diff
@@ -0,0 +1,182 @@
+@deprecate LocationScale(args...; kwargs...) affine(args...)
```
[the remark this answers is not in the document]
ParadaCarleton: @devmotion trying to add it gives me an error about an invalid redefinition of constant `LocationScale`?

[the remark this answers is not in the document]
devmotion: I don't know what you did but it should be deprecated in the same way as `MatrixReshaped`: https://github.com/JuliaStats/Distributions.jl/blob/e94d6704386d1f753b28011067af03990303b14c/src/deprecates.jl#L55-L59 (just be an alias of `AffineDistribution` and call `AffineDistribution` instead)

src/univariate/affine.jl
devmotion: I disagree, I don't think it is useful. Nor should it be exported. A main reason is that `affine` is really a too common name and should not be owned by Distributions. Another reason is that it is much easier to implement specializations for certain distributions if one can implement the shifting and scaling separately.

src/univariate/affine.jl
ParadaCarleton: @devmotion won't the return values for `rand(Affine)` always be `promote(T, eltype(D))`? since we generate them as `shift + scale * rand(D)`, which will automatically promote

ParadaCarleton:
I agree on the array-like variates bit and actually started work on that yesterday.

I agree we should encourage and use `*`, `+`, `/`, and `-` over `affine`, but I think it makes sense to include `affine` for ease of implementation. This way anyone overloading `affine` only has to add 1 method instead of 4.

I'm not sure what benefits a split would give, and it might make type-checking and conversions more of a chore since `Scaled{Shifted}` would be different from `Shifted{Scaled}`. In particular we'd have to add more methods to collapse `Affine` down when performing repeated affine transformations.

src/univariate/affine.jl
devmotion: Currently, this is the case for `LocationScale` but this is not correct - the output of `rand` should not depend on the parameters in general. There's much more explanations in the PR.

devmotion:
> I agree we should encourage and use `*`, `+`, `/`, and `-` over `affine`, but I think it makes sense to include `affine` for ease of implementation. This way anyone overloading `affine` only has to add 1 method instead of 4.

There are default fallbacks for `/` and `-`, so generally one just has to define `x + distribution` and `x * distribution`. I think this is simple enough for developers and not more difficult than defining `affine`. Additionally, having both `affine` and `+`, ... seems confusing - if possible, we should not provide multiple ways but one recommended one for constructing affine transformations.

ParadaCarleton:
> > I agree we should encourage and use `*`, `+`, `/`, and `-` over `affine`, but I think it makes sense to include `affine` for ease of implementation. This way anyone overloading `affine` only has to add 1 method instead of 4.
> 
> There are default fallbacks for `/` and `-`, so generally one just has to define `x + distribution` and `x * distribution`. I think this is simple enough for developers and not more difficult than defining `affine`. Additionally, having both `affine` and `+`, ... seems confusing - if possible, we should not provide multiple ways but one recommended one for constructing affine transformations.

Makes sense, I think; I'll work on that.

ParadaCarleton:
Unrelated to the previous comments, we might want some field name other than `scale`, which seems to be a bit too easy to confuse with the `scale` function.

ParadaCarleton:
@devmotion it turns out that this is already implemented in MeasuresBase. Should we maybe just add MeasuresBase as a dependency?

devmotion:
No, this is not sufficient. We want to create and work with `Distribution`s, `*` etc. should not suddenly lift `Distribution`s to `Measure`s.

mschauer:
If we go with `+` and `-`, the docstring should mention ⊕ from #1445

devmotion:
We have to support `+`, `*` etc. anyway if we want to avoid a breaking change since it is the current API.

To me it seems an additional function would not have any advantage but only extend the API, which is possibly confusing for users, developers, and an additional source for breaking changes in the future. The main argument above for it, it seems, was that it would be easier to implement. Actually, I think this is not correct:
- As the PR for `Normal` etc. shows, one just has to implement two functions for univariate distributions currently
- The more modular design makes it much easier to add specialized implementations for specific types: With `affine` one would have to handle all combinations of the types for scale and shift whereas otherwise one can implement a method for eg a specific type of scales without worrying about types of shifts.

ParadaCarleton:
Got it; I guess I'll rewrite this to handle `+` and `*` directly (or perhaps just keep it an unexported implementation detail?).

ParadaCarleton:
Should we memoize `inv(σ)`, or perhaps precompute it, since there's a good chance it'll be frequently used?

src/matrix/lkj.jl
```diff
@@ -159,7 +159,7 @@ function _marginal(lkj::LKJ)
     d = lkj.d
     η = lkj.η
     α = η + 0.5d - 1
-    LocationScale(-1, 2, Beta(α, α))
+    return 2 * Beta(α, α) - 1
```
mschauer, line 162: Do you mind following the convention, that we "tend to space operators more tightly if they get applied before other nearby operators"
```suggestion
    return 2*Beta(α, α) - 1
```

devmotion:
> Should we memoize `inv(σ)` since there's a good chance it'll be frequently used?

No, I don't think it's worth it. In my experience it just causes problems, eg with AD and undesired promotions (e.g. if sigma is an `Int`, the cached inverse would be a `Float64` and cause promotions in computations with `Float32` that don't occur if we just divide by sigma).

ParadaCarleton:
> > Should we memoize `inv(σ)` since there's a good chance it'll be frequently used?
> 
> No, I don't think it's worth it. In my experience it just causes problems, eg with AD and undesired promotions (e.g. if sigma is an `Int`, the cached inverse would be a `Float64` and cause promotions in computations with `Float32` that don't occur if we just divide by sigma).

Hmm; could we memoize in cases where `base_dist` is continuous, so everything has to be converted to `eltype(base_dist)` (after promoting with the scale+shift params) anyways?

devmotion:
No, `eltype` should (and will) not match the parameters in general. If one wants to e.g. evaluate just the log density it is irrelevant what the `eltype` is - one should be able to provide e.g. a point in `Float32` and obtain a result that is not promoted to `Float64`.

devmotion:
I'm very certain memoization just causes problems here and is not worth it. As an example in a different package, similar issues occurred with memoization of inverses in PDMats and hence in the latest release it was removed.

mschauer:
The right approach is not to memoize but just to not constrain the type of the scale so if needed one can use a wrapper for the inverse scale (similar to a factorization object) as scale object

src/univariate/affine.jl
```diff
@@ -0,0 +1,222 @@
...
+
+```julia
+affine(shift, scale, base_dist)  # location-scale transformed distribution
+params(d)            # Get the parameters, i.e. (shift, scale, base_dist)
+location(d)          # Get the location parameter
+scale(d)             # Get the scale parameter
+```
+"""
+struct AffineDistribution{T<:Real, S<:ValueSupport, D<:UnivariateDistribution{S}} <: UnivariateDistribution{S}
+    shift::T
```
mschauer, line 39: Shift and scale should be different types ("think matrix, vector", but 1d). Why do we restrict to 1-d?

src/univariate/affine.jl
```diff
@@ -0,0 +1,222 @@
...
+    }
+        !check_args || @check_args(AffineDistribution, !iszero(scale))
+        return new{T,S,D}(shift, scale, base_dist)
+    end
+end
+
+
+"""
+    affine(shift, scale, base_dist)
+    affine(base_dist, shift, scale)
```
mschauer, line 53: What's the rationale, why not choose one order?

src/univariate/affine.jl
```diff
@@ -0,0 +1,222 @@
...
+
+#### Parameters
+
+function location(d::AffineDistribution)
+    if hasmethod(location, (typeof(d.base_dist),))
+        return d.shift + location(d.base_dist)
+    else
+        return d.shift
+    end
+end
```
mschauer, line 151:
```suggestion
location(d::AffineDistribution) = d.shift + location(d.base_dist)
```
Having a method error here is fine, something without a location doesn't get one by shifting it.

src/univariate/affine.jl
```diff
@@ -0,0 +1,222 @@
...
+function scale(d::AffineDistribution)
+    if hasmethod(scale, (typeof(d.base_dist),))
+        return d.scale * scale(d.base_dist)
+    else
+        return d.scale
+    end
```
mschauer, line 157: Similar here

src/univariate/affine.jl
```diff
@@ -0,0 +1,222 @@
...
+entropy(d::ContinuousAffine) = entropy(d.base_dist) + log(abs(d.scale))
+entropy(d::DiscreteAffine) = entropy(d.base_dist)
+
+mgf(d::AffineDistribution, t::Real) = exp(d.shift * t) * mgf(d.base_dist, d.scale * t)
+
+
+#### Evaluation & Sampling
+
+pdf(d::ContinuousAffine, x::Real) = pdf(d.base_dist, (x - d.shift) / d.scale) / abs(d.scale)
+pdf(d::DiscreteAffine, x::Real) = pdf(d.base_dist, (x - d.shift) / d.scale)
```
mschauer, line 188: Let's think if this is right for integer valued distributions:
```
julia> x = 1<<60 + 1
1152921504606846977

julia> (x*5)/5 == x
false

julia> (x*5)/5 ≈ x
true
```

devmotion: For integer-valued distributions one has to work with `Rational`s.

devmotion: As done eg in the tests.

mschauer: Ah, nice, let's add a remark to the docs then

.gitignore
```diff
@@ -2,4 +2,4 @@ docs/build
 docs/site
 .Rhistory
 Manifest.toml
-.vscode
+.vscode
```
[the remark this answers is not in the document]
ParadaCarleton: Oh weird, I have no idea why that happened -- I think one of my extensions must have fiddled with it. Actually, I can't even tell what's different -- was `.vscode` not there before?

src/univariate/affine.jl
```diff
@@ -0,0 +1,241 @@
...
+        μ::Tμ, σ::Tσ, base_dist::D; 
+        check_args=true
+        ) where {
+            F<:VariateForm,
+            S<:ValueSupport,
+            D<:Distribution{F, S},
+            Tμ<:ArrayLike,
+            Tσ<:ArrayLike
+        }
+        !check_args || @check_args(AffineDistribution, !iszero(σ))
```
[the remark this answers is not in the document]
ParadaCarleton: `iszero` returns zero for zero matrices --
```
julia> iszero(zeros((10, 10)))
true
```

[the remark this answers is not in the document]
ParadaCarleton: But I do think it makes sense to move this check to `*`.

src/univariate/affine.jl
```diff
@@ -0,0 +1,241 @@
...
+
+"""
+    *(scale_factor, base_dist)
+
+Return a version of `base_dist` that has been scaled by `scale`, where `scale` can be a real
+number or (for a multivariate distribution) matrix.
+"""
+function Base.:*(σ::ArrayLike, base_dist::Distribution)
+    return AffineDistribution(add_identity(σ), σ, base_dist)
+end
```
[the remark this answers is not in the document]
ParadaCarleton: Very confused by the second method defined here -- is this intended to work for a general `ArrayLikeVariate`, and not just multivariate distributions returning vectors? I can't make heads or tails of what's going on with the sizes.

src/univariate/affine.jl
```diff
@@ -0,0 +1,241 @@
...
+
+# Support
+function minimum(d::AffineDistribution)
+    minim = d.σ > 0 ? minimum : maximum
+    return d.μ + d.σ * minim(d.base_dist)
+end
+function maximum(d::AffineDistribution)
+    maxim = d.σ > 0 ? maximum : minimum
+    return d.μ + d.σ * maxim(d.base_dist)
+end
```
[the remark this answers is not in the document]
ParadaCarleton: Will do, but thinking about it, is `extrema` defined for multivariate distributions? If so, what is the return value?

src/univariate/affine.jl
```diff
@@ -0,0 +1,241 @@
...
+    return AffineDistribution(σ * aff_dist.μ, σ * aff_dist.σ, aff_dist.base_dist)
+end
+
+
+Base.:+(d::Distribution, μ::ArrayLike) = μ + d
+Base.:-(d::UnivariateDistribution) = -1 * d
+Base.:-(d::Distribution, μ::ArrayLike) = d + -μ
+Base.:-(μ::ArrayLike, d::Distribution) = μ + -d
+Base.:*(d::Distribution, σ::ArrayLike) = σ * d
+Base.:/(d::Distribution, τ::ArrayLike) = inv(τ) * d
```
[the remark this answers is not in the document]
ParadaCarleton: I see the problems with the other methods being too general, but shouldn't these be fine if I define the other methods more specifically (since these are just syntactic sugar reducing to the other methods)?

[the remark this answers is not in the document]
ParadaCarleton: Oh, perhaps `/` and `*` are too general, since matrix multiplication isn't commutative.

src/univariate/affine.jl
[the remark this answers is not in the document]
ParadaCarleton: (Also, the ability to use `σ::Real, base_dist::Distribution` for both multivariate and univariate distributions is intentional, since scalar multiplication is defined for all array sizes.)

.gitignore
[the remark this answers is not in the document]
devmotion: It seems the newline character at the end of the file was removed.

src/univariate/affine.jl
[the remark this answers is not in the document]
devmotion: I forgot to mention it here but yes, this only works for multi- and matrixvariate functions. While we might want to add scalar scales for arbitrary arrayvariate distributions we don't want to support arbitrary distributions and arbitrary arrays that are compatible with the distribution.

src/univariate/affine.jl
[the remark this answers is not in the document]
devmotion: Generally, `Distribution` is too general.

src/univariate/affine.jl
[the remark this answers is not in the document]
devmotion: Yes, it is, it returns `minimum, maximum`. And these are defined also for higher dimensional distributions, following the Julia definition of them the comparison is performed for one dimension after another. E.g., `minimum(d::Dirichlet) = vcat(zeros(length(d) - 1), 1)` and `maximum(d::Dirichlet) = vcat(1, zeros(length(d) - 1))`.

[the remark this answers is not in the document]
ParadaCarleton: So it returns a vector `[(min1, max1), (min2, max2), ...]`?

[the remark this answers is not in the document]
devmotion: No it returns `minimum(d), maximum(d)`. That's the default definition of `extrema(d)` in base.

ParadaCarleton:
@devmotion working through this, I think we need to document some kind of constraint on the results of `support`. In particular, `f.(support(x))` should return the support of `f(x)`. I can't think of another way to handle the support of arbitrary distributions.

ParadaCarleton:
@devmotion I think the PR is mostly done apart from --
1. Tests for the multivariate case (which I'm writing)
2. Whatever is wrong with the discrete CDF functions, which are failing the tests; if you know what I'm doing wrong I'm all ears.

devmotion (commented):
I added some initial comments, I didn't have time to go through the whole PR.

I think some things should be designed a bit differently. E.g., it is completely fine to translate `Distribution{<:ArrayLikeVariate}` of arbitrary dimension but a scale of `Diagonal(Ones(...))` or `I` only makes sense for `MultivariateDistribution`s and `MatrixDistribution`s. Therefore I think it would be better to split `AffineDistribution` in two different distributions, one that shifts and one that applies linear transformations.

Alternatively, we could use `nothing` or our own internal type to indicate that no linear transformation should be applied.

More generally, it seems have to change some functions to make sure that 1) we only define them for appropriate and matching types and 2) we do not accidentally introduce, or make it easy to run into, method ambiguities. If possible we should also ensure that this PR is non-breaking - e.g., by also introducing a type alias for `LocationScale` (currently only the constructor is - IMO slightly incorrectly - deprecated).

devmotion:
> Whatever is wrong with the discrete CDF functions, which are failing the tests; if you know what I'm doing wrong I'm all ears.

I don't know which error you face, so I can't really help here.

devmotion:
> I think some things should be designed a bit differently. E.g., it is completely fine to translate `Distribution{<:ArrayLikeVariate}` of arbitrary dimension but a scale of `Diagonal(Ones(...))` or `I` only makes sense for `MultivariateDistribution`s and `MatrixDistribution`s. Therefore I think it would be better to split `AffineDistribution` in two different distributions, one that shifts and one that applies linear transformations.

Maybe we could use the following design, which seems to address also your concerns about complicated dispatches and constructions of affine transformations above:
- Add a `ShiftedDistribution`/`TranslatedDistribution` that is the fallback for `x::Real + dist::UnivariateDistribution` and `x::AbstractArray{<:Real,N} + dist::Distribution{ArrayLikeVariate{N}}`, i.e., the distribution of `x + Z` where the law of `Z` is `dist`
- Add a `ScaledDistribution` (not sure about the name) that is the fallback for `x::Real * dist::Distribution{<:ArrayLikeVariate}` and `x::AbstractArray * dist::Distribution{<:ArrayLikeVariate}` (for `x` of the correct last dimension), i.e., the distribution of `x * Z` where the law of `Z` is `dist`
- Add an `AffineDistribution` that is the fallback for `x::Real * dist::ShiftedDistribution`, `x::AbstractArray * dist::ShiftedDistribution`, `x::Real + dist::ScaledDistribution{Univariate}`, and `x::AbstractArray{<:Real,N} + dist::ScaledDistribution{ArrayLikeVariate{N}}`

Then we could support all transformations of the form `a + Z`, `b * Z`, `a + b * Z` where `a` and `b` are of the correct sizes and we don't have to deal with neutral elements of addition and multiplication as in this PR. We could still dispatch on the fallback for affine transformations explicitly and there would be no problems with `ShiftedDistribution` of `ScaledDistribution` versus `ScaledDistribution` of `ShiftedDistribution`.

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+matrix.
+"""
+function Base.:*(σ::Real, ρ::UnivariateDistribution)
+    if iszero(σ)
+        throw(ArgumentError("scale must be non-zero"))
+    end
+    return AffineDistribution(zero(σ), σ, ρ)
+end
+
+function Base.:*(σ::Real, ρ::Distribution)
```
devmotion, line 80: We should restrict it at least to `Distribution{<:ArrayLikeVariate}`.

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+    if iszero(σ)
+        throw(ArgumentError("scale must be non-zero"))
+    end
+    return AffineDistribution(Zeros{eltype(ρ)}(size(ρ)), σ, ρ)
```
devmotion, line 84: I think we should be consistent and base the type on `σ`:
 ```suggestion
    return AffineDistribution(Zeros{typeof(σ)}(size(ρ)), σ, ρ)
```

File: src/univariate/affine.jl [hunk 22]
```diff
@@ -0,0 +1,332 @@
...
+
+function Base.:*(σ::AbstractMatrix{<:Real}, ρ::MultivariateDistribution)
+    if iszero(σ)
+        throw(ArgumentError("scale must be non-zero"))
+    elseif size(σ)[end] ≠ size(ρ)[1]
+        throw(DimensionMismatch(
+            "scaling matrix and distribution have incompatible dimensions" *
+            "$(size(σ)) and $(size(ρ)))"
+        ))
+    end
```
devmotion, line 95: It's not possible to check `iszero` here, I assume? The second check only makes sense if the type is defined more generally and allows to increase or decrease the number of dimensions.

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+    return AffineDistribution(Zeros{eltype(ρ)}(σ), σ, ρ)
+end
+Base.:*(σ::AbstractVector{<:Real}, ρ::MultivariateDistribution) = Diagonal(σ) * ρ
```
devmotion, line 98: I think this should not be defined since `σ * Z` is not defined either where the law of `Z` is `ρ`.
```suggestion
```

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+
+
+# Composing affine distributions
+
+function Base.:+(μ::Real, d::UnivariateAffine)
+    return AffineDistribution(μ + d.μ, d.σ, d.ρ)
+end
```
devmotion, line 105:
```suggestion
Base.:+(μ::Real, d::UnivariateAffine) = AffineDistribution(μ + d.μ, d.σ, d.ρ)
```

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+function Base.:+(μ::AbstractArray, d::MultivariateAffine)
+    return AffineDistribution(μ + d.μ, d.σ, d.ρ)
+end
```
devmotion, line 108: This is too general and has to check the size of `μ`.

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+
+
+function Base.:*(σ::Real, d::UnivariateAffine)
+    return AffineDistribution(σ * d.μ, σ * d.σ, d.ρ)
+end
```
devmotion, line 113:
```suggestion
Base.:*(σ::Real, d::UnivariateAffine) = AffineDistribution(σ * d.μ, σ * d.σ, d.ρ)
```

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+
+function Base.:*(σ::Real, d::MultivariateAffine)
+    return AffineDistribution(σ * d.μ, σ * d.σ, d.ρ)
+end
```
devmotion, line 117:
```suggestion
Base.:*(σ::Real, d::MultivariateAffine) = AffineDistribution(σ * d.μ, σ * d.σ, d.ρ)
```

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+
+function Base.:*(σ::AbstractMatrix, d::MultivariateAffine)
```
devmotion, line 119: Here one has to check the size.

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+    return AffineDistribution(σ * d.μ, σ * d.σ, d.ρ)
+end
+
+
+Base.:+(d::NonMatrixDistribution, μ::ArrayLike) = μ + d
+Base.:-(d::NonMatrixDistribution) = -1 * d
+Base.:-(d::NonMatrixDistribution, μ::ArrayLike) = -μ + d
+Base.:-(μ::ArrayLike, d::NonMatrixDistribution) = μ + -d
```
devmotion, line 127: This is too general.

devmotion: What is a `NonMatrixDistribution`?

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+
+Base.:*(d::NonMatrixDistribution, σ::Real) = σ * d
+Base.:/(d::NonMatrixDistribution, τ::Real) = Base.inv(τ) * d
+Base.:\(τ::ArrayLike, d::NonMatrixDistribution) = Base.inv(τ) * d
```
devmotion, line 131: Again the types should be changed. `inv` does not have to be qualified.

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+function extrema(d::UnivariateAffine)
+    extremes = extrema(d.ρ)
+    extremes = _flip(d.σ, extremes)
+    return @. d.μ + d.σ * extremes
+end
+
+
+## Multivariate
+# when finding the extrema, we need to flip the extremes if the scale is negative
+_flip(signed::Real, tuple::Tuple) = signed > 0 ? tuple : (last(tuple), first(tuple))
```
devmotion, line 157: We should only use broadcasting if necessary, it puts stress on the compiler that should be avoided when not needed.
```suggestion
function extrema(d::UnivariateAffine)
    μ, σ, ρ = params(d)
    min_ρ, max_ρ = extrema(ρ)
    return minmax(μ + σ * min_ρ, μ + σ * max_ρ)
end
```

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+
+function minimum(d::MultivariateAffine)
+    extremes = extrema(d.ρ) |> zip |> collect |> permutedims
+    # The minimum of d.ρ[i] will minimize d[j] if d.σ[i,j] is negative. 
+    # Otherwise the maximum will minimize d[j]
+    extreme_matrix = _flip.(d.σ, extremes)
+    argmins = first.(extreme_matrix)
+    mins = [σ ⋅ argmin for (σ, argmin) in zip(eachrow(d.σ), eachrow(argmins))]
+    return mins
+end
```
devmotion, line 167: This seems inefficient and incorrect - it seems it does not take into account the shift?

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+
+function extrema(d::MultivariateAffine)
+    extremes = extrema(d.ρ) |> zip |> collect |> permutedims
+    extreme_matrix = _flip.(d.σ, extremes)
+    argmins = first.(extreme_matrix)
+    mins = [σ ⋅ argmin for (σ, argmin) in zip(eachrow(d.σ), eachrow(argmins))]
+    argmaxes = last.(extreme_matrix)
+    maxes = [σ ⋅ argmax for (σ, argmax) in zip(eachrow(d.σ), eachrow(argmaxes))]
+    return (mins, maxes)
+end
```
devmotion, line 185: Same here.

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+
+support(d::AffineDistribution) = affine_support(d.μ, d.σ, support(d.ρ))
+affine_support(μ::Real, σ::Real, support) = μ .+ σ .* support
```
devmotion, line 188: This will return the support in the wrong order for discrete sets.

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+
+function affine_support(μ::Real, σ::Real, support::RealInterval) 
+    if σ > 0
+        lower = support.lb
+        upper = support.ub
+    else
+        lower = support.ub
+        upper = support.lb
+    end
+    return RealInterval(μ + σ * lower, μ + σ * upper)
```
devmotion, line 198: Simpler:
```suggestion
    lower, upper = extrema(support)
    return RealInterval(minmax(μ + σ * lower, μ + σ * upper)...)
```

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+
+function Base.eltype(::Type{<:AffineDistribution{F,S,Tμ,Tσ,D}}) where {
+        F<:VariateForm,
+        S<:ValueSupport,
+        D<:Distribution{F, S},
+        Tμ<:ArrayLike,
+        Tσ<:ArrayLike
+    }
+    T = Core.Compiler.return_type(Base.:*, (Tσ, eltype(D)))
+    return Core.Compiler.return_type(Base.:+, (Tμ, T))
```
devmotion, line 223: We should not use `return_type` (generally, usually one shouldn't). It is not guaranteed to return concrete types and e.g. may return `Any` at any point.

Moreover, the general opinion around `eltype` is that it should only be affected by the parameters if absolutely necessary. E.g., for continuous distributions it should not be affected by the parameters.

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+        return d.μ
+    end
+end
+function scale(d::AffineDistribution)
+    if hasmethod(scale, (typeof(d.ρ),))
+        return d.σ * scale(d.ρ)
+    else
+        return d.σ
+    end
+end
```
devmotion, line 242: As @mschauer said, this should just be
```suggestion
location(d::AffineDistribution) = d.μ + d.σ * location(d.ρ)
scale(d::AffineDistribution) = d.σ * scale(d.ρ)
```

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+shape(d::AffineDistribution) = shape(d.ρ)
+ncategories(d::AffineDistribution) = ncategories(d.ρ)
+params(d::AffineDistribution) = (d.μ, d.σ, d.ρ)
+partype(::AffineDistribution{F,S,Tμ,Tσ,D}) where {F,S,Tμ,Tσ,D} = (Tμ, Tσ)
```
devmotion, line 246: The convention is to return a single type (and in particular a type - not a tuple).

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+
+Base.sign(d::UnivariateAffine) = Base.sign(d.σ)
```
devmotion, line 248: IMO this method should be removed. The meaning of `sign(dist)` is unclear - or at least it's unclear (and somewhat a misuse of `Base.sign`) that it denotes the sign of the scale. I would have assumed rather that it refers to signed measures.

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+
+
+#### Statistics
+
+mean(d::AffineDistribution) = d.μ + d.σ * mean(d.ρ)
+median(d::AffineDistribution) = d.μ + d.σ * median(d.ρ)
+mode(d::AffineDistribution) = d.μ + d.σ * mode(d.ρ)
+modes(d::AffineDistribution) = d.μ .+ d.σ .* modes(d.ρ)
```
devmotion, line 256: This is wrong e.g. for `MultivariateDistribution`s.

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+
+var(d::UnivariateAffine) = d.σ^2 * var(d.ρ)
+std(d::UnivariateAffine) = abs(d.σ) * std(d.ρ)
+skewness(d::UnivariateAffine) = Base.sign(d) * skewness(d.ρ)
+kurtosis(d::UnivariateAffine) = kurtosis(d.ρ)
+
+cov(d::AffineDistribution) = d.σ * cov(d.ρ) * d.σ'
```
devmotion, line 263: Probably should be defined only for multivariate distributions, and we should have different implementations based on the type of the scale to reduce allocations.

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+    x = (x - d.μ) / d.σ
+    return d.σ < 0 ? logccdf(d.ρ, x) : logcdf(d.ρ, x)
+end 
+
+function quantile(d::UnivariateAffine, q::Real)
+    q = d.σ < 0 ? (1 - q) : q
+    return d.μ + d.σ * quantile(d.ρ, q)
+end
+
+rand(rng::AbstractRNG, d::AffineDistribution) = d.μ + d.σ * rand(rng, d.ρ)
```
devmotion, line 328: For higher-dimensional distributions we should define `_rand!`.

src/univariate/affine.jl
```diff
@@ -0,0 +1,332 @@
...
+
+function gradlogpdf(d::ContinuousAffine, x::Real)
```
devmotion, line 330: We should restrict it to univariate distributions.

devmotion (commented):
An additional suggestion: Just copy the existing methods for `LocationScale` for univariate `AffineDistribution`s. This makes it also less likely that we accidentally introduce bugs.

src/univariate/affine.jl
ParadaCarleton: You're right about the second check. The first check is possible -- it returns `true` if `σ` is the zero matrix.

src/univariate/affine.jl
ParadaCarleton: Isn't this a default method for e.g. Normal distributions, though?

Actually, maybe it would make sense to define `σ * ρ` as `σ .* ρ` when the dimensions line up? So we can allow scaling the marginals for arbitrary distributions.

src/univariate/affine.jl
ParadaCarleton: @devmotion wouldn't the size be checked automatically when performing the multiplication?

src/univariate/affine.jl
ParadaCarleton: I'm all ears if you have a more efficient way to do it (although you're right that I forgot about the shift).

src/univariate/affine.jl
ParadaCarleton: Why?

src/univariate/affine.jl
devmotion: > Isn't this a default method for e.g. Normal distributions, though?

No, there's no such method, there is only https://github.com/JuliaStats/Distributions.jl/blob/e94d6704386d1f753b28011067af03990303b14c/src/multivariate/mvnormal.jl#L295. Precisely for the reason that I mentioned in my comment: `rand(5) * rand(5)` is not defined but `rand(2, 5) * rand(5)` is.

> Actually, maybe it would make sense to define σ * ρ as σ .* ρ when the dimensions line up? 

There's a hard no from my side. In my experience implicit broadcasting usually is unintuitive, surprising for users, and difficult to handle for developers.

`Diagonal(rand(5)) * rand(5)` is clear and simple. As another example, we deprecated `MvNormal(::AbstractVector)` since the implicit convention `MvNormal(x::AbstractVector) = MvNormal(zeros(length(x)), Diagonal(x .^ 2))` is not clear and caused a lot of confusion among users - the vector could also denote eg. the mean. However, `MvNormal(Diagonal(rand(5)))` is less surprising - it returns a normal distribution with zero mean and the given diagonal covariance matrix. Similarly, I think it is clearer to provide (diagonal) matrices here instead of performing any implicit broadcasting.

src/univariate/affine.jl
ParadaCarleton: If something doesn't have a location parameter, isn't the shift considered the location parameter? For instance, for the T distribution, I'd consider an affinely transformed version to have a location parameter equal to 0.

(This also aligns with the current convention of `LocationScale`, which is to return the shift as the location parameter; removing this behavior would be a breaking change. That being said, I consider the current behavior of treating `LocationScale(0, 1, Normal(2, 1))` as having a location of 0 to be a bug.)

src/univariate/affine.jl
devmotion: The broadcast operations are not compatible, e.g.,
```julia
julia> rand(4) .+ rand(4, 3) .* [rand(3) for _ in 1:5]
ERROR: DimensionMismatch("arrays could not be broadcast to a common size; got a dimension with lengths 4 and 5")
Stacktrace:
  [1] _bcs1
    @ ./broadcast.jl:516 [inlined]
  [2] _bcs
    @ ./broadcast.jl:510 [inlined]
  [3] broadcast_shape
    @ ./broadcast.jl:504 [inlined]
  [4] combine_axes
    @ ./broadcast.jl:499 [inlined]
  [5] _axes
    @ ./broadcast.jl:224 [inlined]
  [6] axes
    @ ./broadcast.jl:222 [inlined]
  [7] combine_axes
    @ ./broadcast.jl:499 [inlined]
  [8] instantiate
    @ ./broadcast.jl:281 [inlined]
  [9] materialize(bc::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Nothing, typeof(+), Tuple{Vector{Float64}, Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Nothing, typeof(*), Tuple{Matrix{Float64}, Vector{Vector{Float64}}}}}})
    @ Base.Broadcast ./broadcast.jl:860
 [10] top-level scope
    @ REPL[1]:1

julia> rand(3) .+ rand() .* [rand(3) for _ in 1:5]
ERROR: DimensionMismatch("arrays could not be broadcast to a common size; got a dimension with lengths 3 and 5")
Stacktrace:
 [1] _bcs1
   @ ./broadcast.jl:516 [inlined]
 [2] _bcs
   @ ./broadcast.jl:510 [inlined]
 [3] broadcast_shape
   @ ./broadcast.jl:504 [inlined]
 [4] combine_axes
   @ ./broadcast.jl:499 [inlined]
 [5] instantiate
   @ ./broadcast.jl:281 [inlined]
 [6] materialize(bc::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Nothing, typeof(+), Tuple{Vector{Float64}, Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Nothing, typeof(*), Tuple{Float64, Vector{Vector{Float64}}}}}})
   @ Base.Broadcast ./broadcast.jl:860
 [7] top-level scope
   @ REPL[2]:1
```

src/univariate/affine.jl
ParadaCarleton: Makes sense. Removed this

src/univariate/affine.jl
ParadaCarleton: We could do that, but do we want to deviate from `Base`'s behavior (where `cov` returns the variance for univariate outcomes)?

src/univariate/affine.jl
ParadaCarleton: Hmm; would `map` work?

src/univariate/affine.jl
ParadaCarleton: `NonMatrix` is defined as either Univariate or Multivariate.

The generality shouldn't make a difference because all of these methods just hand off to another method, so they should throw the same error message and make the same checks if the combination isn't appropriate.

src/univariate/affine.jl
ParadaCarleton: image

src/univariate/affine.jl
ParadaCarleton: What's wrong with it?

src/univariate/affine.jl
ParadaCarleton: In cases where a distribution is made into a location-scale type family by affine transformation, the shift and scale values of the affine transformation are usually considered to be the corresponding location and scale parameters. For instance, the t distribution is usually defined as having one parameter, ν, but if I asked you for the location parameter of a Student t distribution, it would be reasonable to assume I was looking for the shift parameter of an affinely-transformed Student t distribution.

Regardless of whether we consider these to be "True" location and scale parameters, the important question is what the user is likely to want when they call `location` and `scale` on an affinely-shifted distribution. I suspect users getting an error message when they call `location(2*TDist(x) + 1)` will probably be frustrated or confused that this does not return `1` the same way that calling `location(2*Normal() + 1)` does.

In addition, removing `location` and `scale` functions for affine transformations would be a breaking change, as they are currently defined for `LocationScale`-type families.

src/univariate/affine.jl
ParadaCarleton: (Resolving so we can hold conversation in the other comment)

src/univariate/affine.jl
```diff
@@ -0,0 +1,343 @@
...
+"""
+function Base.:+(μ::Real, ρ::UnivariateDistribution)
+    return AffineDistribution(μ, one(μ), ρ)
+end
+
+function Base.:+(μ::AbstractArray{<:Real, N}, ρ::Distribution{ArrayLikeVariate{N}, S}) where {N, S}
+    if size(μ) ≠ size(dist) 
+        throw(DimensionMismatch("array and distribution have different sizes."))
+    end
+    return AffineDistribution(μ, Diagonal(Ones(μ)), ρ)
```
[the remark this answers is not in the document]
ParadaCarleton: It's possible, but `I` can be a pain to deal with because it's not an array. This can throw off code that assumes `scale(x)` returns an array, for instance.

src/univariate/affine.jl
ParadaCarleton: The main problem is that `eltype` actually is affected by the parameters, if `eltype(ρ)` and `typeof(σ)` are different, and it's not clear to me how else to deal with that.

src/univariate/affine.jl
ParadaCarleton: Not clear to me how this can be dealt with; I switched to using `promote` for now, but what should be done if these parameters have different types (which they often will)?

ParadaCarleton:
The bugs that are breaking `cdf` seem to be related to some very weird behavior for transformed discrete distributions, but every time I manage to fix one edge case, another pops up somewhere else. I'm starting to think we should maybe just not allow affine transformations of discrete variables, given I can't think of a use case for an affinely-transformed discrete variable. (With maybe the exceptions of something like `Binomial` or `BetaBinomial` rescaled to the 0, 1 range, but those could be handled on their own pretty easily.)

ParadaCarleton:
All tests now pass, except for `support` when called on normal distributions that have `ForwardDiff.Dual` numbers as arguments. I'm going to be completely honest and say I'm not sure what `support` is supposed to do there.

devmotion:
I hope you don't take this personal but I don't think we can merge the PR in its current state. In my opinion it is difficult to review and, in particular, it is difficult to assess if there are any breaking changes. Distributions is a central package and hence we have to be careful about not breaking downstream packages (since there are so many). To me it still seems there are too many changes and too general definitions, even for the univariate case currently handled by `LocationScale` (indicated by the test failures as well).

As a way forward, I think we should break up this PR (or maybe start from scratch) and
1. make an initial PR that only replaces `LocationScale` with `AffineDistribution` (with less restrictive type parameters), adds a deprecation warning for `LocationScale`, and adds type aliases for `LocationScale` - but without any further changes or generalizations e.g. to multivariate distributions: Tests should still pass and it would be easy to review if the files are not renamed;
2. make a separate PR that allows and tests negative scales for univariate distributions;
2. make a follow-up PR that adds implementations for multi- and possibly even more general arrayvariate distributions (but without changes of the univariate implementations);
3. make additional PRs that introduce `TranslatedDistribution` and `ScaledDistribution` to simplify and enable the handling of `+` and `*`, without changing the implementations for `AffineDistribution` but only changing dispatches of `*` and `+` and combining `ScaledDistribution` and `TranslatedDistribution` to `AffineDistribution`s.

ParadaCarleton:
> I hope you don't take this personal but I don't think we can merge the PR in its current state. In my opinion it is difficult to review and, in particular, it is difficult to assess if there are any breaking changes. Distributions is a central package and hence we have to be careful about not breaking downstream packages (since there are so many). To me it still seems there are too many changes and too general definitions, even for the univariate case currently handled by `LocationScale` (indicated by the test failures as well).
> 
> As a way forward, I think we should break up this PR (or maybe start from scratch) and
> 
> 1. make an initial PR that only replaces `LocationScale` with `AffineDistribution` (with less restrictive type parameters), adds a deprecation warning for `LocationScale`, and adds type aliases for `LocationScale` - but without any further changes or generalizations e.g. to multivariate distributions: Tests should still pass and it would be easy to review if the files are not renamed;
> 2. make a separate PR that allows and tests negative scales for univariate distributions;
> 3. make a follow-up PR that adds implementations for multi- and possibly even more general arrayvariate distributions (but without changes of the univariate implementations);
> 4. make additional PRs that introduce `TranslatedDistribution` and `ScaledDistribution` to simplify and enable the handling of `+` and `*`, without changing the implementations for `AffineDistribution` but only changing dispatches of `*` and `+` and combining `ScaledDistribution` and `TranslatedDistribution` to `AffineDistribution`s.

If you'd like, I can break this into two PRs. The first can add a `UnivariateAffine` which is just `LocationScale` with a possibly negative scale; this would replace `LocationScale`. (This one would be where all the bugs cropped up, and are pretty much all to do with forgetting to handle `P(x=X)` when dealing with CDFs of discrete distributions.) A separate PR can add a `MultivariateAffine` distribution.

ParadaCarleton:
I also would like to point out there have been essentially no changes to the behavior, of `LocationScale`, with the exception of:
1. Bug fixes for the `location` and `scale` methods, which are currently returning incorrect answers. (They always return `locationscale.μ` and `locationscale.σ`, regardless of the location and scale of the underlying distribution. For instance, `LocationScale(0, 1, Normal(2, 4))` currently returns `(0, 1)`.) 
2. The implementations of new constructors and deprecation of the old `LocationScale()`-style constructor, as you requested.

All bugs were related to dealing with negative transformations of discrete distributions, where I forgot to account for the probability that `x` is exactly equal to `X`. There may be bugs regarding multivariate distributions that I haven't caught yet, since I was still writing the tests, but this PR should not contain any breaking changes.

If you'd like to rewrite the code for multivariate affine transformations from scratch, work out a new interface to replace the old `LocationScale` interface from scratch, etc. that's completely reasonable. For now, I'll gladly make a separate PR implementing the code I had ready a week and a half ago. This code made no changes except removing the positivity checks for `LocationScale` and throwing up a couple absolute value bars, before I'd rewritten all the constructors from scratch, deprecated several the old constructors, and implemented several features beyond the original scope of this PR at your request.

devmotion:
> If you'd like, I can break this into two PRs. The first can add a UnivariateAffine which is just LocationScale with a possibly negative scale; this would replace LocationScale. A separate PR can add a MultivariateAffine distribution.

Just to be sure there's no misunderstanding: We should not have separate distributions but at most type aliases (maybe they re not needed though). As mentioned above, it would be preferable to allow negative scales only in a second PR to keep changes in the different PRs minimal and make them easier to review.

> I also would like to point out there have been essentially no changes to the behavior, of LocationScale, with the exception of:

Unfortunately, while the behaviour might not have changed (I don't know) the implementation has to a larger extent than what seemed necessary. During the different rounds of preliminary review I noticed and commented on methods that were modified slightly or generalized (sometimes incorrectly). Due to the renaming and other unrelated changes it is not possible for me to check with certainty if changes are needed, are bug fixes, are necessary or unnecessary generalizations, or are just style changes.

ParadaCarleton:
Ok; if you'd like to implement renaming and all the different changes in a different PR, go ahead, then. I'll make a single PR that only changes LocationScale to allow negative scale values, or else (if you don't like this idea) I'll add reflections directly to my package instead of spending more and more time making changes completely unrelated to this feature at your request.

devmotion:
Sure, it's seemingly simple changes and of course it could be done in fewer PRs - but Distributions has > 300 direct dependents and > 700 indirect ones and hence I won't merge any PRs, in particular no refactorings, for which I feel I can't review them properly and guarantee with reasonable confidence that they don't break anything. Therefore I suggest breaking it up into small PRs that are easy to review.

Issue #1444, opened 2021-11-30
How do I get `Normal(5,1) * 2 + Normal(5, 1)`?


fonsp:
I am trying to do some arithmetic with normally distributed iid variables, but I am not sure how to do that with Distributions.jl.

My goal is to compute
```julia
Normal(5,2) * 2 + Normal(5, 3)
```
and get the `Normal` object as result (the normal distribution is linear), but I can't figure out how to do this. Here is a notebook demonstrating my problem:

https://htmlview.glitch.me/?https://gist.github.com/fonsp/dea12fd7c81fdfab9ae9292f84881b3c

cscherrer:
> As discussed on Slack, `+` is not defined for sum of Distributions but one has to use `convolve`. I'll open a PR that would allow to use `\oplus` which was suggested and seems more convenient.

`\oplus` here will be _very_ nonstandard. As @ablaom has pointed out, `\otimes` and `\oplus` are generally preferred to be the category-theoretic product and sum, respectively.

devmotion:
This PR is really not the right place to discuss Unicode operators but I disagree. `\otimes` is a standard notation for tensor products (and used in this way e.g. in TensorCore and KernelFunctions), and the random variable (and also the density) of the product distribution of independent distributions is a tensor product of the random variables (and the densities) of these underlying distributions, respectively. Similarly, `\oplus` is a standard notation for direct sums, and the random variable corresponding to `convolve(d1, d2)` is a direct sum of the random variables corresponding to `d1` and `d2`.

Issue #142, opened 2013-09-16
Generic sum and product distributions


johnmyleswhite:
At a recent tutorial, someone brought up the following possibility for providing a nice syntax for location-scale distributions:

```
X = Normal(0, 1)
Y = 5X + 1
```

They also asked about adding two distributions together:

```
Z = Normal(1, 1) + Normal(2, 3)
```

I think we should provide a generic univariate `SumDistribution` and `ProductDistribution`, which we can special case for things like Gaussians. I am already working on the location-scale case and will open a PR today/tomorrow for it.

cscherrer:
Hi, just wondering if there's any update on this. I was thinking about how to implement a product distribution, but my intuition was that a product should be in the sense of product measure, so the dimensionalities add, maybe by way of a tuple.

The idea of adding or multiplying distributions as a convolution only works if you can guarantee independent of the components, and I've seen this trip people up a few times. Also, there are only a few [stable distributions](https://en.wikipedia.org/wiki/Stable_distribution); in most cases the sum or product can't be represented in a convenient closed form.

andreasnoack:
I'm afraid that is not the case. It is hard to know how well it could work in practice it would be great if you could do some experiments and report back.

cscherrer:
I'm getting caught up a little bit by the types, still getting used to some fine points of Julia. Rather than having everything be abstract, I was considering having a parameterized `CustomDistribution` type, with every method specified as a component of the defining `struct`. Seems like that could help with type stability. Then maybe a `@productDistribution` macro could assemble things for individual cases. I'll see how it goes.

cscherrer:
Hmm, it might be even simpler than this. It works just fine to build a product as a tuple:
```julia
entropy(d :: Tuple{Distribution, Distribution, Distribution}) =  sum(entropy.(d))
mean(d :: Tuple{Distribution, Distribution, Distribution}) = mean.(d)
```

Then we can write
```julia
Main> d = (Normal(), Uniform(0,1), Poisson(1.3))
(Distributions.Normal{Float64}(μ=0.0, σ=1.0), Distributions.Uniform{Float64}(a=0.0, b=1.0), Distributions.Poisson{Float64}(λ=1.3))

Main> mean(d)
(0.0, 0.5, 1.3)

Main> entropy(d)
2.8753527708838638
```

I'm not sure yet about overhead; maybe the types are too generic.

bhgomes:
[...]

Issue #1438, opened 2021-11-24
Proposal to remove "syntactic sugar" that overloads `*` and `/` for univariate distributions


ablaom:
I propose that the "syntactic sugar" added [here](https://github.com/JuliaStats/Distributions.jl/blob/135f7646eae6b3ff6f266bd520b2a8d4dff77bbc/src/univariate/locationscale.jl#L124) be removed in the next breaking release.

A probability distribution is just a special case of a measure. The transformations implied by the current implementations of `*`, `+`, `-` and `/` so forth do not generalise to arbitrary measures which, moreover, already have well-established meanings for these operations. Eg, for the product of a scalar $x$ with a measure $\mu$ we have $(x\mu)(A) = x \mu(A)$. And these definitions are useful also within probability; for example:

- When constructing a mixture of probability distributions, such as a finite average: The average is a prob measure even though the partial sums are just measures. That is, it's frequently convenient to leave the affine subspace of probability measures using the standard definitions of `+` and `*`. 

- When one wants to avoid normalising the measure representing a probability measure, because it is not needed (eg for generating samples)

My own use case is in ensemble models where I am averaging the probabilistic predictions of multiple classifiers. Computing averages naively is not working smoothly because the forementioned syntactic sugar conflicts with the "usual" ones, preventing me from just doing `mean(....)`.

Thoughts anyone?

@cscherrer

devmotion:
This was added in https://github.com/JuliaStats/Distributions.jl/pull/1217. Affine transformations were already defined for `MvNormal`, and the `LocationScale` definitions could be improved for different univariate distributions as well (https://github.com/JuliaStats/Distributions.jl/pull/1407).

At first glance, these transformations seem completely fine to me: they return the distribution of the transformed random variable. I guess the main concern here is that the use of `+`, `*`, etc. might indicate that the measures or densities are added, multiplied, etc.?

cscherrer:
It might help to consider... When we write `2 + Normal()` or `3 * Normal()`, what kind of objects are `2` and `3`? In the current Distributions notation, this seem to be implicitly converted to Dirac measures, so the operations can be interpreted as convolution. I'm not saying this is good or bad, but it does seem helpful to pin down the semantics.

In MeasureTheory, we've had some discussion of changing to `⊙` for weighted measures. See
https://github.com/cscherrer/MeasureTheory.jl/issues/170

[...]

PR #1445, opened 2021-11-30
Simplify `convolve` for `MvNormal`, add Unicode alias, and add documentation

devmotion:
This PR
- simplifies the implementation of `convolve(::MvNormal, ::MvNormal)`,
- adds a Unicode alias `⊕`, 
- adds documentation for `convolve` and `⊕`.

It seems `convolve` is difficult to discover currently, probably due to the missing documentation. Moreover, I (and apparently some users: https://github.com/JuliaStats/Distributions.jl/issues/1444) think the syntax could be a bit "nicer". The `⊕` was suggested and discussed as an alias for `convolve` before (e.g. https://github.com/JuliaStats/Distributions.jl/issues/142#issuecomment-506470526 and https://github.com/JuliaStats/Distributions.jl/pull/919#issuecomment-510064620), and I think it is a reasonable choice.

### DIFF
--- a/src/convolution.jl
+++ b/src/convolution.jl
@@ -61,27 +61,9 @@ function convolve(d1::Gamma, d2::Gamma)
 end
 
 # continuous multivariate
-# The first two methods exist for performance reasons to avoid unnecessarily converting
-# PDMats to a Matrix
-function convolve(
-    d1::Union{IsoNormal, ZeroMeanIsoNormal, DiagNormal, ZeroMeanDiagNormal},
-    d2::Union{IsoNormal, ZeroMeanIsoNormal, DiagNormal, ZeroMeanDiagNormal},
-    )
-    _check_convolution_shape(d1, d2)
-    return MvNormal(d1.μ .+ d2.μ, d1.Σ + d2.Σ)
-end
-
-function convolve(
-    d1::Union{FullNormal, ZeroMeanFullNormal},
-    d2::Union{FullNormal, ZeroMeanFullNormal},
-    )
-    _check_convolution_shape(d1, d2)
-    return MvNormal(d1.μ .+ d2.μ, d1.Σ.mat + d2.Σ.mat)
-end
-
 function convolve(d1::MvNormal, d2::MvNormal)
     _check_convolution_shape(d1, d2)
-    return MvNormal(d1.μ .+ d2.μ, Matrix(d1.Σ) + Matrix(d2.Σ))
+    return MvNormal(d1.μ + d2.μ, d1.Σ + d2.Σ)
 end
 
 
--- a/src/convolution.jl
+++ b/src/convolution.jl
@@ -1,25 +1,34 @@
 """
-    convolve(d1::T, d2::T) where T<:Distribution -> Distribution
+    convolve(d1::Distribution, d2::Distribution)
+    d1 ⊕ d2
 
-Convolve two distributions of the same type to yield the distribution corresponding to the
-sum of independent random variables drawn from the underlying distributions.
+Convolve two distributions and return the distribution corresponding to the sum of
+independent random variables drawn from the underlying distributions.
 
-The function is only defined in the cases where the convolution has a closed form as
-defined here https://en.wikipedia.org/wiki/List_of_convolutions_of_probability_distributions
+The Unicode operator `⊕` can be typed by `\\oplus<tab>`.
 
-* `Bernoulli`
-* `Binomial`
-* `NegativeBinomial`
-* `Geometric`
-* `Poisson`
-* `Normal`
-* `Cauchy`
-* `Chisq`
-* `Exponential`
-* `Gamma`
-* `MultivariateNormal`
+Currently, the function is only defined in cases where the convolution has a closed form.
+More precisely, the function is defined if the distributions of `d1` and `d2` are the same
+and one of
+* [`Bernoulli`](@ref)
+* [`Binomial`](@ref)
+* [`NegativeBinomial`](@ref)
+* [`Geometric`](@ref)
+* [`Poisson`](@ref)
+* [`Normal`](@ref)
+* [`Cauchy`](@ref)
+* [`Chisq`](@ref)
+* [`Exponential`](@ref)
+* [`Gamma`](@ref)
+* [`MvNormal`](@ref)
+
+External links: [List of convolutions of probability distributions on Wikipedia](https://en.wikipedia.org/wiki/List_of_convolutions_of_probability_distributions)
 """
-function convolve end
+convolve(::Distribution, ::Distribution)
+
+# define Unicode alias and add docstring
+⊕(d1::Distribution, d2::Distribution) = convolve(d1, d2)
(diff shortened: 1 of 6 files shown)


cscherrer:
FWIW in MeasureTheory we're trying to stick to categorical interpretation of `\otimes` and `\oplus`, so the latter will (I think) be a disjoint union. In any case, it seems worth avoiding to have `\otimes` mean one thing and `\oplus` something completely unrelated. We're moving toward `\otimes` for product measure, in part to avoid confusion with Distributions's implicit lifting of `*` and `+` for affine transformations.

I have no real stake in Distributions. But if you want things to be compatible, I hope you'll consider this.

devmotion:
I think there are probably different considerations in MeasureTheory/Base since distribution objects are viewed differently there. Distributions.jl does not work with measures but random variables when transforming distributions, as the package itself does not work with measures (see also https://github.com/JuliaStats/Distributions.jl/issues/1438). Hence I think the first question is: does the notation `\oplus` makes sense for the sum of independent random variables? In my opinion, it does, as I tried to explain in https://github.com/JuliaStats/Distributions.jl/issues/1444#issuecomment-983010605.

---

Side remarks unrelated to this PR:

This PR does not deal with `\otimes` and it can only be considered as alias for `product_distribution(::Tuple)` when the `ProductDistribution` PR is merged. In any case, to me it would seem consistent with `\oplus` in this PR and I don't think it would lead to any inconsistencies (see also https://github.com/JuliaStats/Distributions.jl/issues/1444#issuecomment-983010605).

> in part to avoid confusion with Distributions's implicit lifting of * and + for affine transformations.

As I said before, due to the interpretation of distributions as their corresponding random variables in these expressions there is no implicit lifting. Scalars are just scalars, no implicit Dirac measures.

mschauer:
Do we use oplus in MeasureTheory? I had oplus as convolution in https://github.com/mschauer/GaussianDistributions.jl/blob/c358ae9cd074e5d7a29a68271f0032dceabc432e/src/GaussianDistributions.jl#L106

cscherrer:
We've discussed ([here](https://github.com/cscherrer/MeasureTheory.jl/issues/170#issuecomment-963205784)) the possibility of using `⊗` and `⊕` to mean product measure and superposition, but @ablaom pointed out that it's useful for `⊕` to be the coproduct, which is probably disjoint union of measures (still need to confirm this).

Following this "categorical imperative" :wink: is my personal preference. But as a broader guideline, I'd suggest that whenever `*` and `+` are both defined on a type, they should follow the usual distributive law, and similarly for `⊗` and `⊕`. Violating this just seems to be asking for trouble.

I saw this because I recall discussion of `⊗` being used in Distributions to denote product measure. But I may be misremembering that.

src/convolution.jl
```diff
@@ -26,9 +26,8 @@ External links: [List of convolutions of probability distributions on Wikipedia]
 """
 convolve(::Distribution, ::Distribution)
 
-# define Unicode alias and add docstring
-⊕(d1::Distribution, d2::Distribution) = convolve(d1, d2)
-@doc (@doc convolve) :⊕
+# define Unicode alias
+const ⊕ = convolve
```
mschauer, line 30: That's better, yes.

devmotion: Not sure what I did yesterday, I guess I was too tired 😂😂

devmotion:
It's astonishing that every other PR in Distributions ends up with a general discussion of at least partially unrelated things :smile: Discussions are very important but I think ideally we should keep discussions focused and in the right place since otherwise I fear even small improvements will take a long time.

I don't think this PR is the right place to discuss things like e.g. `\otimes`. It has not even been discussed properly among maintainers of Distributions yet and additionally it is not even possible to define it properly currently, as I wrote above.

I still think some of the more general misunderstanding in the comments above is caused by the different design and interpretation of such operations in Distributions and MeasureTheory: MeasureTheory consistently deals with distributions as measures whereas Distributions does not use this measure-theoretic interpretation. Therefore e.g. in operations with `+` and `*` the distributions are viewed as their corresponding random variables. This is not my special interpretation, this is how the notation was motivated and designed: see e.g. https://github.com/JuliaStats/Distributions.jl/issues/307#issuecomment-429491374 and https://github.com/JuliaStats/Distributions.jl/issues/307#issuecomment-429488430

devmotion:
Test errors are unrelated, known, and caused by the RNG changes in Julia 1.7. Probably can be fixed (IMO ideally in a separate PR) by choosing a different seed or relaxing some tolerances.

cscherrer:
I've tried in good faith to contribute meaningfully to discussions in several repos, and you've repeatedly rebuked me that "this isn't relevant" or "this isn't the place", or "this is redundant". I'll try to stay out of your way, just let me know if you'd value my input at some point.

devmotion:
It's nothing personal from my side, and hence I tried to formulate it as a general comment and to resolve some possible misunderstandings. It's just a general observation that in my experience sometimes slows down development unnecessarily and makes it more difficult to read up on old discussions (if they are scattered in different PRs and issues). I have to blame myself sometimes, e.g., if I bring up additional suggestions or observations that are not needed - and hence not completely relevant - for the PR at hand :slightly_smiling_face:

ParadaCarleton:
I would intuitively expect `+` to be the convolution if we're working with the random variables; it feels a bit weird to have one symbol here and another for something like `AffineDistribution`.

devmotion:
As mentioned above, the general opinion was to not overload `+` for this use case.

> Concerns against + were e.g. that it might look like adding two distributions (#307 (comment)) and does not indicate whether the random variables are independent or not (#307 (comment)).

theogf:
I found this PR a bit randomly but here is another idea:
How about giving a useful error message instead of defining a new unicode operator.
So if someone does `Normal() + Normal()`, they get an error message like
```
Addition of distributions is ambiguous, if you meant to add two random variables use the `convolve` function
```

devmotion:
I like this suggestion 👍 However, it seems orthogonal to this PR, doesn't it? Such an informative error could be added regardless of whether a Unicode alias for `convolve` exists or not.

theogf:
That's true! Just thought this PR could be turned around since there is still no agreement.
Btw, looking quickly at [Wikipedia](https://en.wikipedia.org/wiki/Convolution#Definition), it looks like the correct symbol for convolution is '∗' (\ast)

devmotion:
> Btw, looking quickly at Wikipedia, it looks like the correct symbol for convolution is '∗' (\\ast)\n(I can definitely see how annoylingly similar to '*' it is though)

That's what I did in https://github.com/JuliaStats/Distributions.jl/pull/1455😅

ParadaCarleton:
To keep this issue from tearing our community apart, I propose a compromise: [circled ast](https://www.compart.com/en/unicode/U+229B) :p

(Could actually work, though? It's an asterisk, but the circle visually distinguishes it.)

ParadaCarleton:
Actually, it looks like there's already an alternate notation for convolution -- since it can be thought of as a tensor product:
image

PR #1448, opened 2021-12-05
Fix deprecations and test errors on Julia 1.7

devmotion:
This PR fixes deprecations in the tests and the test failures on Julia 1.7.

Since to me it seemed more reasonable to specify relative tolerances in the failing test (the order of magnitude of the resulting norms of matrices is unknown), I replaced the absolute tolerances with relative ones. I checked locally that smaller tolerances lead to (random) test failures.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -79,7 +79,6 @@ include("testutils.jl")
 
 for t in tests
     @testset "Test $t" begin
-        Random.seed!(345679)
         include("$t.jl")
     end
 end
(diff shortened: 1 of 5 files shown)


test/runtests.jl
```diff
@@ -79,7 +79,6 @@ include("testutils.jl")
 
 for t in tests
     @testset "Test $t" begin
-        Random.seed!(345679)
```
devmotion, line 82: The seed is already set above outside of `@testset` and hence it is not necessary to set it here. From the documentation of `@testset`:
```julia
  Before the execution of the body of a @testset, there is an implicit call to
  Random.seed!(seed) where seed is the current seed of the global RNG. Moreover, after the
  execution of the body, the state of the global RNG is restored to what it was before the
  @testset. This is meant to ease reproducibility in case of failure, and to allow seamless
  re-arrangements of @testsets regardless of their side-effect on the global RNG state.
```

devmotion:
I'll merge this PR since it blocks other PRs and does not affect the package itself but only the tests.

PR #1449, opened 2021-12-08
Change suffstats to take Type{<:MvNormal} instead of Type{MvNormal}

MaltheSR:
I was hit by the fact it's only possible to calculate `suffstats` for `Type{MvNormal}` rather than `Type{<:MvNormal}`. That is, currently,

```julia
x = randn(2, 10)
suffstats(MvNormal, x)
```

runs fine, while e.g.

```julia
suffstats(FullNormal, x)
```

errs with `ERROR: suffstats is not implemented for (FullNormal, Matrix{Float64}).`. This PR fixes this by simply changing `suffstats` to take a `Type{<:MvNormal}` instead, which should fix the problem. I believe this addresses the problem also brought up in #809, where the closing PR #884 seems to have missed `suffstats` for `MvNormal`.

I added a few tests also, I've only recently picked up Julia, however, so I wasn't sure about the most idiomatic way of testing this. I'm happy to make changes if this could be done better.

### DIFF
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -369,7 +369,7 @@ struct MvNormalStats <: SufficientStats
     tw::Float64         # total sample weight
 end
 
-function suffstats(D::Type{MvNormal}, x::AbstractMatrix{Float64})
+function suffstats(::Type{<:MvNormal}, x::AbstractMatrix{Float64})
     d = size(x, 1)
     n = size(x, 2)
     s = vec(sum(x, dims=2))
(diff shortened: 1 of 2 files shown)


devmotion (commented):
Thanks for the PR! It looks good to me, I'll trigger Github action once CI is fixed (there's an open PR waiting for review).

It seems it was decided to allow (not necessarily concrete) subtypes so this PR seems consistent. Personally though I'm not happy about this design choice since, as stated in the documentation, these types are not respected and hence seem misleading.

MaltheSR:
Thanks for the quick feedback.

As you say, this seemed desirable for consistency given decisions in #809 and #884. It also seems to me that something like this is required in order to write code generic over exponential family distributions. At least that was how I ran into this. As I said, however, I'm new to Julia, so perhaps I'm missing a workaround. I see your point that it seems unfortunate to mention e.g. `Float32` and getting `Float64` results, however.

devmotion:
>  It also seems to me that something like this is required in order to write code generic over exponential family distributions. At least that was how I ran into this.

I'm not sure what problem you ran into exactly. Do you have a minimal working example that you can share?

PR #1451, opened 2021-12-08
Make method ambiguities due to `pdf`, `logpdf`, and `rand!` less likely

devmotion:
@oxinabox reported on Slack that https://github.com/JuliaStats/Distributions.jl/pull/1437 caused method ambiguity errors in an internal package. This PR reduces the amount of default fallback implementations and hence should make method ambiguity errors due to `pdf`, `logpdf` and `rand!` less likely.

Of course, I don't know if it would solve the issues that @oxinabox had but the PR would fix e.g. issues with implementations of `pdf(::MyMvNormal, x::AbstractVecOrMat)`, `logpdf(::MyMvNormal, x::AbstractVecOrMat)` or `rand!(::AbstractRNG, ::MyMvNormal, x::AbstractVecOrMat)` for custom multivariate normal distribution of type `MyMvNormal`. With the master branch, calling `pdf(::MyMvNormal, x::AbstractVector)`, `logpdf(::MyMvNormal, x::AbstractVector)` or `rand!(::MyMvNormal, x::AbstractVector)` would cause a method ambiguity error since the default fallback is less specific in the first argument but more specific in the second argument. With this PR, the issue would not occur since the default fallback is less specific in both arguments.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.25.34"
+version = "0.25.35"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/common.jl
+++ b/src/common.jl
@@ -170,12 +170,19 @@ Return the minimum and maximum of the support of `d` as a 2-tuple.
 Base.extrema(d::Distribution) = minimum(d), maximum(d)
 
 """
-    pdf(d::Distribution{ArrayLikeVariate{N}}, x::AbstractArray{<:Real,N}) where {N}
+    pdf(d::Distribution{ArrayLikeVariate{N}}, x) where {N}
 
-Evaluate the probability density function of `d` at `x`.
+Evaluate the probability density function of distribution `d` at `x`.
 
 This function checks if the size of `x` is compatible with distribution `d`. This check can
 be disabled by using `@inbounds`.
+
+Here, `x` can be
+- a single variate of type `AbstractArray{<:Real,N}` with `size(x) = size(d)`,
+- a collection of variates of type `AbstractArray{<:Real,M}` with dimension `M > N` and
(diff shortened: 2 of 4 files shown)


oxinabox:
Here are the errors we had


```
MethodError: logpdf(::Gaussian{Vector{Float64}, Matrix{Float64}}, ::Vector{Float64}) is ambiguous. Candidates:
    logpdf(dist::Union{Gaussian, Node{var"#s92"} where var"#s92"<:Gaussian}, x::Union{AbstractArray, Node{var"#s91"} where var"#s91"<:AbstractArray}) in GPForecasting at /builds/invenia/GPForecasting.jl/src/pdf.jl:36
    logpdf(d::Distribution{ArrayLikeVariate{N}, S} where S<:ValueSupport, x::AbstractArray{var"#s11", M} where var"#s11"<:Real) where {N, M} in Distributions at /root/.julia/packages/Distributions/fBsc5/src/common.jl:282
  Possible fix, define
    logpdf(::Gaussian, ::AbstractArray{var"#s11", M} where var"#s11"<:Real) where M
```
```
MethodError: logpdf(::Gaussian{Zeros{Float64, 1, Tuple{Base.OneTo{Int64}}}, Matrix{Float64}}, ::Vector{Float64}) is ambiguous. Candidates:
    logpdf(dist::Union{Gaussian, Node{var"#s92"} where var"#s92"<:Gaussian}, x::Union{AbstractArray, Node{var"#s91"} where var"#s91"<:AbstractArray}) in GPForecasting at /builds/invenia/GPForecasting.jl/src/pdf.jl:36
    logpdf(d::Distribution{ArrayLikeVariate{N}, S} where S<:ValueSupport, x::AbstractArray{var"#s11", M} where var"#s11"<:Real) where {N, M} in Distributions at /root/.julia/packages/Distributions/fBsc5/src/common.jl:282
  Possible fix, define
    logpdf(::Gaussian, ::AbstractArray{var"#s11", M} where var"#s11"<:Real) where M
```

devmotion:
Hmm these errors wouldn't be fixed by the PR unfortunately. I guess we can hold back this PR until some specific use case pops up that would be solved by these changes.

PR #1453, opened 2021-12-09
Rename LocationScale

ParadaCarleton:
Renames `LocationScale` to `Affine` and aliases


devmotion:
Unfortunately, it's still impossible to see what you changed. Can you please only add the type and alias but in particular not rename any files?

ParadaCarleton:
@devmotion I've reverted the name change for the files. The only things that are different are: 
1. The docstring
2. image
3. These two lines
image

ParadaCarleton:
@mschauer Do you think you could review this so I can get the PRs adding other features ready?

src/Distributions.jl
```diff
@@ -65,6 +65,7 @@ export
     MultivariateMixture,
 
     # distribution types
+    AffineDistribution,
```
devmotion, line 68: I suggest not exporting `AffineDistribution`. It is not intended to be used by users directly, so exporting it should not be necessary. Moreover, non-exported types can be changed without worrying about making breaking changes.
```suggestion
```

src/univariate/locationscale.jl
```diff
@@ -1,129 +1,131 @@
...
-scale(d)             # Get the scale parameter
+d = σ * ρ + μ       # Create location-scale transformed distribution
+params(d)           # Get the parameters, i.e. (μ, σ, ρ)
 ```
-
-External links
-[Location-Scale family on Wikipedia](https://en.wikipedia.org/wiki/Location%E2%80%93scale_family)
 """
-struct LocationScale{T<:Real, S<:ValueSupport, D<:UnivariateDistribution{S}} <: UnivariateDistribution{S}
+struct AffineDistribution{T<:Real, S<:ValueSupport, D<:UnivariateDistribution{S}} <: UnivariateDistribution{S}
```
devmotion, line 36: We will want to allow different types for `mu` and `sigma`. If we don't export `AffineDistribution`, we are allowed to make these changes in one of the follow-up PRs though.

src/univariate/locationscale.jl
```diff
@@ -1,129 +1,131 @@
...
-Base.eltype(::Type{<:LocationScale{T}}) where T = T
+Base.eltype(::Type{<:AffineDistribution{T}}) where T = T
 
-minimum(d::LocationScale) = d.μ + d.σ * minimum(d.ρ)
-maximum(d::LocationScale) = d.μ + d.σ * maximum(d.ρ)
-support(d::LocationScale) = locationscale_support(d.μ, d.σ, support(d.ρ))
-function locationscale_support(μ::Real, σ::Real, support::RealInterval)
+minimum(d::AffineDistribution) = d.μ + d.σ * minimum(d.ρ)
+maximum(d::AffineDistribution) = d.μ + d.σ * maximum(d.ρ)
+support(d::AffineDistribution) = AffineDistribution_support(d.μ, d.σ, support(d.ρ))
```
devmotion, line 65: Can we follow the Julia convention that upper case names are used for types, constructors, and modules only?
```suggestion
support(d::AffineDistribution) = affinedistribution_support(d.μ, d.σ, support(d.ρ))
```

src/univariate/locationscale.jl
```diff
@@ -1,129 +1,131 @@
...
+function AffineDistribution_support(μ::Real, σ::Real, support::RealInterval)
```
devmotion, line 66:
```suggestion
function affinedistribution_support(μ::Real, σ::Real, support::RealInterval)
```

src/univariate/locationscale.jl
```diff
@@ -1,129 +1,131 @@
...
     return RealInterval(μ + σ * support.lb, μ + σ * support.ub)
 end
-locationscale_support(μ::Real, σ::Real, support) = μ .+ σ .* support
+AffineDistribution_support(μ::Real, σ::Real, support) = μ .+ σ .* support
```
devmotion, line 69:
```suggestion
affinedistribution_support(μ::Real, σ::Real, support) = μ .+ σ .* support
```

src/univariate/locationscale.jl
```diff
@@ -1,129 +1,131 @@
...
 
-LocationScale(μ::Real, σ::Real, d::LocationScale) = LocationScale(μ + d.μ * σ, σ * d.σ, d.ρ)
+AffineDistribution(μ::Real, σ::Real, d::AffineDistribution) = AffineDistribution(μ + d.μ * σ, σ * d.σ, d.ρ)
```
devmotion, line 71:
```suggestion
affinedistribution_support(μ::Real, σ::Real, d::AffineDistribution) = AffineDistribution(μ + d.μ * σ, σ * d.σ, d.ρ)
```

devmotion (commented):
Just some minor comments, looks good overall! Thank you for splitting up the changes and keeping the same file names for now, it made the review much easier :slightly_smiling_face:

src/univariate/locationscale.jl
ParadaCarleton: I'll make it in the next one so we can get this one merged :+1:

src/univariate/locationscale.jl
ParadaCarleton: Yes! Sorry about that, I should have used case-sensitive search when I was replacing the names of everything.

test/locationscale.jl
```diff
@@ -2,11 +2,11 @@ function test_location_scale(
     rng::Union{AbstractRNG, Missing},
     μ::Real, σ::Real, ρ::UnivariateDistribution, dref::UnivariateDistribution,
 )
-    d = LocationScale(μ,σ,ρ)
+    d = μ + σ * ρ
```
devmotion, line 5: This should be

```suggestion
    d = Distributions.AffineDistribution(μ, σ, ρ)
```

The main point is to use the constructor here and compare it with the constructions based on `+` and `*` below.

devmotion (changes requested):

src/univariate/locationscale.jl
```diff
@@ -1,129 +1,131 @@
...
+maximum(d::AffineDistribution) = d.μ + d.σ * maximum(d.ρ)
+support(d::AffineDistribution) = affinedistribution_support(d.μ, d.σ, support(d.ρ))
+function affinedistribution_support(μ::Real, σ::Real, support::RealInterval)
     return RealInterval(μ + σ * support.lb, μ + σ * support.ub)
 end
-locationscale_support(μ::Real, σ::Real, support) = μ .+ σ .* support
+affinedistribution_support(μ::Real, σ::Real, support) = μ .+ σ .* support
 
-LocationScale(μ::Real, σ::Real, d::LocationScale) = LocationScale(μ + d.μ * σ, σ * d.σ, d.ρ)
+affinedistribution_support(μ::Real, σ::Real, d::AffineDistribution) = AffineDistribution(μ + d.μ * σ, σ * d.σ, d.ρ)
```
devmotion, line 71:
```suggestion
AffineDistribution(μ::Real, σ::Real, d::AffineDistribution) = AffineDistribution(μ + d.μ * σ, σ * d.σ, d.ρ)
```

mschauer:
There is a warning left,
```
┌ Warning: `LocationScale(args...; kwargs...)` is deprecated, use `AffineDistribution(args...; kwargs...)` instead.
540
│   caller = (::getfield(Main, Symbol("##239#240")))(::Type) at univariate_bounds.jl:60
541
└ @ Main ~/work/Distributions.jl/Distributions.jl/test/univariate_bounds.jl:60
542
┌ Warning: `LocationScale(args...; kwargs...)` is deprecated, use `AffineDistribution(args...; kwargs...)` instead.
543
│   caller = #LocationScale#69(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Type) at deprecated.jl:55
544
└ @ Distributions ./deprecated.jl:55
```
Could not figure out right away where it comes from

ParadaCarleton:
> There is a warning left,
> 
> ```
> ┌ Warning: `LocationScale(args...; kwargs...)` is deprecated, use `AffineDistribution(args...; kwargs...)` instead.
> 540
> │   caller = (::getfield(Main, Symbol("##239#240")))(::Type) at univariate_bounds.jl:60
> 541
> └ @ Main ~/work/Distributions.jl/Distributions.jl/test/univariate_bounds.jl:60
> 542
> ┌ Warning: `LocationScale(args...; kwargs...)` is deprecated, use `AffineDistribution(args...; kwargs...)` instead.
> 543
> │   caller = #LocationScale#69(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Type) at deprecated.jl:55
> 544
> └ @ Distributions ./deprecated.jl:55
> ```
> 
> Could not figure out right away where it comes from

Weird. Running locally I don't see any deprecation warnings:

```
> There is a warning left,
> 
> ```
> ┌ Warning: `LocationScale(args...; kwargs...)` is deprecated, use `AffineDistribution(args...; kwargs...)` instead.
> 540
> │   caller = (::getfield(Main, Symbol("##239#240")))(::Type) at univariate_bounds.jl:60
> 541
> └ @ Main ~/work/Distributions.jl/Distributions.jl/test/univariate_bounds.jl:60
> 542
> ┌ Warning: `LocationScale(args...; kwargs...)` is deprecated, use `AffineDistribution(args...; kwargs...)` instead.
> 543
> │   caller = #LocationScale#69(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Type) at deprecated.jl:55
> 544
> └ @ Distributions ./deprecated.jl:55
> ```
> 
> Could not figure out right away where it comes from

Test Summary:                                                 | Pass  Fail  Broken  Total
Test matrixvariates                                           |  559     3       1    563
  matrixvariates                                              |  559     3       1    563
    MatrixNormal                                              |   89                   89
    Wishart                                                   |  111     1            112
      Testing matrix-variates with rand(...)                  |    2                    2
      Testing matrix-variates with rand(rng, ...)             |    2                    2
      Testing matrix-variates with rand!(rng, ..., false)     |    4                    4
      Testing matrix-variates with rand!(..., false)          |    4                    4
      Testing matrix-variates with rand!(..., true)           |    2                    2
      Testing matrix-variates with rand!(rng, true)           |    2                    2
      deprecations                                            |    4                    4
      meanlogdet                                              |    1                    1
      H ~ W(ν, Σ), a ~ p(a) independent ⟹ a'Ha / a'Σa ~ χ²(ν) |    1                    1
      H ~ W(ν, I) ⟹ H[i, i] ~ χ²(ν)                           |    3                    3
      Check Singular Branch                                   |   34     1             35
    InverseWishart                                            |   68                   68
    MatrixTDist                                               |   85                   85
    MatrixBeta                                                |   56                   56
    MatrixFDist                                               |   53     2       1     56
      Testing matrix-variates with rand(...)                  |    1     1              2
      Testing matrix-variates with rand(rng, ...)             |    1     1              2
      Testing matrix-variates with rand!(rng, ..., false)     |    4                    4
      Testing matrix-variates with rand!(..., false)          |    4                    4
      Testing matrix-variates with rand!(..., true)           |    2                    2
      Testing matrix-variates with rand!(rng, true)           |    2                    2
    LKJ                                                       |   97                   97
ERROR: LoadError: Some tests did not pass: 559 passed, 3 failed, 0 errored, 1 broken.
in expression starting at /home/lime/.julia/dev/Distributions.jl/test/runtests.jl:80
ERROR: Package Distributions errored during testing
```
(From what I can tell the errors are related to Julia replacing its default RNG, so random variates now fail some tests.)

ParadaCarleton:
> > apart from one unrelated to this PR:
> 
> Weird, these should not occur anymore and be removed by my recent PR that fixed test errors and deprecations.

Hmm, maybe I misunderstood the error. Could you try running the tests again?

devmotion:
I don't think this will change anything, they ran basically forever and timed out. I assume the deprecation warnings are caused by the incorrect deprecation of `LocationScale`: Since the deprecation adds a `LocationScale()` method `LocationScale` is not removed from `dists` in https://github.com/ParadaCarleton/Distributions.jl/blob/f31b368a383944ff7357ac195b1730ec2421336d/test/univariate_bounds.jl#L59. Can you fix the deprecation of the constructor? We should only define `LocationScale(mu::Real, sigma::Real, dist::UnivariateDistribution; check_args::Bool=true)`.

ParadaCarleton:
Tests pass! :tada:

devmotion:
Not on Julia 1.0 unfortunately, they don't succeed and are canceled after 6 hours. It seems related to this PR since the last line in the logs is a deprecation warning of LocationScale: https://github.com/JuliaStats/Distributions.jl/runs/4527507045?check_suite_focus=true#step:6:524

ParadaCarleton:
> Not on Julia 1.0 unfortunately, they don't succeed and are canceled after 6 hours. It seems related to this PR since the last line in the logs is a deprecation warning of LocationScale: https://github.com/JuliaStats/Distributions.jl/runs/4527507045?check_suite_focus=true#step:6:524

Do we still intend to support 1.0 now that 1.6 is the LTS?

mschauer:
In any case, there shouldn't be a deprecation warning of `LocationScale` in the test

devmotion:
IMO there's no reason to drop support for old Julia versions (and 1.0 in particular) as long as they work/can be made to work without significant additional effort. And they work on the master branch, so tests with Julia 1.0 should pass with this PR as well since it only adds deprecations and renames stuff 🙂

src/univariate/locationscale.jl
```diff
@@ -1,129 +1,138 @@
...
+    return AffineDistribution{_T,S,D}(_T(μ), _T(σ), ρ; check_args=check_args)
 end
 
-LocationScale(μ::Real, σ::Real, ρ::UnivariateDistribution) = LocationScale(promote(μ, σ)..., ρ)
+AffineDistribution(μ::Real, σ::Real, ρ::UnivariateDistribution) = AffineDistribution(promote(μ, σ)..., ρ)
 
 # aliases
-const ContinuousLocationScale{T<:Real,D<:ContinuousUnivariateDistribution} = LocationScale{T,Continuous,D}
-const DiscreteLocationScale{T<:Real,D<:DiscreteUnivariateDistribution} = LocationScale{T,Discrete,D}
+const LocationScale = AffineDistribution
```
mschauer, line 56:
```suggestion
```

devmotion: We need the type alias since otherwise downstream packages that dispatch on or more generally work with `LocationScale` (it is exported) will break.

mschauer: Yes, but then we cannot `depwarn` in the constructor?

src/univariate/locationscale.jl
```diff
@@ -1,129 +1,138 @@
...
+function LocationScale(μ::Real, σ::Real, ρ::UnivariateDistribution; check_args::Bool=true)
```
mschauer, line 57: This signature does not differ from
```
function AffineDistribution(μ::T, σ::T, ρ::UnivariateDistribution; check_args=true) where {T<:Real}
```
as ```LocationScale===AffineDistribution`

mschauer:
```
julia> f(x::Real) = 2

julia> f(1)
2

julia> f(x::T) where T<: Real = 1

julia> f(1)
1

julia> f(x::Real) = 2

julia> f(1)
2
```

mschauer: Or rather, now I understand:: you only depwarn for non-diagonal argument types and then...

src/univariate/locationscale.jl
devmotion: True, it seems it's not possible in this way at least.

mschauer: Then we keep the alias and remove the alias only during a later iteration?

ParadaCarleton: > Then we keep the alias and remove the alias only during a later iteration?

That feels a bit like it's just kicking the can down the road.

devmotion:
Now you broke the downstream tests it seems :smile:

src/univariate/locationscale.jl
```diff
@@ -1,129 +1,138 @@
...
+function AffineDistribution(μ::T, σ::T, ρ::UnivariateDistribution; check_args=true) where {T<:Real}
     _T = promote_type(eltype(ρ), T)
     D = typeof(ρ)
     S = value_support(D)
-    return LocationScale{_T,S,D}(_T(μ), _T(σ), ρ; check_args=check_args)
+    return AffineDistribution{_T,S,D}(_T(μ), _T(σ), ρ; check_args=check_args)
 end
 
-LocationScale(μ::Real, σ::Real, ρ::UnivariateDistribution) = LocationScale(promote(μ, σ)..., ρ)
+AffineDistribution(μ::Real, σ::Real, ρ::UnivariateDistribution) = AffineDistribution(promote(μ, σ)..., ρ)
```
devmotion, line 53: Maybe this fixes the downstream tests (should be fixed anyway):
```suggestion
function AffineDistribution(μ::Real, σ::Real, ρ::UnivariateDistribution; check_args=true)
    return AffineDistribution(promote(μ, σ)..., ρ; check_args=check_args)
end
```

src/univariate/locationscale.jl
```diff
@@ -1,129 +1,138 @@
...
-    function LocationScale{T,S,D}(μ::T, σ::T, ρ::D; check_args=true) where {T<:Real, S<:ValueSupport, D<:UnivariateDistribution{S}}
-        check_args && @check_args(LocationScale, σ > zero(σ))
+    function AffineDistribution{T,S,D}(μ::T, σ::T, ρ::D; check_args=true) where {T<:Real, S<:ValueSupport, D<:UnivariateDistribution{S}}
+        check_args && @check_args(AffineDistribution, σ > zero(σ))
         new{T, S, D}(μ, σ, ρ)
     end
 end
 
-function LocationScale(μ::T, σ::T, ρ::UnivariateDistribution; check_args=true) where {T<:Real}
+function AffineDistribution(μ::T, σ::T, ρ::UnivariateDistribution; check_args=true) where {T<:Real}
```
devmotion, line 46:
```suggestion
function AffineDistribution(μ::T, σ::T, ρ::UnivariateDistribution; check_args::Bool=true) where {T<:Real}
```

mschauer:
Green lights!

devmotion:
@ParadaCarleton did you build the documentation locally? I assume it displays the documentation for `AffineDistribution` now?

ParadaCarleton:
I did not; how would I do that?

ParadaCarleton:
Ahh, ok, managed to get them built locally; `AffineDistribution` doesn't show up, presumably because it's unexported. What would I have to do to add it to the docs?

devmotion:
That's good actually, since it's not exported and the design is not fixed yet I think it's preferable if it doesn't show up in the docs. But `LocationScale` is still there since we did not remove https://github.com/JuliaStats/Distributions.jl/blob/7670bed8b155d351ad518b28dd345aac1feeec36/docs/src/univariate.md?plain=1#L300-L307?

ParadaCarleton:
> That's good actually, since it's not exported and the design is not fixed yet I think it's preferable if it doesn't show up in the docs. But `LocationScale` is still there since we did not remove https://github.com/JuliaStats/Distributions.jl/blob/7670bed8b155d351ad518b28dd345aac1feeec36/docs/src/univariate.md?plain=1#L300-L307?

Ahh, yes, it keeps the `LocationScale` docstring.

mschauer:
I think it is still wrong. We need to test
```
@test AffineDistribution(1.0, 1, Normal()) isa AffineDistribution{Float64, Continuous, Normal{Float64}}
@test_nowarn AffineDistribution(1.0, 1, Normal())

@test LocationScale(1.0, 1, Normal()) isa LocationScale{Float64, Continuous, Normal{Float64}}
@test_warn "" LocationScale(1.0, 1, Normal())
```

devmotion:
> I think it is still wrong. We need to test
> ```
> @test AffineDistribution(1.0, 1, Normal()) isa AffineDistribution{Float64, Continuous, Normal{Float64}}
> @test_nowarn AffineDistribution(1.0, 1, Normal())
> 
> @test LocationScale(1.0, 1, Normal()) isa LocationScale{Float64, Continuous, Normal{Float64}}
> @test_warn "" LocationScale(1.0, 1, Normal())
> ```

I think this can be simplified to eg.
```julia
@test (@test_deprecated LocationScale(1.0, 1, Normal())) isa LocationScale{Float64, Continuous, Normal{Float64}}
```
Is the other case not covered by the tests yet?

mschauer:
I don't think we test
```
@test_nowarn AffineDistribution(1.0, 1, Normal())
```

devmotion:
It would be fine maybe if we would call the constructor at some point (we checked that there are no deprecation warnings anymore so an explicit test would be a bit redundant) - but I'm not sure if we use arguments of different types in the tests.

mschauer:
That is exactly what I am after

ParadaCarleton:
Added tests

test/locationscale.jl
```diff
@@ -173,4 +173,13 @@ end
         test_location_scale_discretenonparametric(_rng, -1//4, 1//3, (-10):(-1), probs)
         test_location_scale_discretenonparametric(_rng, 6//5, 3//2, 15:24, probs)
     end
+
+    @test_nowarn ad_norm = AffineDistribution(1.0, 1, Normal())
+    @test ad_norm isa AffineDistribution{Float64, Continuous, Normal{Float64}}
+
+    @test_warn ls_norm = LocationScale(1.0, 1, Normal())
+    @test ls_norm isa LocationScale{Float64, Continuous, Normal{Float64}}
+    @test ls_norm isa AffineDistribution{Float64, Continuous, Normal{Float64}}
```
devmotion, line 182: These tests fail, `@test_warn` doesn't exist 🙃 I suggest using my suggestion above:
```suggestion
    @test (@test_deprecated LocationScale(1.0, 1, Normal())) isa LocationScale{Float64, Continuous, Normal{Float64}}
```

devmotion (changes requested):

test/locationscale.jl
mschauer: Ah, test_warn came later than Julia 1.0

devmotion: Ah no, I skimmed through the logs too quickly: It exists but it requires an error message to match: https://docs.julialang.org/en/v1/stdlib/Test/#Test.@test_warn (BTW it doesn't cover logs generated with the logging system but only stderr - does it actually cover deprecation warnings?)

In any case, it is simpler and more direct to use `@test_deprecated`.

test/locationscale.jl
```diff
@@ -173,4 +173,13 @@ end
         test_location_scale_discretenonparametric(_rng, -1//4, 1//3, (-10):(-1), probs)
         test_location_scale_discretenonparametric(_rng, 6//5, 3//2, 15:24, probs)
     end
+
+    @test_nowarn ad_norm = AffineDistribution(1.0, 1, Normal())
+    @test ad_norm isa AffineDistribution{Float64, Continuous, Normal{Float64}}
```
devmotion, line 178: As I assumed, `@test_nowarn` does not cover deprecation warnings. E.g., check the following example (have to start Julia with `julia --depwarn=yes`):
```julia
julia> using Test

julia> f(x, y) = x + y

julia> Base.@deprecate g(x, y) f(x, y)

julia> function h(x, y)
           print(stderr, "`h(x, y)` is deprecated, use `f(x, y)` instead.")
           return f(x, y)
       end
h (generic function with 1 method)

julia> @test_nowarn f(1, 2)
3

julia> @test_nowarn g(1, 2)
┌ Warning: g(x, y) is deprecated, use f(x, y) instead.
│   caller = #14 at Test.jl:801 [inlined]
└ @ Core /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.7/Test/src/Test.jl:801
3

julia> @test_nowarn h(1, 2)
`h(x, y)` is deprecated, use `f(x, y)` instead.Test Failed at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.7/Test/src/Test.jl:806
  Expression: isempty(stderr_content)
   Evaluated: isempty("`h(x, y)` is deprecated, use `f(x, y)` instead.")
```
I.e., `@test_nowarn` passes also for the deprecated method `g` since it only checks `stderr`.

Instead one has to use `@test_logs` here:
```julia
julia> @test_logs f(1, 2)
3

julia> @test_logs g(1, 2)
Log Test Failed at REPL[19]:1
  Expression: g(1, 2)
  Log Pattern:
  Captured Logs:
    Test.LogRecord(Warn, "`g(x, y)` is deprecated, use `f(x, y)` instead.", Main, :depwarn, (Ptr{Nothing} @0x00007f37f8121ba9, :g), "./none", 0, Base.Pairs{Symbol, Any, Tuple{Symbol, Symbol}, NamedTuple{(:caller, :maxlog), Tuple{Base.Stac
kTraces.StackFrame, Int64}}}(:caller => (::var"#33#34")() at none:0, :maxlog => 1))

julia> @test_logs h(1, 2)
`h(x, y)` is deprecated, use `f(x, y)` instead.3
```

Thus we should just check:
```suggestion
    @test (@test_logs Distributions.AffineDistribution(1.0, 1, Normal())) isa Distributions.AffineDistribution{Float64, Continuous, Normal{Float64}}
```
or just 
```suggestion
    @test_logs Distributions.AffineDistribution(1.0, 1, Normal())
```
(IMO it's not particularly useful to check the type). Here `AffineDistribution` has to be qualified since it is not exported.

test/locationscale.jl
```diff
@@ -173,4 +173,13 @@ end
...
+
+    @test_deprecated ls_norm = LocationScale(1.0, 1, Normal())
+    @test ls_norm isa LocationScale{Float64, Continuous, Normal{Float64}}
+    @test ls_norm isa AffineDistribution{Float64, Continuous, Normal{Float64}}
```
devmotion, line 182:
```suggestion
    @test ls_norm isa Distributions.AffineDistribution{Float64, Continuous, Normal{Float64}}
```

test/locationscale.jl
```diff
@@ -173,4 +173,13 @@ end
...
+    
+    @test ls_norm == ad_norm
```
devmotion, line 184: Is this really needed?
```suggestion
```

devmotion (changes requested):

ParadaCarleton:
@devmotion anything left?

src/univariate/locationscale.jl
```diff
@@ -1,129 +1,140 @@
...
-    LocationScale(μ,σ,ρ)
+    AffineDistribution(μ, σ, ρ)
 
-A location-scale transformed distribution with location parameter `μ`,
-scale parameter `σ`, and given univariate distribution `ρ`.
+A shifted and scaled (affinely transformed) version of `ρ`.
 
-If ``Z`` is a random variable with distribution `ρ`, then the distribution of the random
-variable
+If `Z` is a random variable with distribution `ρ`, then `AffineDistribution(μ, σ, ρ)` is
```
devmotion, line 6:
```suggestion
If ``Z`` is a random variable with distribution `ρ`, then `AffineDistribution(μ, σ, ρ)` is
```

src/univariate/locationscale.jl
```diff
@@ -1,129 +1,140 @@
...
+the distribution of the random variable
 ```math
-X = μ + σ Z
+X = μ + σ * Z
```
devmotion, line 9: This is not a code block but an equation:
```suggestion
X = μ + σ Z
```

src/univariate/locationscale.jl
```diff
@@ -1,129 +1,140 @@
...
-the transformed distribution is given by
+If `ρ` is a discrete univariate distribution, the probability mass function of the
+transformed distribution is given by
 ```math
 P(X = x) = P\\left(Z = \\frac{x-μ}{σ} \\right).
 ```
-If `ρ` is a continuous distribution, the probability density function of
-the transformed distribution is given by
+
+If `ρ` is a continuous univariate distribution with probability density function `f_Z`,
```
devmotion, line 18:
```suggestion
If `ρ` is a continuous univariate distribution with probability density function ``f_Z``,
```

src/univariate/locationscale.jl
```diff
@@ -1,129 +1,140 @@
...
+
+Affine transformations of discrete variables are easily affected by rounding errors. If you
+are getting incorrect results, try using exact `Rational` types instead of floats.
+
 ```julia
-LocationScale(μ,σ,ρ) # location-scale transformed distribution
-params(d)            # Get the parameters, i.e. (μ, σ, and the base distribution ρ)
-location(d)          # Get the location parameter
-scale(d)             # Get the scale parameter
+d = σ * ρ + μ       # Create location-scale transformed distribution
```
devmotion, line 32: A bit more consistent with the order of the arguments:
```suggestion
d = μ + σ * ρ       # Create location-scale transformed distribution
```

devmotion (commented):
Looks good, just some minor suggestions regarding the docstring. Can you also merge the latest changes from the master branch (there were some changes of the location scale tests)?

devmotion (approved):

PR #1454, opened 2021-12-09
Log logistic

yunzli:
Hi:

I think it might be good to include LogLogistic distribution in the package.


devmotion:
Thank you for the PR! Before I add more detailed comments, I wonder if we should use [an alternative parameterization](https://en.wikipedia.org/wiki/Log-logistic_distribution#Alternative_parameterization) that is consistent with the logistic distribution. Is there a specific reason for why you chose the parameterization in this PR? An advantage of the other parameterization would be that `LogLogistic` would be consistent with `Logistic`. In all other cases (`LogNormal`, `LogitNormal`, and `LogUniform`) the parameters are consistent with the "original" distributions, i.e., `Normal` and `Uniform`. I assume that consistent parameters will also make it easier to use the existing implementation of `Logistic`.

yunzli:
Hi, using the alternative parameterization is an option. However, to my knowledge, alternative parameterization has not been used widely in research papers. With the current parameterization, the parameter &#952; serves as a scale parameter and median of the distribution. The alternative parameterization doesn't have such interpretations.

ParadaCarleton:
> Hi, using the alternative parameterization is an option. However, to my knowledge, alternative parameterization has not been used widely in research papers. With the current parameterization, the parameter θ serves as a scale parameter and median of the distribution. The alternative parameterization doesn't have such interpretations.

We could use a keyword constructor for the parametrization currently being used (in terms of `θ`).

devmotion:
I think for consistency we should use the same parameters as for `Logistic` here - also in the case of e.g. `LogNormal` we don't use the (possibly) more interpretable median but the mean of the normal distribution it is constructed from.

ParadaCarleton:
> I think for consistency we should use the same parameters as for `Logistic` here - also in the case of e.g. `LogNormal` we don't use the (possibly) more interpretable median but the mean of the normal distribution it is constructed from.

Yep, sorry, I was agreeing; just offering a suggestion about how to add an extra parametrization. (I'd definitely enjoy seeing parametrizations for both the LogNormal and the LogLogistic in terms of their geometric mean/median using keyword constructors.)

ParadaCarleton:
> > I think for consistency we should use the same parameters as for `Logistic` here - also in the case of e.g. `LogNormal` we don't use the (possibly) more interpretable median but the mean of the normal distribution it is constructed from.
> 
> It is possible to make re-parameterization internally so we can use the existing implementation of `Logistic`? My concern is that it is common to use the same parameterization for `LogNormal` and `Normal`, but not for `LogLogistic` and `Logistic`.

I think I've seen this parametrization for the log-logistic before in papers (mostly in economics), so I don't think it's all that uncommon. I assume the popularity of each varies by field.

yunzli:
> > > I think for consistency we should use the same parameters as for `Logistic` here - also in the case of e.g. `LogNormal` we don't use the (possibly) more interpretable median but the mean of the normal distribution it is constructed from.
> > 
> > 
> > It is possible to make re-parameterization internally so we can use the existing implementation of `Logistic`? My concern is that it is common to use the same parameterization for `LogNormal` and `Normal`, but not for `LogLogistic` and `Logistic`.
> 
> I think I've seen this parametrization for the log-logistic before in papers (mostly in economics), so I don't think it's all that uncommon. I assume the popularity of each varies by field.

Hi, thank you for providing that information. Can you give me some references? I couldn't find any. Thank you again.

PR #1455, opened 2021-12-09
Add Unicode alias `\ast` for `convolve`

devmotion:
Since there is no consensus yet on the Unicode alias for `convolve` yet in https://github.com/JuliaStats/Distributions.jl/pull/1445, I put together this PR with an alternative suggestion.

This PR contains the same changes as https://github.com/JuliaStats/Distributions.jl/pull/1445 but uses the alias `∗` (`\ast`) instead of `⊕` (`\oplus`). This alternative was suggested in https://github.com/JuliaStats/Distributions.jl/issues/142#issuecomment-505948261. An advantage is that it matches more closely the ASCII alternative `convolve` as `∗` is a common notation for convolutions. However, disadvantages are that it is arguably less intuitive and looks similar to `*` (multiplication).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.25.35"
+version = "0.25.36"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/docs/src/convolution.md
+++ b/docs/src/convolution.md
@@ -4,7 +4,7 @@ A [convolution of two probability distributions](https://en.wikipedia.org/wiki/L
 is the probability distribution of the sum of two independent random variables that are
 distributed according to these distributions.
 
-The convolution of two distributions can be constructed with [`convolve`](@ref).
+The convolution of two distributions can be constructed with [`convolve`](@ref) or its Unicode alias [`∗`](@ref).
 
 ```@docs
 convolve
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -190,6 +190,7 @@ export
     componentwise_logpdf,   # component-wise logpdf for mixture models
     concentration,      # the concentration parameter
     convolve,           # convolve distributions of the same type
+    ∗,                  # unicode alias for convolve
     dim,                # sample dimension of multivariate distribution
     dof,                # get the degree of freedom
     entropy,            # entropy of distribution in nats
--- a/src/convolution.jl
+++ b/src/convolution.jl
@@ -1,8 +1,11 @@
 """
     convolve(d1::Distribution, d2::Distribution)
+    d1 ∗ d2
 
 Convolve two distributions and return the distribution corresponding to the sum of
 independent random variables drawn from the underlying distributions.
+
+The Unicode operator `∗` can be typed by `\\ast<tab>`.
 
 Currently, the function is only defined in cases where the convolution has a closed form.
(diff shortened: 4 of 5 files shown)


devmotion:
> `x∗y` and `x * y` would mean two very different things.

I agree and therefore mentioned that this might be a problem. However, one should keep in mind that `x * y` (multiplication) is not defined if `x` and `y` are distributions but only if one of them is a scalar, array, etc. So mainly it would be a problem when reading the code - when the code is executed, standard multiplication will just throw a `MethodError`.

ParadaCarleton:
A few more comments:
1. If we don't want `\oplus` but want to keep something that looks like addition, we also have `\boxplus`.
2. `x \ast y` is appropriate if we're thinking of `x` and `y` as being "equal to" their distributions, but in other situations, we've been following the convention of thinking of both `x` and `y` as being "equal to" the random variables. I'd prefer if we stuck to one or the other consistently.
3. If we end up going with `\oplus` or something similar, I think it would make sense to deprecate the old `+` and `*` for affine transformations and use `\oplus` and `\otimes` for this, both for consistency and to avoid confusion with mixtures of distributions (since `a * X + b * Y` can look a lot like a mixture). If we have `\oplus` for sums, it's very likely that someone will notice `+` and assume it means something different from `\oplus`, so it must be taking a mixture.

devmotion:
I disagree with 3., I don't think the choice here should affect the notation for affine transformations. They are separate things, and as discussed in the old issues related to this topic `+` and `*` make sense and are unambiguous for scalars and distributions in the context of Distributions.jl (see e.g. https://github.com/JuliaStats/Distributions.jl/issues/1438 for a discussion about it and references to prior discussions on this topic). We don't support any notation for mixture distributions and currently there are no plans to change this. Interpreting `a * dist1 + b * dist2` as a mixture distribution would only make sense if `a == 1 - b` and not for general expressions of this form. It seems much more generally applicable to view `a * dist1` as an affine transformation (and as mentioned `dist1 + dist2` is not allowed since it is unclear if the two distributions would be independent or not).

mschauer:
`a * dist1 + b * dist2` per se is a good notation to produce an unnormalised mixture of distributions. We should not mix operators for "distributions as distributions" and for "distributions as random variables"

ParadaCarleton:
I have a proposal; how about a macro that lets users clarify whether they want to treat a distribution as its pdf or random variable? `@iid_rv x + y` would return the convolution of `pdf(x)` with `pdf(y)`, while something like `@pdf` could let the user clarify they want to add the pdfs together to get an unnormalized mixture.

devmotion:
I don't think one should add a custom DSL to Distributions, so I think such macros should not be added (in general, I think one should limit the use of metaprogramming). It's also more verbose than `convolve` and the names seem not completely intutive (e.g., why would `@iid_rv` change the meaning of the distributions instead of returning eg a set of iid random variables? Similarly, why does `@pdf` not return or evaluate a probability density function?).

ParadaCarleton:
> I don't think one should add a custom DSL to Distributions, so I think such macros should not be added (in general, I think one should limit the use of metaprogramming). It's also more verbose than `convolve` and the names seem not completely intutive (e.g., why would `@iid_rv` change the meaning of the distributions instead of returning eg a set of iid random variables? Similarly, why does `@pdf` not return or evaluate a probability density function?).

I agree we should limit the use of metaprogramming, but this seems like a sufficiently limited addition that it shouldn’t cause major problems. 

We could alternatively use wrappers around distributions, so that RV(x) + RV(y) returns the convolution.  

(If the names are counterintuitive, maybe `as_rv` and `as_pdf`?

PR #1459, opened 2021-12-23
Add ChainRules derivatives for `logpdf(::Uniform, x)`

devmotion:
This PR adds ChainRules derivatives for `logpdf(::Uniform, x)`. Currently, DistributionsAD uses type piracy and a custom distribution to compute its derivative (see e.g. https://github.com/TuringLang/DistributionsAD.jl/blob/5570ac530ec8c5af3e045297c18d30d1b8819cb9/src/zygote.jl#L3-L5). This PR will allow us to remove the custom distribution and most of the type piracy regarding `Uniform`.

Due to the [recently AD-friendly updated](https://github.com/JuliaStats/Distributions.jl/pull/1411) implementation of `logpdf(::Uniform, x)` Zygote can already compute its derivative automatically. However, a custom derivative is significantly faster:

On the master branch:
```julia
julia> using Distributions, Zygote, BenchmarkTools

julia> f(x) = logpdf(Uniform(x[1], x[2]), x[3])

julia> x = [0.0, 1.0, -0.5];

julia> @btime Zygote.gradient($f, $x);
  9.808 μs (65 allocations: 2.88 KiB)

julia> x = [0.0, 1.0, 0.5];

julia> @btime Zygote.gradient($f, $x);
  11.171 μs (111 allocations: 4.14 KiB)
```

With this PR:
```julia
julia> using Distributions, Zygote, BenchmarkTools

julia> f(x) = logpdf(Uniform(x[1], x[2]), x[3])

julia> x = [0.0, 1.0, -0.5];

julia> @btime Zygote.gradient($f, $x);
  4.480 μs (58 allocations: 2.53 KiB)

julia> x = [0.0, 1.0, 0.5];

julia> @btime Zygote.gradient($f, $x);
  4.511 μs (58 allocations: 2.53 KiB)
```

The derivatives for values outside of the support are defined as zero, in the same way as computed by Zygote currently: local perturbations (of any argument) don't affect the output of the function, it is always `-Inf`. An alternative would be to return NaN in such cases, as done e.g. by finite differencing: the primal is `-Inf` and also local perturbations (of any argument) result in `-Inf`, and thus the computation `-Inf - (-Inf)` in finite differencing leads to NaN values.


mschauer:
I think it should be `NaN` on the boundaries at least, so perhaps `NaN` also outside the support as well for simplicity.

devmotion:
I'm still unsure what the correct approach is here. I think there is still a strong argument for returning zeros: `NaN` is usually not useful, and while it might not matter for points outside of the support which can't be sampled it affects evaluation of the derivatives at the boundaries which can be sampled. Hence in AD in fact derivatives don't necessarily match the mathematical definition but sometimes are defined in more useful ways: https://juliadiff.org/ChainRulesCore.jl/stable/maths/nondiff_points.html

devmotion:
Maybe you have an opinion, @oxinabox? Should we return zero or `NaN`?

mschauer:
Thanks for the link, one can't praise @oxinabox enough for the AD doc work

oxinabox:
I think zero makes sense.
If we are fitting a uniform normal to some data by gradient descent (why?!) and it is getting narrower and narrower then it should stop getting narrower at the point that it shrinks to zero width.
It shouldn't change to giving out NaNs.

I might be thinking about things wrong though

mschauer:
Yes, it's good to have a specific example. So you have samples `X[i] ~ U[0, b]`, `i = 1:n` and consider maximum likelihood estimation. The maximum likelihood estimator is `maximum(X)`: 
Screen Shot 2022-01-12 at 11 38 52

devmotion:
@mschauer, do you still think it should be `NaN`? I tend to 0 on the boundaries for the reasons stated above. Still a bit unsure outside of the support though - I think I tend to 0 as well, for simplicity, since I dislike `NaN`, and since it is a (non-finite) local minimum. For the likelihood/pdf it would be a bit easier, there it is clearly 0 outside of the support.

mschauer (approved):

PR #1460, opened 2021-12-23
Remove documentation for LocationScale

devmotion:
https://github.com/JuliaStats/Distributions.jl/pull/1453 broke the documentation of LocationScale: https://juliastats.org/Distributions.jl/latest/univariate/ This is quite unfortunate since [I explicitly asked whether it still shows up](https://github.com/JuliaStats/Distributions.jl/pull/1453#issuecomment-997017439).

This PR removes the deprecated `LocationScale` from the documentation.

cc: @ParadaCarleton

PR #1462, opened 2021-12-24
Affine doc fix

ParadaCarleton:
Fixes affine distribution docs

### DIFF
--- a/docs/src/univariate.md
+++ b/docs/src/univariate.md
@@ -132,6 +132,13 @@ end
 ```
 
 ```@docs
+Distributions.AffineDistribution
+```
+```@example plotdensity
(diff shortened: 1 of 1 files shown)


ParadaCarleton:
@devmotion sorry about that!

devmotion:
This is a duplicate of https://github.com/JuliaStats/Distributions.jl/pull/1460, it seems? I don't think we should add the docstring for `AffineDistribution` to the webpage since it is not exported and should not be used. We should rather document `*`, `+`, etc properly at some point.

devmotion:
Closing in favour of https://github.com/JuliaStats/Distributions.jl/pull/1460.

PR #1466, opened 2021-12-30
`GeneralizedPareto` improvement

ParadaCarleton:
Clarifies the `GeneralizedPareto` docstring and adds a new constructor, `GeneralizedPareto(ξ)`, which defaults to unit scale and a location of 0.


src/univariate/continuous/generalizedpareto.jl
```diff
@@ -48,7 +49,8 @@ function GeneralizedPareto(μ::Integer, σ::Integer, ξ::Integer)
     GeneralizedPareto(float(μ), float(σ), float(ξ))
 end
 GeneralizedPareto(σ::T, ξ::Real) where {T <: Real} = GeneralizedPareto(zero(T), σ, ξ)
-GeneralizedPareto() = GeneralizedPareto(0.0, 1.0, 1.0, check_args=false)
+GeneralizedPareto(ξ::Real) where {T <: Real} = GeneralizedPareto(zero(T), one(T), ξ)
```
devmotion, line 52: This is wrong, `T` does not exist.

src/univariate/continuous/generalizedpareto.jl
```diff
@@ -15,8 +15,9 @@ f(x; \\mu, \\sigma, \\xi) = \\begin{cases}
 ```
 
 ```julia
-GeneralizedPareto()             # Generalized Pareto distribution with unit shape and unit scale, i.e. GeneralizedPareto(0, 1, 1)
-GeneralizedPareto(μ, σ)         # Generalized Pareto distribution with shape ξ and scale σ, i.e. GeneralizedPareto(0, σ, ξ)
+GeneralizedPareto()             # Standard exponential distribution (GPD with shape 0, location 0, scale 1)
+GeneralizedPareto(ξ)            # Generalized Pareto distribution with shape ξ, location 0, and scale 1
```
devmotion, line 19: The choice for using `ξ` in the one-argument constructor seems a bit arbitrary, is there a specific reason for it? (Of course, the same could be said about the existing two-argument constructor).

src/univariate/continuous/generalizedpareto.jl
```diff
@@ -48,7 +49,8 @@ function GeneralizedPareto(μ::Integer, σ::Integer, ξ::Integer)
     GeneralizedPareto(float(μ), float(σ), float(ξ))
 end
 GeneralizedPareto(σ::T, ξ::Real) where {T <: Real} = GeneralizedPareto(zero(T), σ, ξ)
```
devmotion, line 51: BTW no need for a type parameter here:
```suggestion
GeneralizedPareto(σ::Real, ξ::Real) = GeneralizedPareto(zero(σ), σ, ξ)
```

src/univariate/continuous/generalizedpareto.jl
devmotion:
```suggestion
GeneralizedPareto(ξ::Real) = GeneralizedPareto(zero(ξ), one(ξ), ξ; check_args=false)
```

devmotion (changes requested):

src/univariate/continuous/generalizedpareto.jl
ParadaCarleton: > The choice for using `ξ` in the one-argument constructor seems a bit arbitrary, is there a specific reason for it? (Of course, the same could be said about the existing two-argument constructor).

Parametrizing the Generalized Pareto in terms of scale and shape is pretty common. Parametrizing it only in terms of shape is less common but still done from time to time -- Wikipedia notes:
```
The generalized Pareto distribution (GPD) [...] is specified by three parameters: location {\displaystyle \mu }\mu , scale {\displaystyle \sigma }\sigma , and shape {\displaystyle \xi }\xi .[1][2] Sometimes it is specified by only scale and shape[3] and sometimes only by its shape parameter. 
```

src/univariate/continuous/generalizedpareto.jl
```diff
@@ -47,8 +48,9 @@ GeneralizedPareto(μ::Real, σ::Real, ξ::Real) = GeneralizedPareto(promote(μ,
 function GeneralizedPareto(μ::Integer, σ::Integer, ξ::Integer)
     GeneralizedPareto(float(μ), float(σ), float(ξ))
 end
-GeneralizedPareto(σ::T, ξ::Real) where {T <: Real} = GeneralizedPareto(zero(T), σ, ξ)
-GeneralizedPareto() = GeneralizedPareto(0.0, 1.0, 1.0, check_args=false)
+GeneralizedPareto(σ::Real, ξ::Real) = GeneralizedPareto(zero(σ), σ, ξ)
+GeneralizedPareto(ξ::Real) = GeneralizedPareto(zero(ξ), one(ξ), ξ; check_args=false)
+GeneralizedPareto() = GeneralizedPareto(0.0, 1.0, 1.0; check_args=false)
```
devmotion, line 53: Slightly simpler:

```suggestion
GeneralizedPareto(ξ::Real=1.0) = GeneralizedPareto(zero(ξ), one(ξ), ξ; check_args=false)
```

devmotion (commented):
Can you add some tests?

src/univariate/continuous/generalizedpareto.jl
```diff
@@ -15,8 +15,9 @@ f(x; \\mu, \\sigma, \\xi) = \\begin{cases}
 ```
 
 ```julia
-GeneralizedPareto()             # Generalized Pareto distribution with unit shape and unit scale, i.e. GeneralizedPareto(0, 1, 1)
-GeneralizedPareto(μ, σ)         # Generalized Pareto distribution with shape ξ and scale σ, i.e. GeneralizedPareto(0, σ, ξ)
+GeneralizedPareto()             # Standard exponential distribution (GPD with shape 0, location 0, scale 1)
```
devmotion, line 18: This is wrong. You can just change it back to the current version.
```suggestion
GeneralizedPareto()             # Generalized Pareto distribution with unit shape and unit scale, i.e. GeneralizedPareto(0, 1, 1)
```

src/univariate/continuous/generalizedpareto.jl
devmotion, line 19: This constructor is not defined.

I think it would be good to follow the existing examples and show the corresponding 3-argument constructor as well:
```suggestion
GeneralizedPareto(ξ)            # Generalized Pareto distribution with shape ξ and unit scale, i.e. GeneralizedPareto(0, 1, ξ)
```

src/univariate/continuous/generalizedpareto.jl
```diff
@@ -15,8 +15,9 @@ f(x; \\mu, \\sigma, \\xi) = \\begin{cases}
...
+GeneralizedPareto(ξ)            # Generalized Pareto distribution with shape ξ, location 0, and scale 1
+GeneralizedPareto(σ, ξ)         # Generalized Pareto distribution with shape ξ and scale σ
```
devmotion, line 20:
```suggestion
GeneralizedPareto(σ, ξ)         # Generalized Pareto distribution with shape ξ and scale σ, i.e. GeneralizedPareto(0, σ, ξ)
```

src/univariate/continuous/generalizedpareto.jl
```diff
@@ -52,6 +53,7 @@ end
 function GeneralizedPareto(σ::Real, ξ::Real; check_args::Bool=true)
     GeneralizedPareto(zero(σ), σ, ξ; check_args=check_args)
 end
+
```
devmotion, line 56:
```suggestion
```

devmotion (changes requested):
You removed the one argument constructor it seems. Can you please also address my comment above and add some tests for this new (currently missing) constructor?

It would be nice if you could check next time that the PR is actually ready for review before requesting one.

ParadaCarleton:
> You removed the one argument constructor it seems. Can you please also address my comment above and add some tests for this new (currently missing) constructor?
> 
> It would be nice if you could check next time that the PR is actually ready for review before requesting one.

Sorry about that @devmotion, I meant to request review on a different PR. Do you know which file contains the tests for the GPD? I can't actually find where they are.

devmotion:
They are tested with the standard tests in https://github.com/JuliaStats/Distributions.jl/blob/master/test/univariates.jl. See https://github.com/JuliaStats/Distributions.jl/blob/02bcbf8169160cc499832f05c83e9308ecdb2457/test/ref/continuous_test.ref.json#L1826 etc.

ParadaCarleton:
> They are tested with the standard tests in https://github.com/JuliaStats/Distributions.jl/blob/master/test/univariates.jl. See
> 
> https://github.com/JuliaStats/Distributions.jl/blob/02bcbf8169160cc499832f05c83e9308ecdb2457/test/ref/continuous_test.ref.json#L1826
> 
> etc.

Got it, thanks! I think this should fix it.

test/ref/continuous_test.lst
```diff
@@ -66,6 +66,7 @@ GeneralizedExtremeValue(1.0, 1.0, -1.0)
 GeneralizedExtremeValue(-1.0, 0.5, 0.6)
 
 GeneralizedPareto()
+GeneralizedPareto(1.0)
```
devmotion, line 69: Could we also add something like
```suggestion
GeneralizedPareto(1.0)
GeneralizedPareto(2.0)
```
? That seems a bit safer since `GeneralizedPareto(1.0)` would pass even if e.g. it defaults to `GeneralizedPareto()` or if the argument would set the scale instead of the shape.
?

test/ref/continuous_test.lst
```diff
@@ -66,6 +66,8 @@ GeneralizedExtremeValue(1.0, 1.0, -1.0)
...
+GeneralizedPareto(2.0)
```
devmotion, line 70: AFAICT currently it is not tested. I guess you have to add it to continuous_test.ref.json as well.

ParadaCarleton: > AFAICT currently it is not tested. I guess you have to add it to continuous_test.ref.json as well.

It's already in there, so if it's not being tested I'd be very surprised.

devmotion: The tests (https://github.com/JuliaStats/Distributions.jl/actions/runs/3814509537/jobs/6488867611#step:6:499 and subsequent lines) clearly show that it's not tested. I'm pretty sure it will work if you add a section for `GeneralizedPareto(2.0)` to continuous_test.ref.json.

devmotion: I added reference values for `GeneralizedPareto(2.0)` in https://github.com/JuliaStats/Distributions.jl/pull/1466/commits/6260cebb111d1652b176d41525fb702e9c2fd7bd, let's see if that fixes the tests.

devmotion: It fixed the problem: https://github.com/JuliaStats/Distributions.jl/actions/runs/3878010552/jobs/6613701355#step:6:507

devmotion (commented):
Test seem to be complete and pass. So the final (main) question is: Do we actually want to add more constructors? Maybe it would be better to switch to keyword argument constructors instead (see e.g. https://github.com/JuliaStats/Distributions.jl/issues/584#issuecomment-1376289708 and https://github.com/JuliaStats/Distributions.jl/issues/584#issuecomment-1377003475)?

ParadaCarleton:
> Test seem to be complete and pass. So the final (main) question is: Do we actually want to add more constructors? Maybe it would be better to switch to keyword argument constructors instead (see e.g. [#584 (comment)](https://github.com/JuliaStats/Distributions.jl/issues/584#issuecomment-1376289708) and [#584 (comment)](https://github.com/JuliaStats/Distributions.jl/issues/584#issuecomment-1377003475))?

I'd definitely like to see keyword constructors added to the language, but I don't think that interferes with this PR; this is really just about setting reasonable defaults for the location and scale arguments.

devmotion:
Why was this PR merged?

devmotion:
To explain: I'm not per se against the changes (as commented above) but IMO it is absolutely not OK to merge your own PRs in a repo not owned by you that are not approved and before discussions are resolved.

PR #1467, opened 2022-01-03
Special case Inf and -Inf in truncated

mohamed82008:
This PR fixes a problem that showed up recently but I didn't track down exactly why it shows up now. Currently, the following code produces a `NaN` in the derivative of the log of the normalisation constant of a `Truncated`.
```julia
using ForwardDiff, Distributions

d = truncated(Normal(ForwardDiff.Dual(0.0, 1.0), ForwardDiff.Dual(1.0, 0.0)), 0, Inf)
d.logtp.partials[1] # NaN
```
Since the `logpdf` of a `truncated` subtracts `logtp`, the derivative is always going to be `NaN` at the end. This is clearly wrong and is only an artefact of how ForwardDiff/DiffRules behaviour is undefined for `Inf` values. For example, `x.partials[1]` is `NaN` in:
```julia
x = ForwardDiff.Dual(Inf, 0.0) / ForwardDiff.Dual(0.1, 0.0)
```
causing the `zval`, `(Inf - d.μ) / d.σ`, to give `NaN` in the derivative. This however seems to be the case even in previous versions of ForwardDiff and I haven't tracked down which change in `Distributions.truncated` made this ForwardDiff behaviour significant in the above example.

However, this PR feels like a half fix because:
1. It's not clear which change caused the above issue,
2. There are other similar functions that may need a special case for `Inf` and `-Inf` but I don't need them so I didn't fix their Dual support, and
3. It feels like ForwardDiff or DiffRules should be the place to fix this or special case `Inf` and `-Inf` but that will require a more significant effort.

That said, this PR gets the job done so I am opening it for now to get feedback.

### DIFF
--- a/src/truncate.jl
+++ b/src/truncate.jl
@@ -26,10 +26,17 @@ function truncated(d::UnivariateDistribution, l::T, u::T) where {T <: Real}
     else
         logcdf(d, l)
     end
+    _T = typeof(loglcdf)
+    if isfinite(loglcdf) && l == -Inf
+        loglcdf = convert(_T, -Inf)
+    end
     lcdf = exp(loglcdf)
 
     # (log)ucdf = (log) P(X ≤ u) where X ~ d
     logucdf = logcdf(d, u)
+    if isfinite(logucdf) && l == Inf
+        loglcdf = zero(_T)
+    end
     ucdf = exp(logucdf)
 
     # (log)tp = (log) P(l ≤ X ≤ u) where X ∼ d
--- a/src/truncate.jl
+++ b/src/truncate.jl
@@ -34,8 +34,8 @@ function truncated(d::UnivariateDistribution, l::T, u::T) where {T <: Real}
 
     # (log)ucdf = (log) P(X ≤ u) where X ~ d
     logucdf = logcdf(d, u)
-    if isfinite(logucdf) && l == Inf
-        loglcdf = zero(_T)
+    if isfinite(logucdf) && u == Inf
+        logucdf = zero(_T)
     end
     ucdf = exp(logucdf)
 
--- a/src/univariate/continuous/normal.jl
+++ b/src/univariate/continuous/normal.jl
@@ -155,10 +155,14 @@ end
 
 function logcdf(d::Normal, x::Real)
     # The Inf cases are useful for ForwardDiff.Dual to avoid NaN derivatives
-    if x == Inf
-        return zero(promote_type(typeof(d.μ), typeof(d.σ), typeof(x)))
-    elseif x == -Inf
-        return convert(promote_type(typeof(d.μ), typeof(d.σ), typeof(x)), -Inf)
+    finited = isfinite(d.μ) && isfinite(d.σ)
+    if !isfinite(x) && !isnan(x) && finited
+        T = typeof(log(one(promote_type(typeof(d.μ), typeof(d.σ), typeof(x)))))
+        if x == Inf
+            return zero(T)
+        else
+            return convert(T, -Inf)
+        end
     elseif iszero(d.σ) && x == d.μ
         z = zval(Normal(zero(d.μ), d.σ), one(x))
     else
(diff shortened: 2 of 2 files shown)


devmotion:
Hmm I think it is a bit suboptimal to have to work around such ForwardDiff specific issues since it makes the code more complex without any advantage (?) in other use cases. Does the problem occur also with the NaN-safe setting of ForwardDiff?

mohamed82008:
>Does the problem occur also with the NaN-safe setting of ForwardDiff?

Could you please elaborate on this?

devmotion:
There's a NaN-safe mode: https://juliadiff.org/ForwardDiff.jl/dev/user/advanced/#Fixing-NaN/Inf-Issues It became a bit easier to use with Preferences (before one had to modify ForwardDiff; Chris even created a nansafe branch to make it a bit easier).

mohamed82008:
Same behaviour with the NaN-safe mode.

mohamed82008:
I generalised the fix here to work with any `Truncated`.

src/truncate.jl
```diff
@@ -26,10 +26,17 @@ function truncated(d::UnivariateDistribution, l::T, u::T) where {T <: Real}
     else
         logcdf(d, l)
     end
+    _T = typeof(loglcdf)
+    if isfinite(loglcdf) && l == -Inf
+        loglcdf = convert(_T, -Inf)
```
devmotion, line 31: Can you explain why this fixes the problem? It shouldn't change the value of `loglcdf`?

mohamed82008: It doesn't change its value but it changes the partial to 0 from NaN which gives the correct behaviour.
```julia
julia> using ForwardDiff

julia> d = ForwardDiff.Dual(1.0, 1.0)
Dual{Nothing}(1.0,1.0)

julia> convert(typeof(d), Inf)
Dual{Nothing}(Inf,0.0)
```

mohamed82008:
bump

devmotion:
I'm not happy with this PR as it is a quite strange workaround that only makes sense for ForwardDiff. At least it would be good to add some comments for readers about these checks, add tests for the bug that it fixes, and bump the version.

I'd like to check and see if there's a cleaner approach but it might take a few more days since I'm moving and don't have access to my computer right now.

devmotion:
> Same behaviour with the NaN-safe mode.

I can't reproduce this, NaN-safe mode fixes the issue for me (latest releases of both packages):
```julia
julia> using ForwardDiff, Distributions

julia> d = truncated(Normal(ForwardDiff.Dual(0.0, 1.0), ForwardDiff.Dual(1.0, 0.0)), 0, Inf)
Truncated(Normal{ForwardDiff.Dual{Nothing, Float64, 1}}(μ=Dual{Nothing}(0.0,1.0), σ=Dual{Nothing}(1.0,0.0)), range=(Dual{Nothing}(0.0,0.0), Dual{Nothing}(Inf,0.0)))

julia> d.logtp.partials[1]
0.7978845608028654
```

Maybe you forgot to restart Julia after setting the NaN-safe preference? You can check if it is enabled by running the example from the documentation:
```julia
julia> log(ForwardDiff.Dual{:tag}(0.0, 0.0))
Dual{:tag}(-Inf,0.0)
```
The partial is `NaN` if the NaN-safe mode is not enabled.

I'll close this PR since it is not needed and the problem is fixed by using NaN-safe mode.

mohamed82008:
I think the NaN-safe mode is still a sub-optimal solution to the problem. You are basically taking a performance hit everywhere in the code base to avoid a problem that only happens with truncated distributions or in a few cases.

mohamed82008:
I think the fix in this PR though sub-optimal in a different way still gives a better overall performance.

devmotion:
> I think the NaN-safe mode is still a sub-optimal solution to the problem. You are basically taking a performance hit everywhere in the code base to avoid a problem that only happens with truncated distributions or in a few cases.

It's unclear whether and how large a performance hit it actually causes. Hence it might even become the default at some point, see e.g. https://github.com/JuliaDiff/ForwardDiff.jl/pull/451#issuecomment-943742789. It's really a ForwardDiff issue, and I agree with Andreas that the default setting is just wrong: https://github.com/JuliaDiff/ForwardDiff.jl/pull/451#issuecomment-751721327

Issue #1471, opened 2022-01-12
InverseGamma constructor does not work with integer values and check_args = false


bvdmitri:
```julia
julia> InverseGamma(-1, 2, check_args = false)
ERROR: MethodError: no method matching Gamma(::Int64, ::Float64; check_args=false)
Closest candidates are:
  Gamma(::Real, ::Real) at ~/.julia/packages/Distributions/vR2pk/src/univariate/continuous/gamma.jl:38 got unsupported keyword argument "check_args"
  Gamma(::Integer, ::Integer) at ~/.julia/packages/Distributions/vR2pk/src/univariate/continuous/gamma.jl:39 got unsupported keyword argument "check_args"
  Gamma(::T) where T<:Real at ~/.julia/packages/Distributions/vR2pk/src/univariate/continuous/gamma.jl:40 got unsupported keyword argument "check_args"
  ...
Stacktrace:
 [1] InverseGamma{Int64}(α::Int64, θ::Int64)
   @ Distributions ~/.julia/packages/Distributions/vR2pk/src/univariate/continuous/inversegamma.jl:31
 [2] InverseGamma(α::Int64, θ::Int64; check_args::Bool)
   @ Distributions ~/.julia/packages/Distributions/vR2pk/src/univariate/continuous/inversegamma.jl:36
 [3] top-level scope
   @ REPL[13]:1
```

PR #1472, opened 2022-01-13
Use `check_args` keyword argument consistently

devmotion:
Fixes https://github.com/JuliaStats/Distributions.jl/issues/1471 and many (all?) other constructors where `check_args` keyword argument was only supported for specific types of inputs (but, e.g., often not for arguments of different types or integers).

I've been annoyed by this for quite some time and just commented on it a few days ago in some other repo, but now https://github.com/JuliaStats/Distributions.jl/issues/1471 finally motivated me to go through all distributions and fix it.


devmotion:
The DistributionsAD test error is unrelated and caused by https://github.com/JuliaDiff/ChainRules.jl/pull/559.

src/cholesky/lkjcholesky.jl
```diff
@@ -40,7 +40,7 @@ end
 #  Constructors
 #  -----------------------------------------------------------------------------
 
-function LKJCholesky(d::Int, η::Real, _uplo::Union{Char,Symbol} = 'L'; check_args = true)
+function LKJCholesky(d::Int, η::Real, _uplo::Union{Char,Symbol} = 'L'; check_args::Bool=true)
```
mschauer, line 43: Help me understand, what effect does this have?

devmotion: It ensures that one can only pass `true` or `false` for `check_args`, and hence no weird or surprising things can happen in the `if check_arg` statement in the function body. There's no breaking change since passing anything else would be a bug, it just makes the implementation a bit safer.

mschauer: Ok, so this is just for consistency

src/multivariate/multinomial.jl
```diff
@@ -25,7 +25,7 @@ struct Multinomial{T<:Real, TV<:AbstractVector{T}} <: DiscreteMultivariateDistri
     Multinomial{T, TV}(n::Int, p::TV) where {T <: Real, TV <: AbstractVector{T}} = new{T, TV}(n, p)
 end
 
-function Multinomial(n::Integer, p::TV; check_args=true) where {T <: Real, TV <: AbstractVector{T}}
+function Multinomial(n::Integer, p::AbstractVector{T}; check_args::Bool=true) where {T<:Real}
```
rikhuijzer, line 28: As a reviewer, I'm not really a fan of these kinds of changes in an already massive PR.

test/univariates.jl
```diff
@@ -198,4 +198,10 @@ end
         @test @inferred(ccdf(d, 2.5)) === XFT(1//2)
         @test @inferred(ccdf(d, 3.5)) === zero(XFT)
     end
-end
\ No newline at end of file
+end
+
+# #1471
```
rikhuijzer, line 203:
```suggestion
```
Issue is already mentioned in the name.

rikhuijzer (approved):

src/multivariate/multinomial.jl
devmotion: Which breaking changes? There really shouldn't be any. Here I just removed an unnecessary static type parameter since the line was changed anyway, following the [Julia style guide](https://docs.julialang.org/en/v1/manual/style-guide/index.html#Don't-use-unnecessary-static-parameters). I can move this change  to a separate PR but this seemed a bit unnecessary (and I only noticed this while working on the `check_args` changes, I didn't just include additional changes that I intended to commit).

src/edgeworth.jl
```diff
@@ -77,12 +79,14 @@ end
 struct EdgeworthSum{D<:UnivariateDistribution} <: EdgeworthAbstract
     dist::D
     n::Float64
-    function EdgeworthSum{D}(d::T, n::Real) where {D<:UnivariateDistribution,T<:UnivariateDistribution}
-        @check_args(EdgeworthSum, n > zero(n))
+    function EdgeworthSum{D}(d::UnivariateDistribution, n::Real; check_args::Bool=true) where {D<:UnivariateDistribution}
```
mschauer, line 82: Maybe just
```suggestion
    function EdgeworthSum{D}(d::UnivariateDistribution, n::Real; check_args::Bool=true)
```
?

src/univariate/continuous/beta.jl
```diff
@@ -32,15 +32,18 @@ struct Beta{T<:Real} <: ContinuousUnivariateDistribution
...
-Beta(α::Integer, β::Integer) = Beta(float(α), float(β))
-Beta(α::Real) = Beta(α, α)
-Beta() = Beta(1.0, 1.0, check_args=false)
+Beta(α::Real, β::Real; check_args::Bool=true) = Beta(promote(α, β)...; check_args=check_args)
+Beta(α::Integer, β::Integer; check_args::Bool=true) = Beta(float(α), float(β); check_args=check_args)
+function Beta(α::Real; check_args::Bool=true)
+    check_args && @check_args(Beta, α > zero(α))
+    Beta(α, α; check_args=false)
+end
+Beta() = Beta{Float64}(1.0, 1.0)
```
mschauer, line 46: Yes.

src/univariate/continuous/betaprime.jl
```diff
@@ -32,15 +32,18 @@ struct BetaPrime{T<:Real} <: ContinuousUnivariateDistribution
...
 end
 
-BetaPrime(α::Real, β::Real) = BetaPrime(promote(α, β)...)
-BetaPrime(α::Integer, β::Integer) = BetaPrime(float(α), float(β))
-BetaPrime(α::Real) = BetaPrime(α, α)
-BetaPrime() = BetaPrime(1.0, 1.0, check_args=false)
+BetaPrime(α::Real, β::Real; check_args::Bool=true) = BetaPrime(promote(α, β)...; check_args=check_args)
+BetaPrime(α::Integer, β::Integer; check_args::Bool=true) = BetaPrime(float(α), float(β); check_args=check_args)
+function BetaPrime(α::Real; check_args::Bool=true)
+    check_args && @check_args(BetaPrime, α > zero(α))
```
mschauer, line 43: I think `α > 0` is always save to write?

devmotion: Not always but here probably yes. I just copied it from above though, the check in line 36 already uses `zero` (I got the impression that `x > zero(x)` is a bit more common than `x > 0` in Distributions currently).

devmotion: Not completely sure if `x > 0` is problematic for dual or tracked numbers, it is at least for unitful numbers (which we don't officially support here though).

mschauer: Ah, I see, it fails for unitfuls. It's not crucial, more a curiosity.

src/edgeworth.jl
devmotion: We still need some where clause though but AFAIK the type restriction in the where clause could be dropped. I don't think it matters much, I guess the approach in the PR (and on master) ensures that the constructor can't be called if `D` is not a subtype of `UnivariateDistribution` whereas without the restriction in the where clause the call to `new` would error.

mschauer (approved):

PR #1470, opened 2022-01-12
Add Censored distribution

sethaxen:
This PR fixes #1468 for constructing a right-, left-, or interval-censored form of a distribution.

PR #1473, opened 2022-01-13
Clarify definition of the median

mschauer:
Clarify definition of the median and turn 
```
julia> median(DiscreteUniform(1, 4))
2.5
```
into a bug. Came up in https://github.com/JuliaStats/Distributions.jl/pull/1470#discussion_r783502531

### DIFF
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -182,7 +182,8 @@ std(d::UnivariateDistribution) = sqrt(var(d))
 """
     median(d::UnivariateDistribution)
 
-Return the median value of distribution `d`.
+Return the median value of distribution `d`. The median is the smallest `x` such that `cdf(d, x) ≥ 1/2`.
+Corresponding to this definition as 1/2-quantile, a fallback is provided calling the `quantile` function.
 """
 median(d::UnivariateDistribution) = quantile(d, 1//2)
 


devmotion:
Maybe fix the bug as well?

src/univariates.jl
```diff
@@ -182,7 +182,8 @@ std(d::UnivariateDistribution) = sqrt(var(d))
 """
     median(d::UnivariateDistribution)
 
-Return the median value of distribution `d`.
+Return the median value of distribution `d`. The median is the smallest `x` such that `cdf(d, x) ≥ 1/2`.
```
rikhuijzer, line 185: Can an informal description be added before the formal definition to make it easier to understand? 

Something like "Informally, the median is the value which divides the distribution `d` in two equally sized portions. Formally, the median is the smallest `x` such that `cdf(d, x) ≥ 1/2`."

mschauer:
> Maybe fix the bug as well?

Looks like also quantile is wrong there. Let's merge here and continue with #1474.

devmotion:
Ah, I didn't realize there's more to fix than the median and a separate issue.

devmotion (approved):

src/univariates.jl
mschauer: This is a can of worms, let's avoid explaining medians, people will know to look these things up.

PR #1484, opened 2022-01-18
Add examples to quantile

KronosTheLate:
This PR adds examples to the docstring for quantile.
The main aim of this PR is to make "confidence interval" give a result when searched for in the documentation, as well as to give an example of how to calculate such an interval.

### DIFF
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -367,8 +367,27 @@ logccdf(d::UnivariateDistribution, x::Real) = log(ccdf(d, x))
     quantile(d::UnivariateDistribution, q::Real)
 
 Evaluate the inverse cumulative distribution function at `q`.
+Use `quantile` to calculate quartiles, medians and confidence intervals.
 
 See also: [`cquantile`](@ref), [`invlogcdf`](@ref), and [`invlogccdf`](@ref).
+
+# Examples
+```jldoctest
+julia> d = Normal()
+Normal{Float64}(μ=0.0, σ=1.0)
+
+julia> first_quartile = quantile(d, 0.25)
+-0.6744897501960818
+
+julia> second_quartile = quantile(d, 0.50)
+0.0
+
+julia> third_quartile = quantile(d, 0.75)
+0.6744897501960818
+
+julia> confidence_interval_95_percent = quantile(d, 0.025), quantile(d, 0.975)
+(-1.9599639845400592, 1.9599639845400576)
+```
 """
 quantile(d::UnivariateDistribution, p::Real)
 


src/univariates.jl
```diff
@@ -367,8 +367,27 @@ logccdf(d::UnivariateDistribution, x::Real) = log(ccdf(d, x))
     quantile(d::UnivariateDistribution, q::Real)
 
 Evaluate the inverse cumulative distribution function at `q`.
+Use `quantile` to calculate quartiles, medians and confidence intervals.
```
devmotion, line 370: One shouldn't calculate the median with `quantile`. IMO confidence intervals are not properties of distributions and one usually just wants to use `confint` which is not part of Distributions.

KronosTheLate: >One shouldn't calculate the median with quantile.

Then what is this, from https://github.com/JuliaStats/Distributions.jl/blob/master/src/univariates.jl:
image

> IMO confidence intervals are not properties of distributions...

No they are not a property of the distribution, agreed. Which is why it could be nice with an example of what the function can be used for. To make realizing the potential of the function easier.

> ... and one usually just wants to use confint which is not part of Distributions.

Right. But I found manual computation of confidence intervals very useful in learning about them in a statistics course. And there are LOADS of people taking statistics courses. Which makes me thing that catering a little bit to that audience is justified. Also, there might be other cases where manual computation of confidence intervals would be useful. I am simply suggesting to show how it is done, _in case_ someone needs it. Which I know I did. And which I am confident that someone will need again.

There have been several posts (like these from [discourse](https://discourse.julialang.org/t/package-for-confidence-intervals/33902) and [stackoverflow](https://stackoverflow.com/questions/20425473/calculate-confidence-interval-in-julia)) asking for how to compute confidence intervals. I also had that question, without making a post, suggesting to me that many people have had this question. I think that there is clearly some degree of need for this documentation, and I don't really see the downside...

mschauer: > Then what is this

Well, a "fallback" is something you only do if you don't have anything better/specific.

KronosTheLate: Snap, my bad. Thanks for the patience.

Alright, then I'll remove the suggestion to calculate medians this way.

sethaxen: > Then what is this, from https://github.com/JuliaStats/Distributions.jl/blob/master/src/univariates.jl:

As is noted in the docstring, this is a fallback. Users who want `median` should call `median` directly, since there might be a more efficient overload for that distribution. e.g. for `Normal`, the mean is the median and can be directly returned without calling the inverse error function as `quantile` does.

I should note that it's normal to just define a quantile function as the inverse CDF as is done here. e.g. see [documentation for `scipy.stats.norm.ppf`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.norm.html), or the [docstring for `qnorm` in R](https://stat.ethz.ch/R-manual/R-devel/library/stats/html/Normal.html). The quantile function has many uses, and it seems unnecessary to inform the user of them in my opinion.

KronosTheLate: I agree with your points @sethaxen. Do you see a solution where users would still get a result and some pointers when searching for `confidence interval` in the documentation? Because as a user, I felt that was lacking, and that is the underlying problem I am trying to fix.

sethaxen: I don't think it's lacking for the same reason I don't think it's a problem that in the docstring for `logpdf` we don't say that it could, for example, be used to sample from a distribution using Markov Chain Monte Carlo. These are very basic ingredient functions, and trying to document what they can be used for is an endless task. So IMO, if a user doesn't know how a CI relates to a quantile function, then they shouldn't be trying to compute a CI.

And FWIW, if you search "confidence interval scipy", for example, your hits will all be from StackOverflow detailing how to use SciPy's ingredient functions, not from the SciPy docs itself.

KronosTheLate: Fair points. So perhaps posing the question on Discourse, and answering it myself, is the way to go...

src/univariates.jl
```diff
@@ -367,8 +367,27 @@ logccdf(d::UnivariateDistribution, x::Real) = log(ccdf(d, x))
...
+julia> first_quartile = quantile(d, 0.25)
+-0.6744897501960818
+
+julia> second_quartile = quantile(d, 0.50)
+0.0
+
+julia> third_quartile = quantile(d, 0.75)
+0.6744897501960818
+
+julia> confidence_interval_95_percent = quantile(d, 0.025), quantile(d, 0.975)
```
mschauer, line 388:
```suggestion
julia> quantile(d, 0.025), quantile(d, 0.975) # 95%-probability interval 
```
(there is no parameter, so this is a *probability interval*)

src/univariates.jl
mschauer: Still let’s add examples, that’s always good, even an probability interval if we want.

KronosTheLate: Thanks to you both for your conciderations ^_^

KronosTheLate: > Fair points. So perhaps posing the question on Discourse, and answering it myself, is the way to go...

I ended up making the following post, for googlability:
https://discourse.julialang.org/t/how-to-compute-calculate-a-confidence-interval-ci/74794/2

src/univariates.jl
```diff
@@ -369,6 +369,24 @@ logccdf(d::UnivariateDistribution, x::Real) = log(ccdf(d, x))
...
+
+julia> second_quartile = quantile(d, 0.50)
+0.0
+
+julia> third_quartile = quantile(d, 0.75)
+0.6744897501960818
+
+julia> quantile(d, 0.025), quantile(d, 0.975) # 95%-probability interval 
+(-1.9599639845400592, 1.9599639845400576)
+```
```
devmotion, line 389: Maybe add:

````suggestion

See also [`median`](@ref)
```
````
And possibly also use it in the doctests, something like `quantile(d, 0.5) ≈ median(d)`.

KronosTheLate: Sounds like a good idea. I am not sure how doctests work - can I include `julia>` and the output without problem, just like I have?

Issue #1067, opened 2020-02-13
Make Distributions.jl GPU friendly


darsnack:
An increasing number of ML applications require sampling and reusing Distributions.jl instead of rolling your own sampler would be a major step forward. I have already forked this package to fix issues where parametric types are hardcoded to `Float64` instead of generic (to support a ML project I am working on). At this point, being able to sample on the GPU instead of moving data back and forth is pretty important to me. I am willing to put some effort into making this happen, but Distributions.jl covers a lot of stuff that I am not familiar with.

To that end, I'd like some help coming up with the list of changes that need to bring us towards this GPU friendly goal. Here's what I have so far:

- [ ] Replace hardcoded types (e.g. `Float64`) with generics (e.g. `AbstractFloat`/`Real`)
- [ ] Replicate tests currently intended for CPU to GPU
- [ ] A system to pack sampling over an array of distributions into a single operation

andreasnoack:
> I have already forked this package to fix issues where parametric types are hardcoded to Float64 instead of generic (to support a ML project I am working on).

Many distributions are already generic. It would be great if you could prepare PRs for the ones that aren't.

> A system to pack sampling over an array of distributions into a single operation

Can't you use broadcasting for this?

darsnack:
> Many distributions are already generic. It would be great if you could prepare PRs for the ones that aren't.

Yeah I'll start looking into this.

> Can't you use broadcasting for this?

Yes, we'll need to define broadcasting styles to do this packing.

Also another thing I thought of yesterday — the sampling code will need to be modified to not use scalar indexing. This would be very slow on the GPU, and most people has scalar indexing disabled when loading CuArrays.jl.

johnurbanik:
Curious if this is likely to be prioritized any time soon!

darsnack:
I got pulled into other projects that prevents me from focusing on this issue right now. But I know @srinjoyganguly is working on a DistributionsGPU.jl package.

johnurbanik:
Thanks for the prompt response! 

Including a link to srinjoyganguly's repo for anyone who ends up here: [DistributionsGPU.jl](https://github.com/srinjoyganguly/DistributionsGPU.jl)

azev77:
Are GPU features from DistributionsGPU.jl going to eventually be ported into Distributions.jl?
I really hope so

srinjoyganguly:
[...]

PR #1487, opened 2022-01-22
Use Julia implementations of `normpdf` etc. in StatsFuns

devmotion:
This PR replaces the implementations of `pdf(::Normal, ::Real)` etc. with calls to `StatsFuns.normpdf` etc. which are basically the same Julia implementations.

This removes code duplication (which e.g. means that one would not have to define ChainRules derivatives in both packages but it is sufficient to do so in StatsFuns) and it actually fixes subtle bugs such as reported in https://github.com/JuliaStats/Distributions.jl/issues/1067#issuecomment-1018545600 and arising from undesired promotions in `Normal(mu, sigma)` calls in the current implementation in Distributions.

Tests require https://github.com/JuliaStats/StatsFuns.jl/pull/132 which fixes a bug in the StatsFuns implementation covered by our tests.


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.25.41"
+version = "0.25.42"
```
devmotion, line 4: Should we bump the StatsFuns compat to ensure that https://github.com/JuliaStats/StatsFuns.jl/pull/132 is fixed and users don't run into this issue?

src/univariate/continuous/normal.jl
```diff
@@ -85,163 +85,10 @@ end
...
-            return oftype(q, Inf)
-        elseif isone(_p)
-            return oftype(q, -Inf)
-        else
-            return oftype(q, _μ)
-        end
-    end
-    return q
-end
+gradlogpdf(d::Normal, x::Real) = -(x - d.μ) / d.σ^2
```
sethaxen, line 91: Perhaps
```suggestion
gradlogpdf(d::Normal, x::Real) = (d.μ - x) / abs2(d.σ)
```
?

Project.toml
sethaxen: Yes I think so, once it's released.

sethaxen (commented):
This seems a good idea in general. Although it looks like StatsBase's normal quantile function promotes to `Float64` unnecessarily (see failures).

src/univariate/continuous/normal.jl
devmotion: I think `abs2` or `^` should not matter, last time I checked they compiled to exactly the same thing. Or do you know any reason why it should be preferred? I'll definitely simplify the calculations though, I just plugged in the definition of `zval` :+1:

devmotion:
```suggestion
gradlogpdf(d::Normal, x::Real) = (d.μ - x) / d.σ^2
```

sethaxen: > I think `abs2` or `^` should not matter, last time I checked they compiled to exactly the same thing. Or do you know any reason why it should be preferred?

While they do compile to the same thing, `abs2` is more informative for symbolics systems or an AD. e.g. ForwardDiff and ChainRules ADs will end up adding two floats with `abs2` vs multiplying two floats with `^`:

```julia
julia> using ForwardDiff

julia> @code_typed ForwardDiff.derivative(abs2, 3.5)
CodeInfo(
1 ─ %1 = Base.add_float(x, x)::Float64
│   %2 = Base.mul_float(1.0, %1)::Float64
└──      return %2
) => Float64

julia> @code_typed ForwardDiff.derivative(x -> x^2, 3.5)
CodeInfo(
1 ─ %1 = Base.mul_float(2.0, x)::Float64
│   %2 = Base.mul_float(1.0, %1)::Float64
└──      return %2
) => Float64
```

It's not a difference that I've been able to make appear in benchmarks, but this is why I generally use `abs2`.

devmotion: Interesting, I think one would just have to handle `Val{1}` and `Val{2}` cases in https://github.com/JuliaDiff/ForwardDiff.jl/blob/29332b12075e66e056843d3b4207d856f98772b3/src/dual.jl#L506-L516 without the multiplication in https://github.com/JuliaDiff/ForwardDiff.jl/blob/29332b12075e66e056843d3b4207d856f98772b3/src/dual.jl#L513. Due to `literal_pow` `abs2` and `^ 2` should be equally informative.

devmotion:
Should we bump the Julia bound to 1.3 or 1.6? The test errors on Julia 1.0 are caused by an ancient version of SpecialFunctios (0.8 :open_mouth:) that does not contain implementations of `erfcinv(::Rational)` (introduced in SpecialFunctions 1.2.0). All more recent versions of SpecialFunctions require Julia 1.3, hence for this specific issue it would be sufficient to drop Julia < 1.2.

devmotion:
Alternatively, we could just not run these failing tests on Julia < 1.3.

src/univariate/continuous/normal.jl
devmotion: I'll stick with `^2` for now. As mentioned, it is equally informative as `abs2`, and ForwardDiff even contains specialized derivatives for it. The difference here just comes from the fact that the implementation for `literal_pow` uses `2 * x` whereas DiffRules decided to implement the derivative of `abs2` with `x + x` (https://github.com/JuliaDiff/DiffRules.jl/blob/7b1c31ed5b026ea445fd42c9674391a45dc9c536/src/rules.jl#L12).

In fact, ChainRules also uses multiplication for `abs2`, as can seen by the rules (https://github.com/JuliaDiff/ChainRules.jl/blob/13a362c4b5069c6a83f936d9d9d8821e566232bb/src/rulesets/Base/fastmath_able.jl#L84-L94) or running this example:
```julia
julia> using ChainRules

julia> @code_typed last(ChainRules.rrule(abs2, 3.5))(1.0)
CodeInfo(
1 ─ %1 = Base.mul_float(2.0, ΔΩ)::Float64
│   %2 = Core.getfield(#self#, :z)::Float64
│   %3 = Base.mul_float(%1, %2)::Float64
│   %4 = Core.tuple($(QuoteNode(ChainRulesCore.NoTangent())), %3)::Tuple{ChainRulesCore.NoTangent, Float64}
└──      return %4
) => Tuple{ChainRulesCore.NoTangent, Float64}

julia> @code_typed ChainRules.frule((ChainRules.NoTangent(), 0.0), abs2, 3.5)
CodeInfo(
1 ─ %1 = Base.getfield(@_2, 2)::Float64
│   %2 = Base.mul_float(z, z)::Float64
│   %3 = Base.mul_float(z, %1)::Float64
│   %4 = Base.mul_float(2.0, %3)::Float64
│   %5 = Core.tuple(%2, %4)::Tuple{Float64, Float64}
└──      return %5
) => Tuple{Float64, Float64}
```
And CR also implements rules for `literal_pow` (https://github.com/JuliaDiff/ChainRules.jl/blob/13a362c4b5069c6a83f936d9d9d8821e566232bb/src/rulesets/Base/base.jl#L184-L200):
```julia
julia> @code_typed last(ChainRules.rrule(Base.literal_pow, ^, 3.5, Val(2)))(1.0)
CodeInfo(
1 ─ %1 = Core.getfield(#self#, :x)::Float64
│   %2 = Base.mul_float(2.0, %1)::Float64
│   %3 = Base.mul_float(%2, dy)::Float64
│   %4 = Core.tuple($(QuoteNode(ChainRulesCore.NoTangent())), $(QuoteNode(ChainRulesCore.NoTangent())), %3, $(QuoteNode(ChainRulesCore.NoTangent())))::Tuple{ChainRulesCore.NoTangent, ChainRulesCore.NoTangent, Float64, ChainRulesCore.NoTangent}
└──      return %4
) => Tuple{ChainRulesCore.NoTangent, ChainRulesCore.NoTangent, Float64, ChainRulesCore.NoTangent}

julia> @code_typed ChainRules.frule((ChainRules.NoTangent(), ChainRules.NoTangent(), 0.0, ChainRules.NoTangent()), Base.literal_pow, ^, 3.5, Val(2))
CodeInfo(
1 ─ %1 = Base.getfield(@_2, 3)::Float64
│   %2 = Base.mul_float(x, x)::Float64
│   %3 = Base.mul_float(2.0, x)::Float64
│   %4 = Base.mul_float(%3, %1)::Float64
│   %5 = Core.tuple(%2, %4)::Tuple{Float64, Float64}
└──      return %5
) => Tuple{Float64, Float64}
```

So I think it's fine to keep the more natural `^2`. I guess the compiler could decide to generate the same code for `x + x` and `2 * x` but this is something that we shouldn't have to take care in this package, I think.

devmotion:
OK, this PR is ready for review. I decided to change the Julia bound to 1.3 and SpecialFunctions to 1.2. These changes ensure that there are no regressions for environments that use older Julia or older SpecialFunctions versions.

devmotion:
Thanks for merging the PR @mschauer!

I guess Distributions doesn't follow the ColPrac guidelines officially but maybe we should adopt the policy that PR authors with merge rights merge their own PRs. Even after a PR is approved I sometimes notice an issue that has to be addressed or something minor that I'd like to change (wasn't the case here :slightly_smiling_face:), and then it's nice if the PR is not merged yet and I can update it.

alyst:
Is it intentional that this PR removed `zval()`, `xval()` and potentially something else?

devmotion:
Yes. `xval` and `zval` are not used anymore.

alyst:
But are they still exported and defined for many other distributions?

devmotion:
They are just internal helpers that simplify the code a bit and make it more readable. BTW StatsFuns uses internal helpers of the same name: https://github.com/JuliaStats/StatsFuns.jl/blob/13e231a0a22e716426b73cb87ff3b8b24e33aaf1/src/distrs/norm.jl#L3

devmotion:
I don't understand, why would you you want them to be imported? They are not exported by StatsFuns because they are an internal implementation detail. So if you want to use them you could just call `StatsFuns.xval` etc., with the risk of them being removed in a future non-breaking release. Additionally, the version in StatsFuns is defined for real arguments whereas the ones in Distributions are defined for `Distribution`s. But also in Distributions they are only internal functions and not part of the API - they may be removed at any time and they are only defined for a handful of functions.

alyst:
I don't want `StatsFuns.zval` to be imported.
In my code, I was using the ones from *Distributions.jl* that have the signature `zval(d::Distribution, x::Number)` (that's what I mean by "polymorphic").
So I was wondering if it makes sense to restore 'zval(d::Normal, x::Number)` here.
I understand that non-exported things can break at any time, I just wonder what is your recommendation to get the z-value for a given distribution (symmetric continuous univariate)?
For me, `zval(d::Distribution, x::Number)` was a convenient interface for that. Obviously, for some distributions it doesn't make sense to define `zval()`.

As for the ones in *StatsFuns.jl*, since they are not exported, the name clash with *Distributions.jl* is not critical, but to be consistent with the other normal-related functions there, they could be renamed into *normzval/normxval()*.

devmotion:
Taking one step back, what exactly do you expect that `zval(d::Distribution, x::Number)` should return? In Distributions it is used as an auxiliary function and also for some asymmetric distributions in a way that allows us to simplify the `logpdf` etc. implementations. Many definitions are identical to `zval(d, x) = (x - location(d)) / scale(d)` but this is not satisfied in general.

alyst:
> Taking one step back, what exactly do you expect that zval(d::Distribution, x::Number) should return?
> Many definitions are identical to zval(d, x) = (x - location(d)) / scale(d) but this is not satisfied in general.

That's what I expect (for the transformed distribution the mean is zero, the std is one), and that's why I specify that it might make sense to define it only for symmetric continuous univariate.

devmotion:
In that case, can't you define and use this function `myzval(d, x) = (x - location(d)) / scale(d)` in your code? `scale` and `location` are part of the API, so this seems a clean and safe solution.

alyst:
I can, I was just wondering if it could be implemented by the package, given that `zval()` is already defined for a dozen of distributions.

PR #1489, opened 2022-01-22
Allow specifying non-existant bounds to truncated using nothing

### LINKED ISSUE #1488: Supporting partially truncated distributions
#1470 adopts an approach for  a `censored` of using `nothing` instead of `Inf` to indicate "unbounded in a direction". As discussed on Slack, there are some advantages to doing something similar for `truncated`: 
1. it lets us compile away some computation
2. it avoids cases like the one raised on Slack where a (log)cdf function is non-differentiable at infinity, so providing an infinite bound prevents one from using AD (without workarounds)
3. It allows for providing special overloads for distributions with partial truncations. e.g. a left-truncation of ~`GeneralizedPareto`~, `Pareto`, and `Exponential` is a distribution of the same type (Edit: or a translation of it). Currently we can't type-stably special-case the return value here, so a `Truncated` is returned.
4. It's just cleaner. If a distribution isn't truncated in a direction then there is no bound; not that there is a bound but it's infinite.

Changing `Truncated` to accept `Nothing` bounds would be a breaking change and should be saved for the next breaking release, but we can prep for this in a non-breaking way with the following overloads:
```julia
function truncated(
    d::UnivariateDistribution;
    lower::Union{Real,Nothing}=nothing,
    upper::Union{Real,Nothing}=nothing,
)
    return truncated(d, lower, upper)
end
function truncated(d::UnivariateDistribution, ::Nothing, u::Real)
    # (log)ucdf = (log)tp = (log) P(X ≤ u) where X ~ d
    logucdf = logtp = logcdf(d, u)
    ucdf = tp = exp(logucdf)
    Truncated(d, promote(oftype(float(u), -Inf), u, zero(ucdf), ucdf, tp, logtp)...)
end
function truncated(d::UnivariateDistribution, l::Real, ::Nothing)
    # (log)lcdf = (log) P(X < l) where X ~ d
    loglcdf = if value_support(typeof(d)) === Discrete
        logsubexp(logcdf(d, l), logpdf(d, l))
    else
        logcdf(d, l)
    end
    lcdf = exp(loglcdf)
    # (log)tp = (log) P(l ≤ X) where X ∼ d
    logtp = log1mexp(loglcdf)
    tp = exp(logtp)
    Truncated(d, promote(l, u, lcdf, one(lcdf), tp, logtp)...)
end
```

Then the following works:
```julia
julia> truncated(Normal(0.0, 1.0); lower=0)
Truncated(Normal{Float64}(μ=0.0, σ=1.0), range=(0.0, Inf))

julia> truncated(Normal(); lower=0)
Truncated(Normal{Float64}(μ=0.0, σ=1.0), range=(0.0, Inf))

julia> truncated(Normal(); upper=0)
Truncated(Normal{Float64}(μ=0.0, σ=1.0), range=(-Inf, 0.0))

julia> truncated(Normal(), -2, 2)
Truncated(Normal{Float64}(μ=0.0, σ=1.0), range=(-2.0, 2.0))

julia> truncated(Normal(), 0, Inf)
┌ Warning: `truncated(d, l, Inf)` is deprecated, use `truncated(d; lower=l)` instead.
│   caller = ip:0x0
└ @ Core :-1
Truncated(Normal{Float64}(μ=0.0, σ=1.0), range=(0.0, Inf))
```

The latter case presumes we also add a forced `Base.depwarn` call to the default fallback of `truncated`, which would smoothen the transition when `Truncated` is changed.

sethaxen:
Fixes #1488

### DIFF
--- a/src/truncate.jl
+++ b/src/truncate.jl
@@ -64,6 +64,8 @@ end
 truncated(d::UnivariateDistribution, ::Nothing, ::Nothing) = d
 function truncated(d::UnivariateDistribution, l::T, u::T) where {T <: Real}
     l <= u || error("the lower bound must be less or equal than the upper bound")
+    l == -Inf && Base.depwarn("`truncated(d, Inf, u)` is deprecated, use `truncated(d; upper=u)` instead.", nothing; force=true)
+    u == Inf && Base.depwarn("`truncated(d, l, Inf)` is deprecated, use `truncated(d; lower=l)` instead.", nothing; force=true)
 
     # (log)lcdf = (log) P(X < l) where X ~ d
     loglcdf = if value_support(typeof(d)) === Discrete
--- a/src/truncate.jl
+++ b/src/truncate.jl
@@ -72,7 +72,7 @@ function truncated(d::UnivariateDistribution, l::T, u::T) where {T <: Real}
         force_kwarg...
     )
     u == Inf && Base.depwarn(
-        "`truncated(d, l, Inf)` is deprecated. Please `truncated(d; lower=l)` instead.",
+        "`truncated(d, l, Inf)` is deprecated. Please use `truncated(d; lower=l)` instead.",
         :truncated;
         force_kwarg...
     )
--- a/src/truncate.jl
+++ b/src/truncate.jl
@@ -23,7 +23,29 @@ function truncated(
 )
     return truncated(d, lower, upper)
 end
+function truncated(d::UnivariateDistribution, ::Nothing, u::Real)
+    # (log)ucdf = (log)tp = (log) P(X ≤ u) where X ~ d
+    logucdf = logtp = logcdf(d, u)
+    ucdf = tp = exp(logucdf)
 
+    Truncated(d, promote(oftype(float(u), -Inf), u, zero(ucdf), ucdf, tp, logtp)...)
+end
+function truncated(d::UnivariateDistribution, l::Real, ::Nothing)
+    # (log)lcdf = (log) P(X < l) where X ~ d
+    loglcdf = if value_support(typeof(d)) === Discrete
+        logsubexp(logcdf(d, l), logpdf(d, l))
+    else
+        logcdf(d, l)
+    end
+    lcdf = exp(loglcdf)
+
+    # (log)tp = (log) P(l ≤ X) where X ∼ d
+    logtp = log1mexp(loglcdf)
+    tp = exp(logtp)
+
+    Truncated(d, promote(l, oftype(float(l), Inf), lcdf, one(lcdf), tp, logtp)...)
+end
+truncated(d::UnivariateDistribution, ::Nothing, ::Nothing) = d
 function truncated(d::UnivariateDistribution, l::T, u::T) where {T <: Real}
     l <= u || error("the lower bound must be less or equal than the upper bound")
 
--- a/src/truncate.jl
+++ b/src/truncate.jl
@@ -64,15 +64,16 @@ end
 truncated(d::UnivariateDistribution, ::Nothing, ::Nothing) = d
 function truncated(d::UnivariateDistribution, l::T, u::T) where {T <: Real}
     l <= u || error("the lower bound must be less or equal than the upper bound")
+    force_kwarg = VERSION ≥ v"1.5.2" ? (force=true,) : ()
     l == -Inf && Base.depwarn(
         "`truncated(d, -Inf, u)` is deprecated. Please use `truncated(d; upper=u)` instead.",
         :truncated;
-        force=true,
+        force_kwarg...
     )
     u == Inf && Base.depwarn(
         "`truncated(d, l, Inf)` is deprecated. Please `truncated(d; lower=l)` instead.",
         :truncated;
-        force=true,
+        force_kwarg...
     )
 
     # (log)lcdf = (log) P(X < l) where X ~ d
--- a/src/truncate.jl
+++ b/src/truncate.jl
@@ -65,16 +65,13 @@ end
 truncated(d::UnivariateDistribution, ::Nothing, ::Nothing) = d
 function truncated(d::UnivariateDistribution, l::T, u::T) where {T <: Real}
     l <= u || error("the lower bound must be less or equal than the upper bound")
-    force_kwarg = VERSION ≥ v"1.5.2" ? (force=true,) : ()
     l == -Inf && Base.depwarn(
         "`truncated(d, -Inf, u)` is deprecated. Please use `truncated(d; upper=u)` instead.",
-        :truncated;
-        force_kwarg...
+        :truncated,
     )
     u == Inf && Base.depwarn(
         "`truncated(d, l, Inf)` is deprecated. Please use `truncated(d; lower=l)` instead.",
-        :truncated;
-        force_kwarg...
+        :truncated,
     )
 
     # (log)lcdf = (log) P(X < l) where X ~ d
--- a/src/truncate.jl
+++ b/src/truncate.jl
@@ -13,6 +13,7 @@ The function falls back to constructing a [`Truncated`](@ref) wrapper.
 To implement a specialized truncated form for distributions of type `D`, the method
 `truncated(d::D, l::T, u::T) where {T <: Real}` should be implemented.
 """
+truncated
 function truncated(d::UnivariateDistribution, l::Real, u::Real)
     return truncated(d, promote(l, u)...)
 end
--- a/src/truncate.jl
+++ b/src/truncate.jl
@@ -23,13 +23,13 @@ The function falls back to constructing a [`Truncated`](@ref) wrapper.
 
 # Implementation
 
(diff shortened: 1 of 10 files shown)


src/truncate.jl
```diff
@@ -1,24 +1,71 @@
 """
-    truncated(d::UnivariateDistribution, l::Real, u::Real)
+    truncated(d0::UnivariateDistribution; [lower::Real], [upper::Real])
+    truncated(d0::UnivariateDistribution, lower::Real, upper::Real)
```
devmotion, line 3: Should we remove this from the docstring to encourage the use of the kwargs version? Or maybe even remove it and use a `_truncated(d, lower, upper)` for dispatch?

sethaxen: > Should we remove this from the docstring to encourage the use of the kwargs version?

I would prefer it if a user who used to write `truncated(Normal(), -2, 2)` didn't now have to use the more verbose `truncated(Normal(); lower=-2, upper=2)`. Not only is it more verbose, but the former was easy to understand, and now every usage of `truncated` before now needs to be updated. We can't really avoid that for the left- and right-truncated versions if we want to discontinue using `Inf` to mean "unbounded".

I don't see using the args version as being a problem; it's using `Inf` to mean "unbounded" that's a problem. The docstring includes the `nothing` signatures, so they are free to just replace all `Inf`s with `nothing`, but I suspect they will eventually adopt the more explicit notation in the left- or right-truncated cases, and hence that's what we recommend in the examples.

Currently the deprecation warning is very strong (which may not be ideal), and I expect users using partial bounds will instantly switch to the new syntax to avoid the annoying warnings. I suppose we could add a note to the documentation about deprecated behavior.

> Or maybe even remove it and use a `_truncated(d, lower, upper)` for dispatch?

I thought about that, but then those wanting to specialize `truncated` for their distribution need to overload something that looks like a private function. Plus there are the points given above.

devmotion: Sounds reasonable. I'm also not a fan of making unexported `_...` functions part of the interface (as unfortunately still the case with `_rand!`...).

src/truncate.jl
```diff
@@ -1,24 +1,81 @@
...
+```
 
 The function falls back to constructing a [`Truncated`](@ref) wrapper.
 
 # Implementation
 
-To implement a specialized truncated form for distributions of type `D`, the method
-`truncated(d::D, l::T, u::T) where {T <: Real}` should be implemented.
+To implement a specialized truncated form for distributions of type `D`, one of the
+following methods should be implemented:
```
devmotion, line 27:
```suggestion
To implement a specialized truncated form for distributions of type `D`, one or more of the
following methods should be implemented:
```

src/truncate.jl
```diff
@@ -1,24 +1,81 @@
...
+- `truncated(d0::D, l::T, u::T) where {T <: Real}`: interval-truncated
+- `truncated(d0::D, ::Nothing, u::Real)`: right-truncated
+- `truncated(d0::D, l::Real, u::Nothing)`: left-truncated
 """
+truncated
```
devmotion, line 32: According to the [documentation](https://docs.julialang.org/en/v1/manual/documentation/#Functions-and-Methods-2),
```suggestion
function truncated end
```
is preferred.

devmotion (commented):
Can you test the deprecations (probably sufficient to add `@test_deprecated` to deprecated calls)?

sethaxen:
> Can you test the deprecations (probably sufficient to add `@test_deprecated` to deprecated calls)?

I don't think `@test_deprecated` works with just `Base.depwarn`:

```julia
julia> function foo(x)
           Base.depwarn("blah", :foo)
           x
       end
foo (generic function with 1 method)

julia> bar(x) = x
bar (generic function with 1 method)

julia> baz(x) = x
baz (generic function with 1 method)

julia> @deprecate bar(x) baz(x)
bar (generic function with 1 method)

julia> using Test

julia> @test_deprecated bar(1.0)
1.0

julia> @test_deprecated baz(1.0)
Log Test Failed at REPL[25]:1
  Expression: $(Expr(:escape, :(baz(1.0))))
  Log Pattern: (:warn, r"deprecated"i, Ignored(), :depwarn) match_mode = :any
  Captured Logs: 

julia> @test_deprecated foo(1.0)
Log Test Failed at REPL[26]:1
  Expression: $(Expr(:escape, :(foo(1.0))))
  Log Pattern: (:warn, r"deprecated"i, Ignored(), :depwarn) match_mode = :any
  Captured Logs: 
    Test.LogRecord(Warn, "blah", Main, :depwarn, (Ptr{Nothing} @0x00000001090b8b05, :foo), "./none", 0, Base.Pairs{Symbol, Any, Tuple{Symbol, Symbol}, NamedTuple{(:caller, :maxlog), Tuple{Base.StackTraces.StackFrame, Int64}}}(:caller => (::var"#11#12")() at none:0, :maxlog => 1))

```

But since the warning is forced, I can catch it and test for it.

devmotion:
> I don't think @test_deprecated works with just Base.depwarn:

It does but the message has to match some regex (something like "is deprecated").

devmotion (approved):
Looks good to me, anything that you would like to change or add?

sethaxen:
Nothing else for this PR. I plan to make another PR later with extra `truncated` overloads for some distributions.

src/truncate.jl
```diff
@@ -1,24 +1,81 @@
...
+```
+where ``f_{d_0}(x)`` is the probability density (mass) function of ``d_0``.
+
+The function throws an error if ``l > u``.
+
+```julia
+truncated(d0; lower=l)           # d0 left-censored to the interval [l, Inf)
+truncated(d0; upper=u)           # d0 right-censored to the interval (-Inf, u]
+truncated(d0; lower=l, upper=u)  # d0 interval-censored to the interval [l, u]
+truncated(d0, l, u)              # d0 interval-censored to the interval [l, u]
```
devmotion, line 19:
```suggestion
truncated(d0; lower=l)           # d0 left-truncated to the interval [l, Inf)
truncated(d0; upper=u)           # d0 right-truncated to the interval (-Inf, u]
truncated(d0; lower=l, upper=u)  # d0 truncated to the interval [l, u]
truncated(d0, l, u)              # d0 truncated to the interval [l, u]
```

src/truncate.jl
```diff
@@ -1,24 +1,81 @@
...
+    l == -Inf && Base.depwarn(
+        "`truncated(d, -Inf, u)` is deprecated. Please use `truncated(d; upper=u)` instead.",
+        :truncated;
+        force_kwarg...
+    )
+    u == Inf && Base.depwarn(
+        "`truncated(d, l, Inf)` is deprecated. Please use `truncated(d; lower=l)` instead.",
+        :truncated;
+        force_kwarg...
+    )
```
devmotion, line 78: Do we really want to force the deprecation warnings? This could affect performance quite severely, e.g., typically sampling with Turing is basically impossible if there are deprecation in the model. Since deprecation warnings are only enabled in tests by default in recent Julia versions, it will be less disruptive if we don't use `force`. Usually (e.g., also for `Truncated`) we use regular deprecations.

devmotion (commented):
Sorry, I completely forgot this PR.

I still have one question regarding the deprecations I noticed.

src/truncate.jl
sethaxen: I was thinking in this case it's mostly end users who will need to update their code, not package maintainers. e.g. a lot of Turing models and tutorials. If we don't force the deprecation warning, then likely very few of those users will be aware of the change before the next breaking release. I'm not so much concerned about performance, since all they have to do is switch to the new recommended syntax to avoid the deprecation warnings entirely. And the forced warning will fill their REPL pretty quickly, so I doubt there will be many hold-outs.

But maybe this is not a special case, and there's no point deviating from the norm in this package.

devmotion: I can see the possible benefits but they would apply to many other deprecations, also in other packages. The default, however, in this package and the `@deprecate` macro in general is to not enforce deprecation warnings. And I don't think there's a completely convincing reason, at least not for me, why we should deviate from this behaviour here (but e.g. not when deprecating `Truncated`).

sethaxen: Okay, I removed the forcing.

devmotion:
The DistributionsAD test failure is unrelated, it also shows up in other PRs and is caused by a recent change (projection of UniformScaling) in CR.

devmotion:
Thanks Seth!

devmotion:
@sethaxen On a second thought, maybe we should revert the deprecations (but keep everything else as is). It seems unnecessary to disallow these choices explicitly, as eg in an optimization algorithm you might converge to infinity. Of course, it should still be recommended to use nothing or kwargs if the infinite/non-existing bound is fixed.

sethaxen:
That's fair. But then we should probably add back some tests changed to the new syntax so that we catch regressions.

torfjelde:
You're the best @sethaxen

PR #1490, opened 2022-01-23
Add fast path for `loglikelihood` of `Uniform`

devmotion:
On master:
```julia

julia> a = rand();

julia> b = a + rand();

julia> x = rand(Uniform(a, b), 10_000);

julia> @btime loglikelihood($(Uniform(a, b)), $x);
  77.560 μs (1 allocation: 16 bytes)
```
With this PR:
```julia

julia> a = rand();

julia> b = a + rand();

julia> x = rand(Uniform(a, b), 10_000);

julia> @btime loglikelihood($(Uniform(a, b)), $x);
  8.064 μs (1 allocation: 16 bytes)
```

### DIFF
--- a/src/univariate/continuous/uniform.jl
+++ b/src/univariate/continuous/uniform.jl
@@ -109,6 +109,13 @@ function cf(d::Uniform, t::Real)
     cis(v) * (sin(u) / u)
 end
 
+#### Fast path for `loglikelihood`
+
+function loglikelihood(d::Uniform, x::AbstractArray{<:Real})
+    diff = d.b - d.a
+    return all(Base.Fix1(insupport, d), x) ? -length(x) * log(diff) : log(zero(diff))
+end
+
 #### Affine transformations
 
 Base.:+(d::Uniform, c::Real) = Uniform(d.a + c, d.b + c)
@@ -144,6 +151,7 @@ end
 
 # ChainRules definitions
 
+## logpdf
 function ChainRulesCore.frule((_, Δd, _), ::typeof(logpdf), d::Uniform, x::Real)
     # Compute log probability
     a, b = params(d)
@@ -157,7 +165,6 @@ function ChainRulesCore.frule((_, Δd, _), ::typeof(logpdf), d::Uniform, x::Real
 
     return Ω, ΔΩ
 end
-
 function ChainRulesCore.rrule(::typeof(logpdf), d::Uniform, x::Real)
     # Compute log probability
     a, b = params(d)
@@ -168,7 +175,7 @@ function ChainRulesCore.rrule(::typeof(logpdf), d::Uniform, x::Real)
     # Define pullback
     function logpdf_Uniform_pullback(Δ)
         Δa = Δ / diff
-        Δd = if insupport 
+        Δd = if insupport
             ChainRulesCore.Tangent{typeof(d)}(; a=Δa, b=-Δa)
         else
             ChainRulesCore.Tangent{typeof(d)}(; a=zero(Δa), b=zero(Δa))
@@ -179,3 +186,39 @@ function ChainRulesCore.rrule(::typeof(logpdf), d::Uniform, x::Real)
     return Ω, logpdf_Uniform_pullback
 end
 
+## loglikelihood
+function ChainRulesCore.frule((_, Δd, _), ::typeof(loglikelihood), d::Uniform, x::AbstractArray{<:Real})
+    # Compute log likelihood
+    a, b = params(d)
+    n = length(x)
+    all_insupport = all(Base.Fix1(insupport, d), x)
+    diff = b - a
+    Ω = all_insupport ? -n * log(diff) : log(zero(diff))
+
+    # Compute tangent
+    Δdiff = n * (Δd.a - Δd.b)
+    ΔΩ = (all_insupport ? Δdiff : zero(Δdiff)) / diff
+
+    return Ω, ΔΩ
+end
+function ChainRulesCore.rrule(::typeof(loglikelihood), d::Uniform, x::AbstractArray{<:Real})
+    # Compute log likelihood
+    a, b = params(d)
+    n = length(x)
+    all_insupport = all(Base.Fix1(insupport, d), x)
+    diff = b - a
+    Ω = all_insupport ? -n * log(diff) : log(zero(diff))
+
+    # Define pullback
+    function loglikelihood_Uniform_pullback(Δ)
+        Δa = n * Δ / diff
+        Δd = if all_insupport
+            ChainRulesCore.Tangent{typeof(d)}(; a=Δa, b=-Δa)
+        else
+            ChainRulesCore.Tangent{typeof(d)}(; a=zero(Δa), b=zero(Δa))
+        end
+        return ChainRulesCore.NoTangent(), Δd, ChainRulesCore.ZeroTangent()
+    end
+
+    return Ω, loglikelihood_Uniform_pullback
+end
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.25.41"
+version = "0.25.42"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/test/uniform.jl
+++ b/test/uniform.jl
@@ -6,6 +6,19 @@ using Test
     # affine transformations
     test_affine_transformations(Uniform, rand(), 4 + rand())
 
+    @testset "loglikelihood" begin
+        dist = Uniform(rand(), 2 + rand())
+        for dims in ((10,), (2, 5), (1, 2, 3, 1))
+            # all values in support
+            x = rand(dist, dims...)
+            @test @inferred(loglikelihood(dist, x)) ≈ sum(logpdf.(dist, x))
+
+            # one value not in support
+            x[rand(1:length(x))] = -1
+            @test @inferred(loglikelihood(dist, x)) == -Inf
+        end
+    end
+
     @testset "ChainRules" begin
         # run test suite for values in the support
         dist = Uniform(- 1 - rand(), 1 + rand())
@@ -13,6 +26,11 @@ using Test
         for x in (rand(), -rand())
             test_frule(logpdf, dist ⊢ tangent, x)
             test_rrule(logpdf, dist ⊢ tangent, x)
+        end
+        for dim in ((10,), (2, 5), (1, 2, 3, 1))
+            x = 2 .* rand(dim...) .- 1
+            test_frule(loglikelihood, dist ⊢ tangent, x)
+            test_rrule(loglikelihood, dist ⊢ tangent, x)
         end
 
         # check manually that otherwise derivatives are zero (FiniteDifferences returns NaN)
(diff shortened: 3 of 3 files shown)


src/univariate/continuous/uniform.jl
```diff
@@ -109,6 +109,13 @@ function cf(d::Uniform, t::Real)
     cis(v) * (sin(u) / u)
 end
 
+#### Fast path for `loglikelihood`
+
+function loglikelihood(d::Uniform, x::AbstractArray{<:Real})
+    diff = d.b - d.a
+    return all(Base.Fix1(insupport, d), x) ? -length(x) * log(diff) : log(zero(diff))
```
sethaxen, line 116: Why not
```suggestion
    all_insupport = ChainRulesCore.@ignore_derivatives all(Base.Fix1(insupport, d), x)
    return all_insupport ? -length(x) * log(diff) : log(zero(diff))
```
? Then the AD only needs to differentiate `log` and the ternary, and a custom rule is not needed (Zygote will still be a little slower since hand-derived rules are almost always faster than Zygote's generated codes, especially if there's control flow).

devmotion: I considered something similar first 😄

Since the code that still has to be differentiated is basically like `logpdf` we  should include a custom rule though - we do for `logpdf` and hence it seems we should also for `loglikelihood` which is an (at least as) integral part of e.g. Turing. The rule for `logpdf` was included in DistributionsAD (well, still is but there's a PR that fixes it) - as type piracy, of course, but it is only possible to fix it without performance regression if we define it in Distributions instead.

devmotion: Oh, and even more importantly it's not clear that AD returns the desired derivatives for the boundaries and outside of the support.

sethaxen: I guess if this is part of transferring from DistributionsAD to here, then it should be here.



> Oh, and even more importantly it's not clear that AD returns the desired derivatives for the boundaries and outside of the support.

Why would it not? As far as I can tell the rules included here just manually do what the AD would do.

devmotion: It's exactly what Zygote would do. I'm not sure about general AD though, eg. also FiniteDifferences returns `NaN` instead in these cases.

sethaxen: > FiniteDifferences returns `NaN` instead in these cases.

FD is not an AD. 🙂

devmotion: I'm not worried about FD specifically, just worried that the same NaN issues would occur with AD. But maybe that can't happen?

devmotion: Ah, there's another problem and something I forgot to fix in https://github.com/JuliaStats/Distributions.jl/pull/1459: We want to return 0 also at the boundary. This is a bug in the current implementation, and also not what you get from Zygote without rule.

devmotion: I fixed the bugs, added more tests, and updated the benchmarks above.

sethaxen: Ah yes, I remember reading that discussion. Interesting, I would have thought returning the same thing at the boundary as for the rest of the support would be best, but from the perspective of someone maximizing the likelihood over the parameters starting with bounds that include all `x`, this would result in the maximum being achieved with a non-zero gradient (the bounds being the extrema of `x`), the next step being to shorten the bounds to not include all `x`. So yes there's a case to be made for returning 0 at the boundary. Still not 100% convinced this is good enough of a reason to override the default AD behavior.

devmotion: Yeah, it's clearly not the only choice.

To recap, if we neglect `x` values outside the support completely, then we have to study `f(a) = -log(b - a)`, defined on `(-Inf, x]` (`b` and `x <= b` fixed), and `g(b) = -log(b - a)`, defined on `[x, Inf)` (`a` and `x >= a` fixed). Clearly, for `a < x` we have `f'(a) = 1/ (b - a) > 0` and for `b > x` we have `g'(b) = 1 / (a - b) < 0`.

A straightforward, and I think reasonable, way to define "derivatives" at the non-differentiable points `a = x` and `b = x` would to use the limit from the left and the right, respectively - as also mentioned as one approach in the CR docs. Ie., to set `f'(x) = 1 / (b - a)` and `g'(x) = 1 / (a - b)`. However, this means that the derivatives at the global maxima of `f` and `g` would be non-zero - and hence not satisfy the property, useful in practice and also mentioned in the CR docs, that derivatives at local maxima and minima should be zero.

Thus another approach would be to set `f'(x) = g'(x) = 0`. This is what we wanted to do and what this PR implements. It satisfy the "zero at extrema" property and hence I still think it is the more useful definition in practice as eg. it won't push you over the edge in a gradient descent algorithm. If one uses the convention to set the derivatives of `f` and `g` to 0 if `x` is not in the support, then it even corresponds to the limits from the right and the left, respectively. And I think the increased usefulness in practice outweighs the inconvenience of having to define custom rules.

(Another alternative would be to error or set the derivative to `NaN` but I agree with the CR docs that this is the least useful approach.)

devmotion: The last comment made me even more certain that we should define a rule in any case: Regardless of the convention we use, since there is no mathematically correct definition of these derivatives we don't have any guarantees that AD will necessarily adhere to our convention. Similar to the `abs(x)` example in the CR docs, slightly different implementations of the primal or internals of some AD backend may yield different derivatives following a different convention.

devmotion: And my comment made me realize that, of course, we have to distinguish between `x = a` and `x = b`. And hence the derivatives don't follow our desired convention yet 😂

mschauer: Something we didn't use as a criterion is that the derivative we return could well be an convenient element of the set of sub gradients.

mschauer: That would suggest `0` and not `NaN` for `abs` for example.

devmotion: I actually considered it and mentioned it in the original PR, I just didn't mention it above because I hadn't thought about it carefully and noticed that only `f` is a convex function.

I'm used to definitions of subderivatives in open sets, but anyway in the case of `f` a subderivative `c` at `x` would have to satisfy `f(a) - f(x) >= c (a - x)` for all `a <= x`. We have `f(a) - f(x) = log(b - x) - log(b - a) < 0` for `a < x`, and hence 0 would not be a subderivative (if we assume it makes sense to talk about subderivatives at the boundary). However, since `f(a) - f(x) = log((b - x) / (b - a)) >= 1 - (b - a) / (b - x) = (a - x) / (b - x)` for all `a <= x` (logarithm inequality), `1 / (b - x)` is a subderivative at `x` (and `[1 / (b - x), Inf)` is a subset of subderivatives) [assuming `b != x`].

Similarly, for `g` (which is not convex, another property that's often assumed) a subderivative `c` at `x` would have to satisfy `g(b) - g(x) = log(x - a) - log(b - a) >= c(b - x)` for all `b >= x`. Similarly, `log(x - a) - log(b - a) < 0` for `b > x` and hence 0 is not a subderivative. We have `g(b) - g(x) >= 1 - (b - a) / (x - a) = (b - x) / (a - x)` for all `x >= b`, and thus `1 / (a - x)` is a subderivative at `x` (and `(-Inf, 1 / (a - x)]` is a set of subderivatives) [assuming `a != x`].

So unsurprisingly the subderivatives at the boundary `x` would include the limits of `f'` and `g'` from the left and right side, respectively. Whereas 0 is no subderivative.

The set of subderivatives also provides only a rough guidance - for the (convex!) function `abs` the set of subderivatives at 0 is [-1, 1], so it's not a clear argument for choosing 0 but mainly not for not choosing `NaN`. Which is a bad idea anyway usually.

sethaxen: I'm not certain subderivatives are reasonable concepts for non-convex functions. e.g. the geometric picture of the subderivative is the set of slopes of all tangent lines touching the curve at the desired point such that the tangent lines are completely beneath the curve, which is not possible here, since we have a step function at the bounds. But the subderivative is just the special case for convex functions of the more general "use the derivatives at either side of the point to define a set", in which case we basically have `[0, 1/(b-a)]` being the set of reasonable options. I think we've agreed `NaN` is a bad option. `Inf` is almost as bad. @devmotion it sounds like you still favor 0, while AD would default to `1/(b-a)`. From the perspective of maximizing the log-likelihood, I agree that `0` makes the most sense, but this set also includes what the AD would do by default, so I still am not convinced we should override that.

> The last comment made me even more certain that we should define a rule in any case: Regardless of the convention we use, since there is no mathematically correct definition of these derivatives we don't have any guarantees that AD will necessarily adhere to our convention. Similar to the `abs(x)` example in the CR docs, slightly different implementations of the primal or internals of some AD backend may yield different derivatives following a different convention.

By this reasoning, every conditional expression in Julia needs a custom rule where someone has injected semantic information about what constitutes the most reasonable derivative for optimization purposes. But AD is widely used, and that doesn't seem to be necessary for most functions. But for extremely common functions, and especially those that have discontinuities at a region that is likely to be the target of optimization, these are useful discussions to have. The question is, is anyone going to (reasonably) try to maximize the likelihood of a uniform distribution using the gradient, when the MLE is known in closed form? Will this be regularly done? And by this same reasoning, must we add custom rules for all distributions with bounded support to special-case the bounds? My concern is that this seems like over-engineering.

sethaxen: > And by this same reasoning, must we add custom rules for all distributions with bounded support to special-case the bounds?

To clarify, I'm referring to distributions whose support is determined by their parameters. e.g. `AffineDistribution` or `Pareto`.

mschauer: I think the discussion of sub gradients points in the right direction: It is difficult to come with a rule and guarantee what derivative is returned at these points, but it might be possible to give a guarantee that we always return an (arbitrary) element of a well-described set, and if something is differentiable, that set should contain only one point.

mschauer: > The question is, is anyone going to (reasonably) try to maximize the likelihood of a uniform distribution using the gradient, when the MLE is known in closed form?

A stray uniform might show up somewhere in a more complicated composite likelihood from a PPL model?

devmotion: Yeah, I'm still in favour of 0s. There's no correct choice and hence I think we should choose out of all somewhat reasonable values the ones that are most useful in practice.

There seems to also be a misunderstanding about why I think that a rule should be defined in any case. My point is, if a correct derivative exists and AD returns something else, then it's a bug of the AD - we might fix it by adding a CR definition but there's no need to define a rule in advance for correctness (might want to do it for performance, mutation issues, etc. though). However, if there's no correct derivative, then AD is allowed to do follow different conventions, possibly depending on the implementation, but it's not a bug of AD if it's something else than what we prefer. So a rule is the only way we can ensure that we always get the values we want to get. But it's not relevant if the code contains `if` statements, this argument is solely based on the non-existence of the derivative.

The rule here exists both for performance and correctness reasons. Initially, it was defined in DistributionsAD (and used NaN 🙈), and there's still one eg for Tracker (IIRC there was also one for ForwardDiff at some point, not sure what it returns, maybe `NaN` outside the support?). In any case its removal would lead to performance regressions and mean we don't have control over the values at the boundary and outside the support anymore.

It's not a completely unusual example either, actually for multiple distributions we already define CR definitions, both for correctness and performance: https://github.com/JuliaStats/StatsFuns.jl/blob/master/src/chainrules.jl

So I think the main question is, do we stick with the 0s as discussed previously? Or do we use something else. If this is decided, I can finish the PR.

matbesancon:
@devmotion fixed the conflict here, what's the status?

devmotion:
Thanks for the ping!

I postponed the PR because I wanted to think about it carefully once more. I came to the conclusion that
- in my opinion it is useful to add definitions for the derivatives of the log pdf and loglikelihood, both to ensure that the computation is correct in the way we define it, regardless of the AD-backend specific quirks, and to ensure that computations are fast for such an important distribution,
- in my opinion, one should define the derivatives as NaN for values outside of the support, regardless of how (not) useful NaN is in general. Otherwise if one defines the derivative of `logpdf(::Uniform, x::Real)` and `loglikelihood(::Uniform, xs::AbstractArray{<:Real})` as zero if `x` or at least one element of `xs` is outside the support, as in this PR, additivity is violated: the derivative of `sum(logpdf.(::Uniform, xs))` will be non-zero and hence different from the derivative of `loglikelihood`. Thus it seems we want a "destructive" value as derivative instead, and (a type-stable) NaN seems the natural choice.
- I still think it makes sense to define the derivatives at the boundaries, and in particular ensure that they are finite (ie not NaN). This helps, eg, to avoid surprising results if one is numerically too close to the boundary. I think the one-sided limits are a bit more natural choice. I can still see that zero is an appealing choice as well since it avoids problems with gradient descent algorithms (as discussed in this PR and the previous one), but personally I think this special case and the discontinuity it creates are somehow unsatisfying. Moreover, arguably optimization algorithms/packages are responsible for taking care of and respecting domain constraints, and it should not be encoded implicitly in the derivatives.

Since this PR does not respect these considerations, I think it should not be merged. I guess it's cleaner to open a new separate PR based on these suggestions if others agree that they seem reasonable than updating the PR. I could open such a PR right away but I'd like to have some feedback first 🙂

matbesancon:
the argument for NaN outside of the support makes sense yes

PR #1491, opened 2022-01-23
Remove dynamic call in `check_args` macro

devmotion:
@oschulz This PR fixes your example in https://github.com/JuliaStats/Distributions.jl/issues/1067#issuecomment-1018545600 for me. More concretely, I get
```julia
julia> using Distributions, CUDA

julia> Mu = cu(rand(10));

julia> Sigma = cu(rand(10));

julia> D = Normal.(Mu, Sigma)
10-element CuArray{Normal{Float32}, 1, CUDA.Mem.DeviceBuffer}:
 Normal{Float32}(μ=0.521244f0, σ=0.21886574f0)
 Normal{Float32}(μ=0.71777487f0, σ=0.7293551f0)
 Normal{Float32}(μ=0.9532782f0, σ=0.64151007f0)
 Normal{Float32}(μ=0.053326122f0, σ=0.9042364f0)
 Normal{Float32}(μ=0.7432136f0, σ=0.8857779f0)
 Normal{Float32}(μ=0.12811048f0, σ=0.61526275f0)
 Normal{Float32}(μ=0.25544932f0, σ=0.8564379f0)
 Normal{Float32}(μ=0.71731555f0, σ=0.57268643f0)
 Normal{Float32}(μ=0.17898592f0, σ=0.40761828f0)
 Normal{Float32}(μ=0.71395373f0, σ=0.39995775f0)

julia> X = cu(rand(10));

julia> logpdf.(D, X)
10-element CuArray{Float32, 1, CUDA.Mem.DeviceBuffer}:
 -0.26363182
 -0.6177307
 -0.52780163
 -0.8318763
 -0.980844
 -0.451747
 -0.9199794
 -0.5523126
 -0.7440649
 -0.033901572
```

I don't know how to test this without GPUs though... Maybe we should add some GPU-specific tests and run them with buildkite (I got help with setting it up for OptimalTransport.jl a while ago, so I could add a similar setup here)?

PR #1492, opened 2022-01-23
Explicitly ignore derivatives of argument checks

devmotion:
@willtebbutt noticed in https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/256#issuecomment-1006695214 that the seemingly simple argument checks cause a massive slowdown in AD, specifically with Zygote. Hence this PR ignores derivatives of such checks explicitly.

@willtebbutt's example with `Normal` shows that with this PR AD is very performant, there is basically no overhead compared with the primal and zero allocations, whereas on master it is roughly 60 times slower:

On master:
```julia
julia> using BenchmarkTools, Distributions, Zygote

julia> @benchmark Normal(randn(), rand() + 1)
BenchmarkTools.Trial: 10000 samples with 999 evaluations.
 Range (min … max):  7.032 ns … 42.774 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     7.482 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   7.749 ns ±  0.827 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

        ▄▅▇▇█▆▃▁
  ▁▂▂▄▆█████████▇▅▄▃▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▂▂▂▂▃▃▃▃▃▃▃▂▂▂▂▂▂▁▁ ▃
  7.03 ns        Histogram: frequency by time        9.44 ns <

julia> @benchmark Zygote._pullback($(Zygote.Context()), Normal, randn(), rand() + 1)
BenchmarkTools.Trial: 10000 samples with 198 evaluations.
 Range (min … max):  435.172 ns … 85.840 μs  ┊ GC (min … max):  0.00% … 98.94%
 Time  (median):     454.141 ns              ┊ GC (median):     0.00%
 Time  (mean ± σ):   628.774 ns ±  2.690 μs  ┊ GC (mean ± σ):  25.31% ±  5.91%

  ▃██▇▇▅▄▄▃▃▂▂▂▂▁▂▁▁▂▂▁                                        ▂
  ████████████████████████▇▆▅▆▅▆▆▆▆▆▆▇▄▅▆▆▄▄▁▅▄▆▄▃▅▄▅▆▅▄▃▄▄▄▄▄ █
  435 ns        Histogram: log(frequency) by time       797 ns <

 Memory estimate: 848 bytes, allocs estimate: 17.
```

This PR:
```julia
julia> using BenchmarkTools, Distributions, Zygote

julia> @benchmark Normal(randn(), rand() + 1)
BenchmarkTools.Trial: 10000 samples with 999 evaluations.
 Range (min … max):  6.989 ns … 40.634 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     7.468 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   7.734 ns ±  0.801 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

         ▃▆▇█▆▄▂
  ▁▁▁▂▄▆█████████▆▄▃▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▂▂▂▂▃▃▃▃▃▃▃▃▂▂▂▂▁▁▁ ▃
  6.99 ns        Histogram: frequency by time         9.4 ns <

julia> @benchmark Zygote._pullback($(Zygote.Context()), Normal, randn(), rand() + 1)
BenchmarkTools.Trial: 10000 samples with 999 evaluations.
 Range (min … max):  7.010 ns … 40.456 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     7.466 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   7.718 ns ±  0.789 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

        ▃▆███▆▃▁
  ▁▁▂▃▅█████████▆▅▃▃▂▂▂▁▂▁▁▁▂▁▁▁▁▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▁ ▃
  7.01 ns        Histogram: frequency by time        9.45 ns <

```

### DIFF
--- a/src/multivariate/dirichlet.jl
+++ b/src/multivariate/dirichlet.jl
@@ -26,8 +26,10 @@ struct Dirichlet{T<:Real,Ts<:AbstractVector{T},S<:Real} <: ContinuousMultivariat
     lmnB::S
 
     function Dirichlet{T}(alpha::AbstractVector{T}; check_args::Bool=true) where T
-        if check_args && !all(x -> x > zero(x), alpha)
-            throw(ArgumentError("Dirichlet: alpha must be a positive vector."))
+        ChainRulesCore.ignore_derivatives() do
+            if check_args && !all(x -> x > zero(x), alpha)
+                throw(ArgumentError("Dirichlet: alpha must be a positive vector."))
+            end
         end
         alpha0 = sum(alpha)
         lmnB = sum(loggamma, alpha) - loggamma(alpha0)
@@ -39,7 +41,9 @@ function Dirichlet(alpha::AbstractVector{T}; check_args::Bool=true) where {T<:Re
     Dirichlet{T}(alpha; check_args=check_args)
 end
 function Dirichlet(d::Integer, alpha::Real; check_args::Bool=true)
-    check_args && @check_args(Dirichlet, d > zero(d) && alpha > zero(alpha))
+    ChainRulesCore.ignore_derivatives() do
+        check_args && @check_args(Dirichlet, d > zero(d) && alpha > zero(alpha))
+    end
     return Dirichlet{typeof(alpha)}(Fill(alpha, d); check_args=false)
 end
 
--- a/src/multivariate/dirichlet.jl
+++ b/src/multivariate/dirichlet.jl
@@ -28,7 +28,7 @@ struct Dirichlet{T<:Real,Ts<:AbstractVector{T},S<:Real} <: ContinuousMultivariat
     function Dirichlet{T}(alpha::AbstractVector{T}; check_args::Bool=true) where T
         @check_args(
             Dirichlet,
-            (!all(x -> x > zero(x), alpha), "alpha must be a positive vector."),
+            (all(x -> x > zero(x), alpha), "alpha must be a positive vector."),
         )
         alpha0 = sum(alpha)
         lmnB = sum(loggamma, alpha) - loggamma(alpha0)
--- a/src/multivariate/dirichlet.jl
+++ b/src/multivariate/dirichlet.jl
@@ -28,7 +28,7 @@ struct Dirichlet{T<:Real,Ts<:AbstractVector{T},S<:Real} <: ContinuousMultivariat
     function Dirichlet{T}(alpha::AbstractVector{T}; check_args::Bool=true) where T
         @check_args(
             Dirichlet,
-            (all(x -> x > zero(x), alpha), "alpha must be a positive vector."),
+            (alpha, all(x -> x > zero(x), alpha), "alpha must be a positive vector."),
         )
         alpha0 = sum(alpha)
         lmnB = sum(loggamma, alpha) - loggamma(alpha0)
@@ -40,7 +40,7 @@ function Dirichlet(alpha::AbstractVector{T}; check_args::Bool=true) where {T<:Re
     Dirichlet{T}(alpha; check_args=check_args)
 end
 function Dirichlet(d::Integer, alpha::Real; check_args::Bool=true)
-    @check_args(Dirichlet, d > zero(d), alpha > zero(alpha))
+    @check_args Dirichlet (d, d > zero(d)) (alpha, alpha > zero(alpha))
     return Dirichlet{typeof(alpha)}(Fill(alpha, d); check_args=false)
 end
 
--- a/src/multivariate/dirichlet.jl
+++ b/src/multivariate/dirichlet.jl
@@ -26,11 +26,10 @@ struct Dirichlet{T<:Real,Ts<:AbstractVector{T},S<:Real} <: ContinuousMultivariat
     lmnB::S
 
     function Dirichlet{T}(alpha::AbstractVector{T}; check_args::Bool=true) where T
-        ChainRulesCore.ignore_derivatives() do
-            if check_args && !all(x -> x > zero(x), alpha)
-                throw(ArgumentError("Dirichlet: alpha must be a positive vector."))
-            end
-        end
+        @check_args(
+            Dirichlet,
+            (!all(x -> x > zero(x), alpha), "alpha must be a positive vector."),
+        )
         alpha0 = sum(alpha)
         lmnB = sum(loggamma, alpha) - loggamma(alpha0)
         new{T,typeof(alpha),typeof(lmnB)}(alpha, alpha0, lmnB)
@@ -41,9 +40,7 @@ function Dirichlet(alpha::AbstractVector{T}; check_args::Bool=true) where {T<:Re
     Dirichlet{T}(alpha; check_args=check_args)
 end
 function Dirichlet(d::Integer, alpha::Real; check_args::Bool=true)
-    ChainRulesCore.ignore_derivatives() do
-        check_args && @check_args(Dirichlet, d > zero(d) && alpha > zero(alpha))
-    end
+    @check_args(Dirichlet, d > zero(d), alpha > zero(alpha))
     return Dirichlet{typeof(alpha)}(Fill(alpha, d); check_args=false)
 end
 
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -13,10 +13,10 @@ end
         ...,
     )
 
-A convenience macro that generates AD-compatible checks of arguments for a distribution of
-type `D`.
+A convenience macro that generates checks of arguments for a distribution of type `D`.
 
-More concretely, it generates the following Julia code:
+The macro expects that a boolean variable of name `check_args` is defined and generates
+the following Julia code:
 ```julia
 Distributions.check_args(check_args) do
     \$(statements...)
@@ -91,7 +91,6 @@ function check_args(f::F, check::Bool) where {F}
 end
 
 ChainRulesCore.@non_differentiable check_args(::Any, ::Bool)
-
 
 ##### Utility functions
 
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1,10 +1,60 @@
 ## macro for argument checking
 
-macro check_args(D, cond)
+"""
+    @check_args(
+        D,
+        @setup(statements...),
+        (cond₁, message₁),
+        (cond₂, message₂),
+        ...,
+    )
+
+A convenience macro that generates AD-compatible checks of arguments for a distribution of
+type `D`.
+
+More concretely, it generates the following Julia code:
+```julia
+ChainRulesCore.ignore_derivatives() do
+    if check_args
+        \$(statements...)
+        cond₁ || throw(ArgumentError(\$(string(D, ": ", message₁))))
+        cond₂ || throw(ArgumentError(\$(string(D, ": ", message₂))))
+        ...
+    end
+end
+```
+
+The `@setup` argument can be elided if no setup code is needed. Moreover, error messages
+can be omitted. In this case the message `"the condition \$(cond) is not satisfied."` is
+used.
+"""
+macro check_args(D, setup_or_cond, conds...)
+    # Extract setup statements
+    if Meta.isexpr(setup_or_cond, :macrocall) && setup_or_cond.args[1] == Symbol("@setup")
+        setup_stmts = Any[esc(ex) for ex in setup_or_cond.args[3:end]]
+    else
+        setup_stmts = []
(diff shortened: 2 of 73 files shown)


sethaxen:
I'm not a big fan of how much ChainRules-specific code is now being included in these functions.  What if the `@check_args` macro included `ChainRulesCore.ignore_derivatives` internally? Then you would only need to call it for cases where a constructor doesn't use `@check_args`. Also, what if `@check_args` was generalized to support multiple conditions and custom error messages, which would I think allow it to be used everywhere `@check_args` currently isn't used?

devmotion:
> What if the @check_args macro included ChainRulesCore.ignore_derivatives internally? Then you would only need to call it for cases where a constructor doesn't use @check_args. 

This was my initial approach but unfortunately it is not sufficient if one does not ignore the whole blocks of checks, including `if check_args`.

devmotion:
> Also, what if @check_args was generalized to support multiple conditions and custom error messages, which would I think allow it to be used everywhere @check_args currently isn't used?

I considered this as well but it felt like if it becomes completely general and flexible it would become easier at some point to not use a macro but just implement the checks explicitly. It's also not only used in constructors and IIRC there's at least one case where some variables are unpacked/defined in the `check_args` block. And it felt a bit weird to include `if check_args` in the macro which would be necessary as well.

So basically my impressoon was that the macro should be able to generate something like
```julia
ChainRulesCore.ignore_derivatives() do
    if some_var # or only if check_args
        some_expr
        cond1 || throw(ArgumentError(cond1_msg))
        cond2 || ...
    end
end
```
and it was not immediately clear if and what a short but descriptive syntax could look that would provide a major advantage over implementing the block direv

mschauer:
> I'm not a big fan of how much ChainRules-specific code is now being included in these functions.

This looks like fixing a problem in the wrong place.
This would now work for AD, but every _other_  nonstandard evaluation besides AD (particles, uncertainty intervals, things we don’t anticipate)  has to deal with ChainRules specific code.

devmotion:
It is exactly designed for such a use case: https://juliadiff.org/ChainRulesCore.jl/dev/rule_author/tips_for_packages.html#Ignoring-gradients-for-certain-expressions And I don't see why/how it should interact with other non-AD packages, it just executes the function in the primal: https://github.com/JuliaDiff/ChainRulesCore.jl/blob/699e61f1539fdba362fff5a1b438fbccf32370f0/src/ignore_derivatives.jl#L26

devmotion:
The benchmarks also show that there is no overhead for non-AD use.

sethaxen:
@devmotion I can't think of a nice syntax for the macro, and your points are all valid.

I guess another approach would be to define a `_maybe_check_args(::Type{<:Distribution}, check::Bool, ps...)` function where all checks are performed for each distribution (each with their own overload). Then mark it globally as `ChainRulesCore.@non_differentiable`. But I suspect that would only work for ChainRules, since for operator-overloading ADs, one would need to dispatch on the param values, so there would be unresolved ambiguity.

devmotion:
I thought about moving the checks to a separate function but this seemed unnecessarily complicated - for `ignore_derivatives` one already defines separate distribution specific functions anyway, without adding more methods to the module and hence without potential ambiguity issues.

To summarize, I think it's correct to address these performance issues in Distributions but it would be good to do it in a more developer friendly and "less special" way. I don't think it's an AD backend issue since I don't think it can be expected in general that AD neglects these checks - without us telling it that it should.

I thought about a more convenient macro. Maybe we could use something like `@check_args(Beta, a > zero(a), (b > zero(b)) => "b should be positive!") `. This would not include any additional expressions but maybe that's fine for now (even though it's a bit annoying that AD won't ignore them if they are outside of the function...).

devmotion:
I modified the `@check_args` macro, inspired by `ChainRulesCore.@scalar_rule`:
````julia
    @check_args(
        D,
        @setup(statements...),
        (cond₁, message₁),
        (cond₂, message₂),
        ...,
    )

A convenience macro that generates AD-compatible checks of arguments for a distribution of
type `D`.

More concretely, it generates the following Julia code:
```julia
ChainRulesCore.ignore_derivatives() do
    if check_args
        \$(statements...)
        cond₁ || throw(ArgumentError(\$(string(D, ": ", message₁))))
        cond₂ || throw(ArgumentError(\$(string(D, ": ", message₂))))
        ...
    end
end
```

The `@setup` argument can be elided if no setup code is needed. Moreover, error messages
can be omitted. In this case the message `"the condition \$(cond) is not satisfied."` is
used.
````

I wonder though if it is too surprising that one has to define a boolean variable `check_args`. Maybe its name should be passed in the macro call, i.e., something like `@check_args(check_args, D, ...)`? On the other hand, this would feel a bit redundant.

File: src/multivariate/dirichlet.jl [hunk 1]
```diff
@@ -26,9 +26,10 @@ struct Dirichlet{T<:Real,Ts<:AbstractVector{T},S<:Real} <: ContinuousMultivariat
     lmnB::S
 
     function Dirichlet{T}(alpha::AbstractVector{T}; check_args::Bool=true) where T
-        if check_args && !all(x -> x > zero(x), alpha)
-            throw(ArgumentError("Dirichlet: alpha must be a positive vector."))
-        end
+        @check_args(
+            Dirichlet,
+            (!all(x -> x > zero(x), alpha), "alpha must be a positive vector."),
```
devmotion, line 31:
```suggestion
            (all(x -> x > zero(x), alpha), "alpha must be a positive vector."),
```

mschauer:
Why not just having a dedicated function
```julia
check_arguments(checkargs, f::Function) = checkargs && f()
ChainRulesCore.@non_differentiable check_arguments(checkargs, f)
```
This is positively saying what happens ("checking arguments") instead of negatively "can't differentiate the following code").

devmotion:
Sure, this could be done but it seems equivalent to what `ignore_derivatives` does. In the current state of the PR `ignore_derivatives` only shows up in the code of the macro - and similarly the call to a `check_args` function would only show up there if we don't want to remove the macro and implement all checks explicitly.

mschauer:
We would own `check_arguments ` so on the path of standard evaluation we would not call chainrules code at all and we are free to add any methods

mschauer:
We can thus depend on a hypothetical `UncertaintyCore.jl` and define 
```
UncertaintyCore.@deterministic check_arguments(checkargs, f)
```
if that is required.

devmotion:
I don't see a problem with calling ChainRules - it's documented and guarenteed to not affect the primal computation. Its sole purpose is to not have to define separate functions and mark them as non-differentiable if you want to mark parts of a function body as non-differentiable.

Can you explain your example? What is `@deterministic` supposed to do? In any case if the function you proposed would be owned by Distributions it would be impossible to extend it in other packages - it would be type piracy and if the functions would be defined locally, eg with a do block, they would not even be available for dispatch. I expect similar problems if another package would define such a function and we would like to extend it.

mschauer:
It is a hypothetical example. There is not only automatic differentiation, but also other things we want to do in an automatic fashion in Julia, for example automatic uncertainty propagation etc.
Now we have seen that declaring the argument check `ChainRulesCore.@non_differentiable` is necessary for automatic differentiability performance, so we should anticipate the need to declare the argument check also to be ignored for other use cases in other automatic frameworks. For example that it adds no uncertainty in uncertainty propagation or whatever else might coming up later.

devmotion:
> we should anticipate the need to declare the argument check also to be ignored for other use cases in other automatic frameworks. For example that it adds no uncertainty in uncertainty propagation or whatever else might coming up later.

Ah, OK, so in your example `@deterministic` would be similar to `@non_differentiable`. I thought you wanted to extend the function in some other package or make it an extension of a function in another package, which would both be impossible I think.

I don't have a strong preference, it's simple to switch from one approach to the other since they are only called in the macro. If the consensus is to use a custom function, I can just copy the implementation of `ignore_derivatives` and use it instead 🤷‍♂️

sethaxen:
I don't have a strong opinion wrt function vs macro, but from a design perspective I do prefer both options over calling `ChainRules.ignore_derivative` directly, and this way those contributing new distributions can just use our internal function/macro without even needing to think about ChainRules and how ADs handle argument checking.

src/utils.jl
```diff
@@ -1,14 +1,98 @@
...
+                message = string(D, ": the condition ", cond, " is not satisfied.")
+                return :(($(esc(cond))) || throw(DomainError($(esc(arg)), $message)))
+            end
+        else
+            # only condition specified
+            cond = check
+            message = string(D, ": the condition ", cond, " is not satisfied.")
+            return :(($(esc(cond))) || throw(ArgumentError($message)))
+        end
+    end
```
devmotion, line 70: The idea here is to throw a `DomainError` if possible since it yields more descriptive error messages that also display the problematic value without having to construct strings dynamically which breaks GPU support (https://github.com/JuliaStats/Distributions.jl/pull/1491).

src/utils.jl
```diff
@@ -1,14 +1,98 @@
...
+
+    return @strip_linenos quote
+        Distributions.check_args($(esc(:check_args))) do
+            $(__source__)
```
devmotion, line 74: `@strip_linenos` and `__source__` improve the stacktraces by pointing to the line where the macro was called instead of its definition here.

devmotion:
I tried to incorporate all comments, the PR should be ready for a proper review.

devmotion:
@mschauer @sethaxen I would appreciate if you could review this PR and check if I managed to address your comments and suggestions.

src/utils.jl
```diff
@@ -1,14 +1,98 @@
...
+        (cond₂, message₂),
+        ...,
+    )
+
+A convenience macro that generates AD-compatible checks of arguments for a distribution of
+type `D`.
+
+More concretely, it generates the following Julia code:
+```julia
+Distributions.check_args(check_args) do
```
sethaxen, line 21: Perhaps document that it expects a boolean variable `check_args` to be defined?

src/utils.jl
```diff
@@ -1,14 +1,98 @@
...
+"""
+    @check_args(
+        D,
+        @setup(statements...),
+        (arg₁, cond₁, message₁),
+        (cond₂, message₂),
+        ...,
+    )
+
+A convenience macro that generates AD-compatible checks of arguments for a distribution of
```
sethaxen, line 16: I don't think one needs to say it generates AD-compatible checks. That's an implementation detail. You can just say it standardizes checking of arguments.

test/matrixvariates.jl
```diff
@@ -413,7 +413,7 @@ end
 function test_special(dist::Type{LKJ})
     @testset "LKJ mode" begin
         @test mode(LKJ(5, 1.5)) == mean(LKJ(5, 1.5))
-        @test_throws ArgumentError mode( LKJ(5, 0.5) )
+        @test_throws DomainError mode( LKJ(5, 0.5) )
```
sethaxen, line 416: Is this change considered breaking?

devmotion: No, not according to ColPrac: https://colprac.sciml.ai/#changes-that-are-not-considered-breaking

src/utils.jl
devmotion: Alternatively, one could allow users/developers to provide the name of the boolean variable. I thought about it since it's more obvious - but it seemed unnecessarily verbose since we only use `check_args`.

sethaxen: Yeah, it's only internal, so I don't see the benefit of making it configurable. And this way it actually enforces standardizing the keyword argument.

devmotion: I updated the docstring.

src/utils.jl
devmotion: Sure, I removed it now.

sethaxen (approved):
LGTM!

devmotion:
@mschauer are you happy with the approach in this PR? I added a custom non-differentiable function and generalized the macro.

PR #1494, opened 2022-01-26
Build docs better on local machine

sethaxen:
By default, Documenter builds the docs with `prettyurls=true`. Locally, this makes the docs non-browsable. This PR adds the [recommended check](https://juliadocs.github.io/Documenter.jl/stable/man/guide/) of the `CI` variable to only turn on `prettyurls` when building the docs on CI.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -4,7 +4,7 @@ import Random: AbstractRNG, rand!
 makedocs(
     sitename = "Distributions.jl",
     modules  = [Distributions],
-    doctest  = false,
+    format   = Documenter.HTML(; prettyurls = get(ENV, "CI", nothing) == "true"),
     pages    = [
         "index.md",
         "starting.md",


devmotion:
> Locally, this makes the docs non-browsable. 

This is only an issue if you browse the directory, isn't it? I always used the first of the recommended options and ran a live web server (usually with `python -m http.sever` :stuck_out_tongue:), and can't remember any issues with browsing the docs locally.

sethaxen:
Totally, but this is the most contributor-friendly way. i.e. one doesn't need to read the Documenter docs or install another Julia package or use Python to make it work, just `open docs/build/index.html`.

devmotion:
As I said, I've never done it this way, I always use a web server :shrug: Are the two options mutually exclusive? I guess one should still be able to start a live server even if the URL structure is different?

devmotion:
The only thing I'm unhappy about is that this means that the URLs will be different locally it seems. Otherwise I don't mind if it does not affect my usual workflow.

I also checked some repos quickly, and all possible settings show up: SpecialFunctions, HypothesisTests, and Literate (even with explicit `true`) use pretty URLs both locally and on Github, Zygote uses pretty URLs only online, and ChainRules uses pretty URLs neither locally nor online.

sethaxen:
No, they're not mutually exclusive; it's just that the local version won't have pretty URLs. But I think that's only a problem if the docs have absolute links to themselves, which they shouldn't.

sethaxen:
I understand not every repo takes this approach. I prefer it because it lowers the barrier toward a user who wants to contribute non-trivial documentation doing so. And the ecosystem can definitely use more documentation.

devmotion:
Sure, let's try it, more and better docs are always good. And I can always open an issue or PR if I notice any problems :slightly_smiling_face:

devmotion:
(BTW I think the most helpful feature would be if doc previews would also work for PRs from forks)

sethaxen:
Agreed! It's too bad that doesn't work.

PR #1498, opened 2022-01-30
Fix Chi pdf for x < 0

### LINKED ISSUE #1497: No bounds checking in Chi distribution
From discourse [here](https://discourse.julialang.org/t/plotting-two-distributions-on-same-plot/75475)

```

julia> pdf(Chi(3), -.5)
ERROR: DomainError with -0.5:
log will only return a complex result if called with a complex argument. Try log(Complex(x)).
```

Even though pdf should be 0 for `x < 0`. See [wiki](https://en.wikipedia.org/wiki/Chi_distribution). 

The problem is a `logpdf` function that does not check the bounds. See [here](https://github.com/JuliaStats/Distributions.jl/blob/8aea3cc82ee2f8ffe1e8cd754e7fcd99369c7a1c/src/univariate/continuous/chi.jl#L83)

```
logpdf(d::Chi, x::Real) = (ν = d.ν;
    (1 - ν/2) * logtwo + (ν - 1) * log(x) - x^2/2 - loggamma(ν/2)
)
```

pdeffebach:
Closes #1497

### DIFF
--- a/src/univariate/continuous/chi.jl
+++ b/src/univariate/continuous/chi.jl
@@ -80,9 +80,11 @@ end
 
 #### Evaluation
 
-logpdf(d::Chi, x::Real) = (ν = d.ν;
+function logpdf(d::Chi, x::Real)
+    x < 0 && return 0.0
+    ν = d.ν
     (1 - ν/2) * logtwo + (ν - 1) * log(x) - x^2/2 - loggamma(ν/2)
-)
+end
 
 gradlogpdf(d::Chi{T}, x::Real) where {T<:Real} = x >= 0 ? (d.ν - 1) / x - x : zero(T)
 
--- a/src/univariate/continuous/chi.jl
+++ b/src/univariate/continuous/chi.jl
@@ -81,9 +81,10 @@ end
 #### Evaluation
 
 function logpdf(d::Chi, x::Real)
-    x < 0 && return 0.0
-    ν = d.ν
-    (1 - ν/2) * logtwo + (ν - 1) * log(x) - x^2/2 - loggamma(ν/2)
+    ν, _x = promote(d.ν, x)
+    xsq = _x^2
+    val = (xlogy(ν - 1, xsq / 2) - xsq + logtwo) / 2 - loggamma(ν / 2)
+    return x < zero(x) ? oftype(val, -Inf) : val
 end
 
 gradlogpdf(d::Chi{T}, x::Real) where {T<:Real} = x >= 0 ? (d.ν - 1) / x - x : zero(T)
--- a/test/chi.jl
+++ b/test/chi.jl
@@ -1,2 +1,5 @@
 # see #1305
 @test isfinite(mean(Chi(1000)))
+
+# see #1497
+@test pdf(Chi(3), -.5) == 0.0
--- a/test/chi.jl
+++ b/test/chi.jl
@@ -2,4 +2,6 @@
 @test isfinite(mean(Chi(1000)))
 
 # see #1497
-@test pdf(Chi(3), -.5) == 0.0
+for x in (-0.5, -1, -1.5f0), ν in (3, 3f0, 3.0)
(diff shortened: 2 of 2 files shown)


test/chi.jl
```diff
@@ -1,2 +1,5 @@
 # see #1305
-@test isfinite(mean(Chi(1000)))
\ No newline at end of file
+@test isfinite(mean(Chi(1000)))
+
+# see #1497
+@test pdf(Chi(3), -.5) == 0.0
```
rikhuijzer, line 5:
```suggestion
@test pdf(Chi(3), -0.5) == 0.0
```

src/univariate/continuous/chi.jl
```diff
@@ -80,9 +80,11 @@ end
 
 #### Evaluation
 
-logpdf(d::Chi, x::Real) = (ν = d.ν;
+function logpdf(d::Chi, x::Real)
+    x < 0 && return 0.0
```
devmotion, line 84: This is type unstable, better to compute the result as much as possible (as done for other distributions).
```suggestion
```

rikhuijzer (approved):
Also much better to use multi line function here indeed instead of the one-liner with bracket

Note that I don't have merge rights here, but more eyes is more better I guess

test/chi.jl
devmotion, line 5: It seems more natural to check the output of `logpdf` as this PR changes `logpdf`. Also good to check type stability for different types of `x` and `ν`:
```suggestion
for x in (-0.5, -1, -1.5f0), ν in (3, 3f0, 3.0)
    @test @inferred(logpdf(Chi(ν), x)) == -Inf
end 
```

src/univariate/continuous/chi.jl
devmotion: BTW one should also not return `0` but `-Inf` (of the correct type).

src/univariate/continuous/chi.jl
```diff
@@ -80,9 +80,11 @@ end
...
+    ν = d.ν
     (1 - ν/2) * logtwo + (ν - 1) * log(x) - x^2/2 - loggamma(ν/2)
```
devmotion, line 86: One possible implementation would be e.g., similar to https://github.com/JuliaStats/StatsFuns.jl/blob/13e231a0a22e716426b73cb87ff3b8b24e33aaf1/src/distrs/gamma.jl#L19-L24 and
```suggestion
    ν, _x = promote(d.ν, x)
    xsq = _x^2
    val = (xlogy(ν - 1, xsq / 2) - xsq + logtwo) / 2 - loggamma(ν / 2)
    return x < zero(x) ? oftype(val, -Inf) : val
```

src/univariate/continuous/chi.jl
pdeffebach: Ah sorry, was thinking of `pdf`.

pdeffebach:
Thanks for the detailed in-line changes. I merged them and confirmed they work locally. 

I think this can be merged!

devmotion:
Unfortunately, it seems to break downstream tests of Tracker in DistributionsAD: e.g., https://github.com/JuliaStats/Distributions.jl/runs/5009289497?check_suite_focus=true#step:6:322 We have to check if this is a downstream problem or an actual issue with this PR.

devmotion:
Apart from that the PR looks good :slightly_smiling_face:

devmotion:
Seems to be an issue in Tracker since it does not define derivatives for LogExpFunctions yet. It's an easy fix, I'll open a PR. Let's hold back with merging this PR to avoid breaking downstream packages and rerun tests (and merge if they pass) once a new version of Tracker is released.

devmotion:
https://github.com/FluxML/Tracker.jl/pull/124

devmotion:
The downstream issues are fixed :slightly_smiling_face: 

Thanks a lot for the PR @pdeffebach!

Issue #1502, opened 2022-02-02
`mgf` does not take `Complex` numbers


theogf:
The moment generating function is technically homomorphic in the complex space therefore restricting the input to be real does not make sense.
Would a PR doing `t::Real` -> `t::Number` bet accepted?

devmotion:
Sure. I guess the same could be done for (some) characteristic functions as well (https://en.wikipedia.org/wiki/Characteristic_function_(probability_theory)#Entire_characteristic_functions).

PR #1504, opened 2022-02-02
Use `Number` instead of `Real` for `mgf` and `cf`

theogf:
As pointed out in #1502 , `mgf` and `cf` should be allowed to take `Complex` numbers and the second argument should be therefore relaxed.
All functions are also made type-stable.
**Breaking?**, when the argument for the `mgf` is given outside of the domain definition a `DomainError` is thrown.

### DIFF
--- a/src/univariate/continuous/biweight.jl
+++ b/src/univariate/continuous/biweight.jl
@@ -58,13 +58,14 @@ function mgf(d::Biweight, t::Number)
     a = d.σ * t
     abs(real(a)) < 1 || throw(DomainError(t, "|σ ⋅ real(t)| has to be smaller than 1"))
     a2 = a^2
-    iszero(a) ? one(Base.promote_typeof(a, d.μ)) :
-    15exp(d.μ * t) * (-3cosh(a) + (a + 3/a) * sinh(a)) / (a2^2)
+    result = 15exp(d.μ * t) * (-3cosh(a) + (a + 3/a) * sinh(a)) / (a2^2)
+    return iszero(a) ? one(result) : result
+    
 end
 
 function cf(d::Biweight, t::Number)
     a = d.σ * t
     a2 = a^2
-    iszero(a) ? complex(one(Base.promote_typeof(a, d.μ))) :
-    -15cis(d.μ * t) * (3cos(a) + (a - 3/a) * sin(a)) / (a2^2)
+    result = -15cis(d.μ * t) * (3cos(a) + (a - 3/a) * sin(a)) / (a2^2)
+    return iszero(a) ? complex(one(result)) : result
 end
--- a/src/univariate/continuous/biweight.jl
+++ b/src/univariate/continuous/biweight.jl
@@ -54,16 +54,16 @@ end
 
 @quantile_newton Biweight
 
-function mgf(d::Biweight{T}, t::Number) where T<:Real
+function mgf(d::Biweight, t::Number)
     a = d.σ * t
     a2 = a^2
-    iszero(a) ? one(a) :
+    iszero(a) ? one(Base.promote_typeof(a, d.μ)) :
     15exp(d.μ * t) * (-3cosh(a) + (a + 3/a) * sinh(a)) / (a2^2)
 end
 
 function cf(d::Biweight{T}, t::Number) where T<:Real
     a = d.σ * t
     a2 = a^2
-    iszero(a) ? complex(one(a)) :
+    iszero(a) ? complex(one(Base.promote_typeof(a, d.μ))) :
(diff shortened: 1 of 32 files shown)


theogf:
Thanks for the monster review! I tried to address your comments and integrate them. The inference tests should tell us where things go wrong :)

theogf:
MGF are not necessarily defined on the whole Real/Complex domain, what should be returned when getting a value out of bound? NaN or ArgumentError ?

src/univariate/continuous/biweight.jl
```diff
@@ -54,16 +54,16 @@ end
 
 @quantile_newton Biweight
 
-function mgf(d::Biweight{T}, t::Real) where T<:Real
-    a = d.σ*t
+function mgf(d::Biweight, t::Number)
+    a = d.σ * t
     a2 = a^2
-    a == 0 ? one(T) :
+    iszero(a) ? one(Base.promote_typeof(a, d.μ)) :
```
devmotion, line 60: This is still type unstable (and similar cases below) if e.g. all involved values are `Int`s or `Rational`s. Maybe it's easiest to just compute the result and return `one(result)` in this special case.

src/univariate/continuous/biweight.jl
```diff
@@ -54,16 +54,16 @@ end
...
     15exp(d.μ * t) * (-3cosh(a) + (a + 3/a) * sinh(a)) / (a2^2)
```
devmotion, line 61: So my suggestion would be to just use
```suggestion
    result = 15 * exp(d.μ * t) * (-3 * cosh(a) + (a + 3/a) * sinh(a)) / (a2^2)
    return iszero(a) ? one(result) : result
```

src/univariate/continuous/biweight.jl
```diff
@@ -54,16 +54,16 @@ end
...
 end
 
-function cf(d::Biweight{T}, t::Real) where T<:Real
+function cf(d::Biweight, t::Number)
     a = d.σ * t
     a2 = a^2
-    a == 0 ? one(T)+zero(T)*im :
+    iszero(a) ? complex(one(Base.promote_typeof(a, d.μ))) :
     -15cis(d.μ * t) * (3cos(a) + (a - 3/a) * sin(a)) / (a2^2)
```
devmotion, line 68: Similarly,
```suggestion
    result = -15 * cis(d.μ * t) * (3 * cos(a) + (a - 3/a) * sin(a)) / (a2^2)
    return iszero(a) ? one(result) : result
```

devmotion:
> MGF are not necessarily defined on the whole Real/Complex domain, what should be returned when getting a value out of bound? NaN or ArgumentError ?

I was wondering the same. I'm a bit unsure, do you have a concrete example? Then it might be easier to see.

src/univariate/continuous/biweight.jl
theogf: That's not type stable either no? If `a2 == 0` then you might get some `Inf` or `NaN`

src/univariate/continuous/biweight.jl
theogf: Same here

src/univariate/discrete/negativebinomial.jl
```diff
@@ -109,12 +109,14 @@ invlogccdf(d::NegativeBinomial, lq::Real) = convert(Int, nbinominvlogccdf(d.r, d
 ## sampling
 rand(rng::AbstractRNG, d::NegativeBinomial) = rand(rng, Poisson(rand(rng, Gamma(d.r, (1 - d.p)/d.p))))
 
-function mgf(d::NegativeBinomial, t::Real)
+function mgf(d::NegativeBinomial, t::Number)
     r, p = params(d)
-    return ((1 - p) * exp(t))^r / (1 - p * exp(t))^r
+    real(t) < -log1p(-p) || return Base.promote_typeof(p, t)(NaN)
```
theogf, line 114: Regarding the domains, here is an example where  the domain of `t` is bounded.

src/univariate/continuous/biweight.jl
devmotion: This will always be type stable (and is therefore the common approach in Distributions): it returns either `result` or `one(result)` (in the case `iszero(a)` or equivalently `iszero(a2)`), which are both of type `typeof(result)` (we don't work with unitful numbers, `oneunit(result)` would be even more correct).

src/univariate/continuous/biweight.jl
devmotion: Type stable as well 🙂

theogf:
This has gone very very stale so I will just close it.

Issue #1514, opened 2022-03-06
Incorrect  `convert` implementation for `InverseGamma`


KDr2:
Run:

```
using Distributions

 mutable struct T
     m::InverseGamma{Float64}
     T(v::InverseGamma{Float64}) = new(v) # conversion occurs
 end


# the line below fixes this
# Base.convert(::Type{InverseGamma{T}}, d::InverseGamma{T}) where {T <: Real} = d

m1 = InverseGamma{Float64}(0.0, 0.0)
t1 = T(m1)
t1.m = InverseGamma{Float64}(0.0, 0.0)  # conversion occurs
```
will get an error:
```
ERROR: LoadError: DomainError with 0.0:
InverseGamma: the condition α > zero(α) is not satisfied.
Stacktrace:
 [1] #307
   @ ~/.julia/packages/Distributions/T2SAc/src/univariate/continuous/inversegamma.jl:35 [inlined]
 [2] check_args
   @ ~/.julia/packages/Distributions/T2SAc/src/utils.jl:89 [inlined]
 [3] #InverseGamma#306
   @ ~/.julia/packages/Distributions/T2SAc/src/univariate/continuous/inversegamma.jl:35 [inlined]
 [4] InverseGamma
   @ ~/.julia/packages/Distributions/T2SAc/src/univariate/continuous/inversegamma.jl:35 [inlined]
 [5] convert
   @ ~/.julia/packages/Distributions/T2SAc/src/univariate/continuous/inversegamma.jl:48 [inlined]
 [6] T(v::InverseGamma{Float64})
```

due to an unnecessary conversion from `InverseGamma{T}` to the same type.

Relevant discussion: https://discourse.julialang.org/t/conversion-while-constructing-object/77401

devmotion:
Yeah, we should the no-op `convert`, it's missing for some other distributions as well.

I guess nobody has fixed it yet since it's not an issue if one uses valid parameters. But the parameters in your example are not (`InverseGamma(0.0, 0.0)` errors in the same way) and can only be used with the inner constructor.

Issue #1485, opened 2022-01-19
Convert does not respect type of Normal


sethaxen:
#1472 made some changes to the constructors of `Normal`, so that now we get the following:

```julia
julia> d = Normal{Int64}(0, 1)
Normal{Int64}(μ=0, σ=1)

julia> convert(Normal{Int}, d)
Normal{Float64}(μ=0.0, σ=1.0)

julia> truncated(d, 0, 1)
ERROR: TypeError: in new, expected Normal{Int64}, got a value of type Normal{Float64}
Stacktrace:
 [1] Truncated
   @ ~/projects/Distributions.jl/src/truncate.jl:57 [inlined]
 [2] truncated(d::Normal{Int64}, l::Int64, u::Int64)
   @ Distributions ~/projects/Distributions.jl/src/truncate.jl:39
 [3] top-level scope
   @ REPL[36]:1
```

The latter happens I guess because the constructor for `Truncated` implicitly calls `convert(Normal{Int}, d)` and throws an error when it gets the wrong type.

sethaxen:
Same happens now for `Uniform`:
```julia
julia> convert(Uniform{Int}, Uniform{Int}(0, 1))
Uniform{Float64}(a=0.0, b=1.0)
```

devmotion:
#1472 did not change `convert`, it seems previously it just worked by accident since it did not actually call the constructor for integers in e.g. the case of `Normal` which - for historic reasons that I think should be changed - always promotes integers to floating point numbers.

So possible solutions would be to e.g. use more standard definitions of `convert` that use the inner constructors directly, without relying or falling back on any promotions of the outer constructors, or to remove any such promotions from integers to floating point numbers in the constructors. In my opinion, both things should be done - but since not promoting integers might require changes in other methods it is probably easier to start with using the inner constructors in `convert`.

devmotion:
In some sense it was even worse before #1472 since e.g. both `convert` definitions for `Normal` were inconsistent (example with Distributions 0.25.38):
```julia
julia> convert(Normal{Int}, 0, 1)
Normal{Float64}(μ=0.0, σ=1.0)

julia> convert(Normal{Int}, Normal{Int}(0, 1))
Normal{Int64}(μ=0, σ=1)
```

devmotion:
[...]

PR #1515, opened 2022-03-07
Fix and optimize `convert`

devmotion:
This PR fixes two bugs of the current `convert` definitions:
- `convert(::Type{MyDist{T}}, d::MyDist)` is not always of type `MyDist{T}` (see e.g. https://github.com/JuliaStats/Distributions.jl/issues/1485)
- `convert(::Type{MyDist{T}}, d::MyDist{T})` is not always a no-op (see e.g. https://github.com/JuliaStats/Distributions.jl/issues/1514)

I went through all `convert` implementations, fixed them and added missing definitions, and added tests.

Fixes https://github.com/JuliaStats/Distributions.jl/issues/1514. Fixes https://github.com/JuliaStats/Distributions.jl/issues/1485.

cc: @sethaxen @KDr2

### DIFF
--- a/src/univariate/continuous/chisq.jl
+++ b/src/univariate/continuous/chisq.jl
@@ -42,8 +42,8 @@ params(d::Chisq) = (d.ν,)
 
 ### Conversions
 convert(::Type{Chisq{T}}, ν::Real) where {T<:Real} = Chisq(T(ν))
-convert(::Type{Chisq{T}}, d::Chisq{S}) where {T <: Real, S <: Real} = Chisq(T(d.ν))
-
+Base.convert(::Type{Chisq{T}}, d::Chisq) where {T<:Real} = Chisq{T}(T(d.ν))
+Base.convert(::Type{Chisq{T}}, d::Chisq{T}) where {T<:Real} = d
 
 #### Statistics
 
--- a/src/univariate/discrete/negativebinomial.jl
+++ b/src/univariate/discrete/negativebinomial.jl
@@ -54,9 +54,10 @@ NegativeBinomial() = NegativeBinomial{Float64}(1.0, 0.5)
 function convert(::Type{NegativeBinomial{T}}, r::Real, p::Real) where {T<:Real}
     return NegativeBinomial(T(r), T(p))
 end
-function convert(::Type{NegativeBinomial{T}}, d::NegativeBinomial{S}) where {T <: Real, S <: Real}
-    return NegativeBinomial(T(d.r), T(d.p), check_args=false)
+function Base.convert(::Type{NegativeBinomial{T}}, d::NegativeBinomial) where {T<:Real}
+    return NegativeBinomial{T}(T(d.r), T(d.p))
 end
+Base.convert(::Type{NegativeBinomial{T}}, d::NegativeBinomial{T}) where {T<:Real} = d
 
 #### Parameters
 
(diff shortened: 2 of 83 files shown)


devmotion:
@sethaxen this would fix an issue that you reported a while ago, maybe you could have a look at the PR?

src/univariate/continuous/chisq.jl
```diff
@@ -42,8 +42,8 @@ params(d::Chisq) = (d.ν,)
 
 ### Conversions
 convert(::Type{Chisq{T}}, ν::Real) where {T<:Real} = Chisq(T(ν))
-convert(::Type{Chisq{T}}, d::Chisq{S}) where {T <: Real, S <: Real} = Chisq(T(d.ν))
-
+Base.convert(::Type{Chisq{T}}, d::Chisq) where {T<:Real} = Chisq{T}(T(d.ν))
```
mschauer, line 45: A bit confusing to have `Base.convert` and `convert` side by side, but of course the same given the import

devmotion: I didn't want to touch the `convert(::Type, ::Real...)` methods in this PR since
1. they are broken in so far as they don't ensure that the returned distribution of the desired type, and
2. I think they should be deprecated and eventually be removed.

In particular methods such as `convert(::Type{Normal{T}}, mu, sigma)` seem strange and I don't see a good reason for why one should prefer those over `Normal(mu, sigma)` or `Normal{T}(mu, sigma)`.

mschauer: Then there is no reason to worry about their cosmetics.

src/univariate/locationscale.jl
```diff
@@ -81,7 +86,10 @@ AffineDistribution(μ::Real, σ::Real, d::AffineDistribution) = AffineDistributi
 #### Conversions
 
 convert(::Type{AffineDistribution{T}}, μ::Real, σ::Real, ρ::D) where {T<:Real, D<:UnivariateDistribution} = AffineDistribution(T(μ),T(σ),ρ)
-convert(::Type{AffineDistribution{T}}, d::AffineDistribution{S}) where {T<:Real, S<:Real} = AffineDistribution(T(d.μ),T(d.σ),d.ρ, check_args=false)
+function Base.convert(::Type{AffineDistribution{T}}, d::AffineDistribution) where {T<:Real}
+    AffineDistribution{T}(T(d.μ),T(d.σ),d.ρ)
```
mschauer, line 90:
```suggestion
    AffineDistribution{T}(T(d.μ), T(d.σ), d.ρ)
```

src/univariate/discrete/negativebinomial.jl
```diff
@@ -54,9 +54,10 @@ NegativeBinomial() = NegativeBinomial{Float64}(1.0, 0.5)
 function convert(::Type{NegativeBinomial{T}}, r::Real, p::Real) where {T<:Real}
     return NegativeBinomial(T(r), T(p))
 end
-function convert(::Type{NegativeBinomial{T}}, d::NegativeBinomial{S}) where {T <: Real, S <: Real}
-    return NegativeBinomial(T(d.r), T(d.p), check_args=false)
+function Base.convert(::Type{NegativeBinomial{T}}, d::NegativeBinomial) where {T<:Real}
+    return NegativeBinomial{T}(T(d.r), T(d.p))
```
mschauer, line 58: yes

mschauer (approved):

mschauer:
Some cosmetics, but at least cosmetics helps focusing on reviewing repetitive changes. Thanks!

PR #1517, opened 2022-03-14
Support MvNormal holding Symmetric{<:Real,<:Diagonal} or Hermitian

chriselrod:
On master:
```julia
julia> using Distributions, LinearAlgebra, FillArrays

julia> MvNormal(rand(2), Symmetric(Diagonal([1.0,2.0])))
ERROR: MethodError: Cannot `convert` an object of type Symmetric{Float64, Diagonal{Float64, Vector{Float64}}} to an object of type Diagonal{Float64, Vector{Float64}}

julia> MvNormal(Zeros(2), Symmetric(Diagonal([1.0,2.0])))
ERROR: MethodError: Cannot `convert` an object of type Symmetric{Float64, Diagonal{Float64, Vector{Float64}}} to an object of type Diagonal{Float64, Vector{Float64}}
```
On this PR:
```julia
julia> using Distributions, LinearAlgebra, FillArrays

julia> MvNormal(rand(2), Symmetric(Diagonal([1.0,2.0])))
DiagNormal(
dim: 2
μ: [0.1851389943338715, 0.9573799942189535]
Σ: [1.0 0.0; 0.0 2.0]
)

julia> MvNormal(Zeros(2), Symmetric(Diagonal([1.0,2.0])))
ZeroMeanDiagNormal(
dim: 2
μ: Zeros(2)
Σ: [1.0 0.0; 0.0 2.0]
)
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.25.49"
+version = "0.25.50"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -200,6 +200,7 @@ Construct a multivariate normal distribution with mean `μ` and covariance matri
 """
 MvNormal(μ::AbstractVector{<:Real}, Σ::AbstractMatrix{<:Real}) = MvNormal(μ, PDMat(Σ))
 MvNormal(μ::AbstractVector{<:Real}, Σ::Diagonal{<:Real}) = MvNormal(μ, PDiagMat(Σ.diag))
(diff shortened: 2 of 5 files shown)


devmotion:
Ideally we would use a generic method in PDMats that takes care of these constructions but until this is fixed upstream this PR seems like useful workaround.

Can you apply the same fixes for `MvLogNormal` and `MvNormalCanon`, if needed?

devmotion:
The other question is, why do you end up with a `Symmetric{<:Real,<:Diagonal}`? And should Julia avoid this seemingly unnecessary wrapper by defining `Symmetric(d::Diagonal) = d`? The downside would be though the same as in the discussion in PDMats, the output would not be of type of `Symmetric` anymore.

chriselrod:
> Can you apply the same fixes for `MvLogNormal` and `MvNormalCanon`, if needed?

Done.

> The other question is, why do you end up with a `Symmetric{<:Real,<:Diagonal}`?

There is another package that is explicitly testing this.
The test used to pass, but recently started failing. I am not sure what changed.

I'm not a purist with respect to constructors returning objects of the same type as given by the name, so I'd personally be fine with `Symmetric(::Diagonal)` returning a `Diagonal` or `PDMat(::Diagonal)` returning a `PDiagMat`.

devmotion (approved):
Thanks @chriselrod!

cossio:
Perhaps related: https://github.com/JuliaLang/julia/issues/44618

PR #1519, opened 2022-03-16
Fix Typo for Binomial API

khosravipasha:
Fix typo in `starting.md`. Binomial distribution needs both `n` and `p`.


devmotion (approved):
Thanks, good catch!

> Binomial distribution needs both n and p.

Actually, it is fine to only provide `n` or none of them. In these cases `p = 0.5` and `n = 1`, `p = 0.5` are used: https://github.com/JuliaStats/Distributions.jl/blob/eff3f1c456f53ce1e01cd69ac09b378a8da57248/src/univariate/discrete/binomial.jl#L38-L42

PR #1521, opened 2022-03-17
Fix rand for Product(Fill(...)) distributions

chriselrod:
`Product(Fill(...))` is useful for iid distributions.
```julia
julia> using Random, Distributions, FillArrays

julia> rng = Random.default_rng();

julia> x = Vector{Float64}(undef, 4);

julia> x .= rand.(rng, Fill(Exponential(2.3), 4))
4-element Vector{Float64}:
 1.9539101912937518
 1.9539101912937518
 1.9539101912937518
 1.9539101912937518

julia> map!(Base.Fix1(rand, rng), x, Fill(Exponential(2.3), 4))
4-element Vector{Float64}:
 8.827816210804809
 0.4896979114931074
 1.6137874001903105
 1.0617829479264902
```
I haven't really looked into the problem, but am guessing that `FillArray`s overloads broadcasts, causing the same value of `rand` to fill all values of the result. Using `map!` instead avoids this problem.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.25.51"
+version = "0.25.52"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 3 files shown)


devmotion (approved):
Yeah, this is a known bug of `Product`. There are other issues and limitations of it as well. https://github.com/JuliaStats/Distributions.jl/pull/1391 fixes most of them but unfortunately is breaking and not AD-friendly, and therefore is not merged yet.

Probably good to fix the issue until the generalization is available.

PR #1523, opened 2022-03-22
Add convolve for DiscreteNonParametric

iampritishpatil:
DiscreteNonParametric convolution has a very nice trivial closed form. It was not implemented.

This pull request implements it.


src/convolution.jl
```diff
@@ -12,13 +12,16 @@ and one of
 * [`NegativeBinomial`](@ref)
 * [`Geometric`](@ref)
 * [`Poisson`](@ref)
+* [`DiscreteNonParametric`](@ref)
 * [`Normal`](@ref)
 * [`Cauchy`](@ref)
 * [`Chisq`](@ref)
 * [`Exponential`](@ref)
 * [`Gamma`](@ref)
 * [`MvNormal`](@ref)
```
devmotion, line 22:
```suggestion
```

src/convolution.jl
```diff
@@ -47,6 +50,20 @@ end
 convolve(d1::Poisson, d2::Poisson) =  Poisson(d1.λ + d2.λ)
 
 
+
+function convolve(d1::DiscreteNonParametric, d2::DiscreteNonParametric)
+    support_conv = vcat([d1.support .+ v for v in d2.support]...) |> unique
```
devmotion, line 55: This seems a bit inefficient: due to the allocations, splatting and constructing the full vector. Probably better to use a `Set`:

```suggestion
    support_conv = collect(Set(v1 + v2 for v1 in support(d1), v2 in support(d2)))
```

src/convolution.jl
```diff
@@ -47,6 +50,20 @@ end
 convolve(d1::Poisson, d2::Poisson) =  Poisson(d1.λ + d2.λ)
 
 
+
+function convolve(d1::DiscreteNonParametric, d2::DiscreteNonParametric)
+    support_conv = vcat([d1.support .+ v for v in d2.support]...) |> unique 
+    sort!(support_conv) #for fast index finding below
+    p_conv = zeros(length(support_conv))
```
devmotion, line 57: This should not necessarily be a vector of `Float64`. Maybe
```suggestion
    probs1 = probs(d1)
    probs2 = probs(d2)
    p_conv = zeros(Base.promote_eltype(probs1, probs2), length(support_conv)) 
```
even though this would also not preserve the array type, e.g. if they would be static arrays.

src/convolution.jl
```diff
@@ -47,6 +50,20 @@ end
 convolve(d1::Poisson, d2::Poisson) =  Poisson(d1.λ + d2.λ)
 
 
+
+function convolve(d1::DiscreteNonParametric, d2::DiscreteNonParametric)
+    support_conv = vcat([d1.support .+ v for v in d2.support]...) |> unique 
+    sort!(support_conv) #for fast index finding below
+    p_conv = zeros(length(support_conv)) 
+    for (s1,p1) in zip(d1.support, d1.p)
+        for (s2,p2) in zip(d2.support, d2.p)
```
devmotion, line 59:
```suggestion
    for (s1, p1) in zip(support(d1), probs(d1)), (s2, p2) in zip(support(d2), probs(d2))
```

src/convolution.jl
```diff
@@ -47,6 +50,20 @@ end
...
+            idx = searchsortedfirst(support_conv, s1+s2)
```
devmotion, line 60: I wonder if this (and the `unique` above) is problematic due to floating point inaccuracies.

src/convolution.jl
```diff
@@ -77,3 +94,6 @@ end
 function _check_convolution_shape(d1, d2)
     length(d1) == length(d2) || throw(ArgumentError("$d1 and $d2 are not the same size"))
 end
+
+    
+
```
devmotion, line 99:
```suggestion
```

src/convolution.jl
```diff
@@ -47,6 +50,20 @@ end
...
+    support_conv = vcat([d1.support .+ v for v in d2.support]...) |> unique 
+    sort!(support_conv) #for fast index finding below
+    p_conv = zeros(length(support_conv)) 
+    for (s1,p1) in zip(d1.support, d1.p)
+        for (s2,p2) in zip(d2.support, d2.p)
+            idx = searchsortedfirst(support_conv, s1+s2)
+            p_conv[idx] += p1*p2
+        end
+    end
+    DiscreteNonParametric(support_conv, p_conv)
```
devmotion, line 64: We should not resort the support again.

devmotion (commented):
Thank you for the PR! I made some suggestions. Can you also add some tests, also of more numerically tricky edge cases?

src/convolution.jl
iampritishpatil: Added check args to fix it.

src/convolution.jl
iampritishpatil: It shouldn't not matter as long as we do the same operations s1+s2 exactly both in creation os support_conv and in finding idx. 
Not sure how the compiler optimization might matter in this. 

Added a test with eps(Float64) to see any trivial inconsistencies.

iampritishpatil:
Thanks for the suggestions, they were, all on the point.

I think I addressed all of it. If I run into any more floating-point tricky cases, I'll add them, but as far as I can tell, it shouldn't cause errors, apart from any unexpected compiler optimization which causes different things to happen in the two loops. 

Let me know if anything more is needed.

test/convolution.jl
```diff
@@ -66,6 +66,27 @@ using Test
         @test d3 isa Poisson
         @test d3.λ == 0.5
     end
+
+    @testset "DiscreteNonParametric" begin
+        d1 = DiscreteNonParametric([0,1],[0.5,0.5])
+        d2 = DiscreteNonParametric([1,2],[0.5,0.5])
+        d_eps= DiscreteNonParametric([-1* eps(Float64),0.0,eps(Float64),1.0],[1//4,1//4,1//4,1//4])
```
devmotion, line 73: Maybe use
```suggestion
        d_eps= DiscreteNonParametric([prevfloat(0.0),0.0,nextfloat(0.0),1.0], fill(1//4, 4))
```

test/convolution.jl
```diff
@@ -66,6 +66,27 @@ using Test
...
+        d10= DiscreteNonParametric(collect(1:10)//10,ones(Int,10)//10)
```
devmotion, line 74:
```suggestion
        d10= DiscreteNonParametric(1//10:1//10:1, fill(1//10, 10))
```

src/convolution.jl
```diff
@@ -47,6 +48,20 @@ end
 convolve(d1::Poisson, d2::Poisson) =  Poisson(d1.λ + d2.λ)
 
 
+
```
devmotion, line 51:
```suggestion
```

test/convolution.jl
```diff
@@ -66,6 +66,27 @@ using Test
...
+        
+        d_int_simple = @inferred(convolve(d1, d2))
+        @test d_int_simple isa DiscreteNonParametric
+        @test support(d_int_simple) == [1,2,3]
+        @test probs(d_int_simple) == [0.25,0.5,0.25]
+               
+        d_rat = convolve(d10, d10)
+        @test support(d_rat)[1] == 1//5
```
devmotion, line 82: I guess this could be made more precise?
```suggestion
        @test support(d_rat) == 1//5:1//10:2
```

test/convolution.jl
```diff
@@ -66,6 +66,27 @@ using Test
...
+        @test probs(d_rat)[1] == 1//100
+
+        d_float_supp = convolve(d_eps, d_eps)
+        @test support(d_float_supp)[3] == 0.0
```
devmotion, line 86: I think this will be
```suggestion
        @test support(d_float_supp) == [2 * prevfloat(0.0), prevfloat(0.0), 0.0, nextfloat(0.0), 2 * nextfloat(0.0), 1.0, 2.0]
```
since numerically `prevfloat(0.0) + 1 = 1 = nextfloat(0.0) + 1`.

mharradon:
I think this would be a nice PR to merge. I implemented the latest suggestions from @devmotion and PR'd at #1850. Happy to either see this PR fixed up or merge that.

Thank you!

devmotion:
Fixed by #1850.

PR #1524, opened 2022-03-23
Fix MvNormal subtraction bug

ParadaCarleton:


src/univariate/locationscale.jl
```diff
@@ -139,9 +139,30 @@ gradlogpdf(d::ContinuousAffineDistribution, x::Real) = gradlogpdf(d.ρ,(x-d.μ)/
 
 #### Syntactic sugar for simple transforms of distributions, e.g., d + x, d - x, and so on
 
+"""
+    +(d::Distribution, μ)
+
+Return a distribution that has been shifted by `μ`. If ``X ~ d``, this returns the distribution of ``μ + X``.
+"""
```
devmotion, line 146: This is unrelated and should be part of an improvement of the docs. There are already some open PRs that discuss what would be desired.

```suggestion
```

(BTW I think mu seems to suggest incorrectly that it's the mean or location of the resulting distribution, I think eg `x` would be better.)

devmotion, line 146: This is unrelated and should be part of an improvement of the docs. There are already some open PRs that discuss what would be desired.

```suggestion
```

(BTW I think mu seems to suggest incorrectly that it's the mean or location of the resulting distribution, I think eg `x` would be better.)

src/univariate/locationscale.jl
```diff
@@ -139,9 +139,30 @@ gradlogpdf(d::ContinuousAffineDistribution, x::Real) = gradlogpdf(d.ρ,(x-d.μ)/
...
 Base.:+(d::UnivariateDistribution, x::Real) = AffineDistribution(x, one(x), d)
 Base.:+(x::Real, d::UnivariateDistribution) = d + x
+
+"""
+    *(d::Distribution, σ)
+
+Return a distribution that has been scaled by `σ`. If ``X ~ d``, this returns the distribution of ``σ X``.
+"""
```
devmotion, line 154: Same as above:
```suggestion
```

src/univariate/locationscale.jl
```diff
@@ -139,9 +139,30 @@ gradlogpdf(d::ContinuousAffineDistribution, x::Real) = gradlogpdf(d.ρ,(x-d.μ)/
...
 Base.:*(x::Real, d::UnivariateDistribution) = AffineDistribution(zero(x), x, d)
-Base.:*(d::UnivariateDistribution, x::Real) = x * d
-Base.:-(d::UnivariateDistribution, x::Real) = d + -x
-Base.:/(d::UnivariateDistribution, x::Real) = inv(x) * d
```
devmotion, line 147: These methods should not be removed. Can you revert it?

src/univariate/locationscale.jl
```diff
@@ -139,9 +139,30 @@ gradlogpdf(d::ContinuousAffineDistribution, x::Real) = gradlogpdf(d.ρ,(x-d.μ)/
...
+    -(d::Distribution, μ)
+
+Return a distribution that has been shifted by `-μ`.
+
+See also: [`+`](@ref)
+"""
+Base.:-(d::Distribution, x) = d + -x
+
+
+Base.:/(d::UnivariateDistribution, x) = inv(x) * d
```
devmotion, line 168: This is too general, we only want addition etc. of some specific types. Eg.,

```suggestion
Base.:-(d::MultivariateDistribution, x::AbstractVector{<:Real}) = d + (-x)
```
or
```suggestion
Base.:-(d::Distribution{<:ValueSupport,ArrayLikeVariate{N}}, x::AbstractArray{<:Real,N}) where {N} = d + (-x)
```

devmotion (changes requested):

src/univariate/locationscale.jl
ParadaCarleton: What's wrong with them being a bit more general? In this case, the method not being general is what caused the bug. Switching types to be "As specific as possible without breaking anything" broke the old code (which was more weakly typed). Trying to be ultra-specific with the types like this by only patching the exact place we found a bug feels like an accident waiting to happen.

ParadaCarleton: If we have two distributions, then trying to take `dist_1 - dist_2` will still throw an easy-to-interpret method error when `-dist_2` is undefined, or if it is defined, when addition of distributions is undefined. If, on the other hand, addition of distributions has been well-defined, I would expect subtraction of distributions to work just as well in the same way.

devmotion: We only want to support precisely the subtractions and additions I mentioned above. There's no good reason to be more general here, there are only disadvantages without any benefit it seems.

ParadaCarleton: > We only want to support precisely the subtractions and additions I mentioned above. There's no good reason to be more general here, there are only disadvantages without any benefit it seems.

I outlined the benefits above -- `dist_1 - dist_2` should be defined because we want this to be equivalent to `dist_1 + -dist_2`. In addition, using overly-strict types leads to bugs, like the one this PR fixes.

In fact, I tried using the approach you suggested and ran the tests. The tests broke because in trying to define the change as narrowly as possible, we forgot to define what it means to subtract a distribution from a scalar, as in `x - dist_1`.

ParadaCarleton: Relevant -- it looks like there's a [consensus](https://github.com/JuliaLang/julia/issues/43903) that defining a fallback `-(x, y) = x + -y` is a good idea, and it might be done in 1.9. The definition I proposed tries to hold to that as closely as we can without type piracy, whenever we work with distributions.

devmotion: I was aware of the Julia issue but my impression was a bit different and that some people try to push it. I'm not in favour of it since I don't think it's better to support it only whenever desired for the desired types.

I am also aware of your comments but I still haven't seen any convincing argument for the general definitions. To be clear I don't think I'll approve a PR with these definitions.

We also don't want to support addition of multivariate distributions with real numbers - it's not even supported for `MvNormal` and would require implicit broadcasting.

devmotion:
I'll close this PR in favour of https://github.com/JuliaStats/Distributions.jl/pull/1525. The general definitions in this PR are not correct and not desired in general, and not needed to fix the existing issue. Moreover, the proposed fix here is suboptimal for the issue and causes unnecessary allocations.

PR #1532, opened 2022-04-12
Collect a `MixtureModel` of `DiscreteNonParametric`s into a `DiscreteNonParametric`

yuvalwas:
It is sometimes convenient to "collect" a `MixtureModel` of `DiscreteNonParametric`s into a single `DiscreteNonParametric` object. This PR implements this as an additional constructor but I guess can also be a function with a different name if that is preferred. I was not sure whether to put it in `discretenonparametric.jl` or `mixturemodel.jl`, to me it seemed more specific to `DiscreteNonParametric` than to `MixtureModel`. The implementation is quite trivial however I benchmarked several similar ways of collecting the support:
```
f1(mm) = unique(vcat(support.(components(mm))...))
f2(mm) = unique(reduce(vcat, support.(components(mm))))
f3(mm) = unique(reduce(vcat, (support(m) for m in components(mm)))) # eventually chosen

N = 100000
# create some overlapping support
n = 5000
a1 = sample(1:N, n, replace=false, ordered=true)
a2 = sample(1:N, n, replace=false)
idx = rand(1:n, 2500)
a2 = [a2; a1[idx]] |> unique |> sort
# create some pdf, not actually relevant
b1 = rand(n)
b1 /= sum(b1)
b2 = rand(length(a2))
b2 /= sum(b2)
d1 = DiscreteNonParametric(a1,b1)
d2 = DiscreteNonParametric(a2,b2)
mm = MixtureModel([d1,d2])


julia> @btime f1($mm);
  438.500 μs (34 allocations: 612.31 KiB)

julia> @btime f2($mm);
  438.100 μs (35 allocations: 612.39 KiB)

julia> @btime f3($mm);
  437.200 μs (33 allocations: 612.25 KiB)
```

### DIFF
--- a/src/univariate/discrete/discretenonparametric.jl
+++ b/src/univariate/discrete/discretenonparametric.jl
@@ -312,3 +312,9 @@ end
 fit_mle(::Type{<:DiscreteNonParametric},
         ss::DiscreteNonParametricStats{T,W,Ts,Ws}) where {T,W,Ts,Ws} =
     DiscreteNonParametric{T,W,Ts,Ws}(ss.support, normalize!(copy(ss.freq), 1), check_args=false)
+
+# Collect a MixtureModel of DiscreteNonParametrics
+function Distributions.DiscreteNonParametric(mm::MixtureModel{<:VariateForm,Discrete,DiscreteNonParametric{T,P,Ts,Ps}}) where {T,P,Ts,Ps}
+	supp = unique(reduce(vcat, (support(m) for m in components(mm))))
+	DiscreteNonParametric(supp, pdf.(mm, supp))
+end


yuvalwas:
I now see that inside `discretenonparametric.jl`, `MixtureModel` is undefined. Perhaps it should be inside `MixtureModel.jl`?

src/univariate/discrete/discretenonparametric.jl
```diff
@@ -312,3 +312,9 @@ end
 fit_mle(::Type{<:DiscreteNonParametric},
         ss::DiscreteNonParametricStats{T,W,Ts,Ws}) where {T,W,Ts,Ws} =
     DiscreteNonParametric{T,W,Ts,Ws}(ss.support, normalize!(copy(ss.freq), 1), check_args=false)
+
+# Collect a MixtureModel of DiscreteNonParametrics
+function DiscreteNonParametric(mm::MixtureModel{<:VariateForm,Discrete,DiscreteNonParametric{T,P,Ts,Ps}}) where {T,P,Ts,Ps}
```
devmotion, line 317: I wonder if it would be more appropriate to define a `convert` method than a constructor in this case (https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#Conversion-vs.-Construction).

In any case, I think probably we could make it less restrictive:
```suggestion
function DiscreteNonParametric(mm::UnivariateMixture{Discrete,<:DiscreteNonParametric)
```

src/univariate/discrete/discretenonparametric.jl
```diff
@@ -312,3 +312,9 @@ end
...
+	supp = unique(reduce(vcat, (support(m) for m in components(mm))))
+	DiscreteNonParametric{T,P,Ts,Ps}(supp, pdf.(mm, supp))
```
devmotion, line 319: This can be improved. E.g., the following seems to be faster:
```julia

julia> using BenchmarkTools

julia> f(xs) = unique(reduce(vcat, xs))

julia> g(xs) = collect(Set(Iterators.flatten(xs)))

julia> xs = ntuple(_ -> randn(rand(1:10)), 5);

julia> Set(f(xs)) == Set(g(xs))
true

julia> @btime f($xs);
  986.100 ns (14 allocations: 2.39 KiB)

julia> @btime g($xs);
  586.717 ns (8 allocations: 1.38 KiB)
```
In this setting here one can improve performance further by not allocating a vector of all components and sorting the array in-place.

```suggestion
    supports = (support(component(mm, i)) for i in 1:ncomponents(mm))
    supp = sort!(collect(Set(Iterators.flatten(supports))))
    probs = map(Base.Fix1(pdf, mm), supp)
    return DiscreteNonParametric(supp, probs; check_args=false)
```

devmotion (commented):
Thanks!

It would be good to add some tests. I think it's also possible to improve performance a bit more.

src/univariate/discrete/discretenonparametric.jl
yuvalwas: Thanks for the feedback! Definitelty learned some new stuff. After a thorough investigation, I have some comments:

First I tried again several ways without `Set`:
```
f1(mm) = unique(Iterators.flatten(support(m) for m in components(mm)))
f2(mm) = unique(Iterators.flatten(support(component(mm, i)) for i in 1:ncomponents(mm)))
f3(mm) = unique(vcat(support.(components(mm))...))
f4(mm) = unique(reduce(vcat, support.(components(mm)))) 
f5(mm) = unique(reduce(vcat, (support(m) for m in components(mm))))
f6(mm) = unique(reduce(vcat, Iterators.flatten(support(m) for m in components(mm))))
f7(mm) = unique(reduce(vcat, Iterators.flatten(support(component(mm, i)) for i in 1:ncomponents(mm))))

@assert sort(f1(mm)) == sort(f2(mm)) == sort(f3(mm)) == sort(f4(mm)) == sort(f5(mm)) == sort(f6(mm)) == sort(f7(mm))

@btime f1($mm);
@btime f2($mm);
@btime f3($mm);
@btime f4($mm);
@btime f5($mm);
@btime f6($mm);
@btime f7($mm);
  320.462 μs (31 allocations: 519.89 KiB)
  327.561 μs (31 allocations: 519.89 KiB)
  330.518 μs (34 allocations: 612.81 KiB)
  331.338 μs (35 allocations: 612.89 KiB)
  330.477 μs (33 allocations: 612.75 KiB)
  58.838 ms (21740 allocations: 539.85 MiB)
  59.363 ms (21740 allocations: 539.85 MiB)
```
And indeed using `Set` is faster without taking sorting into account:
```
# without Iterators.flatten it will not give the correct result
f1set(mm) = collect(Set(Iterators.flatten(support(m) for m in components(mm))))
f2set(mm) = collect(Set(Iterators.flatten(support(component(mm, i)) for i in 1:ncomponents(mm))))

@assert sort(f1set(mm)) == sort(f2set(mm)) == sort(f1(mm))

@btime f1set($mm);
@btime f2set($mm);
  294.174 μs (24 allocations: 269.58 KiB)
  296.421 μs (24 allocations: 269.58 KiB)
```
(Note that in any case it is a bit faster to iterate over the components than the indexes)
However, when adding the sorting, `Set` is slower as it is much more unsorted:
```
function f1sort(mm) 
    supp = f1(mm)
    sort!(supp)
    supp
end

function f1setsort(mm) 
    supp = f1set(mm)
    sort!(supp)
    supp
end

@btime f1sort($mm);
@btime f1setsort($mm);
  525.044 μs (31 allocations: 519.89 KiB)
  669.321 μs (24 allocations: 269.58 KiB)
```
`map` does not seem to be faster than broadcasting, and a sorted support speeds up `pdf`:
```
supp1 = f1set(mm) # more unsorted
supp2 = f1(mm) # less unsorted
supp3 = sort(supp1) # sorted
p1(supp, mm) = map(Base.Fix1(pdf, mm), supp)
p2(supp, mm) = pdf.(mm, supp)

@assert p1(supp1, mm) == p2(supp1, mm)

@btime p1($supp1, $mm);
@btime p2($supp1, $mm);
@btime p1($supp2, $mm);
@btime p2($supp2, $mm);
@btime p1($supp3, $mm);
@btime p2($supp3, $mm);
  1.284 ms (2 allocations: 76.23 KiB)
  1.280 ms (2 allocations: 76.23 KiB)
  864.864 μs (2 allocations: 76.23 KiB)
  856.627 μs (2 allocations: 76.23 KiB)
  818.825 μs (2 allocations: 76.23 KiB)
  817.067 μs (2 allocations: 76.23 KiB)
```
What I didn't to understand, is why sorting in place is slower than letting `DiscreteNonParametric` sort upon construction:
```
supp1 = f1set(mm) # more unsorted
supp2 = f1(mm) # less unsorted

h1(supp, mm) = (sort!(supp); ps = pdf.(mm, supp); DiscreteNonParametric(supp, ps))
h2(supp, mm) = (ps = pdf.(mm, supp); DiscreteNonParametric(supp, ps))

@assert h1(supp1, mm) == h2(supp1, mm) 
    
@btime h1($supp1, $mm);
@btime h2($supp1, $mm);
@btime h1($supp2, $mm);
@btime h2($supp2, $mm);
  1.145 ms (31 allocations: 498.30 KiB)
  1.103 ms (31 allocations: 498.30 KiB)
  1.144 ms (31 allocations: 498.30 KiB)
  1.104 ms (31 allocations: 498.30 KiB)
```
So finally:
```
function fun1(mm::UnivariateMixture{Discrete,<:DiscreteNonParametric})
	supp = unique(Iterators.flatten(support(m) for m in components(mm)))
	DiscreteNonParametric(supp, pdf.(mm, supp))
end
function fun2(mm::UnivariateMixture{Discrete,<:DiscreteNonParametric})
    supp = unique(Iterators.flatten(support(m) for m in components(mm)))
    sort!(supp)
    DiscreteNonParametric(supp, pdf.(mm, supp), check_args=false)
end
function fun3(mm::UnivariateMixture{Discrete,<:DiscreteNonParametric})
    supports = (support(component(mm, i)) for i in 1:ncomponents(mm))
    supp = sort!(collect(Set(Iterators.flatten(supports))))
    probs = map(Base.Fix1(pdf, mm), supp)
    DiscreteNonParametric(supp, probs; check_args=false)
end
@assert fun1(mm) == fun2(mm) == fun3(mm)
@btime fun1($mm);
@btime fun2($mm);
@btime fun3($mm);
  1.697 ms (62 allocations: 1018.19 KiB)
  1.339 ms (33 allocations: 596.12 KiB)
  1.485 ms (26 allocations: 345.81 KiB)
```

src/univariate/discrete/discretenonparametric.jl
yuvalwas: Regarding `convert` - tell me whatever you think is best. `convert` makes sense although to me it seems a bit less convenient, as you need to specify the type `DiscreteNonParametric` although there is no really any other option that I can think of. Initially to me it felt like a `collect` but eventually I wrote it as a constructor.

PR #1534, opened 2022-04-24
Differentiation Dirichlet

matbesancon:
This adds the ChainRules rules for Dirichlet, constructor and logpdf


matbesancon:
Some tests are performed with a manual finite diff with specific tangents that remain in the domain

src/multivariate/dirichlet.jl
```diff
@@ -26,10 +26,12 @@ struct Dirichlet{T<:Real,Ts<:AbstractVector{T},S<:Real} <: ContinuousMultivariat
     lmnB::S
 
     function Dirichlet{T}(alpha::AbstractVector{T}; check_args::Bool=true) where T
-        @check_args(
-            Dirichlet,
-            (alpha, all(x -> x > zero(x), alpha), "alpha must be a positive vector."),
-        )
+        if check_args
```
devmotion, line 29: This should not be needed, I think? IIRC `@check_args` already performs this check. The macro was added specifically to improve AD support.

src/multivariate/dirichlet.jl
```diff
@@ -40,7 +42,9 @@ function Dirichlet(alpha::AbstractVector{T}; check_args::Bool=true) where {T<:Re
     Dirichlet{T}(alpha; check_args=check_args)
 end
 function Dirichlet(d::Integer, alpha::Real; check_args::Bool=true)
-    @check_args Dirichlet (d, d > zero(d)) (alpha, alpha > zero(alpha))
+    if check_args
```
devmotion, line 45: Same here.

devmotion (commented):
Thanks! I'll do a proper review later, here just two quick initial comments.

src/multivariate/dirichlet.jl
matbesancon: ah my bad I thought the check_args argument had been forgotten here

matbesancon: done

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,63 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+
+## Differentiation
+function ChainRulesCore.frule((_, Δalpha)::Tuple{Any,Any}, DT::Union{Type{Dirichlet{T}}, Type{Dirichlet}}, alpha::AbstractVector{T}; check_args = true) where {T}
+    d = DT(alpha; check_args=check_args)
+    Δalpha = ChainRulesCore.unthunk(Δalpha)
+    ∂alpha0 = sum(Δalpha)
+    ∂lmnB::typeof(∂alpha0) = sum(Δalpha[i] * (SpecialFunctions.digamma(alpha[i]) - SpecialFunctions.digamma(d.alpha0)) for i in eachindex(alpha))
+    backing = (alpha=Δalpha, alpha0=∂alpha0, lmnB=∂lmnB)
+    t = ChainRulesCore.Tangent{typeof(d), NamedTuple{(:alpha, :alpha0, :lmnB), Tuple{typeof(alpha), typeof(d.alpha0), typeof(d.lmnB)}}}(backing)
+    return d, t
```
devmotion, line 387: This can be simplified to
```suggestion
    tangent = ChainRulesCore.Tangent{typeof(d)}(; alpha=Δalpha, alpha0=∂alpha0, lmnB=∂lmnB)
    return d, tangent
```

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,63 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+end
+
+function ChainRulesCore.rrule(DT::Union{Type{Dirichlet{T}}, Type{Dirichlet}}, alpha::AbstractVector{T}; check_args = true) where {T}
```
devmotion, line 390: This could be changed to
```suggestion
function ChainRulesCore.rrule(::Type{DT}, alpha::AbstractVector{T}; check_args::Bool = true) where {T<:Real, DT<:Union{Dirichlet, Dirichlet{T}}}
```

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,63 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+    d = DT(alpha; check_args=check_args)
+    function dirichlet_pullback(d_dir)
+        d_dir = ChainRulesCore.unthunk(d_dir)
+        ∂l = d_dir.lmnB * (SpecialFunctions.digamma.(alpha) .- SpecialFunctions.digamma.(d.alpha0))
+        return (ChainRulesCore.NoTangent(), d_dir.alpha .+ d_dir.alpha0 .+ ∂l)
```
devmotion, line 395: Reduces some allocations:
```suggestion
        dalpha = d_dir.alpha .+ d_dir.alpha0 .+ d_dir.lmnB .* (SpecialFunctions.digamma.(alpha) .- SpecialFunctions.digamma.(d.alpha0))
        return ChainRulesCore.NoTangent(), dalpha
```

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,63 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+    end
+    return d, dirichlet_pullback
+end
+
+function ChainRulesCore.frule((_, Δd, Δx), ::typeof(_logpdf), d::Dirichlet, x::AbstractVector{T}) where {T}
```
devmotion, line 400: The existing rules in Distributions are defined for `logpdf` but not `_logpdf`. Is there a reason to work with `_logpdf` here?

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,63 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+    lp = _logpdf(d, x)
+    if !insupport(d, x)
+        return (lp, zero(lp))
```
devmotion, line 403: This can cause a type instability since it does not consider the types of the tangents. Maybe we can just run the code below with `max(x[i], zero(x[i]))` instead of `x[i]` and return `zero(tangent)` after performing all computations? This would be similar to how we avoid type instabilities in other methods in Distributions such as `logpdf` itself.

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,63 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+    end
+    ∂α = sum(Δd.alpha[i] * log(x[i]) for i in eachindex(x))
+    ∂l = - Δd.lmnB
+    ∂x = sum((d.alpha[i] - 1) * Δx[i] / x[i] for i in eachindex(x))
+    return (lp, ∂α + ∂l + ∂x)
```
devmotion, line 408: It seems one can combine the two loops. Again, indices might be different. One could use `zip` or pairwise summation with
```suggestion
    dlp = sum(Broadcast.instantiate(Broadcast.broadcasted(d.alpha, Δd.alpha, x, Δx) do alpha_i, Δalpha_i, x_i, Δx_i
        Δalpha_i * log(x_i) + (alpha_i - 1) * Δx_i / x_i
    end)) - Δd.lmnB
    return (lp, dlp)
```

Additionally, it seems the case `x[i] = 0` is problematic and maybe should be handled. `xlogy` could be useful for the first term, not sure about the division in the second one.

src/multivariate/dirichlet.jl
devmotion: Also, we only support `T<:Real` and `T` is not used below it seems:
```suggestion
function ChainRulesCore.frule((_, Δd, Δx), ::typeof(_logpdf), d::Dirichlet, x::AbstractVector{<:Real})
```

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,63 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+end
+
+function ChainRulesCore.rrule(::typeof(_logpdf), d::Dirichlet, x::AbstractVector{T}) where {T}
```
devmotion, line 411: Same question here, is there a reason to use `_logpdf` instead of `logpdf`?
```suggestion
function ChainRulesCore.rrule(::typeof(_logpdf), d::Dirichlet, x::AbstractVector{<:Real})
```

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,63 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+    y = _logpdf(d, x)
+    function Dirichlet_logpdf_pullback(dy)
+        if !isfinite(y)
+            backing = (alpha = zero(d.alpha), alpha0 = ChainRulesCore.ZeroTangent(), lmnB=zero(d.lmnB))
+            ∂d = ChainRulesCore.Tangent{typeof(d), typeof(backing)}(backing)
+            ∂x = zero(d.alpha + x)
+            return (ChainRulesCore.NoTangent(), ∂d, ∂x)
+        end
```
devmotion, line 419: Again this can cause type instabilities and maybe could be handled below by working with non-negative `x` instead.

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,63 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+        ∂alpha = dy * log.(x)
```
devmotion, line 420: Saves allocations:
```suggestion
        ∂alpha = dy .* log.(x)
```
Additionally, maybe for the case `dy = x = 0` it would be useful to use `xlogy`?

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,63 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+        ∂l = -dy
+        ∂x = dy * (d.alpha .-1) ./ x
```
devmotion, line 422:
```suggestion
        ∂x = dy .* (d.alpha .-1) ./ x
```
What about `x[i] = 0`?

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,63 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+        backing = (alpha = ∂alpha, alpha0 = ChainRulesCore.ZeroTangent(), lmnB=∂l)
+        ∂d = ChainRulesCore.Tangent{typeof(d), typeof(backing)}(backing)
```
devmotion, line 424: The tangent seems wrong? `alpha0 = sum(alpha)`, and hence the derivative for `alpha0` should be `sum(∂alpha)`, shouldn't it?
Additionally, a bit simpler:
```suggestion
        ∂d = ChainRulesCore.Tangent{typeof(d)}(;  alpha=∂alpha, alpha0=sum(∂alpha), lmnB=∂l)
```

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,63 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+    return (y, Dirichlet_logpdf_pullback)
+end
+
+function _logpdf(d::Dirichlet, x::AbstractVector{<:Real})
+    if !insupport(d, x)
+        return xlogy(one(eltype(d.alpha)), zero(eltype(x))) - d.lmnB
+    end
+    a = d.alpha
+    s = sum(xlogy(αi - 1, xi) for (αi, xi) in zip(d.alpha, x))
+    return s - d.lmnB
```
devmotion, line 436: This is already defined and causes a redefinition (https://github.com/JuliaStats/Distributions.jl/runs/6145567730?check_suite_focus=true#step:6:41):
```suggestion
```

test/dirichlet.jl
```diff
@@ -127,3 +128,54 @@ end
     @test entropy(Dirichlet(N, 1)) ≈ -loggamma(N)
     @test entropy(Dirichlet(ones(N))) ≈ -loggamma(N)
 end
+
+@testset "Dirichlet differentiation $n" for n in (2, 10)
+    alpha = rand(n)
+    Δalpha = randn(n)
+    d, ∂d = ChainRulesCore.frule((nothing, Δalpha), Dirichlet, alpha)
+    ChainRulesTestUtils.test_frule(Dirichlet ⊢ ChainRulesCore.NoTangent(), alpha ⊢ Δalpha)
```
devmotion, line 136: This causes errors with finite differencing since `alpha` can become negative (https://github.com/JuliaStats/Distributions.jl/runs/6145567730?check_suite_focus=true#step:6:503). I guess using `forward_fdm` instead of `central_fdm` should fix the problem (https://github.com/JuliaDiff/FiniteDifferences.jl#dealing-with-singularities).

```suggestion
    ChainRulesTestUtils.test_frule(Dirichlet ⊢ ChainRulesCore.NoTangent(), alpha ⊢ Δalpha; fdm=FiniteDifferences.forward_fdm(5, 1))
```

test/dirichlet.jl
```diff
@@ -127,3 +128,54 @@ end
...
+    _, dp = ChainRulesCore.rrule(Dirichlet, alpha)
```
devmotion, line 137: Not used anywhere?
```suggestion
```

test/dirichlet.jl
```diff
@@ -127,3 +128,54 @@ end
...
+    ChainRulesTestUtils.test_rrule(Dirichlet{Float64} ⊢ ChainRulesCore.NoTangent(), alpha)
+    x = rand(n)
+    x ./= sum(x)
+    Δx = 0.05 * rand(n)
+    Δx .-= mean(Δx)
+    # such that x ∈ Δ, x + Δx ∈ Δ
```
devmotion, line 143: Is the restriction on `Δx` needed? It seems unnecessary for the `test_frule` below.

test/dirichlet.jl
devmotion: Also `NoTangent` should be the default here (https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/5f523e963c0d86018074beadf9019c2147975c46/src/rand_tangent.jl#L55):
```suggestion
    ChainRulesTestUtils.test_frule(Dirichlet, alpha ⊢ Δalpha; fdm=FiniteDifferences.forward_fdm(5, 1))
```

test/dirichlet.jl
```diff
@@ -127,3 +128,54 @@ end
...
     @test entropy(Dirichlet(ones(N))) ≈ -loggamma(N)
 end
+
+@testset "Dirichlet differentiation $n" for n in (2, 10)
+    alpha = rand(n)
+    Δalpha = randn(n)
+    d, ∂d = ChainRulesCore.frule((nothing, Δalpha), Dirichlet, alpha)
+    ChainRulesTestUtils.test_frule(Dirichlet ⊢ ChainRulesCore.NoTangent(), alpha ⊢ Δalpha)
+    _, dp = ChainRulesCore.rrule(Dirichlet, alpha)
+    ChainRulesTestUtils.test_rrule(Dirichlet{Float64} ⊢ ChainRulesCore.NoTangent(), alpha)
```
devmotion, line 138: Same here, I assume `NoTangent` is the default:
```suggestion
    ChainRulesTestUtils.test_rrule(Dirichlet{Float64}, alpha)
```

test/dirichlet.jl
```diff
@@ -127,3 +128,54 @@ end
...
+    x = rand(n)
+    x ./= sum(x)
+    Δx = 0.05 * rand(n)
+    Δx .-= mean(Δx)
+    # such that x ∈ Δ, x + Δx ∈ Δ
+    ChainRulesTestUtils.test_frule(Distributions._logpdf ⊢ ChainRulesCore.NoTangent(), d, x ⊢ Δx)
```
devmotion, line 144:
```suggestion
    ChainRulesTestUtils.test_frule(Distributions._logpdf, d, x ⊢ Δx)
```

test/dirichlet.jl
```diff
@@ -127,3 +128,54 @@ end
...
+    @testset "finite diff f/r-rule logpdf" begin
```
devmotion, line 145: Why are these manual tests needed? Can't we use eg `test_frule` and `test_rrule` with `forward_fdm`?

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,63 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
     elogp = mean_logp(suffstats(Dirichlet, P, w))
     fit_dirichlet!(elogp, α; maxiter=maxiter, tol=tol, debug=debug)
 end
+
+## Differentiation
+function ChainRulesCore.frule((_, Δalpha)::Tuple{Any,Any}, DT::Union{Type{Dirichlet{T}}, Type{Dirichlet}}, alpha::AbstractVector{T}; check_args = true) where {T}
```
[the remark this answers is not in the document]
matbesancon: I'll check but I think I had type instability issues on the tuple for some reason

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,63 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+    d = DT(alpha; check_args=check_args)
+    Δalpha = ChainRulesCore.unthunk(Δalpha)
+    ∂alpha0 = sum(Δalpha)
+    ∂lmnB::typeof(∂alpha0) = sum(Δalpha[i] * (SpecialFunctions.digamma(alpha[i]) - SpecialFunctions.digamma(d.alpha0)) for i in eachindex(alpha))
```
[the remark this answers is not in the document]
matbesancon: through the change of type, it seems now that the inference works so the type annotation is not needed anymore.
For the indices I'm not sure I follow, it doesn't seem correct if alpha and Deltaalpha don't have the same indices?

src/multivariate/dirichlet.jl
matbesancon: this currently fails inference for some reason :/ hence the weird rewrite

src/multivariate/dirichlet.jl
matbesancon: done

src/multivariate/dirichlet.jl
matbesancon: I added the zeros of the different eltypes, I think this should do it for most type combinations

matbesancon:
On why `_logpdf` and not `logpdf` directly, it was to be matching the implementation signature.
We could add an rrule and frule for logpdf on all distributions that would just call into that of _logpdf?

src/multivariate/dirichlet.jl
matbesancon: done in a single sum using the do notation, I think a broadcast would have allocated right?

matbesancon:
For the tests, the rrule one fails even with forward finite diff, the finite diff version ends ups being NaN

src/multivariate/dirichlet.jl
devmotion: Weird, this should be exactly the same: https://github.com/JuliaDiff/ChainRulesCore.jl/blob/2d75b4be102bb41ba3ac6df6dec8bb9617b20f0f/src/tangent_types/tangent.jl#L42-L45

src/multivariate/dirichlet.jl
devmotion: No, in contrast to regular `broadacst` `Broadcast.instantiate(Broadcast.broadcasted(...))` doesn't allocate. It's a "trick" that is used e.g. in StatsBase but also Distributions to use pairwise summation without allocating intermediate arrays. E.g., compare
```julia
julia> using BenchmarkTools, LinearAlgebra

julia> A = vcat(1f0, fill(1f-8, 10^8));

julia> w = ones(Float32, length(A));

julia> f_dot(A, w) = dot(A, w)
f_dot (generic function with 1 method)

julia> f_sum_zip(A, w) = sum(Ai * wi for (Ai, wi) in zip(A, w))
f_sum_zip (generic function with 1 method)

julia> f_sum_broadcasted(A, w) = sum(Broadcast.instantiate(Broadcast.broadcasted(*, A, w)))
f_sum_broadcasted (generic function with 1 method)

julia> f_dot(A, w)
1.9625133f0

julia> f_sum_zip(A, w)
1.0f0

julia> f_sum_broadcasted(A, w)
1.9999989f0

julia> @btime f_dot($A, $w);
  42.887 ms (0 allocations: 0 bytes)

julia> @btime f_sum_zip($A, $w);
  140.985 ms (0 allocations: 0 bytes)

julia> @btime f_sum_broadcasted($A, $w);
  46.262 ms (0 allocations: 0 bytes)
```

src/multivariate/dirichlet.jl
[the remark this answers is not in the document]
devmotion: I meant that indexing `alpha` and `Deltaalpha` with the same indices `i` is not necessarily correct since one might use e.g. offset indices whereas the other doesn't. So `zip`, or even better `Broadcast.instantiate(Broadcast.broadcasted(...))` with pairwise summation, seems better.

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,56 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
     fit_dirichlet!(elogp, α; maxiter=maxiter, tol=tol, debug=debug)
 end
+
+## Differentiation
+function ChainRulesCore.frule((_, Δalpha), ::Type{DT}, alpha::AbstractVector{T}; check_args::Bool = true) where {T <: Real, DT <: Union{Dirichlet{T}, Dirichlet}}
+    d = DT(alpha; check_args=check_args)
+    Δalpha = ChainRulesCore.unthunk(Δalpha)
+    ∂alpha0 = sum(Δalpha)
+    digamma_alpha0 = SpecialFunctions.digamma(d.alpha0)
+    ∂lmnB = sum(Δalpha[i] * (SpecialFunctions.digamma(alpha[i]) - digamma_alpha0) for i in eachindex(alpha))
```
devmotion: As mentioned in my comment above, I think it woud be better to use `Broadcast.instantiate(Broadcast.broadcasted(...))`.

src/multivariate/dirichlet.jl
devmotion: Does type inference fail only on some Julia versions or is it version independent?

test/dirichlet.jl
matbesancon: removed in latest commits

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,56 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+        return ChainRulesCore.NoTangent(), dalpha
+    end
+    return d, dirichlet_pullback
+end
+
+function ChainRulesCore.frule((_, Δd, Δx), ::typeof(_logpdf), d::Dirichlet, x::AbstractVector{<:Real})
+    lp = _logpdf(d, x)
+    if !insupport(d, x)
+        return (lp, zero(lp) + zero(eltype(Δx)) + zero(eltype(Δd.alpha)) + zero(eltype(Δd.lmnB)))
+    end
```
devmotion, line 408: It still feels somewhat unsatisfying to me since it is different from the computation in the common case `insupport(d, x)` and hence possibly in some special cases could lead to different types. It is also different from the usual approach taken in Distributions to ensure type stability.

Additionally, I think we should not return `0` but `NaN` (of the correct type) here: https://github.com/JuliaStats/Distributions.jl/pull/1490#issuecomment-1101305239

Generally, I think this branch could be removed if one uses `xlogy(Δd.alpha[i], max(zero(x[i]), x[i]))` (or some optimized version of it) below and returns `(lp, (insupport(d, x) ? ∂α_x + ∂l : oftype(∂α_x + ∂l, NaN))` (probably slightly simplified).

(In any case, `zero(eltype(Δd.lmnB))` could be replaced with `zero(Δd.lmnB)`.)

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,56 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+
+function ChainRulesCore.rrule(::typeof(_logpdf), d::Dirichlet, x::AbstractVector{<:Real})
+    y = _logpdf(d, x)
+    function Dirichlet_logpdf_pullback(dy)
+        if !isfinite(y)
+            backing = (alpha = zero(d.alpha), alpha0 = ChainRulesCore.ZeroTangent(), lmnB=zero(d.lmnB))
+            ∂d = ChainRulesCore.Tangent{typeof(d), typeof(backing)}(backing)
+            ∂x = zero(d.alpha + x)
+            return (ChainRulesCore.NoTangent(), ∂d, ∂x)
+        end
```
devmotion, line 424: Also here I think the values are incorrect and should be `NaN`, as in finite differencing. Moreover, also here the branch is unfortunate.

src/multivariate/dirichlet.jl
devmotion: Would be good to get rid of the allocations :slightly_smiling_face:

src/multivariate/dirichlet.jl
devmotion: Also this should be fixed :slightly_smiling_face:

devmotion:
> For the tests, the rrule one fails even with forward finite diff, the finite diff version ends ups being NaN

I assume the problem with finite differencing in general is that perturbing `x` leads to values that are not in the support anymore which therefore yield infinite log probabilities.

src/multivariate/dirichlet.jl
matbesancon: changed to the broadcast version in latest commits

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,59 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+end
+
+function ChainRulesCore.frule((_, Δd, Δx), ::typeof(_logpdf), d::Dirichlet, x::AbstractVector{<:Real})
+    lp = _logpdf(d, x)
+    if !insupport(d, x)
+        return (lp, zero(lp) + zero(eltype(Δx)) + zero(eltype(Δd.alpha)) + zero(eltype(Δd.lmnB)))
+    end
+    ∂α_x = sum(Broadcast.broadcasted(Δd.alpha, Δx, d.alpha, x)) do Δalpha_i, Δx_i, alpha_i, x_i
+        xlogy(Δalpha_i, x_i) + (alpha_i - 1) * Δx_i / x_i
+    end
```
devmotion, line 411: It's only fast with `Broadcast.instantiate` IIRC.
```suggestion
    ∂α_x = sum(Broadcast.instantiate(Broadcast.broadcasted(Δd.alpha, Δx, d.alpha, x)) do Δalpha_i, Δx_i, alpha_i, x_i
        xlogy(Δalpha_i, x_i) + (alpha_i - 1) * Δx_i / x_i
    end))
```

src/multivariate/dirichlet.jl
matbesancon: yeah I read the argument for NaN values and it makes sense here

src/multivariate/dirichlet.jl
matbesancon: done

src/multivariate/dirichlet.jl
matbesancon: done in latest commits

src/multivariate/dirichlet.jl
devmotion: Can you fix this?
```suggestion
        ∂x = dy .* (d.alpha .-1) ./ x
```

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,60 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+    return (lp, ∂α_x + ∂l)
+end
+
+function ChainRulesCore.rrule(::typeof(_logpdf), d::Dirichlet, x::AbstractVector{<:Real})
+    y = _logpdf(d, x)
+    function Dirichlet_logpdf_pullback(dy)
+        ∂alpha = xlogy.(dy, x)
+        ∂l = -dy
+        ∂x = dy * (d.alpha .-1) ./ x
+        ∂alpha0 = 0.0
```
devmotion, line 422: I think this should be
```suggestion
        ∂alpha0 = sum(∂alpha)
```
?

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,60 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+        if !isfinite(y)
+            ∂alpha .= NaN
```
devmotion, line 424: I'm not sure if we can assume that the derivative is mutable. If e.g. `dy` and `x` are StaticArrays, it won't be mutable.

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,60 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+            ∂l = oftype(∂l, NaN)
+            ∂x .= NaN
```
devmotion, line 426: Same here, maybe we should not require mutability.

src/multivariate/dirichlet.jl
```diff
@@ -375,3 +375,60 @@ function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
...
+            ∂alpha0 = NaN
```
devmotion, line 427: If we don't use the Float64 literal above:
```suggestion
            ∂alpha0 = oftype(∂alpha0, NaN)
```

src/multivariate/dirichlet.jl
matbesancon: it failed at least on 1.8

test/dirichlet.jl
matbesancon: thought so too but otherwise your second point is outside the domain with probability 1
```
  Expression: isapprox(actual, expected; kwargs...)
   Evaluated: isapprox(1.0589089695661067, NaN; rtol = 1.0e-9, atol = 1.0e-9)
```

matbesancon: this is also the case with fdm set to forward

src/multivariate/dirichlet.jl
matbesancon: does it make sense here? I thought the gradient wrt to each field is computed with how directly involved it is in the function, not with their interdependence

matbesancon: bump here @devmotion, this implies that we integrate the constructor in the rule here (defining ∂alpha0 as the one from the constructor, and same thing for lmnB

src/multivariate/dirichlet.jl
devmotion: I just checked locally and there are no type inference issues with the suggestion above. I'll run the full tests as well.

devmotion: I could confirm my impression, tests pass with older and recent Julia versions: https://github.com/JuliaStats/Distributions.jl/pull/1555

src/multivariate/dirichlet.jl
devmotion: I see, I think I agree with you regarding alpha0 as it does not enter the `logpdf` computation and hence probably should be neglected here. I removed it in #1555, without defining it explicitly it will be `ZeroTangent` automatically.  However, the derivative for lmnB has to be defined since it is part of the logpdf computations.

matbesancon: yes agreed, that's the case already I believe

devmotion (approved):
Looks good to me, is there anything else you would like to include in this PR?

matbesancon:
setting it as draft to not merge it yet, I have to sort out some things before

matbesancon:
Merging as already discussed and approved

matbesancon:
No all good on my side thanks a lot!

PR #1536, opened 2022-04-25
ajr/negativebinomialvariants

andrewjradcliffe:
This introduces 3 new distribution types, corresponding to
re-parameterizations of the NegativeBinomial distribution as
1. location-scale [Negative binomial distribution (alternative paramerization), Stan](https://mc-stan.org/docs/2_29/functions-reference/nbalt.html)
2. loglocation-scale [Negative binomial distribution (log alternative parameterization), Stan](https://mc-stan.org/docs/2_29/functions-reference/neg-binom-2-log.html)
3. shape-scale [Negative binomial distribution, Bayesian Data Analysis (3rd edition), Appendix A](http://www.stat.columbia.edu/~gelman/book/BDA3.pdf)

These have a number of uses. In particular, 1. and 2. emphasize the
that the negative binomial distribution is a robust form of the Poisson,
which has noted utility for negative binomial regression involving
offsets (a model frequently employed in epidemiology, failure
modeling). The shape-scale parameterization confers superior
numerical stability, which can be useful when working with rare
events (particularly when one wishes to draw samples).

In particular, for rare events, the alternative parameterizations
are robust in the presence of values which might otherwise suffer
from floating point roundoff using the `NegativeBinomial`.
This can be demonstrated by considering the conversion of a
location-scale `NegativeBinomial2` to `NegativeBinomial`

```
using Plots
gr(size=(600,400))
f(μ, ϕ) = ϕ / (μ + ϕ)
p1 = contour(1e-16:5e-17:1e-15, 1.0:0.01:10.0, (x, y) -> log(f(x, y)), ylabel="ϕ", xlabel="μ", right_margin=40*Plots.px);
p2 = heatmap(1e-16:5e-17:1e-15, 1.0:0.01:10.0, (x, y) -> log(f(x, y)), ylabel="ϕ", xlabel="μ", right_margin=40*Plots.px);
```

I write the above not to malign the choice of the `NegativeBinomial`
in terms of (`r`, `p`), but to emphasize that there are real
advantages to using the alternative parameterizations. Naturally,
conversions are provided between the parameterizations, so that
the experience is seamless.

Notes:
- `] test Distributions` revealed no issues.
- Documentation additions seem to work without issue.


docs/src/univariate.md
```diff
@@ -481,6 +481,9 @@ DiscreteNonParametric
 Geometric
 Hypergeometric
 NegativeBinomial
+NegativeBinomial2
+NegativeBinomial2Log
+NegativeBinomial3
```
devmotion, line 486: I would prefer more descriptive names, similar to `Normal`/`NormalCanon`.

src/univariate/discrete/negativebinomial.jl
```diff
@@ -59,6 +59,19 @@ function Base.convert(::Type{NegativeBinomial{T}}, d::NegativeBinomial) where {T
...
+#     NegativeBinomial{T}(T(d.ϕ), T(d.ϕ / (d.μ + d.ϕ)))
+# end
+# function convert(::Type{NegativeBinomial{T}}, d::NegativeBinomial2Log) where {T<:Real}
+#     NegativeBinomial{T}(T(d.ϕ), T(d.ϕ / (exp(d.η) + d.ϕ)))
+# end
+# function convert(::Type{NegativeBinomial{T}}, d::NegativeBinomial3) where {T<:Real}
+#     NegativeBinomial{T}(T(d.α), T(one(T) / (d.β + one(T))))
+# end
+
+
```
devmotion, line 74: Maybe add a comment that conversion methods are defined in src/conversion.jl but I don't think it's helpful to copy code from there in a comment. The implementation might diverge and it creates additional but not really needed changes in this PR.

devmotion:
```suggestion
```

src/univariate/discrete/negativebinomial2.jl
```diff
@@ -0,0 +1,137 @@
...
+# function convert(::Type{NegativeBinomial2{T}}, d::NegativeBinomial) where {T<:Real}
+#     NegativeBinomial2{T}(T(d.r * (1 - d.p) / d.p), T(d.r))
+# end
+# function convert(::Type{NegativeBinomial2{T}}, d::NegativeBinomial2Log) where {T<:Real}
+#     NegativeBinomial2{T}(T(exp(d.η)), T(d.ϕ))
+# end
+# function convert(::Type{NegativeBinomial2{T}}, d::NegativeBinomial3) where {T<:Real}
+#     NegativeBinomial2{T}(T(d.α * d.β), T(d.α))
+# end
+
```
devmotion, line 75: Same here:
```suggestion
```

src/univariate/discrete/negativebinomial2.jl
```diff
@@ -0,0 +1,137 @@
...
+    NegativeBinomial2{T}(μ::T, ϕ::T) where {T<:Real} = new{T}(μ, ϕ)
+end
+
+function NegativeBinomial2(μ::T, ϕ::T; check_args::Bool=true) where {T<:Real}
+    @check_args NegativeBinomial2 (μ, μ > zero(μ)) (ϕ, ϕ > zero(ϕ))
+    return NegativeBinomial2{T}(μ, ϕ)
+end
+
+NegativeBinomial2(μ::Real, ϕ::Real; check_args::Bool=true) = NegativeBinomial2(promote(μ, ϕ)...; check_args=check_args)
+NegativeBinomial2(μ::Integer, ϕ::Integer; check_args::Bool=true) = NegativeBinomial2(float(μ), float(ϕ); check_args=check_args)
```
devmotion, line 47: Is this needed? Integer-valued parameters are fine in principle, but for historical reasons often converted to floats in Distributions.
```suggestion
```

src/univariate/discrete/negativebinomial2.jl
```diff
@@ -0,0 +1,137 @@
...
+
+@distr_support NegativeBinomial2 0 Inf
+
+insupport(d::NegativeBinomial2, x::Real) = false
+insupport(d::NegativeBinomial2, x::T) where {T<:Integer} = x ≥ 0
+#### Conversions
+
+function convert(::Type{NegativeBinomial2{T}}, μ::Real, ϕ::Real) where {T<:Real}
+    NegativeBinomial2(T(μ), T(ϕ))
+end
```
devmotion, line 59: Actually I think the existing conversion methods of this type should be deprecated and eventually removed - I don't see any benefit over the constructors and they are quite unidiomatic. So maybe we don't have to add them for the new types:
```suggestion
```

src/univariate/discrete/negativebinomial2log.jl
```diff
@@ -0,0 +1,141 @@
...
+
+@distr_support NegativeBinomial2Log 0 Inf
+
+insupport(d::NegativeBinomial2Log, x::Real) = false
+insupport(d::NegativeBinomial2Log, x::T) where {T<:Integer} = x ≥ 0
+#### Conversions
+
+function convert(::Type{NegativeBinomial2Log{T}}, η::Real, ϕ::Real) where {T<:Real}
+    NegativeBinomial2Log(T(η), T(ϕ))
+end
```
devmotion, line 56:
```suggestion
```

src/univariate/discrete/negativebinomial2.jl
```diff
@@ -0,0 +1,137 @@
...
+# function convert(::Type{NegativeBinomial2{T}}, d::NegativeBinomial3) where {T<:Real}
+#     NegativeBinomial2{T}(T(d.α * d.β), T(d.α))
+# end
+
+#### Parameters
+
+params(d::NegativeBinomial2) = (d.μ, d.ϕ)
+partype(::NegativeBinomial2{T}) where {T} = T
+
+succprob(d::NegativeBinomial2{T}) where {T} = d.ϕ / (d.μ + d.ϕ)
```
devmotion, line 81: `T` not used:
```suggestion
succprob(d::NegativeBinomial2) = d.ϕ / (d.μ + d.ϕ)
```

src/univariate/discrete/negativebinomial2.jl
```diff
@@ -0,0 +1,137 @@
...
+failprob(d::NegativeBinomial2{T}) where {T} = d.μ / (d.μ + d.ϕ)
+
+#### Statistics
+
+mean(d::NegativeBinomial2{T}) where {T} = d.μ
```
devmotion, line 86:
```suggestion
mean(d::NegativeBinomial2) = d.μ
```

src/univariate/discrete/negativebinomial3.jl
```diff
@@ -0,0 +1,134 @@
...
+
+@distr_support NegativeBinomial3 0 Inf
+
+insupport(d::NegativeBinomial3, x::Real) = false
+insupport(d::NegativeBinomial3, x::T) where {T<:Integer} = x ≥ 0
+#### Conversions
+
+function convert(::Type{NegativeBinomial3{T}}, α::Real, β::Real) where {T<:Real}
+    NegativeBinomial3(T(α), T(β))
+end
```
devmotion, line 56:
```suggestion
```

src/univariate/discrete/negativebinomial3.jl
```diff
@@ -0,0 +1,134 @@
...
+# function convert(::Type{NegativeBinomial3{T}}, d::NegativeBinomial) where {T<:Real}
+#     NegativeBinomial3{T}(T(d.r), T((1 - d.p) / d.p))
+# end
+# function convert(::Type{NegativeBinomial3{T}}, d::NegativeBinomial2) where {T<:Real}
+#     NegativeBinomial3{T}(T(d.ϕ), T(d.μ / d.ϕ))
+# end
+# function convert(::Type{NegativeBinomial3{T}}, d::NegativeBinomial2Log) where {T<:Real}
+#     NegativeBinomial3{T}(T(d.ϕ), T(exp(d.η) / d.ϕ))
+# end
+
```
devmotion, line 72:
```suggestion
```

src/univariate/discrete/negativebinomial3.jl
```diff
@@ -0,0 +1,134 @@
...
+params(d::NegativeBinomial3) = (d.α, d.β)
+partype(::NegativeBinomial3{T}) where {T} = T
+
+succprob(d::NegativeBinomial3{T}) where {T} = inv(d.β + one(T))
+failprob(d::NegativeBinomial3{T}) where {T} = (β = d.β; β / (β + one(T)))
+
+#### Statistics
+
+mean(d::NegativeBinomial3{T}) where {T} = d.α * d.β
```
devmotion, line 83:
```suggestion
mean(d::NegativeBinomial3) = d.α * d.β
```

src/univariate/discrete/negativebinomial2.jl
andrewjradcliffe: This one was a holdover from the extant negativebinomial.jl; I wasn't certain as to why it was there, but thought: might as well carry it forward. Happily removed.

src/univariate/discrete/negativebinomial2.jl
andrewjradcliffe: I agree. I was carrying them forward, but happy to remove them.

src/univariate/discrete/negativebinomial2log.jl
```diff
@@ -0,0 +1,141 @@
...
+    NegativeBinomial2Log{T}(η::T, ϕ::T) where {T<:Real} = new{T}(η, ϕ)
+end
+
+function NegativeBinomial2Log(η::T, ϕ::T; check_args::Bool=true) where {T<:Real}
+    @check_args NegativeBinomial2Log (η, η > -Inf) (ϕ, ϕ > zero(ϕ))
+    return NegativeBinomial2Log{T}(η, ϕ)
+end
+
+NegativeBinomial2Log(η::Real, ϕ::Real; check_args::Bool=true) = NegativeBinomial2Log(promote(η, ϕ)...; check_args=check_args)
+NegativeBinomial2Log(η::Integer, ϕ::Integer; check_args::Bool=true) = NegativeBinomial2Log(float(η), float(ϕ); check_args=check_args)
```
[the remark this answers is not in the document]
andrewjradcliffe: Analogously to the other cases involving float(...), I agree. Better to simplify the interface and leave the work with promote(...)

src/univariate/discrete/negativebinomial3.jl
andrewjradcliffe: It's true, this `{T} where {T}` should not be here.

src/univariate/discrete/negativebinomial2.jl
```diff
@@ -0,0 +1,137 @@
...
+
+NegativeBinomial2(μ::Real, ϕ::Real; check_args::Bool=true) = NegativeBinomial2(promote(μ, ϕ)...; check_args=check_args)
+NegativeBinomial2(μ::Integer, ϕ::Integer; check_args::Bool=true) = NegativeBinomial2(float(μ), float(ϕ); check_args=check_args)
+NegativeBinomial2(μ::Real; check_args::Bool=true) = NegativeBinomial2(μ, 1.0; check_args=check_args)
+NegativeBinomial2() = NegativeBinomial2{Float64}(1.0, 1.0)
+
+@distr_support NegativeBinomial2 0 Inf
+
+insupport(d::NegativeBinomial2, x::Real) = false
+insupport(d::NegativeBinomial2, x::T) where {T<:Integer} = x ≥ 0
```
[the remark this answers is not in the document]
andrewjradcliffe: Agree. This is a more elegant way of using the type system (via `isinteger`, which has the same effect as the two dispatches).

src/univariate/discrete/negativebinomial2log.jl
```diff
@@ -0,0 +1,141 @@
...
+
+#### Evaluation & Sampling
+@inline binomial_log(n, k) = loggamma(n + 1) - loggamma(k + 1) - loggamma(n - k + 1)
+# binomial_log2(n, k) = log(n) - log((n - k) * k) - logbeta(n - k, k)
+
+function logpdf(d::NegativeBinomial2Log, n::Real)
+    η, ϕ = params(d)
+    # ϕₘ₁ = ϕ - 1
+    # c = log(n + ϕₘ₁) - log(n * ϕₘ₁) - logbeta(n, ϕₘ₁)
+    # c = binomial_log(n + ϕₘ₁, n) # safer; TO DO: to create branches to use logbeta
```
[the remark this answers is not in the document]
andrewjradcliffe: Removed.

docs/src/univariate.md
andrewjradcliffe: I agree, more descriptive names would be nice. A believe a reasonable argument could be made for re-naming `NegativeBinomial3` as `PoissonGamma`. 

For the location-overdispersion parameterization (and the respective log version), it could be re-named uncreatively as `NegativeBinomialAlt` (and `NegativeBinomialAltLog`).

andrewjradcliffe:
> Additionally, I think the tests should be extended such that most/all newly added lines are covered.

With respect to tests, a quick inspection of the `NegativeBinomial` reveals that it mostly relies on R for the tests. The readme.md file in `/test/ref/` seems to indicate that extending the R tests to these additional distributions should proceed analogously to `NormalCanon`?

Thoughts:
- What sorts of additional tests should I add in Julia?
- The R tests are nicely set up, but for this example, will merely verify the math specifying the conversions between parameterizations. Any other suggestions for tests?

docs/src/univariate.md
devmotion: Unfortunately, I am not very creative here either and don't have much better suggestions right now :smile: 

Maybe one could use `NegativeBinomialMean` instead of `NegativeBinomial2` to highlight that it uses the mean instead of the number of successes. But I guess the name could maybe be misinterpreted as the mean of a negative binomial distribution.

devmotion:
Yeah, unfortunately the test setup is a bit of a mess, mostly due to historical reasons.

Since `NegativeBinomial` is already checked with the values from R, I think a good start would be to check that the alternative parameterizations are consistent with `NegativeBinomial` in Julia. Currently, in this PR only the `logpdf` values are compared it seems but actually ideally we would check all implemented functions. Sampling should be checked as well, e.g., by computing summary statistics (not sure how it's done for `NegativeBinomial`).

Regarding R, I guess one could copy the existing references for `NegativeBinomial` and replace the Julia type and its parameters with the ones from the alternative parameterizations (ie. update the Julia side without changing the R part).

docs/src/univariate.md
st--: How about
```suggestion
NegativeBinomialLocationScale
NegativeBinomialLogLocationScale
NegativeBinomialShapeScale
```
? that's how you explain them in the PR description, which seems sensible and fairly self-explanatory.

src/univariate/discrete/negativebinomial2.jl
```diff
@@ -0,0 +1,121 @@
...
+end
+
+function NegativeBinomial2(μ::T, ϕ::T; check_args::Bool=true) where {T<:Real}
+    @check_args NegativeBinomial2 (μ, μ > zero(μ)) (ϕ, ϕ > zero(ϕ))
+    return NegativeBinomial2{T}(μ, ϕ)
+end
+
+NegativeBinomial2(μ::Real, ϕ::Real; check_args::Bool=true) = NegativeBinomial2(promote(μ, ϕ)...; check_args=check_args)
+NegativeBinomial2(μ::Real; check_args::Bool=true) = NegativeBinomial2(μ, one(μ); check_args=check_args)
+NegativeBinomial2() = NegativeBinomial2{Float64}(1.0, 1.0)
```
st--, line 48: For distributions with a "standard" form, e.g. `Normal() == Normal(0, 1)` makes sense, but here I'm not sure in what situation you'd want to use a NegativeBinomial with mean 1 and variance 2... from a user point of view it seems better to say "you need to provide these arguments" than to hard-code these?
```suggestion
```

src/univariate/discrete/negativebinomial2.jl
```diff
@@ -0,0 +1,121 @@
...
+P(X = n) = {n + \\phi - 1 \\choose n} (\\frac{\\mu}{\\mu + \\phi})^n (\\frac{\\phi}{\\mu + \\phi})^\\phi, \\quad \\text{for } n = 0,1,2,\\ldots
+```
+
+In terms of the mixture definition, this corresponds to:
+```math
+n \\sim \\text{Poisson}(z), \\quad
+z \\sim \\text{Gamma}(\\phi, \\frac{\\mu}{\\phi})
+```
+This provides a clear interpretation of the negative binomial as an overdispersed
+Poisson with expected location 𝔼[z] = μ and var[z] = μ + μ²/ϕ. It also has advantages in that
```
st--, line 15: Is there a specific reason why you're choosing this parametrisation, as opposed to `var[z] = μ + μ² * ϕ`?
this would make more sense in relation to calling ϕ the overdispersion parameter, as then larger ϕ -> larger dispersion, and ϕ = 0 would actually recover the `Poisson` distribution.

docs/src/univariate.md
devmotion: An option would be also to add an alias `const PoissonGamma = NegativeBinomialPoissonGamma` for people that prefer the shorter name.

andrewjradcliffe: With respect to the definition of 'scale parameter', I agree -- ϕ is actually a dispersion parameter in the location parameterizations. Consequently, I will update the wording in the docstrings to reflect this.

Also, I think that the combined proposal of:
```
NegativeBinomialLocation
NegativeBinomialLogLocation
NegativeBinomialPoissonGamma
const PoissonGamma = NegativeBinomialPoissonGamma
```
achieves the best clarity.

src/univariate/discrete/negativebinomial3.jl
```diff
@@ -0,0 +1,118 @@
...
+
+In terms of the mixture definition, this corresponds to:
+```math
+n \\sim \\text{Poisson}(z), \\quad
+z \\sim \\text{Gamma}(\\alpha, \\beta)
+```
+
+```julia
+NegativeBinomial3()        # distribution with α = 1.0 and β = 1.0
+NegativeBinomial3(α, β)    # distribution with shape α and scale β
```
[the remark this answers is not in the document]
andrewjradcliffe: Indeed, BDA3's `Gamma(α, β)` is in terms of shape and inverse scale. The pmf for the above is obtained by substituting `β_dist = 1 / β` into the BDA definition. The β in the docstring refers to scale; perhaps it would be a good idea to re-name it to θ for clarity.

src/univariate/discrete/negativebinomial2.jl
andrewjradcliffe: This is a good point, which, incidentally, applies to all 3 new definitions. I added the zero-arg methods in analogy to `NegativeBinomial`'s zero-arg method. I am not particularly attached to them, hence, we could remove them? Thoughts @devmotion ?

devmotion: Seems reasonable. We can always add constructors with default arguments at a later point if we think they are really needed and/or useful.

src/univariate/discrete/negativebinomial2.jl
andrewjradcliffe: Thank you for this suggestion. At first, I thought: why deviate from Stan? But then remembered [this](https://github.com/stan-dev/stan/wiki/Prior-Choice-Recommendations#story-when-the-generic-prior-fails-the-case-of-the-negative-binomial). Clearly, it's preferable to use `var[z] = μ + μ² * ϕ`, rather than `var[z] = μ + μ²/ϕ`, as one can then reason about overdispersion without invoking its inverse. Incidentally, Dan Simpson's [remarks](https://statmodeling.stat.columbia.edu/2018/04/03/justify-my-love/) support this line of reasoning.

But why might the Stan implementation prefer the inverse? Nothing that I can see from a numerical stability perspective. Other than a historical precedent, my guess might be that the pmf is uglier when parameterized such that `var[z] = μ + μ² * ϕ`.

DoktorMike:
Just a friendly ping of this PR since I indeed would like to use these alternative parameterizations of the NegativeBinomial. Is there anything major holding a merge back?

PR #1547, opened 2022-05-12
Do not include docstring for deprecated `TruncatedNormal`

devmotion:
`TruncatedNormal` is deprecated, hence it seems a bit strange (and possibly misleading) to include it in the documentation.


devmotion:
I'll merge since this does not affect the package and users are not supposed to use this constructor.

Issue #1265, opened 2021-01-23
Incorrect uses of @inbounds cause miscalculation of statistics


yurivish:
There are many incorrect uses of `@inbounds` in this package, leading to incorrect results when fitting distributions and doing other calculations on arrays with offset axes. 

I came across this issue [previously](https://github.com/JuliaStats/Distributions.jl/issues/1253) with `DiscreteUniform` but did not realize how widespread the problem was until I ran a [search](https://github.com/JuliaStats/Distributions.jl/search?q=inbounds) for `@inbounds` across the code in this package. 

Many uses of `@inbounds` in this package are in functions that accept arbitrary abstract vectors, matrices, or arrays, then disable bounds checks and index into the array with unsafe one-based indices.

One example is fitting a Normal distribution:

```julia
julia> using Distributions, OffsetArrays

julia> a = collect(-5:5);

julia> b = OffsetArray(a, -5:5);

julia> suffstats(Normal, a)
Distributions.NormalStats(0.0, 0.0, 110.0, 11.0)

julia> fit(Normal, a)
Normal{Float64}(μ=0.0, σ=3.1622776601683795)

julia> suffstats(Normal, b)
Distributions.NormalStats(7.39673056934e11, 6.7243005175818184e10, 1.5901446116594743e23, 11.0)

julia> fit(Normal, b)
Normal{Float64}(μ=6.7243005175818184e10, σ=1.2023252515852448e11)
```

https://github.com/JuliaStats/Distributions.jl/blob/863844c88e4153af13996f571fcc612d159de542/src/univariate/continuous/normal.jl#L254-L271

```
julia> versioninfo()
Julia Version 1.5.3
Commit 788b2c77c1 (2020-11-09 13:37 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin18.7.0)
  CPU: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)
```

ParadaCarleton:
Are there any uses of inbounds left after [this PR](https://github.com/JuliaStats/Distributions.jl/pull/1742), @itsdebartha?

itsdebartha:
[...]

PR #1550, opened 2022-05-16
use `axes()` and `eachindex()` to better support arbitrary `AbstractArray`

palday:
Starts the process of fixing #1265.

### DIFF
--- a/src/mixtures/mixturemodel.jl
+++ b/src/mixtures/mixturemodel.jl
@@ -377,8 +377,8 @@ _logpdf!(r::AbstractArray, d::MultivariateMixture, x::AbstractMatrix) = _mixlogp
 function _cwise_pdf1!(r::AbstractVector, d::AbstractMixtureModel, x)
     K = ncomponents(d)
     length(r) == K || error("The length of r should match the number of components.")
-    for i = 1:K
-        r[i] = pdf(component(d, i), x)
+    for (ri, i) in zip(axes(r,1), 1:K)
+        r[ri] = pdf(component(d, i), x)
     end
     r
 end
@@ -386,8 +386,8 @@ end
 function _cwise_logpdf1!(r::AbstractVector, d::AbstractMixtureModel, x)
     K = ncomponents(d)
     length(r) == K || error("The length of r should match the number of components.")
-    for i = 1:K
-        r[i] = logpdf(component(d, i), x)
+    for (ri, i) in zip(axes(r,1), 1:K)
+        r[ri] = logpdf(component(d, i), x)
     end
     r
 end
@@ -396,11 +396,11 @@ function _cwise_pdf!(r::AbstractMatrix, d::AbstractMixtureModel, X)
     K = ncomponents(d)
     n = size(X, ndims(X))
     size(r) == (n, K) || error("The size of r is incorrect.")
-    for i = 1:K
+    for (ri, i) in zip(axes(r, 2), 1:K)
         if d isa UnivariateMixture
-            view(r,:,i) .= pdf.(Ref(component(d, i)), X)
+            view(r,:,ri) .= pdf.(Ref(component(d, i)), X)
         else
-            pdf!(view(r,:,i),component(d, i), X)
+            pdf!(view(r,:,ri),component(d, i), X)
         end
     end
     r
@@ -410,11 +410,11 @@ function _cwise_logpdf!(r::AbstractMatrix, d::AbstractMixtureModel, X)
     K = ncomponents(d)
     n = size(X, ndims(X))
     size(r) == (n, K) || error("The size of r is incorrect.")
-    for i = 1:K
+    for (ri, i) in zip(axes(r, 2), 1:K)
         if d isa UnivariateMixture
-            view(r,:,i) .= logpdf.(Ref(component(d, i)), X)
+            view(r,:,ri) .= logpdf.(Ref(component(d, i)), X)
         else
-            logpdf!(view(r,:,i), component(d, i), X)
+            logpdf!(view(r,:,ri), component(d, i), X)
         end
     end
     r
--- a/src/mixtures/mixturemodel.jl
+++ b/src/mixtures/mixturemodel.jl
@@ -318,7 +318,7 @@ function _mixlogpdf!(r::AbstractArray, d::AbstractMixtureModel, x)
     p = probs(d)
     n = length(r)
     Lp = Matrix{eltype(p)}(undef, n, K)
-    m = fill(-Inf, n)
+    m = similar(r)
     @inbounds for i in eachindex(p)
         pi = p[i]
         if pi > 0.0
@@ -334,7 +334,7 @@ function _mixlogpdf!(r::AbstractArray, d::AbstractMixtureModel, x)
 
             # in the mean time, add log(prior) to lp and
             # update the maximum for each sample
-            for j = 1:n
+            for j in axes(r, 1)
                 lp_i[j] += lpri
                 if lp_i[j] > m[j]
                     m[j] = lp_i[j]
@@ -347,13 +347,13 @@ function _mixlogpdf!(r::AbstractArray, d::AbstractMixtureModel, x)
     @inbounds for i = 1:K
         if p[i] > 0.0
             lp_i = view(Lp, :, i)
-            for j = 1:n
-                r[j] += exp(lp_i[j] - m[j])
+            for (j, lp_ij) in zip(axes(r, 1), lp_i)
+                r[j] += exp(lp_ij - m[j])
             end
         end
     end
 
-    @inbounds for j = 1:n
+    @inbounds for j in axes(r, 1)
         r[j] = log(r[j]) + m[j]
     end
     return r
--- a/src/multivariates.jl
+++ b/src/multivariates.jl
@@ -37,7 +37,7 @@ function insupport!(r::AbstractArray, d::Union{D,Type{D}}, X::AbstractMatrix) wh
     n = length(r)
     size(X) == (length(d),n) ||
         throw(DimensionMismatch("Inconsistent array dimensions."))
-    for i in 1:n
+    for i in axes(r, 1)
         @inbounds r[i] = insupport(d, view(X, :, i))
     end
     return r


src/mixtures/mixturemodel.jl
```diff
@@ -318,7 +318,7 @@ function _mixlogpdf!(r::AbstractArray, d::AbstractMixtureModel, x)
     p = probs(d)
     n = length(r)
     Lp = Matrix{eltype(p)}(undef, n, K)
-    m = fill(-Inf, n)
+    m = similar(r)
```
devmotion, line 321: I think this will cause incorrect results, the algorithm below is based on the fact that `m` is filled with `-Inf`.

src/mixtures/mixturemodel.jl
```diff
@@ -334,7 +334,7 @@ function _mixlogpdf!(r::AbstractArray, d::AbstractMixtureModel, x)
 
             # in the mean time, add log(prior) to lp and
             # update the maximum for each sample
-            for j = 1:n
+            for j in axes(r, 1)
```
devmotion, line 337: This should be
```suggestion
            for j in eachindex(r)
```
it seems.

src/mixtures/mixturemodel.jl
devmotion: Maybe use
```suggestion
    m = similar(r)
    fill!(m, -Inf)
```

src/mixtures/mixturemodel.jl
```diff
@@ -347,13 +347,13 @@ function _mixlogpdf!(r::AbstractArray, d::AbstractMixtureModel, x)
     @inbounds for i = 1:K
         if p[i] > 0.0
             lp_i = view(Lp, :, i)
-            for j = 1:n
-                r[j] += exp(lp_i[j] - m[j])
+            for (j, lp_ij) in zip(axes(r, 1), lp_i)
```
devmotion, line 350: This seems wrong as well? Maybe rather
```suggestion
            for (j, lp_ij) in zip(eachindex(r), lp_i)
```

src/mixtures/mixturemodel.jl
```diff
@@ -347,13 +347,13 @@ function _mixlogpdf!(r::AbstractArray, d::AbstractMixtureModel, x)
...
+                r[j] += exp(lp_ij - m[j])
             end
         end
     end
 
-    @inbounds for j = 1:n
+    @inbounds for j in axes(r, 1)
```
devmotion, line 356:
```suggestion
    @inbounds for j in eachindex(r)
```

src/mixtures/mixturemodel.jl
```diff
@@ -377,17 +377,17 @@ _logpdf!(r::AbstractArray, d::MultivariateMixture, x::AbstractMatrix) = _mixlogp
 function _cwise_pdf1!(r::AbstractVector, d::AbstractMixtureModel, x)
     K = ncomponents(d)
     length(r) == K || error("The length of r should match the number of components.")
-    for i = 1:K
-        r[i] = pdf(component(d, i), x)
+    for (ri, i) in zip(axes(r,1), 1:K)
```
devmotion, line 380:
```suggestion
    for (i, ri) in enumerate(eachindex(r))
```

src/mixtures/mixturemodel.jl
```diff
@@ -377,17 +377,17 @@ _logpdf!(r::AbstractArray, d::MultivariateMixture, x::AbstractMatrix) = _mixlogp
...
     end
     r
 end
 
 function _cwise_logpdf1!(r::AbstractVector, d::AbstractMixtureModel, x)
     K = ncomponents(d)
     length(r) == K || error("The length of r should match the number of components.")
-    for i = 1:K
-        r[i] = logpdf(component(d, i), x)
+    for (ri, i) in zip(axes(r,1), 1:K)
```
devmotion, line 389:
```suggestion
    for (i, ri) in enumerate(eachindex(r))
```

src/mixtures/mixturemodel.jl
```diff
@@ -396,11 +396,11 @@ function _cwise_pdf!(r::AbstractMatrix, d::AbstractMixtureModel, X)
     K = ncomponents(d)
     n = size(X, ndims(X))
     size(r) == (n, K) || error("The size of r is incorrect.")
-    for i = 1:K
+    for (ri, i) in zip(axes(r, 2), 1:K)
```
devmotion, line 399:
```suggestion
    for (i, ri) in enumerate(axes(r, 2))
```

src/mixtures/mixturemodel.jl
```diff
@@ -410,11 +410,11 @@ function _cwise_logpdf!(r::AbstractMatrix, d::AbstractMixtureModel, X)
...
```
devmotion, line 413:
```suggestion
    for (i, ri) in enumerate(axes(r, 2))
```

src/multivariate/dirichlet.jl
```diff
@@ -191,30 +191,30 @@ function suffstats(::Type{<:Dirichlet}, P::AbstractMatrix{Float64})
...
     n = size(P, 2)
     slogp = zeros(K)
-    for i = 1:n
-        for k = 1:K
-            @inbounds slogp[k] += log(P[k,i])
+    for i in axes(P, 2)
+        for (k, kP) in zip(1:K, axes(P,1))
+            @inbounds slogp[k] += log(P[kP,i])
         end
     end
```
devmotion, line 198: Simpler:
```suggestion
    n = size(P, 2)
    slog = vec(sum(log, P; dims=2))
```

src/multivariate/dirichlet.jl
```diff
@@ -191,30 +191,30 @@ function suffstats(::Type{<:Dirichlet}, P::AbstractMatrix{Float64})
...
     DirichletStats(slogp, n)
 end
 
 function suffstats(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
-                   w::AbstractArray{Float64})
+                   w::AbstractArray{Float64,1})
```
devmotion, line 203: Why this change? This seems breaking?

src/multivariate/dirichlet.jl
```diff
@@ -191,30 +191,30 @@ function suffstats(::Type{<:Dirichlet}, P::AbstractMatrix{Float64})
...
     K = size(P, 1)
     n = size(P, 2)
     if length(w) != n
         throw(DimensionMismatch("Inconsistent argument dimensions."))
     end
 
-    tw = 0.
+    tw = zero(Float64)
```
devmotion, line 210: This seems unnecessarily complicated and does not change the result compared with the current implementation?

src/multivariate/dirichlet.jl
```diff
@@ -191,30 +191,30 @@ function suffstats(::Type{<:Dirichlet}, P::AbstractMatrix{Float64})
...
     slogp = zeros(K)
 
-    for i = 1:n
+    for (i, iP) in zip(axes(w, 1), axes(P, 2))
         @inbounds wi = w[i]
```
devmotion, line 214: Simpler:
```suggestion
    for (iP, wi) in zip(axes(P, 2), w)
```

src/multivariate/dirichlet.jl
```diff
@@ -191,30 +191,30 @@ function suffstats(::Type{<:Dirichlet}, P::AbstractMatrix{Float64})
...
         tw += wi
-        for k = 1:K
-            @inbounds slogp[k] += log(P[k,i]) * wi
+        for (k, kP) in zip(1:K, axes(P,1))
```
devmotion, line 216:
```suggestion
        for (k, kP) in enumerate(axes(P, 1))
```

src/multivariate/dirichlet.jl
```diff
@@ -253,19 +255,21 @@ function dirichlet_mle_init(P::AbstractMatrix{Float64})
     _dirichlet_mle_init2(μ, γ)
 end
 
-function dirichlet_mle_init(P::AbstractMatrix{Float64}, w::AbstractArray{Float64})
+function dirichlet_mle_init(P::AbstractMatrix{Float64}, w::AbstractVector{Float64})
```
devmotion, line 258: Again, this seems breaking?

devmotion:
```suggestion
function dirichlet_mle_init(P::AbstractMatrix{Float64}, w::AbstractArray{Float64})
```

src/multivariate/dirichlet.jl
devmotion:
```suggestion
                   w::AbstractArray{Float64})
```

src/multivariate/dirichlet.jl
devmotion:
```suggestion
    tw = 0.
```

src/multivariate/dirichlet.jl
```diff
@@ -253,19 +255,21 @@ function dirichlet_mle_init(P::AbstractMatrix{Float64})
...
 
+    n == length(w) || throw(DimensionMismatch("Inconsistent argument dimensions."))
+
     μ = zeros(K)  # E[p]
     γ = zeros(K)  # E[p^2]
     tw = 0.0
 
-    for i in 1:n
+    for (i, iP) in zip(axes(w, 1), axes(P, 2))
         @inbounds wi = w[i]
```
devmotion, line 269:
```suggestion
    for (iP, wi) in zip(axes(P, 2), w)
```

src/multivariate/dirichlet.jl
```diff
@@ -364,7 +368,7 @@ function fit_mle(::Type{T}, P::AbstractMatrix{Float64};
 end
 
 function fit_mle(::Type{<:Dirichlet}, P::AbstractMatrix{Float64},
-                 w::AbstractArray{Float64};
+                 w::AbstractVector{Float64};
```
devmotion, line 371:
```suggestion
                 w::AbstractArray{Float64};
```

src/multivariate/dirichletmultinomial.jl
```diff
@@ -89,7 +89,7 @@ function suffstats(::Type{<:DirichletMultinomial}, x::Matrix{T}) where T<:Real
     end
     DirichletMultinomialStats(n, s, m)
 end
-function suffstats(::Type{<:DirichletMultinomial}, x::Matrix{T}, w::Array{Float64}) where T<:Real
+function suffstats(::Type{<:DirichletMultinomial}, x::Matrix{T}, w::Vector{Float64}) where T<:Real
```
devmotion, line 92:
```suggestion
function suffstats(::Type{<:DirichletMultinomial}, x::Matrix{T}, w::Array{Float64}) where T<:Real
```

src/multivariates.jl
```diff
@@ -37,7 +37,7 @@ function insupport!(r::AbstractArray, d::Union{D,Type{D}}, X::AbstractMatrix) wh
     n = length(r)
     size(X) == (length(d),n) ||
         throw(DimensionMismatch("Inconsistent array dimensions."))
-    for i in 1:n
+    for i in axes(r, 1)
         @inbounds r[i] = insupport(d, view(X, :, i))
```
devmotion, line 41: This should be
```suggestion
    for (i, Xi) in zip(eachindex(r), axes(X, 2))
        @inbounds r[i] = insupport(d, view(X, :, Xi))
```

devmotion (commented):
Thanks for the PR!

In general, I think it would be good to
- not try to fix all array indexing in one PR, and
- not to mix functional changes (new features or breaking changes) with these bug fixes.

By not trying to go through all of Distributions in one go, fixes and reviews, and hence releases, will be much quicker. Even more so since for every fix we should add a test with e.g. OffsetArrays to make sure we don't accidentally break stuff again at some point in the future.

Keeping the PR focused to these bug fixes will also make reviewing easier and avoid accidentally breaking downstream code.

src/multivariate/dirichlet.jl
palday: The weights should be 1D array, shouldn't they? i.e. a vector with each element corresponding to a row in `P`? Or is there an example where it makes sense to have an array of arbitrary dimension ?

devmotion: Well, I don't know what people use. But the current implementation definitely allows to use higher-dimensional weight arrays (the code also only checks and works with `length(w)`), so any change here is breaking. I assume it's just similar to many functions in StatsBase that allow arrays for arbitrary and non-matching size as inputs if their lengths match.

nalimilan:
@palday Any plans to finish this PR? ;-)

palday:
@nalimilan I started working on this again while other stuff was compiling and running today, but the tests -- even on `master` -- have both failures and errors on Julia 1.9.2 on Apple silicon.

src/mixtures/mixturemodel.jl
nalimilan: But `r` is not required to be a vector so we should at least do `m = similar(r, n)`. I'm not sure it really makes sense to use `similar` here though. Are we sure we don't want a `Float64` eltype?

src/mixtures/mixturemodel.jl
```diff
@@ -334,7 +335,7 @@ function _mixlogpdf!(r::AbstractArray, d::AbstractMixtureModel, x)
 
             # in the mean time, add log(prior) to lp and
             # update the maximum for each sample
-            for j = 1:n
+            for j in eachindex(r)
```
nalimilan, line 338: This seems problematic as `j` is used to index into `lp_i` which is a view of a `Matrix` column. We should use two different indices or `for (j, lp_ij) in zip(eachindex(m), lp_i)` as below. `r` isn't relevant here anyway.

src/mixtures/mixturemodel.jl
```diff
@@ -347,13 +348,13 @@ function _mixlogpdf!(r::AbstractArray, d::AbstractMixtureModel, x)
...
-            for j = 1:n
-                r[j] += exp(lp_i[j] - m[j])
+            for (j, lp_ij) in zip(eachindex(r), lp_i)
+                r[j] += exp(lp_ij - m[j])
             end
         end
     end
 
-    @inbounds for j = 1:n
+    @inbounds for j in eachindex(r)
```
nalimilan, line 357: Same here.

src/multivariate/dirichlet.jl
```diff
@@ -257,15 +252,16 @@ function dirichlet_mle_init(P::AbstractMatrix{Float64}, w::AbstractArray{Float64
...
-    for i in 1:n
-        @inbounds wi = w[i]
+    for (iP, wi) in zip(axes(P, 2), w)
         tw += wi
-        for k in 1:K
-            pk = P[k, i]
+        for k in axes(P, 1)
+            pk = P[k, iP]
             @inbounds μ[k] += pk * wi
             @inbounds γ[k] += pk * pk * wi
```
nalimilan, line 266: Should use a different index for `μ` and `γ` as these may not have the same indices as `P`.

PR #1554, opened 2022-05-23
Differentiating mvnormal

matbesancon:


src/multivariate/mvnormal.jl
```diff
@@ -515,3 +515,95 @@ function fit_mle(D::Type{IsoNormal}, x::AbstractMatrix{Float64}, w::AbstractVect
...
+        (_, ∂d_c0) = c0_pullback(dy)
+        ∂d_c0 = ChainRulesCore.unthunk(∂d_c0)
+        (_, ∂d_sq, ∂x_sq) = sq_pullback(dy)
+        ∂d_sq = ChainRulesCore.unthunk(∂d_sq)
+        ∂x_sq = ChainRulesCore.unthunk(∂x_sq)
+        backing = NamedTuple{(:μ, :Σ), Tuple{typeof(∂d_sq.μ), typeof(∂d_sq.Σ)}}((
+            (∂d_c0.μ - 0.5 * ∂d_sq.μ),
+            (∂d_c0.Σ - 0.5 * ∂d_sq.Σ),
+        ))
+        ∂d = ChainRulesCore.Tangent{typeof(d), typeof(backing)}(backing)
```
devmotion, line 545:
```suggestion
        ∂d = ChainRulesCore.Tangent{typeof(d)}(; μ=∂d_c0.μ .- ∂d_sq.μ ./ 2, Σ=∂d_c0.Σ .- ∂d_sq.Σ ./ 2)
```

src/multivariate/mvnormal.jl
```diff
@@ -515,3 +515,95 @@ function fit_mle(D::Type{IsoNormal}, x::AbstractMatrix{Float64}, w::AbstractVect
...
+        return ChainRulesCore.NoTangent(), ∂d, - 0.5 * ∂x_sq
+    end
+    return c0 - 0.5 * sq, logpdf_MvNormal_pullback
```
devmotion, line 548:
```suggestion
    return c0 - sq / 2, logpdf_MvNormal_pullback
```

src/multivariate/mvnormal.jl
```diff
@@ -515,3 +515,95 @@ function fit_mle(D::Type{IsoNormal}, x::AbstractMatrix{Float64}, w::AbstractVect
...
+end
+
+function ChainRulesCore.frule((_, Δd)::Tuple{Any,Any}, ::typeof(mvnormal_c0), d::MvNormal)
+    y = mvnormal_c0(d)
+    Δy = ChainRulesCore.@thunk(begin
+        Δd = ChainRulesCore.unthunk(Δd)
+        -dot(Δd.Σ, invcov(d)) / 2
+    end)
```
devmotion, line 556: Same here, no thunks.
```suggestion
    Δy = -dot(Δd.Σ, invcov(d)) / 2
```

src/multivariate/mvnormal.jl
```diff
@@ -515,3 +515,95 @@ function fit_mle(D::Type{IsoNormal}, x::AbstractMatrix{Float64}, w::AbstractVect
...
+end
+
+function ChainRulesCore.rrule(::typeof(mvnormal_c0), d::MvNormal)
+    y = mvnormal_c0(d)
+    function mvnormal_c0_pullback(dy)
+        ∂d = ChainRulesCore.@thunk(begin
+            dy = ChainRulesCore.unthunk(dy)
+            ∂Σ = -dy/2 * invcov(d)
+            ChainRulesCore.Tangent{typeof(d)}(μ = ChainRulesCore.ZeroTangent(), Σ = ∂Σ)
+        end)
```
devmotion, line 567: No thunk :slightly_smiling_face:

src/multivariate/mvnormal.jl
```diff
@@ -515,3 +515,95 @@ function fit_mle(D::Type{IsoNormal}, x::AbstractMatrix{Float64}, w::AbstractVect
...
+        (_, Δd, Δx) = dargs
+        Δd = ChainRulesCore.unthunk(Δd)
+        Δx = ChainRulesCore.unthunk(Δx)
+        Σinv = invcov(d)
+        # TODO optimize
+        dΣ = -dot(Σinv * Δd.Σ * Σinv, x * x' - d.μ * x' - x * d.μ' + d.μ * d.μ')
+        dx = 2 * dot(Σinv * (x - d.μ), Δx)
+        dμ = 2 * dot(Σinv * (d.μ - x), Δd.μ)
+        dΣ + dx + dμ
+    end)
```
devmotion, line 585: No thunk.

src/multivariate/mvnormal.jl
```diff
@@ -515,3 +515,95 @@ function fit_mle(D::Type{IsoNormal}, x::AbstractMatrix{Float64}, w::AbstractVect
...
+    return (y, Δy)
+end
+
+function ChainRulesCore.rrule(::typeof(sqmahal), d::MvNormal, x::AbstractVector)
+    y = sqmahal(d, x)
+    function sqmahal_pullback(dy)
+        ∂x = ChainRulesCore.@thunk(begin
+            dy = ChainRulesCore.unthunk(dy)
+            Σinv = invcov(d)
```
devmotion, line 594: This is needed below as well, so shouldn't be put in a thunk.

test/mvnormal.jl
```diff
@@ -302,3 +305,67 @@ end
...
+            # inference broken
+            # (_, ∇s_d, ∇s_x) = @inferred logpdf_MvNormal_pullback(1.0)
+            (_, ∇s_d, ∇s_x) = logpdf_MvNormal_pullback(1.0)
+
+            y2 = Distributions._logpdf(MvNormal(d.μ + t.μ, d.Σ + t.Σ), x + Δx)
+            y3 = Distributions._logpdf(MvNormal(d.μ - t.μ, d.Σ - t.Σ), x - Δx)
+            @test unthunk(Δy) ≈ y - y3 atol = n * 1e-4
+            @test unthunk(Δy) ≈ y2 - y atol = n * 1e-4
+            @test dot(∇s_d.Σ, t.Σ) + dot(∇s_d.μ, t.μ) + dot(∇s_x, Δx) ≈ y2 - y atol = n * 1e-4
+            @test dot(∇s_d.Σ, t.Σ) + dot(∇s_d.μ, t.μ) + dot(∇s_x, Δx) ≈ y - y3 atol = n * 1e-4
```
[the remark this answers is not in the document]
matbesancon: this is nontrivial in the case of the perturbation of the covariance matrix

src/multivariate/mvnormal.jl
matbesancon: type stability issues, like the other PR

devmotion: There are no type stability issues with this approach in the other PR, tests in #1555 check type stability and pass. If there are type inference issues they are caused by something else.

src/multivariate/mvnormal.jl
```diff
@@ -515,3 +515,87 @@ function fit_mle(D::Type{IsoNormal}, x::AbstractMatrix{Float64}, w::AbstractVect
...
+
+## Differentiation
+
+function ChainRulesCore.frule((_, Δd, Δx)::Tuple{Any,Any,Any}, ::typeof(_logpdf), d::AbstractMvNormal, x::AbstractVector)
+    c0, Δc0 = ChainRulesCore.frule((ChainRulesCore.NoTangent(), Δd), mvnormal_c0, d)
+    sq, Δsq = ChainRulesCore.frule((ChainRulesCore.NoTangent(), Δd, Δx), sqmahal, d, x)
+    Δc0 = ChainRulesCore.unthunk(Δc0)
+    Δsq = ChainRulesCore.unthunk(Δsq)
+    return c0 - sq/2, Δc0 - Δsq/2
+end
```
devmotion, line 527: I don't think it's useful to add this definition. This is exactly what AD systems do anyway.
```suggestion
```

src/multivariate/mvnormal.jl
```diff
@@ -515,3 +515,87 @@ function fit_mle(D::Type{IsoNormal}, x::AbstractMatrix{Float64}, w::AbstractVect
...
+        ∂x_sq = ChainRulesCore.unthunk(∂x_sq)
+        backing = NamedTuple{(:μ, :Σ), Tuple{typeof(∂d_sq.μ), typeof(∂d_sq.Σ)}}((
+            (∂d_c0.μ - 0.5 * ∂d_sq.μ),
+            (∂d_c0.Σ - 0.5 * ∂d_sq.Σ),
+        ))
+        ∂d = ChainRulesCore.Tangent{typeof(d), typeof(backing)}(backing)
+        return ChainRulesCore.NoTangent(), ∂d, ∂x_sq / (-2)
+    end
+    return c0 - sq / 2, logpdf_MvNormal_pullback
+end
```
devmotion, line 547: Same here, it seems this is exactly what AD does if no rule is defined.
```suggestion
```

src/multivariate/mvnormal.jl
```diff
@@ -372,7 +372,7 @@ struct MvNormalStats <: SufficientStats
     tw::Float64         # total sample weight
 end
 
-function suffstats(D::Type{MvNormal}, x::AbstractMatrix{Float64})
+function suffstats(::Type{MvNormal}, x::AbstractMatrix{Float64})
```
devmotion, line 375: Maybe move unrelated changes to a separate PR?

test/mvnormal.jl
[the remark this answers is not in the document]
devmotion: Yeah, sometimes one has to add custom tests. But the standard should be `test_frule` and `test_rrule` since they check multiple other parts of the CR interface as well, in addition to numerical accuracy and type inference. And even for cases that are problematic for finite differencing (e.g. due to singularities and domain constraints) it is sometimes possible to use `test_frule` and `test_rrule` by specifying custom finite differencing methods, such as e.g. in #1555.

matbesancon:
```
│   %34 = Base.getproperty(∂d_sq::Tangent{FullNormal}, :μ)::Any
│   %35 = (0.5 * %34)::Any
│   %36 = (%33 - %35)::Any
│   %37 = Base.getproperty(∂d_c0, :Σ)::Matrix{Float64}
│   %38 = Base.getproperty(∂d_sq::Tangent{FullNormal}, :Σ)::Any
```

One of the instability issues seems to be that the type of the members of Tangent{FullNormal} don't seem to be inferrable

src/multivariate/mvnormal.jl
matbesancon: It doesn't cost us much to add this definition and lets us have derivatives built-in, we can also re-add specialized methods for some MvNormal if necessary

devmotion: Actually, it can be quite problematic to define derivatives that overrule the AD system if they are not needed and e.g. to generic (as possibly the case here). I've ran into multiple issues of this kind with ChainRules, which then requires e.g. packages that otherwise would just work (without even knowing about ChainRules) to use `ChainRules.@opt_out` or define their own rules. The rule here will catch every `AbstractMvNormal` and `AbstractVector` which can be problematic e.g. similar to https://github.com/TuringLang/DistributionsAD.jl/issues/180.

So I strongly recommend not adding rules that are not needed.

src/multivariate/mvnormal.jl
matbesancon: ok yes good point, I considered initially that the non-unthunked version should be basically avoiding all expensive operations, but yeah we can expect at least one of them to be performed

src/multivariate/mvnormal.jl
matbesancon: there were all relatively minor things (unused variables)

src/multivariate/mvnormal.jl
```diff
@@ -515,3 +515,87 @@ function fit_mle(D::Type{IsoNormal}, x::AbstractMatrix{Float64}, w::AbstractVect
...
+    Δy = dΣ + dx + dμ
+    return (y, Δy)
+end
+
+function ChainRulesCore.rrule(::typeof(sqmahal), d::MvNormal, x::AbstractVector)
+    y = sqmahal(d, x)
+    function sqmahal_pullback(dy)
+        Σinv = invcov(d)
+        ∂x = ChainRulesCore.@thunk(begin
+            dy = ChainRulesCore.unthunk(dy)
```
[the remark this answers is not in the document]
matbesancon: done

matbesancon:
@devmotion I think everything discussed has been adapted/validated

src/multivariate/mvnormal.jl
```diff
@@ -515,3 +515,57 @@ function fit_mle(D::Type{IsoNormal}, x::AbstractMatrix{Float64}, w::AbstractVect
...
+    Δd = ChainRulesCore.unthunk(Δd)
+    Δy = -dot(Δd.Σ, invcov(d)) / 2
+    return y, Δy
+end
+
+function ChainRulesCore.rrule(::typeof(mvnormal_c0), d::MvNormal)
+    y = mvnormal_c0(d)
+    function mvnormal_c0_pullback(dy)
+        dy = ChainRulesCore.unthunk(dy)
+        ∂Σ = -dy/2 * invcov(d)
```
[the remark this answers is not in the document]
matbesancon: what does it change here?

matbesancon:
not computing the inverse at all would be a bit of a hassle here since it's reused several times in the whole function. I removed the materialization of the inverse as a matrix here to lighten the computations, we are using the inverse of the Cholesky decomposition directly

matbesancon:
That change here: https://github.com/JuliaStats/Distributions.jl/pull/1554/commits/b225298aad7d57f96d560da7606987b3cd41e434
was somehow incorrect

matbesancon:
Reverted the changes, the inverse covariance matrix is needed here, there is no way around it since the derivative is just the invcov scaled

matbesancon:
> Are the rules for sqmahal etc. needed? Would it maybe be sufficient to just make sure that PDMats works and is optimized for ChainRules?

I would say not since the operations done on top are non-trivial and could be costly

matbesancon:
ping @devmotion for another round of review

PR #1557, opened 2022-05-25
Improve Bernoulli and Binomial

FedericoStra:
Some improvements to `Bernoulli`. Other distributions might benefit from similar changes, but I didn't go through all of them before knowing whether such modifications would be appreciated.

### Changes

- replace `1:length(x)` with `eachindex(x)` so that it works with more arrays, for instance `OffsetArrays`
- initialize the counters `c0` and `c1` with `zero(eltype(w))` so that they are type-stable (previously they were `Int32/Float64`)
- fix the creation of domain error: the void method `DomainError()` does not exist
- make `BernoulliStats` generic so that it can collect counters of more general type, for instance `ForwardDiff.Dual`

### Examples

This new version allows for instance

```julia
using OffsetArrays
x = [0, 1]; w = [1., 2.];
ox = OffsetArray(x, 5); ow = OffsetArray(w, 5);
suffstats(Bernoulli, ox, ow) #=> Distributions.BernoulliStats{Float64}(1.0, 2.0)
```

and

```julia
julia> ForwardDiff.gradient([1.,2.]) do w
           mean(fit_mle(Bernoulli, suffstats(Bernoulli, [0,1], w)))
       end
2-element Vector{Float64}:
 -0.2222222222222222
  0.1111111111111111
```

### DIFF
--- a/src/univariate/discrete/bernoulli.jl
+++ b/src/univariate/discrete/bernoulli.jl
@@ -113,44 +113,39 @@ rand(rng::AbstractRNG, d::Bernoulli) = rand(rng) <= succprob(d)
 
 #### MLE fitting
 
-struct BernoulliStats <: SufficientStats
-    cnt0::Float64
-    cnt1::Float64
-
-    BernoulliStats(c0::Real, c1::Real) = new(Float64(c0), Float64(c1))
+struct BernoulliStats{C<:Real} <: SufficientStats
+    cnt0::C
+    cnt1::C
 end
+
+BernoulliStats(c0::Real, c1::Real) = BernoulliStats(promote(c0, c1)...)
 
 fit_mle(::Type{<:Bernoulli}, ss::BernoulliStats) = Bernoulli(ss.cnt1 / (ss.cnt0 + ss.cnt1))
 
 function suffstats(::Type{<:Bernoulli}, x::AbstractArray{T}) where T<:Integer
-    n = length(x)
     c0 = c1 = 0
-    for i = 1:n
-        @inbounds xi = x[i]
+    for xi in x
         if xi == 0
             c0 += 1
         elseif xi == 1
             c1 += 1
         else
-            throw(DomainError())
+            throw(DomainError(xi, "samples must be 0 or 1"))
         end
     end
     BernoulliStats(c0, c1)
 end
 
-function suffstats(::Type{<:Bernoulli}, x::AbstractArray{T}, w::AbstractArray{Float64}) where T<:Integer
-    n = length(x)
-    length(w) == n || throw(DimensionMismatch("Inconsistent argument dimensions."))
-    c0 = c1 = 0
-    for i = 1:n
-        @inbounds xi = x[i]
-        @inbounds wi = w[i]
+function suffstats(::Type{<:Bernoulli}, x::AbstractArray{T}, w::AbstractArray{<:Real}) where T<:Integer
+    length(x) == length(w) || throw(DimensionMismatch("inconsistent argument dimensions"))
+    c0 = c1 = zero(eltype(w))
+    for (xi, wi) in zip(x, w)
         if xi == 0
             c0 += wi
         elseif xi == 1
             c1 += wi
         else
-            throw(DomainError())
+            throw(DomainError(xi, "samples must be 0 or 1"))
         end
     end
     BernoulliStats(c0, c1)
--- a/src/univariate/discrete/bernoulli.jl
+++ b/src/univariate/discrete/bernoulli.jl
@@ -122,7 +122,7 @@ BernoulliStats(c0::Real, c1::Real) = BernoulliStats(promote(c0, c1)...)
 
 fit_mle(::Type{<:Bernoulli}, ss::BernoulliStats) = Bernoulli(ss.cnt1 / (ss.cnt0 + ss.cnt1))
 
-function suffstats(::Type{<:Bernoulli}, x::AbstractArray{T}) where T<:Integer
+function suffstats(::Type{<:Bernoulli}, x::AbstractArray{<:Integer})
     c0 = c1 = 0
     for xi in x
         if xi == 0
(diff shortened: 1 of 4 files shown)


src/univariate/discrete/bernoulli.jl
```diff
@@ -113,44 +113,41 @@ rand(rng::AbstractRNG, d::Bernoulli) = rand(rng) <= succprob(d)
...
+BernoulliStats(c0::Real, c1::Real) = BernoulliStats(promote(c0, c1)...)
+
 fit_mle(::Type{<:Bernoulli}, ss::BernoulliStats) = Bernoulli(ss.cnt1 / (ss.cnt0 + ss.cnt1))
 
 function suffstats(::Type{<:Bernoulli}, x::AbstractArray{T}) where T<:Integer
-    n = length(x)
     c0 = c1 = 0
-    for i = 1:n
+    for i = eachindex(x)
         @inbounds xi = x[i]
```
devmotion, line 128: A bit simpler:
```suggestion
    for xi in x
```

src/univariate/discrete/bernoulli.jl
```diff
@@ -113,44 +113,41 @@ rand(rng::AbstractRNG, d::Bernoulli) = rand(rng) <= succprob(d)
...
         if xi == 0
             c0 += 1
         elseif xi == 1
             c1 += 1
         else
-            throw(DomainError())
+            throw(DomainError(xi, "the sampled value must be 0 or 1"))
```
devmotion, line 134:
```suggestion
            throw(DomainError(xi, "samples must be 0 or 1"))
```

src/univariate/discrete/bernoulli.jl
```diff
@@ -113,44 +113,41 @@ rand(rng::AbstractRNG, d::Bernoulli) = rand(rng) <= succprob(d)
...
+    for i = eachindex(x, w)
         @inbounds xi = x[i]
         @inbounds wi = w[i]
         if xi == 0
             c0 += wi
         elseif xi == 1
             c1 += wi
         else
-            throw(DomainError())
+            throw(DomainError(xi, "the sampled value must be 0 or 1"))
```
devmotion, line 150:
```suggestion
            throw(DomainError(xi, "samples must be 0 or 1"))
```

src/univariate/discrete/bernoulli.jl
```diff
@@ -113,44 +113,41 @@ rand(rng::AbstractRNG, d::Bernoulli) = rand(rng) <= succprob(d)
...
-function suffstats(::Type{<:Bernoulli}, x::AbstractArray{T}, w::AbstractArray{Float64}) where T<:Integer
-    n = length(x)
-    length(w) == n || throw(DimensionMismatch("Inconsistent argument dimensions."))
-    c0 = c1 = 0
-    for i = 1:n
+function suffstats(::Type{<:Bernoulli}, x::AbstractArray{T}, w::AbstractArray{<:Real}) where T<:Integer
+    c0 = c1 = zero(eltype(w))
+    for i = eachindex(x, w)
         @inbounds xi = x[i]
         @inbounds wi = w[i]
```
devmotion, line 144: This would be breaking since currently we don't enforce that `x` and `w` have the same indices (nor that they have the same shape). Instead we should use something like
```suggestion
    length(w) == n || throw(DimensionMismatch("inconsistent argument dimensions"))
    c0 = c1 = zero(eltype(w))
    for xi, wi in zip(x, w)
```

devmotion (commented):
Thanks for the PR!

Can you add some tests for all the fixes?

src/univariate/discrete/bernoulli.jl
FedericoStra: The verbose version is _slightly_ faster...
Similarly,
```julia
for i = eachindex(x, w)
        @inbounds xi = x[i]
        @inbounds wi = w[i]
```
is faster than
```julia
for (xi, wi) in zip(x, w)
```
I can change it though, in which case one could also remove the `::AbstractArray` bound to allow for other iterators.

devmotion: We don't want to remove the `::AbstractArray` types in the second case (at least not currently), it's only about not breaking existing code. `eachindex(x, w)` is too restrictive.

src/univariate/discrete/bernoulli.jl
devmotion: > Would eachindex(vec(x), vec(w)) make it better then?

IMO that's less clear than just iterating over the two arrays.

>  The fact that x and w can have different shapes seems to me a possible source of bugs though.

We might want to restrict it at some point but it's a breaking change and hence should be postponed at least until we prepare the next breaking release.

FedericoStra: Could you give me an example of where `eachindex` fails whereas the old implementation worked?
For instance, `eachindex([1,2,3,4], [1 2; 3 4])` works fine and returns `Base.OneTo(4)`. I can't see an example that breaks.

src/univariate/discrete/bernoulli.jl
```diff
@@ -113,44 +113,39 @@ rand(rng::AbstractRNG, d::Bernoulli) = rand(rng) <= succprob(d)
...
-    n = length(x)
-    length(w) == n || throw(DimensionMismatch("Inconsistent argument dimensions."))
-    c0 = c1 = 0
-    for i = 1:n
-        @inbounds xi = x[i]
-        @inbounds wi = w[i]
+function suffstats(::Type{<:Bernoulli}, x::AbstractArray{T}, w::AbstractArray{<:Real}) where T<:Integer
+    length(x) == length(w) || throw(DimensionMismatch("inconsistent argument dimensions"))
+    c0 = c1 = zero(eltype(w))
+    for xi, wi in zip(x, w)
```
devmotion: Sorry, my fault, this should be
```suggestion
    for (xi, wi) in zip(x, w)
```

FedericoStra:
I ported the same changes to `Binomial`. Now one can do things like
```julia
julia> ForwardDiff.gradient(w -> mean(fit_mle(Binomial, 3, [0, 1, 2, 3], w)), [0.5, 1.5, 3.9, 2.4])
4-element Vector{Float64}:
 -0.2395122659311946
 -0.11903033822035122
  0.00145158949049215
  0.12193351720133552
```

devmotion:
Thanks for the PR @FedericoStra!

FedericoStra:
> Thanks for the PR @FedericoStra!

Thank you! I wish you a good day!

Issue #1544, opened 2022-05-07
Fitting does not respect type parameters


gdalle:
Hi!
In my current project, I would like to fit distributions in reduced precision to speed up my code. I was therefore wondering if the following behavior is intentional:
```julia
julia> using Distributions

julia> fit(Normal{Float32}, [1, 2, 3])
Normal{Float64}(μ=2.0, σ=0.816496580927726)
```

devmotion:
It's known, the type only depends on the data. If you use `Float32[1, 2, 3]` the type should be `Normal{Float32}`.

devmotion:
Of course, the question is still if this should be changed. I think it would be reasonable.

gdalle:
Thanks for your answer. Indeed, when presented like this the current behavior seems logical too, since sufficient statistics are computed from the sample. But I agree that both options can make sense, and it is hard to decide which one should prevail

gdalle:
Hum, after testing it doesn't go the way you promised it would :sweat_smile: 
```julia
julia> using Distributions

julia> fit(Normal{Float32}, Float32[1, 2, 3])
Normal{Float64}(μ=2.0, σ=0.816496580927726)
```

devmotion:
Well, it seems there are some hardcoded Float64 arguments, e.g. in https://github.com/JuliaStats/Distributions.jl/blob/master/src/univariate/continuous/normal.jl#L238. It would be easy to change these type restrictions at least in a first step.

gdalle:
I think one would also need to make sufficient stats parametric, for instance here:
https://github.com/JuliaStats/Distributions.jl/blob/dd6ae8f4eac304f404b0069540a6c3bb1c667f92/src/univariate/continuous/normal.jl#L108-L113

matbesancon:
yeah there are still a lot of F64 hardcoded sprinkled in the package :/

timholy:
In case it helps resolve difficult design decisions,

```julia
help?> fit_mle
search: fit_mle

  fit_mle(D, x)

  Fit a distribution of type D to a given data set x.
```

From this, there's no doubt that the type parameters of `D` should take precedence over the eltype of the data. Either the docs need changing or the behavior should match the docs.

gdalle:
I had given it a first shot months ago in this PR:

https://github.com/JuliaStats/Distributions.jl/pull/1560

Looking back at it, my main issue is that we need to accommodate both

[...]

PR #1558, opened 2022-05-29
Respect type of parameters in `fit` for `Bernoulli`, `Binomial`, and `Uniform`

devmotion:
As  a follow-up to #1551 and #1557, this PR addresses #1544 for `Bernoulli`, `Binomial`, and `Uniform`.

With this PR `fit(Uniform{Float32}, ...)` etc. will actually estimate a distribution of type `Uniform{Float32}`, regardless of the observations. One can still use `fit(Uniform, ...)` if the parameter type should not be fixed but depend on the type of the observations.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.25.61"
+version = "0.25.62"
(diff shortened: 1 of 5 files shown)


matbesancon:
I agree with this in principal, it was either this solution or a promotion mechanism, but using the specified type param makes more sense when provided

gdalle:
Thank you for this @devmotion !

PR #1560, opened 2022-05-30
Make suffstats and fit_mle type generic for Normal{T}

gdalle:
In the spirit of #1558, here is a prototype PR that:

- removes type constraints for `x` and `w` in `fit_mle` (they can both be `AbstractVector{<:Real}`)
- adds a type parameter to `NormalStats` (which becomes `NormalStats{T}`)
- makes sure that `fit_mle(Normal{Float32}, x, w)` will return a `Normal{Float32}`

I haven't added type tests yet, waiting on confirmation from the maintainers that this is a welcome addition. If so, I can implement similar changes for a few more standard distributions.

Note: this PR originated from a discussion with @matbesancon. Looking at `methods(suffstats)` makes it clear that some homogeneization wouldn't hurt there.

### DIFF
--- a/src/univariate/continuous/normal.jl
+++ b/src/univariate/continuous/normal.jl
@@ -105,14 +105,14 @@ rand(rng::AbstractRNG, d::Normal{T}) where {T} = d.μ + d.σ * randn(rng, float(
 
 #### Fitting
 
-struct NormalStats <: SufficientStats
-    s::Float64    # (weighted) sum of x
-    m::Float64    # (weighted) mean of x
-    s2::Float64   # (weighted) sum of (x - μ)^2
-    tw::Float64    # total sample weight
+struct NormalStats{T} <: SufficientStats
+    s::T    # (weighted) sum of x
+    m::T    # (weighted) mean of x
+    s2::T   # (weighted) sum of (x - μ)^2
+    tw::T    # total sample weight
 end
 
-function suffstats(::Type{<:Normal}, x::AbstractArray{T}) where T<:Real
+function suffstats(::Type{Normal{T}}, x::AbstractArray{<:Real}) where {T<:Real}
     n = length(x)
 
     # compute s
@@ -128,10 +128,10 @@ function suffstats(::Type{<:Normal}, x::AbstractArray{T}) where T<:Real
         @inbounds s2 += abs2(x[i] - m)
     end
 
-    NormalStats(s, m, s2, n)
+    NormalStats{T}(s, m, s2, n)
 end
 
-function suffstats(::Type{<:Normal}, x::AbstractArray{T}, w::AbstractArray{Float64}) where T<:Real
+function suffstats(::Type{Normal{T}}, x::AbstractArray{<:Real}, w::AbstractArray{<:Real}) where {T<:Real}
     n = length(x)
 
     # compute s
@@ -150,7 +150,7 @@ function suffstats(::Type{<:Normal}, x::AbstractArray{T}, w::AbstractArray{Float
         @inbounds s2 += w[i] * abs2(x[i] - m)
     end
 
-    NormalStats(s, m, s2, tw)
+    NormalStats{T}(s, m, s2, tw)
 end
 
 # Cases where μ or σ is known
@@ -211,16 +211,16 @@ end
 
 # fit_mle based on sufficient statistics
 
-fit_mle(::Type{<:Normal}, ss::NormalStats) = Normal(ss.m, sqrt(ss.s2 / ss.tw))
+fit_mle(::Type{D}, ss::NormalStats) where {D<:Normal} = D(ss.m, sqrt(ss.s2 / ss.tw))
 fit_mle(g::NormalKnownMu, ss::NormalKnownMuStats) = Normal(g.μ, sqrt(ss.s2 / ss.tw))
 fit_mle(g::NormalKnownSigma, ss::NormalKnownSigmaStats) = Normal(ss.sx / ss.tw, g.σ)
 
 # generic fit_mle methods
 
-function fit_mle(::Type{<:Normal}, x::AbstractArray{T}; mu::Float64=NaN, sigma::Float64=NaN) where T<:Real
+function fit_mle(::Type{D}, x::AbstractArray{<:Real}; mu::Float64=NaN, sigma::Float64=NaN) where {D<:Normal}
     if isnan(mu)
         if isnan(sigma)
-            fit_mle(Normal, suffstats(Normal, x))
+            fit_mle(D, suffstats(Normal, x))
         else
             g = NormalKnownSigma(sigma)
             fit_mle(g, suffstats(g, x))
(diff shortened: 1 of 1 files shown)


matbesancon:
ping @devmotion since it started with your PR

src/univariate/continuous/normal.jl
```diff
@@ -105,14 +105,14 @@ rand(rng::AbstractRNG, d::Normal{T}) where {T} = d.μ + d.σ * randn(rng, float(
...
-    s::Float64    # (weighted) sum of x
-    m::Float64    # (weighted) mean of x
-    s2::Float64   # (weighted) sum of (x - μ)^2
-    tw::Float64    # total sample weight
+struct NormalStats{T} <: SufficientStats
+    s::T    # (weighted) sum of x
+    m::T    # (weighted) mean of x
+    s2::T   # (weighted) sum of (x - μ)^2
+    tw::T    # total sample weight
 end
```
devmotion, line 113: I think this requires a constructor that promotes everything to a common type. See also my comments below and the previous PRs regarding `Bernoulli` and `Binomial`.

src/univariate/continuous/normal.jl
```diff
@@ -105,14 +105,14 @@ rand(rng::AbstractRNG, d::Normal{T}) where {T} = d.μ + d.σ * randn(rng, float(
...
 
-function suffstats(::Type{<:Normal}, x::AbstractArray{T}) where T<:Real
+function suffstats(::Type{Normal{T}}, x::AbstractArray{<:Real}) where {T<:Real}
```
devmotion, line 115: I don't think the type parameter of `Normal` should matter here. The sufficient statistics are very much data-dependent, and it might even cause surprising errors if the type parameter is not "wide" enough (since in general the sufficient statistics don't have to be of the same type as the elements of `x`). `suffstats` is just an intermediate step in the `fit` pipeline, and the type of the resulting distribution is only determined when calling `fit`, either with data or the sufficient statistics.

devmotion:
```suggestion
function suffstats(::Type{<:Normal}, x::AbstractArray{<:Real})
```

src/univariate/continuous/normal.jl
```diff
@@ -128,10 +128,10 @@ function suffstats(::Type{<:Normal}, x::AbstractArray{T}) where T<:Real
         @inbounds s2 += abs2(x[i] - m)
     end
 
-    NormalStats(s, m, s2, n)
+    NormalStats{T}(s, m, s2, n)
```
devmotion, line 131:
```suggestion
    NormalStats(s, m, s2, n)
```

src/univariate/continuous/normal.jl
```diff
@@ -128,10 +128,10 @@ function suffstats(::Type{<:Normal}, x::AbstractArray{T}) where T<:Real
...
 end
 
-function suffstats(::Type{<:Normal}, x::AbstractArray{T}, w::AbstractArray{Float64}) where T<:Real
+function suffstats(::Type{Normal{T}}, x::AbstractArray{<:Real}, w::AbstractArray{<:Real}) where {T<:Real}
```
devmotion, line 134:
```suggestion
function suffstats(::Type{<:Normal}, x::AbstractArray{<:Real}, w::AbstractArray{<:Real})
```

src/univariate/continuous/normal.jl
```diff
@@ -150,7 +150,7 @@ function suffstats(::Type{<:Normal}, x::AbstractArray{T}, w::AbstractArray{Float
         @inbounds s2 += w[i] * abs2(x[i] - m)
     end
 
-    NormalStats(s, m, s2, tw)
+    NormalStats{T}(s, m, s2, tw)
```
devmotion, line 153:
```suggestion
    NormalStats(s, m, s2, tw)
```

src/univariate/continuous/normal.jl
```diff
@@ -211,16 +211,16 @@ end
 
 # fit_mle based on sufficient statistics
 
-fit_mle(::Type{<:Normal}, ss::NormalStats) = Normal(ss.m, sqrt(ss.s2 / ss.tw))
+fit_mle(::Type{D}, ss::NormalStats) where {D<:Normal} = D(ss.m, sqrt(ss.s2 / ss.tw))
 fit_mle(g::NormalKnownMu, ss::NormalKnownMuStats) = Normal(g.μ, sqrt(ss.s2 / ss.tw))
 fit_mle(g::NormalKnownSigma, ss::NormalKnownSigmaStats) = Normal(ss.sx / ss.tw, g.σ)
```
devmotion, line 216: These should be changed as well it seems.

src/univariate/continuous/normal.jl
```diff
@@ -211,16 +211,16 @@ end
...
 
 # generic fit_mle methods
 
-function fit_mle(::Type{<:Normal}, x::AbstractArray{T}; mu::Float64=NaN, sigma::Float64=NaN) where T<:Real
+function fit_mle(::Type{D}, x::AbstractArray{<:Real}; mu::Float64=NaN, sigma::Float64=NaN) where {D<:Normal}
```
devmotion, line 220: The keyword arguments should be relaxed as well. I suggest changing their default values to `nothing`, as this allows a type stable implementation (similar to e.g. `mean` in `var`: https://docs.julialang.org/en/v1/stdlib/Statistics/#Statistics.var) and avoids promotion of e.g. `Float32` data:
```suggestion
function fit_mle(::Type{D}, x::AbstractArray{<:Real}; mu::Union{Nothing,Real}=nothing, sigma::Union{Nothing,Real}=nothing) where {D<:Normal}
```

src/univariate/continuous/normal.jl
```diff
@@ -211,16 +211,16 @@ end
...
     if isnan(mu)
         if isnan(sigma)
```
devmotion, line 222:
```suggestion
    if mu === nothing
        if sigma === nothing
```

src/univariate/continuous/normal.jl
```diff
@@ -211,16 +211,16 @@ end
...
-            fit_mle(Normal, suffstats(Normal, x))
+            fit_mle(D, suffstats(Normal, x))
         else
             g = NormalKnownSigma(sigma)
```
devmotion, line 225: We should find a way to preserve/forward the type of `D` in these cases as well.

src/univariate/continuous/normal.jl
```diff
@@ -230,15 +230,15 @@ function fit_mle(::Type{<:Normal}, x::AbstractArray{T}; mu::Float64=NaN, sigma::
...
             fit_mle(g, suffstats(g, x))
         else
-            Normal(mu, sigma)
+            D(mu, sigma)
         end
     end
 end
 
-function fit_mle(::Type{<:Normal}, x::AbstractArray{T}, w::AbstractArray{Float64}; mu::Float64=NaN, sigma::Float64=NaN) where T<:Real
+function fit_mle(::Type{D}, x::AbstractArray{<:Real}, w::AbstractArray{<:Real}; mu::Float64=NaN, sigma::Float64=NaN) where {D<:Normal}
```
devmotion, line 238:
```suggestion
function fit_mle(::Type{D}, x::AbstractArray{<:Real}, w::AbstractArray{<:Real}; mu::Union{Nothing,Real}=nothing, sigma::Union{Real,Nothing}=nothing) where {D<:Normal}
```

src/univariate/continuous/normal.jl
```diff
@@ -230,15 +230,15 @@ function fit_mle(::Type{<:Normal}, x::AbstractArray{T}; mu::Float64=NaN, sigma::
...
     if isnan(mu)
         if isnan(sigma)
```
devmotion, line 240:
```suggestion
    if mu === nothing
        if sigma === nothing
```

src/univariate/continuous/normal.jl
gdalle: Which eltype should we pick for `D` though, the one of `g` or the one of the suffstats?

gdalle: I added a conversion in my latest commit, what do you think?

gdalle:
Now the behavior is that:
 
- `suffstats` does the necessary promotions
- `fit_mle` brings it back to the type of `Normal{T}` if `T` is specified

PR #1563, opened 2022-06-06
Ported R code for pdf, cdf, rand, and quantile for NoncentralChisq

atkirtland:
I more or less directly ported the R code for computing the pdf, cdf, rand, and quantile functions (via `dnchisq`, `pnchisq`, `rnchisq`, and `qnchisq`, respectively) for the NoncentralChisq distribution. All tests that were passing before seem to pass with the exception of a select number of tests involving the quantile function for the `univariates.jl` Rician distribution test. The `univariates.jl` NoncentralChisq distribution tests pass. The Rician distribution is defined in terms of the NoncentralChisq distribution, and the computed values for the tests that fail differ from the expected values around the seventh decimal place.

### DIFF
--- a/src/univariate/continuous/noncentralchisq.jl
+++ b/src/univariate/continuous/noncentralchisq.jl
@@ -26,10 +26,10 @@ External links
 struct NoncentralChisq{T<:Real} <: ContinuousUnivariateDistribution
     ν::T
     λ::T
-    NoncentralChisq{T}(ν::T, λ::T) where {T <: Real} = new{T}(ν, λ)
+    NoncentralChisq{T}(ν::T, λ::T) where {T<:Real} = new{T}(ν, λ)
 end
 
-function NoncentralChisq(ν::T, λ::T; check_args::Bool=true) where {T <: Real}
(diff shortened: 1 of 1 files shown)


devmotion (commented):
Thanks for the PR!

Some high-level comments:
- These pdf, cdf, etc. implementations for the noncentral chi square distribution should be added to StatsFuns in https://github.com/JuliaStats/StatsFuns.jl/blob/master/src/distrs/nchisq.jl, so that other packages can benefit as well
- Is the R code that you ported GPL-licensed? This would mean that the Julia code has to use the same license (AFAIK) which we want to avoid. One of the reasons from moving from the R to native Julia implementations is to get rid of the GPL parts (others are support for generic number types such as dual numbers and performance as calling the Rmath functions usually has some overhead, it seems).
- It seems there are a few unrelated changes in this PR (eg formatting). Generally, it is easier to review your PR if it only contains a minimal set of relevant changes.

devmotion:
I close the PR since, as mentioned above, the underlying functions are implemented in StatsFuns and there seem to be license issues.

Issue #1564, opened 2022-06-14
Docs not clear on usage of Poisson Object


Aman-Pandey-afk:
https://github.com/JuliaStats/Distributions.jl/blob/f889f9e56b0243d770c195b3eee8baef4880bd2e/src/univariate/discrete/poisson.jl#L1-L22
Hi, I think the docs should be more clear on the Poisson Object that is generated when Poisson ($\lambda$) is executed. In Python we use numpy.random.poisson and give it an argument to generate an iterable over the Poisson Distribution, but here the function just takes one argument and the returned Poisson object has only one field ( $\lambda$ itself). I later found out that the object can be used as an argument for rand to generate the iterable. A newcomer might get stuck as there is no help regarding the nature of the Poisson Object in Docs. 
I think a line like this in the docs might work:
`The Poisson Object is of type UnivariateDistribution which can be passed to the rand() function as an argument along with an integer N, to generate an iterable of size N.`
I will make a pull request if the issue is considerable.

devmotion:
This design is not specific to `Poisson` but applies to any distribution in Distributions.jl. So I guess it might be better to improve the general documentation if this is unclear than adding a line to the docs of `Poisson`. There is an example with normal distributions in the "Getting started" section (https://juliastats.org/Distributions.jl/stable/starting/#Starting-With-a-Normal-Distribution) and the use with `rand` (and other functions) is listed in the section for univariate distributions (https://juliastats.org/Distributions.jl/stable/univariate/). Maybe we should make it even clearer?

Aman-Pandey-afk:
Yeah, adding a line to Poisson will not be right, as it applies for all kinds of distributions. Although, in getting started guide it is only explicitly shown for Normal, and in using other distributions, using rand is not mentioned. But I guess it is not that necessary too, one can follow the working of normal if he/she has generating samples in mind. I can close the issue if it is so. I guess the example is clear, just a reference to it being used in all kind of distributions was what I thought, so specifically I asked for Poisson.

PR #1565, opened 2022-06-17
Add sentence about distribution objects

mschauer:
Follow up on #1564

### DIFF
--- a/docs/src/starting.md
+++ b/docs/src/starting.md
@@ -24,8 +24,8 @@ julia> d = Normal()
 Normal(μ=0.0, σ=1.0)
 ```
 
-The object `d::Distribution` represents a probability distribution, in our
-case the standard-normal distribution. One can query its properties such as the mean:
+The object `d` represents a probability distribution, in our case the standard-normal distribution.
+One can query its properties such as the mean:
 
 ```julia
 julia> mean(d)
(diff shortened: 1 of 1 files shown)


docs/src/starting.md
```diff
@@ -22,7 +22,18 @@ Then, we create a standard-normal distribution `d` and obtain samples using `ran
...
+The object `d::Distribution` represents a probability distribution, in our
+case the standard-normal distribution. One can query its properties such as the mean:
+
+```julia
+julia> mean(d)
+0.0
+```
 
+We can also draw samples from `d` with `rand`.
+```
```
devmotion, line 36:
````suggestion
```julia
````

File: docs/src/starting.md [hunk 2]
```diff
@@ -22,7 +22,18 @@ Then, we create a standard-normal distribution `d` and obtain samples using `ran
 ```julia
 julia> d = Normal()
 Normal(μ=0.0, σ=1.0)
+```
+
+The object `d::Distribution` represents a probability distribution, in our
+case the standard-normal distribution. One can query its properties such as the mean:
```
devmotion, line 28: Maybe the type is not relevant at this point here?
```suggestion
The object `d` represents a probability distribution, in our case the standard-normal distribution.
One can query its properties such as the mean:
```

devmotion (approved):

PR #1572, opened 2022-06-26
Fix deprecations of `dim`

devmotion:
This PR fixes the deprecations introduced in the recent release of PDMats (https://github.com/JuliaStats/PDMats.jl/pull/170). Additionally, it deprecates `dim(::Wishart)` etc. to simplify the transition to the next breaking release of PDMats which might not export `dim` anymore.

### DIFF
--- a/src/matrix/matrixbeta.jl
+++ b/src/matrix/matrixbeta.jl
@@ -72,15 +72,13 @@ end
 #  Properties
 #  -----------------------------------------------------------------------------
 
-dim(d::MatrixBeta) = dim(d.W1)
-
 size(d::MatrixBeta) = size(d.W1)
 
-rank(d::MatrixBeta) = dim(d)
+rank(d::MatrixBeta) = rank(d.W1)
 
 insupport(d::MatrixBeta, U::AbstractMatrix) = isreal(U) && size(U) == size(d) && isposdef(U) && isposdef(I - U)
 
-params(d::MatrixBeta) = (dim(d), d.W1.df, d.W2.df)
+params(d::MatrixBeta) = (size(d, 1), d.W1.df, d.W2.df)
 
 mean(d::MatrixBeta) = ((p, n1, n2) = params(d); Matrix((n1 / (n1 + n2)) * I, p, p))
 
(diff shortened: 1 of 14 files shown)


src/matrix/matrixbeta.jl
```diff
@@ -72,15 +72,13 @@ end
 #  Properties
 #  -----------------------------------------------------------------------------
 
-dim(d::MatrixBeta) = dim(d.W1)
-
 size(d::MatrixBeta) = size(d.W1)
 
-rank(d::MatrixBeta) = dim(d)
+rank(d::MatrixBeta) = rank(d.W1)
```
matbesancon, line 77: Rank or side-dimension?

src/matrix/matrixfdist.jl
```diff
@@ -78,18 +78,16 @@ end
 #  Properties
 #  -----------------------------------------------------------------------------
 
-dim(d::MatrixFDist) = dim(d.W)
-
 size(d::MatrixFDist) = size(d.W)
 
-rank(d::MatrixFDist) = dim(d)
+rank(d::MatrixFDist) = rank(d.W)
```
matbesancon, line 83: same?

src/matrix/matrixbeta.jl
devmotion: I guess I assumed it would be better to forward the call to `W1` in case its samples are not of full rank. By construction that can't happen though, it seems. So I guess it's a matter of personal preference whether to use `rank(d.W1)` or e.g. `size(d, 1)`. What do you prefer?

devmotion: @matbesancon Do you have any preference?

```suggestion
rank(d::MatrixBeta) = size(d, 1)
```

The deprecations cause time outs in the Turing tests downstream, so it would be good to fix them :slightly_smiling_face:

src/matrix/matrixfdist.jl
devmotion:
```suggestion
rank(d::MatrixFDist) = size(d, 1)
```

matbesancon (approved):

PR #1578, opened 2022-07-01
Include and update tests for `BetaBinomial`

### LINKED ISSUE #1576: The test file test/betabinomial.jl is not included in runtests.jl
The list of tests does not include this file. Is this intentional?

devmotion:
Fixes #1576.


devmotion:
I'll go on and merge this PR since it does not the package itself but only adds more (and already existing) tests.

PR #1580, opened 2022-07-03
speed up Semicirle sampling

jw3126:
```julia
# before
julia> @btime rand($(Semicircle(1.0)))
  149.858 ns (0 allocations: 0 bytes)
# after
julia> @btime rand($(Semicircle(1.0)))
  9.858 ns (0 allocations: 0 bytes)
```
Not sure where and what unit tests to add?

### DIFF
--- a/src/univariate/continuous/semicircle.jl
+++ b/src/univariate/continuous/semicircle.jl
@@ -72,4 +72,15 @@ function cdf(d::Semicircle, x::Real)
     end
 end
 
+function rand(rng::AbstractRNG, d::Semicircle)
+    # Idea:
+    # sample polar coodinates r,θ 
+    # of point uniformly distributed on radius d.r half disk
+    # project onto x axis
+    θ = rand(rng)
+    r = d.r*sqrt(rand(rng))
+    x = cos(θ) * r
+    return x
+end
+
 @quantile_newton Semicircle


devmotion (commented):
I'm not familiar with `SemiCircle` but from quick look I think the code might not be completely correct? It seems the support of a `d::SemiCircle` is `[-d.r, d.r]` but AFAICT the proposed `rand` only returns values in the interval `[0, d.r]` as the range of `cos(θ)` is `(cos(1), cos(0)] ≈ (0.54, 1]`.

I'm not sure if there's a standard test suite for samplers or `rand`, I have to check, but I think reasonable tests would be e.g. evaluating statistics such as mean, variance etc. of a larger set of samples and compare them with the true values. Additionally, of course, all samples should be in the support of the distribution and for a large number of samples also minimum and maximum should be close to the true ones.

jw3126:
> I'm not familiar with `SemiCircle` but from quick look I think the code might not be completely correct? 

Good catch I forgot a `pi`.

> I'm not sure if there's a standard test suite for samplers or `rand`, I have to check, but I think reasonable tests would be e.g. evaluating statistics such as mean, variance etc. of a larger set of samples and compare them with the true values. Additionally, of course, all samples should be in the support of the distribution and for a large number of samples also minimum and maximum should be close to the true ones.

These are good suggestions, thanks!

devmotion:
I couldn't find any other more general tests (apart from the ones in test/univariates.jl based on R), so probably best to add new tests to https://github.com/JuliaStats/Distributions.jl/blob/master/test/semicircle.jl.

src/univariate/continuous/semicircle.jl
```diff
@@ -72,4 +72,15 @@ function cdf(d::Semicircle, x::Real)
...
 
+function rand(rng::AbstractRNG, d::Semicircle)
+    # Idea:
+    # sample polar coodinates r,θ 
+    # of point uniformly distributed on radius d.r half disk
+    # project onto x axis
+    θ = pi*rand(rng)
+    r = d.r*sqrt(rand(rng))
+    x = cos(θ) * r
+    return x
```
devmotion, line 83: I think it would be better to use [`cospi`](https://docs.julialang.org/en/v1/base/math/#Base.Math.cospi) here:
```suggestion
    θ = rand(rng) # multiple of π
    r = d.r * sqrt(rand(rng))
    return cospi(θ) * r
```

test/semicircle.jl
```diff
@@ -37,3 +38,36 @@ d = Semicircle(2.0)
 @test quantile(d,  .0) == -2.0
 @test quantile(d,  .5) ==   .0
 @test quantile(d, 1.0) == +2.0
+
+function test_sample_uniform(rng,sample)
```
jw3126, line 42: This is a bit ad hoc... I will check the literature if there are statistical tests for uniform distribution that are easy to implement and powerful.

jw3126: Is it an option to have `HypothesisTests.jl` as a test dependency?

devmotion: No, I would like to avoid that. We removed it a while ago because it was very annoying since it created a circular dependency. So, e.g., it was not possible to test any breaking releases properly.

devmotion: However, you could use `pvalue_kolmogorovsmirnoff` which was added to the tests when the HypothesisTests tests dependency was removed: https://github.com/JuliaStats/Distributions.jl/blob/8864b314c29925d6cf512965ace6818e600675b5/test/testutils.jl#L594-L597

devmotion (approved):
LGTM, thank you!

PR #1581, opened 2022-07-04
Update ksdist.jl

jw3126:
While the sqrt(n) factor is often part of the test statistic this is not what is implemented here. To witness, see

https://github.com/JuliaStats/Distributions.jl/blob/254f27f1b32a72d6529e2f78a3745b68bd38b31e/test/testutils.jl#L597-L607

where there is no sqrt(n) factor.


devmotion (approved):
LGTM, thank you!

It's e.g. also indicated by the support of `KSDist` that the docstring is incorrect.

PR #1583, opened 2022-07-04
Improve the `logpdf` of `NegativeBinomial`

### LINKED ISSUE #1582: ForwardDiff of `logpdf` of `NegativeBinomial` gives wrong results for `p==1`
This was discovered and documented in #1568 via tests, but #1579 reverted those tests. I still think this is worth tracking:

```julia
using Distributions, ForwardDiff
using Distributions: digamma

mydiffp(r, p, k) = isone(p) && iszero(k) ? r : r/p - k/(1 - p)
mydiffr(r, p, k) = log(p) - inv(k + r) - digamma(r) + digamma(r + k + 1)

f1(_p) = logpdf(NegativeBinomial(1.0, _p), 0)
f2(_r) = logpdf(NegativeBinomial(_r, 1.0), 0)

ForwardDiff.derivative(f1, 1.0) ≈ mydiffp(1.0, 1.0, 0) # false
ForwardDiff.derivative(f2, 1.0) ≈ mydiffr(1.0, 1.0, 0) # false
```

devmotion:
Uses `xlogy` and `xlog1py` now which take care of the special case `p == 1` automatically. Moreover, for improved numerical accuracy for all cases where `k == 0` (and a bit improved efficiency) there's a new branch for `k == 0`.

Will fix #1582 once the rules for `xlogy` and `xlog1py` in DiffRules and LogExpFunctions are fixed (will open PRs).

Edit: Fixes #1582.


simsurace:
> Will fix https://github.com/JuliaStats/Distributions.jl/issues/1582

Wouldn't it make sense to add a test for this now?

devmotion:
It's not fixed by the PR alone, so the test would fail.

simsurace:
Now that https://github.com/JuliaDiff/DiffRules.jl/pull/85 and https://github.com/JuliaStats/LogExpFunctions.jl/pull/57 have been merged, this could be revived, right?

devmotion:
FYI I applied similar changes also to the reverse-mode rule, and moved computations of constants outside of the pullback. That leads to a significant performance improvement of the pullback function (of course, only matters if it is evaluated multiple times):

## On master

```julia
julia> using Distributions, ChainRulesCore, BenchmarkTools

julia> d = NegativeBinomial(0.8, 0.5);

julia> x = 1;

julia> @btime rrule($logpdf, $d, $x);
  45.421 ns (0 allocations: 0 bytes)

julia> _, pb = rrule(logpdf, d, x);

julia> @btime $pb($(randn()));
  36.472 ns (0 allocations: 0 bytes)
```

## With this PR

```julia
julia> using Distributions, ChainRulesCore, BenchmarkTools

julia> d = NegativeBinomial(0.8, 0.5);

julia> x = 1;

julia> @btime rrule($logpdf, $d, $x);
  77.026 ns (0 allocations: 0 bytes)

julia> _, pb = rrule(logpdf, d, x);

julia> @btime $pb($(randn()));
  1.501 ns (0 allocations: 0 bytes)
```

src/univariate/discrete/negativebinomial.jl
```diff
@@ -97,13 +97,15 @@ end
...
 function logpdf(d::NegativeBinomial, k::Real)
-    r = d.r * log(d.p) + k * log1p(-d.p)
-    if isone(d.p) && iszero(k)
-        return zero(r)
-    elseif !insupport(d, k)
-        return oftype(r, -Inf)
+    r, p = params(d)
+    z = xlogy(r, p) + xlog1py(k, -p)
+    if iszero(k)
+        # in this case `log(k + r) + logbeta(r, k + 1) == 0` analytically but unfortunately not numerically
```
matbesancon, line 103: maybe add another comment line on why this leads to returning `z`?

src/univariate/discrete/negativebinomial.jl
```diff
@@ -97,13 +97,15 @@ end
...
+        return z
+    elseif insupport(d, k)
```
matbesancon, line 105: I have a preference for early termination with edge cases and finishing with the standard case (as was done before, so a branch for !insupport and the normal computation otherwise

test/univariate/discrete/negativebinomial.jl
```diff
@@ -57,3 +70,11 @@ end
...
     test_rrule(logpdf, dist, 0.0 ⊢ ChainRulesTestUtils.NoTangent(); fdm=fdm)
 end
+
+@testset "issue #1582" begin
+    dp = mydiffp(1.0, 1.0, 0.0)
+    @test ForwardDiff.derivative(p -> logpdf(NegativeBinomial(1.0, p), 0.0), 1.0) == dp == 1.0
+
+    dr = mydiffr(1.0, 1.0, 0.0)
+    @test ForwardDiff.derivative(r -> logpdf(NegativeBinomial(r, 1.0), 0.0), 1.0) == dr == 0.0
+end
```
matbesancon, line 80:
```suggestion
end

```

matbesancon:
two minor comments, shouldn't be a big deal

devmotion:
I tried to address your comments but I was not completely sure what you had in mind. Can you check if I changed it correctly?

devmotion:
I'd like to get this in, so I'll merge since I think I addressed the two comments. I'm happy to open a follow-up PR if something could/should be improved.

PR #1586, opened 2022-07-07
Implement log10-normal distribution

cgarling:
The existing lognormal distribution in Distributions.jl is nice, but for my work I also require a lognormal distribution defined in base 10; e.g., if X ~ Normal(a,b); 10^x ~ Log10Normal(a,b). This is relatively common in astrophysical applications and it seems there are implementations of this in other software ecosystems for applications like risk modelling, so I think it is a worthwhile addition. 

I did my best on the statistical quantities (e.g., skewness, kurtosis) but I'm not a statistician and have never computed these before. 

I based my tests on those already existing for the lognormal distribution. I added a doc entry in docs/src/univariate.md including a plot of the PDF. 

Please let me know if you require any revisions or additions to merge.

### DIFF
--- a/docs/src/fit.md
+++ b/docs/src/fit.md
@@ -44,6 +44,7 @@ The `fit_mle` method has been implemented for the following distributions:
 - [`DiscreteUniform`](@ref)
 - [`Exponential`](@ref)
 - [`LogNormal`](@ref)
+- [`Log10Normal`](@ref)
 - [`Normal`](@ref)
 - [`Gamma`](@ref)
 - [`Geometric`](@ref)
--- a/docs/src/univariate.md
+++ b/docs/src/univariate.md
@@ -319,6 +319,13 @@ plotdensity((0, 5), LogNormal, (0, 1)) # hide
 ```
 
 ```@docs
+Log10Normal
+```
+```@example plotdensity
+plotdensity((0, 20), Log10Normal, (1, 0.2)) # hide
+```
+
+```@docs
 LogUniform
 ```
 ```@example plotdensity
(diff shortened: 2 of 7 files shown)


devmotion:
Thanks for the PR!

I mainly wonder if it would be better to generalize the existing `LogNormal` to other bases (such as 10)?

cgarling:
I think there is a performance benefit to having unique forms for base e and 10; i.e. log10(x) is faster than log(10,x). 

As a user, I would be happy to have LogNormal, Log10Normal, and LogBNormal (for example) for arbitrary base, but I'm not sure what the maintainers would prefer. 

I'll start looking at the derivations in my spare time, as I agree it would be nice to be able to use arbitrary bases. If I make progress, I'll start a new branch.

cgarling:
Reviewing the math, it looks like we could just add a new constructor for LogNormal like

`LogNormal(\mu, \sigma, B) = LogNormal(\mu * log(B), \sigma * log(B))`

as all lognormal distributions are transformable from base `B` to base `e` through the above substitution. All the methods defined for LogNormal should then just work on the transformed distribution. The downside is that it's an invisible transformation to the user, who might be confused if they look at the parameters of the distribution in the REPL and notice they aren't the same as what they provided.

I have an implementation of the general LogBNormal distribution ready on [a new branch of my fork](https://github.com/cgarling/Distributions.jl/tree/cgarling/logbnormal), if you'd rather have a unique distribution. Let me know which you'd rather do.

cgarling:
I agree that an additional constructor sounds like a better solution. I'll open a new PR with that implementation.

PR #1591, opened 2022-07-13
Add alternate LogNormal constructor for arbitrary base

cgarling:
As discussed in [#1586](https://github.com/JuliaStats/Distributions.jl/pull/1586), this PR adds a new constructor for the LogNormal distribution that takes logB-mean \mu and scale \sigma in an arbitrary base B and converts it to a natural base-e LogNormal distribution. I added a few tests for type stability and PDF accuracy along with documentation for the new constructor.

I wasn't sure whether I should change the call signature in the documentation to something like `LogNormal(\sigma,\mu [,B])` but I tried to follow the correct style when updating the docs. Let me know what revisions you'd like to see

### DIFF
--- a/src/univariate/continuous/lognormal.jl
+++ b/src/univariate/continuous/lognormal.jl
@@ -10,10 +10,21 @@ f(x; \\mu, \\sigma) = \\frac{1}{x \\sqrt{2 \\pi \\sigma^2}}
 \\quad x > 0
 ```
 
+The log normal distribution is transformable between different log bases. For example, if ``X \\sim \\operatorname{Normal}(\\mu, \\sigma)``, then ``\\exp(X)`` is log-normally distributed, and more generally ``B^X`` is also log-normally distributed. In this more general case, the probability density function is
+
+```math
+f(x; \\mu, \\sigma, B) = \\frac{1}{x \\log(B) \\sqrt{2 \\pi \\sigma^2}}
+\\exp \\left( - \\frac{(\\log(B,x) - \\mu)^2}{2 \\sigma^2} \\right),
+\\quad x > 0, \\quad B > 0
+```
+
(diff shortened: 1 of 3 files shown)


src/univariate/continuous/lognormal.jl
```diff
@@ -41,6 +52,13 @@ LogNormal(μ::Real, σ::Real; check_args::Bool=true) = LogNormal(promote(μ, σ)
 LogNormal(μ::Integer, σ::Integer; check_args::Bool=true) = LogNormal(float(μ), float(σ); check_args=check_args)
 LogNormal(μ::Real=0.0) = LogNormal(μ, one(μ); check_args=false)
 
+# for arbitrary base
+function LogNormal(μ::Real, σ::Real, B::Real; check_args::Bool=true)
+    @check_args LogNormal (B, B ≥ zero(B))
+    lb = log(B)
+    LogNormal(promote(μ * lb, σ * lb)...; check_args=check_args)
+end
```
devmotion, line 60: This could be simplified to
```suggestion
function LogNormal(μ::Real, σ::Real, B::Real; check_args::Bool=true)
    @check_args LogNormal (B, B ≥ zero(B))
    lb = log(B)
    LogNormal(μ * lb, σ * lb; check_args=check_args)
end
```
as the `LogNormal` constructors already take care of promotion.

However, I wonder if we should promote all arguments first to ensure that e.g. `LogNormal(0f0, 1f0, 10)` returns a `LogNormal{Float32}` (currently it would return a `LogNormal{Float64}`). I.e., maybe we should use
```suggestion
function LogNormal(μ::Real, σ::Real, B::Real; check_args::Bool=true)
    return LogNormal(promote(μ, σ, B)...; check_args=check_args)
end
function LogNormal(μ::T, σ::T, B::T; check_args::Bool=true) where {T<:Real}
    @check_args LogNormal (B, B ≥ zero(B))
    lb = log(B)
    LogNormal(μ * lb, σ * lb; check_args=check_args)
end
```

cgarling: Thanks for the review. Good catch; I prefer the latter solution. Will change this and add a test for this behavior.

cgarling:
Just checking in -- can this be merged or is there something else to be done?

PR #1594, opened 2022-07-19
WIP: Add new distributions (HyperSphericalUniform and PowerSpherical)

marcpabst:
The Power Spherical distribution (https://arxiv.org/abs/2006.04437) is nice (and simpler!) alternative to the von-Mises-Fisher distribution.


src/samplers/powerspherical.jl
```diff
@@ -0,0 +1,37 @@
...
+    t = 2 * z - 1
+    m = sqrt(1 - t ^ 2) * v'
+    y = [t; m]
+    e_1 = [1.; zeros(eltype(spl.μ), spl.d -1)]
+    u = e_1 - spl.μ
+    normalize!(u)
+    x .= (-1) * (Matrix{eltype(spl.μ)}(I, spl.d, spl.d) .- 2*u*u') * y
+end
+
+function _rand(rng::AbstractRNG, spl::PowerSphericalSampler, x::AbstractVector)
```
sethaxen, line 23: Is this method needed? Or does the default call the `_rand!` method above after allocating `x`?

src/multivariate/hypersphericaluniform.jl
```diff
@@ -0,0 +1,28 @@
...
+    x .= randn(rng, T, length(x))
+    normalize!(x)
+end
+
+function _rand(rng::AbstractRNG, ::HyperSphericalUniform, x::AbstractVector{T}) where T<:Real
+    normalize!(randn(rng, T, length(x)))
+end
+
+function StatsBase.entropy(s::HyperSphericalUniform)
+    return 1/2 * s.d * log(pi) - lgamma(s.d/2) + log(2)
```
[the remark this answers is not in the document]
devmotion: The suggestion promotes to Float64 as well since `halfd` is always a Float64. But I don't think that can be avoided, the result will always be `Float64` as there are no unknown types here. `lgamma` is deprecated, instead one should use `loggamma`.
```suggestion
    halfd = s.d / 2
    return halfd * logπ - loggamma(halfd) + logtwo
```

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,56 @@
...
+_rand!(rng::AbstractRNG, d::PowerSpherical, x::AbstractVector) =
+    _rand!(rng, sampler(d), x)
+
+_rand(rng::AbstractRNG, d::PowerSpherical, x::AbstractVector) =
+    _rand(rng, sampler(d), x)
+
+#_logpdf
+function _logpdf(d::PowerSpherical, x::AbstractArray)
+    a, b = (length(d) - 1) / 2. + d.κ, (length(d) - 1) / 2.
+    return log(2) * (-a-b) + lgamma(a+b) - lgamma(a) + b * log(π) + d.κ .* log(d.μ' * x .+ 1)
```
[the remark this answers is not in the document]
devmotion: This won't avoid promotion to `Float64` either and should take into account the type of `x` when working with the irrationals to avoid unnecessary promotion.

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,56 @@
...
+end
+
+# entropy
+function StatsBase.entropy(d::PowerSpherical)
+    a, b = (length(d) - 1) / 2. + d.κ, (length(d) - 1) / 2.
+    logC = -( (a+b) * log(2) + lgamma(a) + b * log(pi) - lgamma(a+b))
+    return -(logC + d.κ * ( log(2) + digamma(a) - digamma(a+b)))
```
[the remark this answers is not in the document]
devmotion: Same here.

src/samplers/powerspherical.jl
marcpabst: I need Zygote to diff through `rand(PowerSpherical)` (so a non-array mutating version would be nice) but I just realised that that only works if you call `rand` without providing an RNG. Maybe we could have a `rsample()` method or something alike?

src/multivariate/hypersphericaluniform.jl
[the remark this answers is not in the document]
marcpabst: One could probably use something like s.d / 2one(s.d), though I'm not sure how efficient that would be....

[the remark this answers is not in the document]
devmotion: Since the type of `s.d` is `Int`, this would not change the type of the output. Division by two is fine here.

src/samplers/powerspherical.jl
marcpabst: I have to correct myself, when correctly implementing a non-mutating `rand()` function, Zygote knows how to handle this. I agree that a `ChainRule` (and potentially a `DiffRule`) would probably be best, but I just don't have time for that at the moment. Can we just have a non-mutating `rand()` version for know?

src/samplers/powerspherical.jl
```diff
@@ -0,0 +1,37 @@
...
+
+struct PowerSphericalSampler{T <: Real}
+    μ::Vector{T}
+    κ::T
+    d::Int
+    dist_b::Beta
+    dist_u::HyperSphericalUniform
+end
+
+function _rand!(rng::AbstractRNG, spl::PowerSphericalSampler, x::AbstractVector)
```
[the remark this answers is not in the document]
marcpabst: I'll make some changes and see what I can do...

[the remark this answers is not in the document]
marcpabst: Not sure if you can do better than this (probably)... Maybe you can have a look?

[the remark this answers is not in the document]
sethaxen: Sure, will look at it in more detail later today. Is it safe to assume `μ` is a unit vector?

src/samplers/powerspherical.jl
```diff
@@ -0,0 +1,43 @@
...
+    t = 2 * z - 1
+    m = sqrt(1 - t ^ 2) * v'
+    y = [t; m]
+    x[1] = 1.
+    x[2:end] .= 0.
+    x .= x - spl.μ
+
+    normalize!(x)
+
+    x .= (-1) * (I(2) .- 2*x*x') * y
```
sethaxen, line 24: Doesn't `I(2)` imply this only works for 2D vectors `x`?

marcpabst: Oh, good catch :) You should be careful when committing your experiments I guess ...

src/samplers/powerspherical.jl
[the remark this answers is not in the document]
marcpabst: Yes! I changed the constructor so it will either normalize `μ` or complain if it is not a unit vector.

src/multivariate/hypersphericaluniform.jl
```diff
@@ -0,0 +1,29 @@
+"""
+Uniform distribution on the hyperspherical unit ball in ``d`` dimensions.
+"""
+struct HyperSphericalUniform <: ContinuousMultivariateDistribution
+    d :: Int
+end
+
+# basic properties
+Base.length(s::HyperSphericalUniform) = s.d
+Base.eltype(::HyperSphericalUniform) = Float32
```
devmotion, line 10: This should be `Float64` but IIRC that's the default already.
```suggestion
```

src/multivariate/hypersphericaluniform.jl
```diff
@@ -0,0 +1,29 @@
...
+
+# sampling
+function _rand!(rng::AbstractRNG, ::HyperSphericalUniform, x::AbstractVector{T}) where T<:Real
+    x .= randn(rng, T, length(x))
```
devmotion, line 14:
```suggestion
function _rand!(rng::AbstractRNG, ::HyperSphericalUniform, x::AbstractVector{<:Real})
    randn!(rng, x)
```

src/multivariate/hypersphericaluniform.jl
```diff
@@ -0,0 +1,29 @@
...
+    normalize!(x)
+end
+
+function _rand(rng::AbstractRNG, ::HyperSphericalUniform, x::AbstractVector{T}) where T<:Real
+    normalize!(randn(rng, T, length(x)))
+end
+
```
devmotion, line 21: This should be removed I think.

```suggestion
```

src/multivariate/hypersphericaluniform.jl
```diff
@@ -0,0 +1,29 @@
+"""
+Uniform distribution on the hyperspherical unit ball in ``d`` dimensions.
+"""
+struct HyperSphericalUniform <: ContinuousMultivariateDistribution
+    d :: Int
+end
+
+# basic properties
```
devmotion, line 8: I guess we also have to define `support` and `insupport`.

src/multivariate/hypersphericaluniform.jl
```diff
@@ -0,0 +1,29 @@
+"""
+Uniform distribution on the hyperspherical unit ball in ``d`` dimensions.
```
devmotion, line 2: Can you extend the docstring and add it to the documentation?

src/multivariate/hypersphericaluniform.jl
```diff
@@ -0,0 +1,29 @@
...
+end
+
+function StatsBase.entropy(s::HyperSphericalUniform)
+    halfd = s.d / 2
+    return halfd * logπ - loggamma(halfd) + logtwo
+end
+
+function _logpdf(s::HyperSphericalUniform, x)
+    return -entropy(s)
+end
```
devmotion, line 29:
```suggestion
function _logpdf(s::HyperSphericalUniform, x::AbstractVector{<:Real})
    y = entropy(s)
    return insupport(s, x) ? -y : oftype(y, -Inf)
end
```

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,77 @@
+"""
+The Power Spherical distribution is useful as a replacement for the von Mises-Fisher distribution.
+
+The probability density function of the Power Spherical distribution parameterized by the mean direction ``\\mu`` and the concentration parameter ``\\kappa`` is given by
+    
+```math
+p_{X}(x ; \\mu, \\kappa)= \\left\\{2^{\\alpha+\\beta} \\pi^{\\beta} \\frac{\\Gamma(\\alpha)}{\\Gamma(\\alpha+\\beta)}\\right\\}^{-1}\\left(1+\\mu^{\\top} x\\right)^{\\kappa}
+```
```
devmotion, line 8: Can you make it consistent with other docstrings, add signatures etc. and add it to the docs?

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,77 @@
...
+"""
+struct PowerSpherical{T <: Real} <: ContinuousMultivariateDistribution
+    μ::Vector{T}
+    κ::T
+
+    function PowerSpherical(μ::Vector{T}, κ::T = one(T); checknorm = true, normalize_μ = true) where {T <: Real}
```
devmotion, line 14: IIRC we use `checkargs` for enabling/disabling checks of arguments.

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,77 @@
...
+        κ > 0 || error("κ must be positive.")
+        normalize_μ || !checknorm || isunitvec(μ) || error("μ must be a unit vector")
+             
+        new{eltype(μ)}(normalize_μ ? normalize(μ) : μ, κ)
+    end
+end
+
+function PowerSpherical(μ::Vector{T}, κ::Real; kwargs...) where {T<:Real}
+    R = promote_type(T, eltype(κ))
+    return PowerSpherical(convert(AbstractArray{R}, μ), convert(R, κ), kwargs...)
```
devmotion, line 25:
```suggestion
    return PowerSpherical(convert(AbstractArray{R}, μ), convert(R, κ); kwargs...)
```

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,77 @@
...
+end
+
+### Basic properties
+Base.length(d::PowerSpherical) = length(d.μ)
+Base.eltype(d::PowerSpherical) = eltype(d.μ)
+meandir(d::PowerSpherical) = d.μ
+concentration(d::PowerSpherical) = d.κ
+insupport(::PowerSpherical, x::AbstractVector{T}) where {T<:Real} = isunitvec(x)
```
devmotion, line 33:
```suggestion
insupport(d::PowerSpherical, x::AbstractVector{<:Real}) = length(x) == length(d) && isunitvec(x)
```

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,77 @@
...
+
+function sampler(d::PowerSpherical)
+    dim = length(d)
+    return PowerSphericalSampler(
+        d.μ,
+        d.κ,
+        dim, 
+        Beta((dim - 1) / 2. + d.κ, (dim - 1) / 2.),
```
devmotion, line 41:
```suggestion
        Beta(beta + d.κ, beta; check_args=false), 
```

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,77 @@
...
+meandir(d::PowerSpherical) = d.μ
+concentration(d::PowerSpherical) = d.κ
+insupport(::PowerSpherical, x::AbstractVector{T}) where {T<:Real} = isunitvec(x)
+
+function sampler(d::PowerSpherical)
+    dim = length(d)
+    return PowerSphericalSampler(
+        d.μ,
+        d.κ,
+        dim,
```
devmotion, line 40: Is this argument needed? The information is already contained in mu.
```suggestion
```

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,77 @@
...
+
+### Basic properties
+Base.length(d::PowerSpherical) = length(d.μ)
+Base.eltype(d::PowerSpherical) = eltype(d.μ)
+meandir(d::PowerSpherical) = d.μ
+concentration(d::PowerSpherical) = d.κ
+insupport(::PowerSpherical, x::AbstractVector{T}) where {T<:Real} = isunitvec(x)
+
+function sampler(d::PowerSpherical)
+    dim = length(d)
```
devmotion, line 36:
```suggestion
    dim = length(d)
    beta = (dim - 1) / 2
```

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,77 @@
...
+        d.μ,
+        d.κ,
+        dim, 
+        Beta((dim - 1) / 2. + d.κ, (dim - 1) / 2.), 
+        HyperSphericalUniform(dim-1)
+    )
+end
+
+_rand!(rng::AbstractRNG, d::PowerSpherical, x::AbstractVector) =
+    _rand!(rng, sampler(d), x)
```
devmotion, line 47: Is this needed or already the default?

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,77 @@
...
+
+rand(rng::AbstractRNG, d::PowerSpherical) =
+    rand(rng, sampler(d))
```
devmotion, line 50:
```suggestion
```

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,77 @@
...
+
+#_logpdf
+function _logpdf(d::PowerSpherical, x::AbstractArray)
```
devmotion, line 53: I would assume this creates method ambiguity issues?

```suggestion
function _logpdf(d::PowerSpherical, x::AbstractVector{<:Real})
```

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,77 @@
...
+    b = (length(d) - 1) // 2
+    a = b + d.κ
+    c = a + b
+
+    return logtwo * (-a-b) + loggamma(c) - loggamma(a) + b * logπ + d.κ .* log1p(d.μ' * x)
```
devmotion, line 58: This promotes to Float64 which might not be desired and causes intermediate allocations. One needs to convert irrationals to something like `float(promote_eltype(d.mu, x))`. Additionally, one has to handle the case `!insupport(d, x)`.

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,77 @@
...
+end
+
+
+# entropy
+function StatsBase.entropy(d::PowerSpherical)
+    b = (length(d) - 1) / 2
```
devmotion, line 64:
```suggestion
    b = (length(d) - 1) // 2
```

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,77 @@
...
+    _rand!(rng, sampler(d), x)
+
+rand(rng::AbstractRNG, d::PowerSpherical) =
+    rand(rng, sampler(d))
+
+#_logpdf
+function _logpdf(d::PowerSpherical, x::AbstractArray)
+    b = (length(d) - 1) // 2
+    a = b + d.κ
+    c = a + b
```
devmotion, line 56: Isn't this just

```suggestion
    c = length(d) - 1 + d.κ
```

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,77 @@
...
+
+    return logtwo * (-a-b) + loggamma(c) - loggamma(a) + b * logπ + d.κ .* log1p(d.μ' * x)
+end
+
+
+# entropy
+function StatsBase.entropy(d::PowerSpherical)
+    b = (length(d) - 1) / 2
+    a = b + d.κ
+    c = a + b
```
devmotion, line 66:
```suggestion
    c = length(d) - 1 + d.κ
```

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,77 @@
...
+
+    logC = -(c * logtwo + loggamma(a) + b * logπ - loggamma(c))
```
devmotion, line 68: This causes undesired promotions. Probably one has to convert irrationals, `b`, and `c` to `typeof(a)` first.

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,77 @@
...
+    return -(logC + d.κ * ( logtwo + digamma(a) - digamma(c)))
+end
+
+# analytical KL divergences
+function kldivergence(p::PowerSpherical, q::HyperSphericalUniform)
+    return -entropy(p) + entropy(q)
```
devmotion, line 74:
```suggestion
    return entropy(q) - entropy(p)
```

src/samplers/powerspherical.jl
```diff
@@ -0,0 +1,44 @@
...
+
+    y = [t; m]
+    e_1 = [1.; zeros(eltype(spl.μ), spl.d -1)]
+
+    û = e_1 - spl.μ
+    u = normalize(û)
+
+    return (-1) * (I(spl.d) .- 2*u*u') * y
+end
+
```
devmotion, line 43:
```suggestion
```

test/powerspherical.jl
```diff
@@ -0,0 +1,10 @@
+@testset "Testing argument promotions" begin
```
devmotion, line 1: Can you extend the tests? Every added line should be tested.

src/multivariate/powerspherical.jl
marcpabst: I this a performance related change?

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,77 @@
...
+end
+
+### Basic properties
+Base.length(d::PowerSpherical) = length(d.μ)
+Base.eltype(d::PowerSpherical) = eltype(d.μ)
+meandir(d::PowerSpherical) = d.μ
+concentration(d::PowerSpherical) = d.κ
+insupport(::PowerSpherical, x::AbstractVector{T}) where {T<:Real} = isunitvec(x)
+
+function sampler(d::PowerSpherical)
```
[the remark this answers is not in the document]
marcpabst: Nope. Tbh, I don't think it makes much of a difference in terms of performance but it keeps the distribution itself lean and provides a nice separation between the parameters and the sampling. Otherwise we would need to store both a `Beta` and a `HyperSphericalUniform` int the `PowerSpherical` type.

src/multivariate/hypersphericaluniform.jl
marcpabst: I can do that, but it is not mentioned in the docs as a requirement for multivariate distributions... :)

src/samplers/powerspherical.jl
```diff
@@ -0,0 +1,27 @@
...
+    m = sqrt(1 - t ^ 2) * v'
+    y = [t; m]
+
+    x[1] = 1.
+    x[2:end] .= 0.
+    x .= x - spl.μ
+
+    normalize!(x)
+
+    x .= (-1) * (I(length(spl)) .- 2*x*x') * y
```
sethaxen, line 24: Here's a completely non-allocating implementation:

```suggestion
    v = @views x[begin+1:end]
    uhat1 = 1 - spl.μ[begin]
    μv = @views spl.μ[begin+1:end]

    z = rand(rng, spl.dist_b)
    rand!(rng, spl.dist_u, v)

    t = 2 * z - 1

    v .*= sqrt(1 - t ^ 2)

    twouuᵀy1 = t * uhat1 - dot(v, μv)
    x[begin] = t - twouuᵀy1
    v .+= μv .* (twouuᵀy1 / uhat1)
    return x
```

src/samplers/powerspherical.jl
```diff
@@ -0,0 +1,27 @@
+# Sampler for Power Spherical
+
+struct PowerSphericalSampler{T <: Real} <: Sampleable{Multivariate,Continuous}
+    μ::Vector{T}
+    κ::T
+    dist_b::Beta
```
sethaxen, line 6: Without a type annotation, the `_rand!` method is type-unstable.
```suggestion
    dist_b::Beta{T}
```

src/samplers/powerspherical.jl
```diff
@@ -0,0 +1,27 @@
...
+    dist_u::HyperSphericalUniform
+end
+
```
sethaxen, line 9: I think it makes sense to add a constructor like this, and then your `sampler` implementation elsewhere can just call the constructor:

```suggestion

function PowerSphericalSampler(μ, κ)
    d = length(μ)
    β = (d - 1) // 2
    dist_b = Beta(β + κ, β)
    dist_u = HyperSphericalUniform(d - 1)
    return PowerSphericalSampler(μ, κ, dist_b, dist_u)
end

```

src/multivariate/powerspherical.jl
```diff
@@ -0,0 +1,69 @@
...
+    dim = length(d)
+    beta = (dim - 1) / 2
+    return PowerSphericalSampler(
+        d.μ,
+        d.κ,
+        Beta(beta + d.κ, beta; check_args=false), 
+        HyperSphericalUniform(dim-1)
+    )
+end
```
sethaxen, line 42: As recommended below, I think it makes more sense to put this in the constructor for `PowerSphericalSampler`.
```suggestion
sampler(d::PowerSpherical) = PowerSphericalSampler(d.μ, d.κ)
```

sethaxen (changes requested):

src/multivariate/powerspherical.jl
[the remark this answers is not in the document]
devmotion: > Otherwise we would need to store both a Beta and a HyperSphericalUniform in the PowerSpherical type.

The usual alternative would rather be to construct these distributions in `_rand!`.

src/multivariate/powerspherical.jl
devmotion: I guess rather something like
```julia
function StatsBase.entropy(d::PowerSpherical)
    _b = (length(d) - 1) // 2
    a = _b + d.κ
    T = float(typeof(a))
    b = T(_b)
    c = T(length(d) - 1 + d.κ)

    _logtwo = T(logtwo)
    _logπ = T(logπ)

    logC = c * _logtwo + loggamma(a) + b * _logπ - loggamma(c)
    return logC - d.κ * ( _logtwo + digamma(a) - digamma(c))
end
```

PR #1598, opened 2022-07-21
Use besselix instead of besseli for better numerical stability.

marcpabst:
This PR improves numerical stability for the von-Mises-Fisher distribution for large values of `κ`.


src/multivariate/vonmisesfisher.jl
```diff
@@ -65,7 +65,7 @@ function _vmflck(p, κ)
     T = typeof(κ)
     hp = T(p/2)
     q = hp - 1
-    q * log(κ) - hp * log2π - log(besseli(q, κ))
+    q * log(κ) - hp * log2π - log(besselix(q, κ)) + κ
```
devmotion, line 68: I think this should be
```suggestion
    q * log(κ) - hp * log2π - log(besselix(q, κ)) - κ
```
?

devmotion (commented):
Thanks! Can you add a test that would fail currently and be fixed by this PR?

src/multivariate/vonmisesfisher.jl
marcpabst: Ahhhh, don't just remove parentheses I guess without thinking :)

marcpabst:
Will do!

marcpabst:
Okay, I hope tests are okay now?

marcpabst:
I think the one failing test is unrelated to my changes?

test/vonmisesfisher.jl
```diff
@@ -6,8 +6,10 @@ using LinearAlgebra, Test
 using SpecialFunctions
 
 vmfCp(p::Int, κ::Real) = (κ ^ (p/2 - 1)) / ((2π)^(p/2) * besseli(p/2-1, κ))
+logvmfCp(p::Int, κ::Real) = 1/2 * ((p - 2) * log(κ) - 2 * (log(besselix(p/2 - 1, κ)) +  κ) + p * (-log(2π)))
```
devmotion, line 9: Personally, I think the amount of brackets makes the code a bit difficult to parse. Maybe use something like
```suggestion
logvmfCp(p::Int, κ::Real) = (p / 2 - 1) * log(κ) - log(besselix(p / 2 - 1, κ) - κ - p / 2 * log(2π)
```
? One could use `log2π` here (from IrrationalConstants/StatsFuns) and avoid some duplicate computations of `p / 2` and `p / 2 - 1` but I don't think it matters since it is only used in the tests.

test/vonmisesfisher.jl
```diff
@@ -6,8 +6,10 @@ using LinearAlgebra, Test
...
 
 safe_vmfpdf(μ::Vector, κ::Real, x::Vector) = vmfCp(length(μ), κ) * exp(κ * dot(μ, x))
```
devmotion, line 11: Can this be removed?

```suggestion
```

test/vonmisesfisher.jl
marcpabst: I agree that the whole `VonMisesFisher` implementation could use a bit of a rewrite. So I would not make too many changes right now...

test/vonmisesfisher.jl
marcpabst: Now it's just used to test the log-variant...

test/vonmisesfisher.jl
```diff
@@ -6,8 +6,10 @@ using LinearAlgebra, Test
 using SpecialFunctions
 
 vmfCp(p::Int, κ::Real) = (κ ^ (p/2 - 1)) / ((2π)^(p/2) * besseli(p/2-1, κ))
```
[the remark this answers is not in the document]
marcpabst: Same as the other non-log-variant.

test/vonmisesfisher.jl
devmotion: This is new and only used in the tests though? So it would be good to ensure that at least the new code is sufficiently clear and simple.

test/vonmisesfisher.jl
devmotion: Then let's remove it.

marcpabst: sure!

test/vonmisesfisher.jl
```diff
@@ -159,6 +161,13 @@ end
     @test d isa VonMisesFisher{Float64}
 end
 
+## Test log functions
+@testset "Testing logvmfCp and safe_logvmfpdf" begin
+    @test isapprox(log(vmfCp(5, 345.)), logvmfCp(5, 345.), atol=10e-10)
+    @test isapprox(log(safe_vmfpdf([1., 0], 500., [1., 7.])), safe_logvmfpdf([1., 0], 500., [1., 7.]), atol=10e-10)
+end
+
+
```
devmotion, line 170:
```suggestion
```

test/vonmisesfisher.jl
```diff
@@ -99,13 +101,13 @@ function test_vonmisesfisher(p::Int, κ::Real, n::Int, ns::Int,
     @test meandir(d2) ≈ μ
     @test concentration(d2) ≈ κ
 
-    @test isapprox(d.logCκ, log(vmfCp(p, κ)), atol=1.0e-12)
+    @test isapprox(d.logCκ, logvmfCp(p, κ), atol=1.0e-12)
```
devmotion, line 104:
```suggestion
    @test d.logCκ ≈ logvmfCp(p, κ) atol=1.0e-12
```

devmotion:
Thank you for the PR!

I wanted to add these improvements, so I applied the suggestions and hope this is fine with you.

PR #1600, opened 2022-07-30
Fix mini typos in docs

svilupp:
As part of the JuliaCon hackathon, I have run all the docs through Grammarly. In case of ambiguity (eg, inplace vs in-place), I have googled for the more popular spelling (often picked based on Wikipedia).

There were no changes to active code, ie, only comments or docs were changed.

Examples:
- log-scale --> on a log scale
- inplace --> in-place
- sub type (was inconsistently used) --> subtype

I'm happy to withdraw this PR if it's not helpful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ A Julia package for probability distributions and associated functions. Particul
 * Moments (e.g mean, variance, skewness, and kurtosis), entropy, and other properties
 * Probability density/mass functions (pdf) and their logarithm (logpdf)
 * Moment generating functions and characteristic functions
-* Sampling from population or from a distribution
+* Sampling a population or a distribution
 * Maximum likelihood estimation
 
 **Note:** The functionalities related to conjugate priors have been moved to the [ConjugatePriors package](https://github.com/JuliaStats/ConjugatePriors.jl).
@@ -32,7 +32,7 @@ Also, for casual conversation and quick questions, there are the channels `#help
 
 ### Reporting issues
 
-* If you need help or an explanation how to use *Distributions* ask in the forum (https://discourse.julialang.org) or, for informal questions, visit the chat (https://julialang.slack.com).
+* If you need help or an explanation of how to use *Distributions* ask in the forum (https://discourse.julialang.org) or, for informal questions, visit the chat (https://julialang.slack.com).
 
 If you have a bug linked with *Distributions*, check that it has
 not been reported yet on the issues of the repository.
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -13,7 +13,7 @@ A Julia package for probability distributions and associated functions. Particul
 * Moments (e.g mean, variance, skewness, and kurtosis), entropy, and other properties
 * Probability density/mass functions (pdf) and their logarithm (logpdf)
 * Moment generating functions and characteristic functions
-* Sampling from population or from a distribution
+* Sampling a population or a distribution
```
devmotion, line 16: I guess this change should be reverted as we're not sampling populations or distributions but drawing samples from a population or distribution:
```suggestion
* Sampling from a population or from a distribution
```

docs/src/multivariate.md
```diff
@@ -35,7 +35,7 @@ pdf(::MultivariateDistribution, ::AbstractArray)
 logpdf(::MultivariateDistribution, ::AbstractArray)
 loglikelihood(::MultivariateDistribution, ::AbstractVector{<:Real})
 ```
-**Note:** For multivariate distributions, the pdf value is usually very small or large, and therefore direct evaluating the pdf may cause numerical problems. It is generally advisable to perform probability computation in log-scale.
+**Note:** For multivariate distributions, the pdf value is usually very small or large, and therefore direct evaluation of the pdf may cause numerical problems. It is generally advisable to perform probability computation on a log scale.
```
devmotion, line 38: Hmm, not sure about "on a log scale". I guess I would prefer the original version:
```suggestion
**Note:** For multivariate distributions, the pdf value is usually very small or large, and therefore direct evaluation of the pdf may cause numerical problems. It is generally advisable to perform probability computation in log scale.
```

src/univariate/continuous/skewedexponentialpower.jl
```diff
@@ -74,7 +74,7 @@ function m_k(d::SkewedExponentialPower, k::Integer)
         loggamma(inv_p) + log(abs((-1)^k * α^(1 + k) + (1 - α)^(1 + k)))
 end
 
-# needed for odd moments on log-scale
+# needed for odd moments on a log scale
```
devmotion, line 77: Same here:
```suggestion
# needed for odd moments in log scale
```

devmotion (commented):
Thanks for the PR! Seems to be an improvement, I only had some minor comments :slightly_smiling_face:

docs/src/multivariate.md
svilupp: "on" is a much more popular preposition in this case, but happy to take your suggestion

devmotion (approved):
Thanks for the PR!

PR #1604, opened 2022-08-13
fix mgf of NegativeBinomial

### LINKED ISSUE #1603: NegativeBinomial wrong mgf
The moment generating function of NegativeBinomial is wrong:
```julia
julia> using Revise, Distributions, ForwardDiff, Test

julia> d = NegativeBinomial(4, 0.2)
NegativeBinomial{Float64}(r=4.0, p=0.2)

julia> @test ForwardDiff.derivative(Base.Fix1(mgf, d), 0) ≈ mean(d)
Test Failed at REPL[3]:1
  Expression: ForwardDiff.derivative(Base.Fix1(mgf, d), 0) ≈ mean(d)
   Evaluated: 5.000000000000001 ≈ 16.0
```
From the code it seems that the implementation of the mgf confuses success and fail probability.

jw3126:
fix #1603

### DIFF
--- a/src/univariate/discrete/geometric.jl
+++ b/src/univariate/discrete/geometric.jl
@@ -122,16 +122,12 @@ function invlogccdf(d::Geometric{T}, lp::Real) where T<:Real
     max(ceil(lp/log1p(-d.p)) - 1, zero(T))
 end
 
-function mgf(d::Geometric, t::Real)
+function laplace_transform(d::Geometric, t)
     p = succprob(d)
-    p / (p - (1 - p) * expm1(t))
+    p / (p - (1 - p) * expm1(-t))
 end
-
-function cf(d::Geometric, t::Real)
-    p = succprob(d)
-    p / (p - (1 - p) * expm1(t*im))
-end
-
+mgf(d::Geometric, t::Real) = laplace_transform(d, -t)
+cf(d::Geometric, t::Real) = laplace_transform(d, -t*im)
 
 ### Sampling
 
--- a/src/univariate/discrete/geometric.jl
+++ b/src/univariate/discrete/geometric.jl
@@ -124,7 +124,7 @@ end
 
 function mgf(d::Geometric, t::Real)
     p = succprob(d)
-    p / (expm1(-t) + p)
+    p / (expm1(-t) + p) - t
 end
 
 function cf(d::Geometric, t::Real)
--- a/src/univariate/discrete/geometric.jl
+++ b/src/univariate/discrete/geometric.jl
@@ -124,7 +124,7 @@ end
 
 function mgf(d::Geometric, t::Real)
     p = succprob(d)
-    p / (expm1(-t) + p) - t
+    p / (1 - (1 - p) * exp(t))
 end
 
 function cf(d::Geometric, t::Real)
--- a/src/univariate/discrete/geometric.jl
+++ b/src/univariate/discrete/geometric.jl
@@ -124,7 +124,7 @@ end
 
 function mgf(d::Geometric, t::Real)
     p = succprob(d)
-    p / (1 - (1 - p) * exp(t))
+    p / (p - (1 - p) * expm1(t))
 end
 
 function cf(d::Geometric, t::Real)
--- a/src/univariate/discrete/geometric.jl
+++ b/src/univariate/discrete/geometric.jl
@@ -129,8 +129,7 @@ end
 
 function cf(d::Geometric, t::Real)
     p = succprob(d)
-    # replace with expm1 when complex version available
-    p / (exp(-t*im) - 1 + p)
+    p / (1 - (1 - p) * cis(t))
 end
 
 
--- a/src/univariate/discrete/geometric.jl
+++ b/src/univariate/discrete/geometric.jl
@@ -129,7 +129,7 @@ end
 
 function cf(d::Geometric, t::Real)
     p = succprob(d)
-    p / (1 - (1 - p) * cis(t))
+    p / (p - (1 - p) * expm1(t*im))
 end
 
 
--- a/src/univariate/discrete/negativebinomial.jl
+++ b/src/univariate/discrete/negativebinomial.jl
@@ -128,12 +128,12 @@ end
 
 function mgf(d::NegativeBinomial, t::Real)
     r, p = params(d)
-    return mgf(Geometric(p), t)^r
+    return mgf(Geometric(p, check_args=false), t)^r
 end
 
 function cf(d::NegativeBinomial, t::Real)
     r, p = params(d)
-    return (((1 - p) * cis(t)) / (1 - p * cis(t)))^r
+    return cf(Geometric(p, check_args=false), t)^r
 end
 
 # ChainRules definitions
--- a/src/univariate/discrete/negativebinomial.jl
+++ b/src/univariate/discrete/negativebinomial.jl
@@ -132,7 +132,7 @@ function laplace_transform(d::NegativeBinomial, t)
 end
 
 mgf(d::NegativeBinomial, t::Real) = laplace_transform(d, -t)
-cf(d::Geometric, t::Real) = laplace_transform(d, -t*im)
+cf(d::NegativeBinomial, t::Real) = laplace_transform(d, -t*im)
 
 # ChainRules definitions
 
--- a/src/univariate/discrete/negativebinomial.jl
+++ b/src/univariate/discrete/negativebinomial.jl
@@ -131,10 +131,6 @@ function laplace_transform(d::NegativeBinomial, t)
     return laplace_transform(Geometric(p, check_args=false), t)^r
 end
 
-function cf(d::NegativeBinomial, t::Real)
-    r, p = params(d)
-    return cf(Geometric(p, check_args=false), t)^r
-end
 mgf(d::NegativeBinomial, t::Real) = laplace_transform(d, -t)
 cf(d::Geometric, t::Real) = laplace_transform(d, -t*im)
 
--- a/src/univariate/discrete/negativebinomial.jl
+++ b/src/univariate/discrete/negativebinomial.jl
@@ -126,15 +126,17 @@ function rand(rng::AbstractRNG, d::NegativeBinomial)
     end
 end
 
-function mgf(d::NegativeBinomial, t::Real)
+function laplace_transform(d::NegativeBinomial, t)
     r, p = params(d)
-    return mgf(Geometric(p, check_args=false), t)^r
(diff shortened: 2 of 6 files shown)


src/univariate/discrete/geometric.jl
```diff
@@ -124,7 +124,7 @@ end
 
 function mgf(d::Geometric, t::Real)
     p = succprob(d)
-    p / (expm1(-t) + p)
+    p / (expm1(-t) + p) - t
```
devmotion, line 127: Shouldn't this be reverted to
```suggestion
    p / (expm1(-t) + p)
```
(see e.g. https://en.wikipedia.org/wiki/Moment-generating_function#Examples)?

src/univariate/discrete/negativebinomial.jl
```diff
@@ -128,7 +128,7 @@ end
 
 function mgf(d::NegativeBinomial, t::Real)
     r, p = params(d)
-    return ((1 - p) * exp(t))^r / (1 - p * exp(t))^r
+    return mgf(Geometric(p), t)^r
```
devmotion, line 131: I think that's wrong (see e.g. https://en.wikipedia.org/wiki/Moment-generating_function#Examples)? My guess would be it should be something like
```suggestion
    return (p / (1 - (1 - p) * exp(t)))^r
```

test/univariate/discrete/negativebinomial.jl
```diff
@@ -7,6 +7,16 @@ using FiniteDifferences
...
 
 mydiffp(r, p, k) = r/p - k/(1 - p)
+@testset "issue #1603" begin
+    d = NegativeBinomial(4, 0.2)
+    @test ForwardDiff.derivative(Base.Fix1(mgf, d), 0) ≈ mean(d)
+    d = NegativeBinomial(1, 0.2)
+    @test ForwardDiff.derivative(Base.Fix1(mgf, d), 0) ≈ mean(d)
+    d = Geometric(0.2)
+    @test mgf(d, 0) ≈ 1
+    @test ForwardDiff.derivative(Base.Fix1(mgf, d), 0) ≈ mean(d)
```
devmotion, line 17: Could this be moved to test/univariate/discrete/geometric.jl instead?

src/univariate/discrete/geometric.jl
devmotion: I guess the tests for `Geometric` pass even with the (IMO) incorrect version since `-t` does not affect the derivative at 0.

src/univariate/discrete/negativebinomial.jl
jw3126: There is a subtle point here and with your other comment. Namely there are two widespread ways to parametrize the Geometric distribution, see https://en.wikipedia.org/wiki/Geometric_distribution

What is called `k-1` in the wikipedia mgf table is called `k` in `Distributions.jl`
With the Distributions.jl parametrizations `NegativeBinomial` is a convolution power in the most straight forward fashion
https://github.com/JuliaStats/Distributions.jl/blob/aca39200d7e2288b571895a09d8e0ec34b8dcf83/src/convolution.jl#L37

This translates into the mgf of NegativeBinomial being a power of that of Geometric.

src/univariate/discrete/geometric.jl
jw3126: `-t` adds a `-1` to the derivative. This corresponds to shifting the expectation value by `1`. Which corresponds to chaning the `Geometric` parametrization from `k-1` to `k`. See also my comments below.

devmotion: Ah, I completely missed the different parameterization here. However, AFAIK the linear transformation Y := X - 1 of a random variable X should result in M_Y(t) = exp(-t) * M_X(t) instead of a shift by -t. So it seems it should be something like in https://github.com/JuliaStats/Distributions.jl/pull/1604/files/1b75633699749011a03a6b0ec85c335a8281073a#r945207774 (without the `^r`), e.g.,
```suggestion
    p / (1 - (1 - p) * exp(t))
```
That would also correspond to the formula given on Wikipedia (https://en.wikipedia.org/wiki/Geometric_distribution, right column).

src/univariate/discrete/negativebinomial.jl
devmotion: Thanks for the explanation. Also with the formula in https://github.com/JuliaStats/Distributions.jl/pull/1604/files/1b75633699749011a03a6b0ec85c335a8281073a#r945344615 both the code in the PR (calling `mgf(Geometric(p), t)`) and the suggestion above would be identical it seems.

test/univariate/discrete/negativebinomial.jl
devmotion: Could you address this comment @jw3126?

src/univariate/discrete/geometric.jl
jw3126: Yes you are right, I was thinking about `cgf = log mgf` where `exp(t)` becomes `t`.

src/univariate/discrete/geometric.jl
```diff
@@ -124,13 +124,12 @@ end
...
     p = succprob(d)
-    p / (expm1(-t) + p)
+    p / (1 - (1 - p) * exp(t))
 end
 
 function cf(d::Geometric, t::Real)
     p = succprob(d)
-    # replace with expm1 when complex version available
-    p / (exp(-t*im) - 1 + p)
+    p / (1 - (1 - p) * cis(t))
```
devmotion, line 132: For the same reason as above, maybe use
```suggestion
    p / (p - (1 - p) * expm1(t*im))
```

test/univariate/discrete/geometric.jl
```diff
@@ -0,0 +1,11 @@
...
+using Test
+using ForwardDiff
+
+@testset "Geometric mgf and k vs k-1 parametrization #1604" begin
+    d = Geometric(0.2)
+    @test mgf(d, 0) ≈ 1 atol = 10eps(Float64) rtol=0
+    @test cf(d, 0)  ≈ 1 atol = 10eps(Float64) rtol=0
+    @test ForwardDiff.derivative(Base.Fix1(mgf, d), 0) ≈ mean(d)
+    @test ForwardDiff.derivative(Base.Fix1(cf, d), 0) ≈ mean(d) * im
+end
```
devmotion, line 11: Maybe we should check derivatives with FiniteDifferences instead since ForwardDiff sometimes silently returns wrong derivatives (https://github.com/JuliaStats/Distributions.jl/issues/1582) and finite differencing seems more reliable. It also avoids issues on old Julia versions mentioned below, I assume.

E.g., with the `expm1` changes above,
```suggestion
using FiniteDifferences

@testset "Geometric mgf and k vs k-1 parametrization #1604" begin
    d = Geometric(0.2)
    @test mgf(d, 0) == 1
    @test cf(d, 0) == 1

    fdm = central_fdm(5, 1)
    @test fdm(Base.Fix1(mgf, d), 0) ≈ mean(d)
    @test fdm(Base.Fix1(cf, d), 0) ≈ mean(d) * im
end
```

test/univariate/discrete/negativebinomial.jl
```diff
@@ -7,6 +7,17 @@ using FiniteDifferences
...
 mydiffp(r, p, k) = r/p - k/(1 - p)
+@testset "issue #1603" begin
+    d = NegativeBinomial(4, 0.2)
+    @test ForwardDiff.derivative(Base.Fix1(mgf, d), 0) ≈ mean(d)
+    d = NegativeBinomial(1, 0.2)
+    @test ForwardDiff.derivative(Base.Fix1(mgf, d), 0) ≈ mean(d)
+    if VERSION >= v"1.6"
+        # ForwardDiff bug in old julia
+        @test ForwardDiff.derivative(Base.Fix1(cf, d), 0) ≈ mean(d) * im
+    end
```
devmotion, line 18: Same here:
```suggestion
    fdm = central_fdm(5, 1)
    @test fdm(Base.Fix1(mgf, d), 0) ≈ mean(d)
    d = NegativeBinomial(1, 0.2)
    @test fdm(Base.Fix1(mgf, d), 0) ≈ mean(d)
    @test fdm(Base.Fix1(cf, d), 0) ≈ mean(d) * im
```

test/univariates.jl
```diff
@@ -127,13 +127,13 @@ function verify_and_test(D::Union{Type,Function}, d::UnivariateDistribution, dct
 
     try
         m = mgf(d,0.0)
-        @test m == 1.0
+        @test m ≈ 1.0 atol=10*eps(Float64) rtol=0
```
devmotion, line 130: I think with the `expm1` suggestions above this could be reverted.
```suggestion
        @test m == 1.0
```

test/univariates.jl
```diff
@@ -127,13 +127,13 @@ function verify_and_test(D::Union{Type,Function}, d::UnivariateDistribution, dct
...
     catch e
         isa(e, MethodError) || throw(e)
     end
     try
         c = cf(d,0.0)
-        @test c == 1.0
+        @test c ≈ 1.0 atol=10*eps(Float64) rtol=0
```
devmotion, line 136: Same here:
```suggestion
        @test c == 1.0
```

src/univariate/discrete/geometric.jl
```diff
@@ -124,13 +124,12 @@ end
 
 function mgf(d::Geometric, t::Real)
     p = succprob(d)
-    p / (expm1(-t) + p)
+    p / (1 - (1 - p) * exp(t))
```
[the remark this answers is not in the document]
jw3126: Awesome this trick is very neat, thanks a lot for all the suggestions.

src/univariate/discrete/negativebinomial.jl
```diff
@@ -126,15 +126,17 @@ function rand(rng::AbstractRNG, d::NegativeBinomial)
...
+    return laplace_transform(Geometric(p, check_args=false), t)^r
 end
 
 function cf(d::NegativeBinomial, t::Real)
     r, p = params(d)
-    return (((1 - p) * cis(t)) / (1 - p * cis(t)))^r
+    return cf(Geometric(p, check_args=false), t)^r
 end
+mgf(d::NegativeBinomial, t::Real) = laplace_transform(d, -t)
+cf(d::Geometric, t::Real) = laplace_transform(d, -t*im)
```
devmotion, line 139:
```suggestion
mgf(d::NegativeBinomial, t::Real) = laplace_transform(d, -t)
cf(d::NegativeBinomial, t::Real) = laplace_transform(d, -t*im)
```

devmotion (approved):
Thank you for these fixes!

PR #1606, opened 2022-08-22
Bug fix for CF and MGF of discretenonparametric

trappmartin:


devmotion:
Thanks for the PR! I wonder why https://github.com/JuliaStats/Distributions.jl/blob/7b3890d6547385ca460c209b993481335a629995/src/univariate/discrete/discretenonparametric.jl#L50 is not sufficient? In any case, could you add some tests that are broken currently and fixed by the PR?

trappmartin:
Good question, no idea.

trappmartin:
>  In any case, could you add some tests that are broken currently and fixed by the PR?
Sure, can give it a try.

trappmartin:
I added a test to the categorical distribution as I encountered the issue when using it.

test/univariate/discrete/categorical.jl
```diff
@@ -57,6 +57,12 @@ for p in Any[
     @test pdf.(d, support(d)) == p
     @test pdf.(d, 1:k) == p
 
+    @test cf(d, 0) ≈ 1.0
+    @test cf(d, 1) ≈ p' * Distributions.cis.(1:length(p))
```
devmotion, line 61: IIRC `cis` is in Base and exported:

```suggestion
    @test cf(d, 1) ≈ p' * cis.(1:length(p))
```

Possibly a bit nicer (also below) would be to use `dot`:

```suggestion
    @test cf(d, 1) ≈ dot(p, cis.(1:length(p))
```
However, possibly this requires to add a `using LinearAlgebra`.

devmotion (approved):
Thank you!

Issue #1607, opened 2022-08-23
Linear transforms of GenericMvTDist


AlexRobson:
In the univariate case we can do: 

```
julia> μ + TDist(ν)*σ
LocationScale{Float64, Distributions.Continuous, TDist{Float64}}(
μ: 1.0
σ: 5.0
ρ: TDist{Float64}(ν=4.5)
)
```

However, we can't do this in the multivariate case, although we can for MvNormal: 

```
# MvNormal
rand(2,) + Diagonal(rand(2,)) * MvNormal(zeros(2), I(2))
FullNormal(
dim: 2
μ: [0.7341432738772181, 0.8191044540263914]
Σ: [0.08121349453484361 0.0; 0.0 0.6136564428526856]
)
# MvTDist
rand(2,) + Diagonal(rand(2,)) * GenericMvTDist(4.5, zeros(2,), PDMat(I(2)))
ERROR: MethodError: no method matching *(::Diagonal{Float64, Vector{Float64}}, ::GenericMvTDist{Float64, PDMat{Float64, Diagonal{Float64, Vector{Float64}}}, Vector{Float64}})
```

Implementing these isn't difficult, I have them locally, however, I'm now a bit confused whether this is the right approach, or whether I should be using `LocationScale` or `AffineDistribution`. 

Otherwise, i can make an MR with the following: 

```
Base.:+(d::GenericMvTDist, c::AbstractVector) = mvtdist(d.df, d.μ + c, d.Σ)
Base.:+(c::AbstractVector, d::GenericMvTDist) = d + c
Base.:-(d::GenericMvTDist, c::AbstractVector) = mvtdist(d.df, d.μ - c, d.Σ)

Base.:*(B::AbstractMatrix, d::GenericMvTDist) = mvtdist(d.df, B * d.μ, X_A_Xt(d.Σ, B))
Base.:*(B::GenericMvTDist, d::AbstractMatrix) = mvtdist(d.df, B * d.μ, X_A_Xt(d.Σ, B))
```

What's the correct approach here?

PR #1608, opened 2022-08-24
Add transforms for genericmvtdist

AlexRobson:
Solves #1607 

Similar to MvNormal, extends * + and - to the GenericMvTDist. This will allow for transformations of the form: 

```
[1,2] + Diagonal([2,2]) * GenericMvTDist(4.5, zeros(2,), PDMat(I(2)))
Distributions.MvTDist(
df: 4.5
dim: 2
μ: [1.0, 2.0]
Σ: [4.0 0.0; 0.0 4.0]
)
```

However, as discussed in the issue, I was unclear whether to use some of the other machinry in Distributions. Tests are essentially copied from the ones for MvNormal with appropriate changes. 

These changes don't effect the degress of freedom: see e.g. [here](http://users.isy.liu.se/en/rt/roth/student.pdf)

### DIFF
--- a/src/multivariate/mvtdist.jl
+++ b/src/multivariate/mvtdist.jl
@@ -1,7 +1,13 @@
 # Multivariate t-distribution
 
 ## Generic multivariate t-distribution class
-
+"""
+Multivariate Student-T distribution support affine transformations:
+```julia
+d = GenericMvTDist(ν, μ, Σ)
+c + B * d    # == GenericMvTDist(d.df, B * μ + c, B * Σ * B')
+```
+"""
 abstract type AbstractMvTDist <: ContinuousMultivariateDistribution end
 
 struct GenericMvTDist{T<:Real, Cov<:AbstractPDMat, Mean<:AbstractVector} <: AbstractMvTDist
@@ -172,3 +178,11 @@ function _rand!(rng::AbstractRNG, d::GenericMvTDist, x::AbstractMatrix{T}) where
     x .= x ./ sqrt.(y ./ d.df) .+ d.μ
     x
 end
+
+### Affine transformations
+# Base.:+(d::GenericMvTDist, c::AbstractVector) = mvtdist(d.df, d.μ + c, d.Σ)
+# Base.:+(c::AbstractVector, d::GenericMvTDist) = d + c
+# Base.:-(d::GenericMvTDist, c::AbstractVector) = mvtdist(d.df, d.μ - c, d.Σ)
+
+# Base.:*(B::AbstractMatrix, d::GenericMvTDist) = mvtdist(d.df, B * d.μ, X_A_Xt(d.Σ, B))
+# Base.:*(B::GenericMvTDist, d::AbstractMatrix) = mvtdist(d.df, B * d.μ, X_A_Xt(d.Σ, B))
--- a/src/multivariate/mvtdist.jl
+++ b/src/multivariate/mvtdist.jl
@@ -180,9 +180,9 @@ function _rand!(rng::AbstractRNG, d::GenericMvTDist, x::AbstractMatrix{T}) where
 end
 
 ### Affine transformations
-# Base.:+(d::GenericMvTDist, c::AbstractVector) = mvtdist(d.df, d.μ + c, d.Σ)
-# Base.:+(c::AbstractVector, d::GenericMvTDist) = d + c
-# Base.:-(d::GenericMvTDist, c::AbstractVector) = mvtdist(d.df, d.μ - c, d.Σ)
+Base.:+(d::GenericMvTDist, c::AbstractVector) = mvtdist(d.df, d.μ + c, d.Σ)
+Base.:+(c::AbstractVector, d::GenericMvTDist) = d + c
+Base.:-(d::GenericMvTDist, c::AbstractVector) = mvtdist(d.df, d.μ - c, d.Σ)
 
-# Base.:*(B::AbstractMatrix, d::GenericMvTDist) = mvtdist(d.df, B * d.μ, X_A_Xt(d.Σ, B))
-# Base.:*(B::GenericMvTDist, d::AbstractMatrix) = mvtdist(d.df, B * d.μ, X_A_Xt(d.Σ, B))
+Base.:*(B::AbstractMatrix, d::GenericMvTDist) = mvtdist(d.df, B * d.μ, X_A_Xt(d.Σ, B))
+Base.:*(B::GenericMvTDist, d::AbstractMatrix) = mvtdist(d.df, B * d.μ, X_A_Xt(d.Σ, B))
--- a/test/multivariate/mvtdist.jl
+++ b/test/multivariate/mvtdist.jl
@@ -86,4 +86,50 @@ end
     x = rand(X_implicit)
     @test logpdf(X_implicit, x) ≈ logpdf(X_expicit, x)
 end
+
+@testset "MvNormal affine tranformations" begin
+    @testset "moment identities" begin
+        for n in 1:5                       # dimension
+            # distribution
+            μ = randn(n)
+            ν = 4
+            for Σ in (randn(n, n) |> A -> A*A',  # dense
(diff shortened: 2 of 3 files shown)


test/multivariate/mvtdist.jl
```diff
@@ -86,4 +86,50 @@ end
     x = rand(X_implicit)
     @test logpdf(X_implicit, x) ≈ logpdf(X_expicit, x)
 end
+
+@testset "MvNormal affine tranformations" begin
+    @testset "moment identities" begin
```
AlexRobson, line 91: Largely taken from here: 

https://github.com/JuliaStats/Distributions.jl/blob/506ec162912a2edff1b78fc06f274a815f3acd76/test/multivariate/mvnormal.jl#L244

Though if a different form was preferred happy to change them.

src/multivariate/mvtdist.jl
```diff
@@ -172,3 +178,11 @@ function _rand!(rng::AbstractRNG, d::GenericMvTDist, x::AbstractMatrix{T}) where
...
     x
 end
+
+### Affine transformations
+Base.:+(d::GenericMvTDist, c::AbstractVector) = mvtdist(d.df, d.μ + c, d.Σ)
+Base.:+(c::AbstractVector, d::GenericMvTDist) = d + c
+Base.:-(d::GenericMvTDist, c::AbstractVector) = mvtdist(d.df, d.μ - c, d.Σ)
+
+Base.:*(B::AbstractMatrix, d::GenericMvTDist) = mvtdist(d.df, B * d.μ, X_A_Xt(d.Σ, B))
+Base.:*(B::GenericMvTDist, d::AbstractMatrix) = mvtdist(d.df, B * d.μ, X_A_Xt(d.Σ, B))
```
AlexRobson, line 188: From https://github.com/JuliaStats/Distributions.jl/blob/master/src/multivariate/mvtdist.jl#L72 this should support different types of abstractpdmat, and not just pdmat

src/multivariate/mvtdist.jl
```diff
@@ -1,7 +1,13 @@
...
 
 ## Generic multivariate t-distribution class
-
+"""
+Multivariate Student-T distribution support affine transformations:
+```julia
+d = GenericMvTDist(ν, μ, Σ)
+c + B * d    # == GenericMvTDist(d.df, B * μ + c, B * Σ * B')
+```
+"""
```
devmotion, line 10: That should be moved to `GenericMvTDist` I think? (And be part of a larger more detailed docstring, similar to other distributions - seems it is missing completely currently.)

src/multivariate/mvtdist.jl
devmotion, line 188: I think this should be removed:
```suggestion
```

test/multivariate/mvtdist.jl
```diff
@@ -86,4 +86,50 @@ end
     x = rand(X_implicit)
     @test logpdf(X_implicit, x) ≈ logpdf(X_expicit, x)
 end
+
+@testset "MvNormal affine tranformations" begin
```
devmotion, line 90:
```suggestion
@testset "GenericMvTDist affine tranformations" begin
```

devmotion (commented):
IMO it's reasonable to add these optimizations. `AffineDistribution` is supposed to be a generic fallback but specializations/optimizations are desired if they exist.

src/multivariate/mvtdist.jl
```diff
@@ -72,11 +101,14 @@ IsoTDist(df::Real, d::Int, σ::Real) = GenericMvTDist(df, ScalMat(d, abs2(σ)))
...
 mvtdist(df::Real, d::Int, σ::Real) = GenericMvTDist(df, μ, ScalMat(d, abs2(σ)))
-mvtdist(df::Real, μ::Vector, σ::Vector) = GenericMvTDist(df, μ, PDiagMat(abs2.(σ)))
-mvtdist(df::Real, μ::Vector, Σ::Matrix) = GenericMvTDist(df, μ, PDMat(Σ))
-mvtdist(df::Real, Σ::Matrix) = GenericMvTDist(df, PDMat(Σ))
+mvtdist(df::Real, μ::AbstractVector, σ::Vector) = GenericMvTDist(df, μ, PDiagMat(abs2.(σ)))
+mvtdist(df::Real, μ::AbstractVector, Σ::AbstractMatrix) = GenericMvTDist(df, μ, PDMat(Σ))
+mvtdist(df::Real, Σ::AbstractMatrix) = GenericMvTDist(df, PDMat(Σ))
+
+mvtdist(df::Real, μ::AbstractVector, Σ::UniformScaling{<:Real}) = GenericMvTDist(df, μ, ScalMat(length(μ), Σ.λ))
+
```
devmotion, line 111: This is not part of the API and should probably be removed completely? I don't think there's any advantage over just using constructors. And we don't want to support vectors or scalars for the scale matrix anyway (similar to the deprecations in `MvNormal` etc.) as it only causes confusion (is it squared or not?).

AlexRobson: Sorry, can you clarify this comment - are you referring to the `mvtdist`? functions? And they should be removed?

devmotion: Well, not in this PR as it is not about these functions. But in general, yes. We should just follow the same approach as in `MvNormal` and other distributions - we should just call the constructor. And we should not operate with scalars or vectors of scale matrices.

AlexRobson: Actually here the only reason I worked with these utilities rahter than calling the constructor directly was to handle the densification of the output of `X_A_Xt`. Although the output (for these test cases) is a PDMatable, th X_A_Xt call densifies it. In retrospect, defining: 

```
GenericMvTDist(df::Real, μ::AbstractVector, Σ::AbstractMatrix) = GenericMvTDist(df, μ, PDMat(Σ))
```

is more appropriate here, rather than using these helper functions. I'll follow up with this change and these changes should disappear from the change set. 

But agree with your point - these changes make sense.

src/multivariate/mvtdist.jl
```diff
@@ -30,6 +59,8 @@ function GenericMvTDist(df::Real, Σ::AbstractPDMat)
     GenericMvTDist(df, Zeros{R}(size(Σ, 1)), Σ)
 end
 
+GenericMvTDist(df::Real, μ::AbstractVector, Σ::AbstractMatrix) = GenericMvTDist(df, μ, PDMat(Σ))
```
AlexRobson, line 62: New constructor added so no longer using the `mvtdist` ones which should be removed.

AlexRobson:
This should be ready I believe. 

I could also include a deprecation to `mvtdist` because, as you point out, we should be using the constructor. Thanks for your review comments!

src/multivariate/mvtdist.jl
devmotion: That's a bit suboptimal as it will create a dense representation (in `PDMat`) also for `UniformScaling`, `Diagonal` matrices, etc. that should be converted to `ScalMat` and `PDiagMat`. I guess we want the same logic here as in e.g. `MvNormal` that creates these optimized `AbstractPDMat` types (and probably in the long run this should be supported by PDMats e.g. via a `AbstractPDMat` constructor).

AlexRobson: Sorry for the delay - I've been away. 

Good point! I've added these in along with some tests.

Crown421:
What is the status of this PR? Can I help get it merged? 
I was part way through implementing my own MvTDist before noticing that there is something implemented but not documented.

AlexRobson:
I havne't looked at this for a while. That said, last I left it, I thought it was ready. I've moved well out of the context of what I was looking at originally, but I can pick it up again - looks like it rebased cleanly so there isn't any code drift to deal with on main.

Crown421:
It also seemed pretty complete to me, but I did not give it a super close look. I can have a more thorough on the weekend, and then it will be an issue of fixing the CI failures, and lastly getting maintainer attention to merge.

PR #1609, opened 2022-08-24
add cgf

jw3126:


Project.toml
```diff
@@ -8,6 +8,7 @@ ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 DensityInterface = "b429d917-457f-4dbc-8f4c-0cc954292b1d"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+LogExpFunctions = "2ab3a3ac-af41-5b50-aa03-7779005ae688"
```
jw3126, line 11: Note that this is a transitive dependency of Distributions.jl via StatsBase anyway.

src/cgf.jl
```diff
@@ -0,0 +1,90 @@
...
+    -ν/2*log1p(-2*t)
+end
+function cgf(d::Chisq, t)
+    ν, = params(d)
+    cgf_Chisq(ν,t)
+end
+function cgf(d::NoncentralChisq, t)
+    ν, λ = params(d)
+    λ*t/(1-2t) + cgf_Chisq(ν,t)
+end
```
devmotion, line 90:
```suggestion
function cgf(d::Chisq, t)
    ν = dof(d)
    return -ν/2 * log1p(-2*t)
end

function cgf(d::NoncentralChisq, t)
    ν, λ = params(d)
    return λ*t/(1 - 2*t) + cgf(Chisq{typeof(ν)}(ν), t)
end
```

devmotion (commented):
Thanks!

Since the function is so closely related to `mgf`, I think it would be good to move the generic definition/docstring just below `mgf` in https://github.com/JuliaStats/Distributions.jl/blob/e71089949396ba2d5c7ed2042b80576e483e9049/src/univariates.jl#L276 and the definitions for different distributions next to the definition of `mgf` in their respective files. Similarly, probably we should then also move the tests to the respective distributions. Could you change this?

And could you add it to the documentation as well (again probably best next to `mgf`)?

src/cgf.jl
```diff
@@ -0,0 +1,90 @@
...
+
+function cgf(d::Uniform, t)
+    a,b = params(d)
+    # log((exp(t*b) - exp(t*a))/ (t*(b-a)))
+    x = t*(b-a)
+    if abs(x) < sqrt(eps(float(one(x))))
+        t*a + log(expfd0_taylor(x))
+    else
+        logsubexp(t*b, t*a) - log(abs(x))
+    end
```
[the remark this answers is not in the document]
jw3126: Not that I am aware of, I derived this myself. Should I add some comments about the derivation or do you have an idea for an alternative more "official" implementation?

jw3126:
> Thanks!
> 
> Since the function is so closely related to `mgf`, I think it would be good to move the generic definition/docstring 

There could be a fallback `cgf = log mgf` should we have this fallback or not? I think there are arguments either way.

> just below `mgf` in
> 
> https://github.com/JuliaStats/Distributions.jl/blob/e71089949396ba2d5c7ed2042b80576e483e9049/src/univariates.jl#L276
> and the definitions for different distributions next to the definition of `mgf` in their respective files. Similarly, probably we should then also move the tests to the respective distributions. Could you change this?

Sure sounds good.
> 
> And could you add it to the documentation as well (again probably best next to `mgf`)?

jw3126:
> Similarly, probably we should then also move the tests to the respective distributions. 

Actually, I think it might be nicer to keep the `cgf` tests in a single place. Otherwise one either needs to duplicate quite a bit of code.

src/univariate/continuous/uniform.jl
```diff
@@ -101,6 +101,27 @@ function mgf(d::Uniform, t::Real)
...
+function expfd0_taylor(x)
+    # taylor series of (exp(x) - 1) / x
+    T = typeof(x)
+    a0 = inv(T(1))
+    a1 = inv(T(2))
+    a2 = inv(T(6))
+    a3 = inv(T(24))
+    a4 = inv(T(120))
+    @evalpoly(x, a0, a1, a2, a3, a4)
+end
```
devmotion, line 113: I am worried that this is not sufficiently accurate for non-standard (i.e. non-`Float64`) number types.

src/univariate/discrete/binomial.jl
```diff
@@ -147,6 +147,10 @@ function mgf(d::Binomial, t::Real)
     n, p = params(d)
     (one(p) - p + p * exp(t)) ^ n
 end
+function cgf(d::Binomial, t)
+    n,p = params(d)
+    n*cgf(Bernoulli(p),t)
```
devmotion, line 152: To avoid checks or promotions of the argument:
```suggestion
    n * cgf(Bernoulli{typeof(p)}(p), t)
```

src/univariate/discrete/binomial.jl
```diff
@@ -147,6 +147,10 @@ function mgf(d::Binomial, t::Real)
     n, p = params(d)
     (one(p) - p + p * exp(t)) ^ n
 end
+function cgf(d::Binomial, t)
+    n,p = params(d)
```
devmotion, line 151:
```suggestion
    n, p = params(d)
```

src/univariate/discrete/dirac.jl
```diff
@@ -50,8 +50,10 @@ logccdf(d::Dirac, x::Real) = x < d.value ? 0.0 : isnan(x) ? NaN : -Inf
 quantile(d::Dirac{T}, p::Real) where {T} = 0 <= p <= 1 ? d.value : T(NaN)
 
 mgf(d::Dirac, t) = exp(t * d.value)
+cgf(d::Dirac, t) = t*d.value
 cf(d::Dirac, t) = cis(t * d.value)
 
+
```
devmotion, line 56:
```suggestion
```

src/univariate/discrete/geometric.jl
```diff
@@ -127,6 +127,11 @@ function laplace_transform(d::Geometric, t)
     p / (p - (1 - p) * expm1(-t))
 end
 mgf(d::Geometric, t::Real) = laplace_transform(d, -t)
+function cgf(d::Geometric, t)
+    p, = params(d)
```
devmotion, line 131:
```suggestion
    p = succprob(d)
```

src/univariate/discrete/geometric.jl
```diff
@@ -127,6 +127,11 @@ function laplace_transform(d::Geometric, t)
...
+    # log(p / (1 - (1-p) * exp(t)))
+    log(p) - logsubexp(0, t + log1p(-p))
```
devmotion, line 133:
```suggestion
    log(p) - log1mexp(t + log1p(-p))
```

src/univariate/discrete/negativebinomial.jl
```diff
@@ -132,6 +132,10 @@ function laplace_transform(d::NegativeBinomial, t)
 end
 
 mgf(d::NegativeBinomial, t::Real) = laplace_transform(d, -t)
+function cgf(d::NegativeBinomial, t)
+    r,p = params(d)
```
devmotion, line 136:
```suggestion
    r, p = params(d)
```

src/univariate/discrete/negativebinomial.jl
```diff
@@ -132,6 +132,10 @@ function laplace_transform(d::NegativeBinomial, t)
...
+    r*cgf(Geometric(p),t)
```
devmotion, line 137:
```suggestion
    r * cgf(Geometric{typeof(p)}(p), t)
```

src/univariate/continuous/uniform.jl
jw3126: Note that we only use this internal function it in the region ` sqrt(eps(float(one(x))))`. So the `x^4` is less than `eps(x)^2`. In fact I think ` @evalpoly(x, a0, a1, a2)` would already reach machine precision in this range. I included the higher order terms only to also get automatic differentiation accurate.

src/univariate/discrete/negativebinomial.jl
jw3126: Thanks a lot I was not aware of this trick!

devmotion:
> or factor the tests into a function which makes them slightly less straightforward to read.

Testing utilities of the interface are a long-term goal anyway. Basically a generalization of what's available for `MvNormal` here: https://github.com/JuliaStats/Distributions.jl/blob/master/src/test_utils.jl We could add a `test_cgf` there (which at some point should be integrated into a larger set of tests for univariate distributions) but without any ForwardDiff dependency.

src/univariate/continuous/uniform.jl
devmotion: Sure, I assumed that you tested it with Float64 and that it is sufficiently accurate there. But it might still be too inaccurate for e.g. higher-order differentiation. Or BigFloat. Or Float128. And for Float16 or Float32 it might be a bit inefficient since possibly one could remove terms. Hence I think usually such series approximations should either be computed dynamically (e.g. until numerical convergence) or be restricted to the specific type that they were designed for.

src/univariate/continuous/uniform.jl
```diff
@@ -101,6 +101,36 @@ function mgf(d::Uniform, t::Real)
...
     exp(v) * (sinh(u) / u)
 end
+function cgf_uniform_around_zero_kernel(x)
+    # taylor series of (exp(x) - x - 1) / x
+    T = typeof(x)
+    a0 = inv(T(2))
+    a1 = inv(T(6))
+    a2 = inv(T(24))
+    a3 = inv(T(120))
+    x*@evalpoly(x, a0, a1, a2, a3)
```
jw3126, line 111: I changed a bit what to put in the kernel and I believe this is sufficiently accurate for any floating point number type in the region where we apply it. 
The region of applicability is `abs(x) < sqrt(eps(T))` and I will assume `x > 0` for readability. The error from truncating the Taylor series is `err(x) ~ O(x^5)`, so we find:
```
err(x) 
~ O(x^5)
~ O(x^2) * O(x) * O(x^2) 
< O(x^2) * O(x) * sqrt(eps(T))^2 # region assumption
~ O(x^2) * O(eps(x))                   # eps(x)  ≈ x * eps(T)`
```
Since `cgf_uniform_around_zero_kernel(x) = x + O(x^2)` this error is way below machine precision `eps(x + O(x^2))`. This argument is not specific to `Float64`, it only assumes `eps(x)  ≈ x * eps(T)` which should be true for any floating point type I think. See also tests against `BigFloat` below.
First and second order derivatives are also fine by similar argument. 
Of course, higher order derivatives will at some point become zero since this is a finite degree polynomial.

So I would like to leave the kernel as is. If people want to optimize it further or better support higher order AD that can be a future PR, what do you think @devmotion ?

test/univariate/continuous/exponential.jl
```diff
@@ -0,0 +1,4 @@
+
+test_cgf(Exponential(1), Any[0.9, -1, -100f0, -1e6])
```
devmotion, line 2: Maybe it would simpler to use tuples here?
```suggestion
test_cgf(Exponential(1), (0.9, -1, -100f0, -1e6))
```

src/univariate/continuous/uniform.jl
devmotion: I'm still wondering if there's another approach that would avoid that we have to work with the approximations and branching in Distributions directly (instead of reusing existing functionality in LogExpFunctions, SpecialFunctions etc.). But I think it's the best we can do for now at least.

jw3126: I agree, it would be nice if we could pull this function in from somewhere else. I did look through LogExpFunctions before coding this, but could not see a way to use LogExpFunctions here. For SpecialFunctions I don't have high hopes either, since our function is not really special, but expressible in terms of `exp` and polynomials.

devmotion (approved):
Looks good to me, thanks a lot for the work on this PR!

devmotion (approved):
Looks good to me, thanks a lot for the work on this PR!

PR #1610, opened 2022-08-27
Return -Inf in logpdf of LKJCholesky when out of support

mohamed82008:
Given the docstrings of `DomainError` and `ArgumentError` shown below, I think a `DomainError` makes more sense here.

```julia
  DomainError(val)
  DomainError(val, msg)

  The argument val to a function or constructor is outside the valid domain.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> sqrt(-1)
  ERROR: DomainError with -1.0:
  sqrt will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).
  Stacktrace:
  [...]
```

```julia
  ArgumentError(msg)

  The parameters to a function call do not match a valid signature. Argument msg is a descriptive error string.
```

### DIFF
--- a/src/cholesky/lkjcholesky.jl
+++ b/src/cholesky/lkjcholesky.jl
@@ -136,7 +136,7 @@ function logkernel(d::LKJCholesky, R::LinearAlgebra.Cholesky)
 end
 
 function logpdf(d::LKJCholesky, R::LinearAlgebra.Cholesky)
-    insupport(d, R) || throw(DomainError(R, "provided point is not in the support"))
+    insupport(d, R) || return oftype(_logpdf(d, LinearAlgebra.Cholesky(zeros(eltype(R.factors), d.d, d.d), R.uplo, R.info)), -Inf)
     return _logpdf(d, R)
 end
 
--- a/src/cholesky/lkjcholesky.jl
+++ b/src/cholesky/lkjcholesky.jl
@@ -136,7 +136,7 @@ function logkernel(d::LKJCholesky, R::LinearAlgebra.Cholesky)
 end
 
 function logpdf(d::LKJCholesky, R::LinearAlgebra.Cholesky)
-    insupport(d, R) || throw(ArgumentError("provided point is not in the support"))
+    insupport(d, R) || throw(DomainError(R, "provided point is not in the support"))
     return _logpdf(d, R)
 end
 
--- a/src/cholesky/lkjcholesky.jl
+++ b/src/cholesky/lkjcholesky.jl
@@ -136,8 +136,8 @@ function logkernel(d::LKJCholesky, R::LinearAlgebra.Cholesky)
 end
 
 function logpdf(d::LKJCholesky, R::LinearAlgebra.Cholesky)
-    insupport(d, R) || return oftype(_logpdf(d, LinearAlgebra.Cholesky(zeros(eltype(R.factors), d.d, d.d), R.uplo, R.info)), -Inf)
-    return _logpdf(d, R)
+    lp = _logpdf(d, R)
+    return insupport(d, R) ? lp : oftype(lp, -Inf)
 end
 
 _logpdf(d::LKJCholesky, R::LinearAlgebra.Cholesky) = logkernel(d, R) + d.logc0
--- a/src/matrix/lkj.jl
+++ b/src/matrix/lkj.jl
@@ -195,8 +195,6 @@ function lkj_onion_loginvconst(d::Integer, η::T) where {T <: Real}
     for k in 3:d - 1
         sumlogs += (h * k) * logπ + loggamma(η + h * (d - 1 - k))
     end
-    α = η + h * d - 1
-    loginvconst = (2η + d - 3)*logtwo + logbeta(α, α) + sumlogs - (d - 2) * loggamma(η + h * (d - 1))
     return loginvconst
 end
 
--- a/src/matrix/lkj.jl
+++ b/src/matrix/lkj.jl
@@ -188,7 +188,8 @@ end
 
 function lkj_onion_loginvconst(d::Integer, η::T) where {T <: Real}
     #  Equation (17) in LKJ (2009 JMA)
-    h = convert(float(T), 0.5)
+    T = float(Base.promote_typeof(d, η))
+    h = T(1//2)
     sumlogs = logπ + loggamma(η + h * (d - 3))
     sumlogs = ifelse(d >= 3, sumlogs, zero(sumlogs))
     for k in 3:d - 1
--- a/src/matrix/lkj.jl
+++ b/src/matrix/lkj.jl
@@ -100,12 +100,13 @@ params(d::LKJ) = (d.d, d.η)
 #  -----------------------------------------------------------------------------
 
 function lkj_logc0(d::Integer, η::Real)
+    T = float(Base.promote_typeof(d, η))
     d > 1 || return zero(η)
     if isone(η)
         if iseven(d)
-            logc0 = -lkj_onion_loginvconst_uniform_even(d)
+            logc0 = -lkj_onion_loginvconst_uniform_even(d, T)
         else
-            logc0 = -lkj_onion_loginvconst_uniform_odd(d)
+            logc0 = -lkj_onion_loginvconst_uniform_odd(d, T)
         end
     else
         logc0 = -lkj_onion_loginvconst(d, η)
@@ -198,23 +199,27 @@ function lkj_onion_loginvconst(d::Integer, η::Real)
     return loginvconst
 end
 
-function lkj_onion_loginvconst_uniform_odd(d::Integer)
+function lkj_onion_loginvconst_uniform_odd(d::Integer, ::Type{T} = Float64) where {T}
     #  Theorem 5 in LKJ (2009 JMA)
-    sumlogs = 0.0
+    sumlogs = zero(T)
     for k in 1:div(d - 1, 2)
-        sumlogs += loggamma(2k)
+        sumlogs += loggamma(T(2k))
     end
-    loginvconst = 0.25(d^2 - 1)*logπ + sumlogs - 0.25(d - 1)^2*logtwo - (d - 1)*loggamma(0.5(d + 1))
+    h = T(1//2)
+    q = T(1//4)
+    loginvconst = q*(d^2 - 1)*logπ + sumlogs - q*(d - 1)^2*logtwo - (d - 1)*loggamma(h*(d + 1))
     return loginvconst
 end
 
-function lkj_onion_loginvconst_uniform_even(d::Integer)
+function lkj_onion_loginvconst_uniform_even(d::Integer, ::Type{T} = Float64) where {T}
     #  Theorem 5 in LKJ (2009 JMA)
-    sumlogs = 0.0
+    sumlogs = zero(T)
     for k in 1:div(d - 2, 2)
-        sumlogs += loggamma(2k)
+        sumlogs += loggamma(T(2k))
     end
-    loginvconst = 0.25d*(d - 2)*logπ + 0.25(3d^2 - 4d)*logtwo + d*loggamma(0.5d) + sumlogs - (d - 1)*loggamma(d)
+    h = T(1//2)
+    q = T(1//4)
+    return q*d*(d - 2)*logπ + q*(3d^2 - 4d)*logtwo + d*loggamma(h*d) + sumlogs - (d - 1)*loggamma(T(d))
 end
 
 function lkj_vine_loginvconst(d::Integer, η::Real)
--- a/src/matrix/lkj.jl
+++ b/src/matrix/lkj.jl
@@ -199,27 +199,24 @@ function lkj_onion_loginvconst(d::Integer, η::Real)
     return loginvconst
 end
 
-function lkj_onion_loginvconst_uniform_odd(d::Integer, ::Type{T} = Float64) where {T}
+function lkj_onion_loginvconst_uniform_odd(d::Integer, ::Type{T}) where {T <: Real}
     #  Theorem 5 in LKJ (2009 JMA)
-    sumlogs = zero(T)
-    for k in 1:div(d - 1, 2)
-        sumlogs += loggamma(T(2k))
-    end
     h = T(1//2)
-    q = T(1//4)
-    loginvconst = q*(d^2 - 1)*logπ + sumlogs - q*(d - 1)^2*logtwo - (d - 1)*loggamma(h*(d + 1))
+    loginvconst = (d - 1) * ((d + 1) * (T(logπ) / 4) - (d - 1) * (T(logtwo) / 4) - loggamma(h * (d + 1)))
+    for k in 2:2:(d - 1)
+        loginvconst += loggamma(T(k))
+    end
     return loginvconst
 end
 
-function lkj_onion_loginvconst_uniform_even(d::Integer, ::Type{T} = Float64) where {T}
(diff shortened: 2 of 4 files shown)


devmotion:
We should just return `-Inf` outside of the support, as we do for the other distributions.

mohamed82008:
How would you make it type stable?

mohamed82008:
I noticed some implementations are not type stable.
```julia
_logpdf(d::MvLogNormal, x::AbstractVecOrMat{T}) where {T<:Real} = insupport(d, x) ? (_logpdf(d.normal, log.(x)) - sum(log.(x))) : -Inf
```

devmotion:
Yes, this is ancient code that should be improved. Most univariate distributions should be fixed by now. The common and safest approach is to just compute everything but eg clamp inputs if needed, and then return `oftype(result, -Inf)` outside of the support.

mohamed82008:
Is `insupport(d, R) || _logpdf(d, LinearAlgebra.Cholesky(zero(R.data), R.uplo, R.info))` reasonable?

mohamed82008:
or the -Inf of that output type

mohamed82008:
I pushed a commit that returns a -Inf of the correct type.

src/cholesky/lkjcholesky.jl
```diff
@@ -136,7 +136,7 @@ function logkernel(d::LKJCholesky, R::LinearAlgebra.Cholesky)
 end
 
 function logpdf(d::LKJCholesky, R::LinearAlgebra.Cholesky)
-    insupport(d, R) || throw(ArgumentError("provided point is not in the support"))
+    insupport(d, R) || return oftype(_logpdf(d, LinearAlgebra.Cholesky(zeros(eltype(R.factors), d.d, d.d), R.uplo, R.info)), -Inf)
     return _logpdf(d, R)
```
devmotion, line 140: The only problem I can think of here is that it does not necessarily preserve the type of `R.factors` (e.g., if it has some special structure or is some special matrix type).

It's not clear to me though why constructing a new `Cholesky` is needed here at all. `_logpdf` just sums `logkernel` and the `logc0` constant (independent of the input), and the `logkernel` function just above should work (or at least not error) even if `Cholesky` has the wrong size. The only problem there would be a `Cholesky` with negative values on the diagonal of `R.factors`, and that should not be possible (if one does not circumvent the error checks at construction time).

So I think maybe (untested) we could just use
```suggestion
    lp = _logpdf(d, R)
    return insupport(d, R) ? lp : oftype(lp, -Inf) 
```

test/cholesky/lkjcholesky.jl
```diff
@@ -141,6 +141,7 @@ using FiniteDifferences
             @test insupport(LKJCholesky(40, 2, 'U'), cholesky(rand(LKJ(40, 2))))
             @test insupport(LKJCholesky(40, 2), cholesky(rand(LKJ(40, 2))))
             @test !insupport(LKJCholesky(40, 2), cholesky(rand(LKJ(41, 2))))
+            @test logpdf(LKJCholesky(40, 2), cholesky(rand(LKJ(41, 2)))) == -Inf
```
devmotion, line 144: We should make sure that everything is indeed type stable:
```suggestion
            @test @inferred(logpdf(LKJCholesky(40, 2), cholesky(rand(LKJ(41, 2))))) == -Inf
```

In fact, however, most distributions are type stable for `Float64` though, even if implemented incorrectly, so we should add some more tests with other types (`Float32` maybe?) as well.

src/cholesky/lkjcholesky.jl
mohamed82008: Done. It seems to work fine.

test/cholesky/lkjcholesky.jl
mohamed82008: Done. I had to generalise `lkj_onion_loginvconst `.

src/matrix/lkj.jl
```diff
@@ -186,14 +186,15 @@ end
 #  for 1 / c₀, even if they say that it is not.
 #  -----------------------------------------------------------------------------
 
-function lkj_onion_loginvconst(d::Integer, η::Real)
+function lkj_onion_loginvconst(d::Integer, η::T) where {T <: Real}
     #  Equation (17) in LKJ (2009 JMA)
     sumlogs = zero(η)
```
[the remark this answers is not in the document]
mohamed82008: I pushed a reasonable fix imo.

src/matrix/lkj.jl
```diff
@@ -186,14 +186,16 @@ end
 #  for 1 / c₀, even if they say that it is not.
 #  -----------------------------------------------------------------------------
 
-function lkj_onion_loginvconst(d::Integer, η::Real)
+function lkj_onion_loginvconst(d::Integer, η::T) where {T <: Real}
```
devmotion, line 189: The functions below `lkj_onion_loginvconst_uniform_even` and `lkj_onion_loginvconst_uniform_odd` have to be fixed as well, otherwise `lkj_logc0` becomes type unstable it seems.

src/matrix/lkj.jl
```diff
@@ -186,14 +186,16 @@ end
...
     #  Equation (17) in LKJ (2009 JMA)
-    sumlogs = zero(η)
-    for k in 2:d - 1
-        sumlogs += 0.5k*logπ + loggamma(η + 0.5(d - 1 - k))
+    h = convert(float(T), 0.5)
```
devmotion, line 191: There are still problems below with e.g. `(2η + d - 3)*logtwo` - that will be of type `Float64` if `η` is an integer. Hence some more changes are needed. It also seems we should take into account the type of `d` here - if it is eg. a `BigInt` then we would want to work with `BigFloat` since e.g. `::BigInt * ::Float64 -> ::BigFloat`.
Maybe we need to convert irrationals etc. to something like
```suggestion
    T = float(Base.promote_typeof(d, η))
    h = T(1//2)
```
(`T` in the function signature could be reverted to `Real`)

src/matrix/lkj.jl
```diff
@@ -186,14 +186,16 @@ end
...
+    sumlogs = logπ + loggamma(η + h * (d - 3))
+    sumlogs = ifelse(d >= 3, sumlogs, zero(sumlogs))
+    for k in 3:d - 1
+        sumlogs += (h * k) * logπ + loggamma(η + h * (d - 1 - k))
```
devmotion, line 195: As I wrote in my previous comment, I think it would be simpler to just initialize with the other terms in `loginvconst` below. Moreover, it would be more efficient to move the `logπ` computation out of the loop. With the definition of `T` suggested above one could write
```suggestion
    α = η + h * d - 1
    loginvconst = (2*η + d - 3)*oftype(T, logtwo) + (oftype(T, logπ) / 4) * (d * (d - 1) - 2) + logbeta(α, α) - (d - 2) * loggamma(η + h * (d - 1))
    for k in 2:(d - 1)
        loginvconst += loggamma(η + h * (d - 1 - k))
```

src/matrix/lkj.jl
```diff
@@ -186,14 +186,16 @@ end
...
     end
-    α = η + 0.5d - 1
-    loginvconst = (2η + d - 3)*logtwo + logbeta(α, α) + sumlogs - (d - 2) * loggamma(η + 0.5(d - 1))
+    α = η + h * d - 1
+    loginvconst = (2η + d - 3)*logtwo + logbeta(α, α) + sumlogs - (d - 2) * loggamma(η + h * (d - 1))
```
devmotion, line 198: See above:
```suggestion
```

src/matrix/lkj.jl
```diff
@@ -186,14 +186,15 @@ end
 #  for 1 / c₀, even if they say that it is not.
 #  -----------------------------------------------------------------------------
 
-function lkj_onion_loginvconst(d::Integer, η::Real)
+function lkj_onion_loginvconst(d::Integer, η::T) where {T <: Real}
```
devmotion, line 189:
```suggestion
function lkj_onion_loginvconst(d::Integer, η::Real)
```

src/matrix/lkj.jl
```diff
@@ -188,12 +188,13 @@ end
 
 function lkj_onion_loginvconst(d::Integer, η::Real)
     #  Equation (17) in LKJ (2009 JMA)
-    sumlogs = zero(η)
-    for k in 2:d - 1
-        sumlogs += 0.5k*logπ + loggamma(η + 0.5(d - 1 - k))
+    T = float(Base.promote_typeof(d, η))
+    h = T(1//2)
+    α = η + h * d - 1
+    loginvconst = (2*η + d - 3)*oftype(T, logtwo) + (oftype(T, logπ) / 4) * (d * (d - 1) - 2) + logbeta(α, α) - (d - 2) * loggamma(η + h * (d - 1))
```
devmotion, line 194:
```suggestion
    loginvconst = (2*η + d - 3)*T(logtwo) + (T(logπ) / 4) * (d * (d - 1) - 2) + logbeta(α, α) - (d - 2) * loggamma(η + h * (d - 1))
```

mohamed82008:
Tests pass now.

devmotion:
Looks good, I think the only thing left to do is to address https://github.com/JuliaStats/Distributions.jl/pull/1610#discussion_r956777053 and add corresponding tests with eta = 1 and d even and d odd.

src/matrix/lkj.jl
```diff
@@ -188,32 +189,37 @@ end
...
+    for k in 2:(d - 1)
+        loginvconst += loggamma(η + h * (d - 1 - k))
     end
-    α = η + 0.5d - 1
-    loginvconst = (2η + d - 3)*logtwo + logbeta(α, α) + sumlogs - (d - 2) * loggamma(η + 0.5(d - 1))
     return loginvconst
 end
 
-function lkj_onion_loginvconst_uniform_odd(d::Integer)
+function lkj_onion_loginvconst_uniform_odd(d::Integer, ::Type{T} = Float64) where {T}
```
devmotion, line 202: Do we really want a default value here? The function is neither exported nor part of the API, so we don't need one for backward compatibility.
```suggestion
function lkj_onion_loginvconst_uniform_odd(d::Integer, ::Type{T}) where {T<:Real}
```

src/matrix/lkj.jl
```diff
@@ -188,32 +189,37 @@ end
...
-    sumlogs = 0.0
+    sumlogs = zero(T)
     for k in 1:div(d - 1, 2)
-        sumlogs += loggamma(2k)
+        sumlogs += loggamma(T(2k))
     end
-    loginvconst = 0.25(d^2 - 1)*logπ + sumlogs - 0.25(d - 1)^2*logtwo - (d - 1)*loggamma(0.5(d + 1))
+    h = T(1//2)
+    q = T(1//4)
+    loginvconst = q*(d^2 - 1)*logπ + sumlogs - q*(d - 1)^2*logtwo - (d - 1)*loggamma(h*(d + 1))
```
devmotion, line 210: Similar to the implementation above we should ensure that be more careful with irrationals and can remove `sumlogs` and the multiplications with `2`:
```suggestion
    h = T(1//2)
    loginvconst = (d - 1) * ((d + 1) * (T(logπ) / 4) - (d - 1) * (T(logtwo) / 4) - loggamma(h * (d + 1)))
    for k in 2:2:(d - 1)
        loginvconst += loggamma(T(k))
    end
```

src/matrix/lkj.jl
```diff
@@ -188,32 +189,37 @@ end
...
     return loginvconst
 end
 
-function lkj_onion_loginvconst_uniform_even(d::Integer)
+function lkj_onion_loginvconst_uniform_even(d::Integer, ::Type{T} = Float64) where {T}
```
devmotion, line 214:
```suggestion
function lkj_onion_loginvconst_uniform_even(d::Integer, ::Type{T}) where {T<:Real}
```

src/matrix/lkj.jl
```diff
@@ -188,32 +189,37 @@ end
...
-    sumlogs = 0.0
+    sumlogs = zero(T)
     for k in 1:div(d - 2, 2)
-        sumlogs += loggamma(2k)
+        sumlogs += loggamma(T(2k))
     end
-    loginvconst = 0.25d*(d - 2)*logπ + 0.25(3d^2 - 4d)*logtwo + d*loggamma(0.5d) + sumlogs - (d - 1)*loggamma(d)
+    h = T(1//2)
+    q = T(1//4)
+    return q*d*(d - 2)*logπ + q*(3d^2 - 4d)*logtwo + d*loggamma(h*d) + sumlogs - (d - 1)*loggamma(T(d))
```
devmotion, line 222: Same here:
```suggestion
    h = T(1//2)
    loginvconst = d * ((d - 2) * (T(logπ) / 4) + (3 * d - 4) * (T(logtwo) / 4) + loggamma(h * d)) - (d - 1) * loggamma(T(d))
    for k in 2:2:(d - 2)
        loginvconst += loggamma(k)
    end
    return loginvconst
end
```

test/cholesky/lkjcholesky.jl
```diff
@@ -135,12 +135,14 @@ using FiniteDifferences
             @test m isa Cholesky{eltype(d)}
             @test Matrix(m) ≈ I
         end
-        @test_broken partype(LKJCholesky(2, 4f0)) <: Float32
+        @test partype(LKJCholesky(2, 4f0)) <: Float32
```
devmotion, line 138: We should add tests for the other cases as well, e.g.,
```suggestion
        for (d, η) in ((2, 4), (2, 1), (3, 1)), T in (Float32, Float64)
            @test @inferred(partype(LKJCholesky(d, T(η)))) === T
        end
```

test/cholesky/lkjcholesky.jl
```diff
@@ -135,12 +135,14 @@ using FiniteDifferences
...
 
         @testset "insupport" begin
             @test insupport(LKJCholesky(40, 2, 'U'), cholesky(rand(LKJ(40, 2))))
             @test insupport(LKJCholesky(40, 2), cholesky(rand(LKJ(40, 2))))
             @test !insupport(LKJCholesky(40, 2), cholesky(rand(LKJ(41, 2))))
+            @test @inferred(logpdf(LKJCholesky(40, 2), cholesky(rand(LKJ(41, 2))))) === -Inf
+            @test @inferred(logpdf(LKJCholesky(40, 2f0), cholesky(Float32.(rand(LKJ(41, 2)))))) === -Inf32
```
devmotion, line 145: Similarly, here we should also test:
```suggestion
           for (d, η) in ((2, 4), (2, 1), (3, 1)), T in (Float32, Float64)
                @test @inferred(logpdf(LKJCholesky(40, T(2)), cholesky(T.(rand(LKJ(41, 2)))))) === T(-Inf)
            end
```

src/matrix/lkj.jl
mohamed82008: Done. Thanks.

src/matrix/lkj.jl
mohamed82008: Done. Thanks.

src/matrix/lkj.jl
mohamed82008: Done. Thanks.

src/matrix/lkj.jl
mohamed82008: Done. Thanks.

devmotion (approved):
LGTM, thanks!

mohamed82008:
Thank you for all your inputs.

PR #1613, opened 2022-09-06
Docs: Creating matrix variate

mschauer:
Is `rand(::AbstractRNG,...)` the right one here?

### DIFF
--- a/docs/src/extends.md
+++ b/docs/src/extends.md
@@ -154,13 +154,13 @@ It is also recommended that one also implements the following statistics functio
 - [`entropy(d::MultivariateDistribution)`](@ref)
 - [`cov(d::MultivariateDistribution)`](@ref)
 
-## Create a Matrix-variate Distribution
+## Create a matrix-variate Distribution
 
-A multivariate distribution type should be defined as a subtype of `DiscreteMatrixDistribution` or `ContinuousMatrixDistribution`.
+A matrix-variate distribution type should be defined as a subtype of `DiscreteMatrixDistribution` or `ContinuousMatrixDistribution`.
 
 The following methods need to be implemented for each matrix-variate distribution type:
 
 - [`size(d::MatrixDistribution)`](@ref)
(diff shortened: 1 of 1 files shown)


docs/src/extends.md
```diff
@@ -154,13 +154,13 @@ It is also recommended that one also implements the following statistics functio
...
+## Create a matrix-variate Distribution
 
-A multivariate distribution type should be defined as a subtype of `DiscreteMatrixDistribution` or `ContinuousMatrixDistribution`.
+A matrix-variate distribution type should be defined as a subtype of `DiscreteMatrixDistribution` or `ContinuousMatrixDistribution`.
 
 The following methods need to be implemented for each matrix-variate distribution type:
 
 - [`size(d::MatrixDistribution)`](@ref)
-- [`rand(d::MatrixDistribution)`](@ref)
+- [`rand(rng::AbstractRNG, d::MatrixDistribution)`](@ref)
```
devmotion, line 164: This should be (similar to the multivariate case)
```suggestion
- [`Distributions._rand!(rng::AbstractRNG, d::MatrixDistribution, A::AbstractMatrix)`](@ref)
```
see e.g https://github.com/JuliaStats/Distributions.jl/blob/371a427205b605df38eb7d4f2aedc2ecd0d9047b/src/matrix/inversewishart.jl#L129

docs/src/extends.md
```diff
@@ -154,13 +154,13 @@ It is also recommended that one also implements the following statistics functio
 - [`entropy(d::MultivariateDistribution)`](@ref)
 - [`cov(d::MultivariateDistribution)`](@ref)
 
-## Create a Matrix-variate Distribution
+## Create a matrix-variate Distribution
```
devmotion, line 157: It seems the whole page uses title case, so I guess we should rather change it to
```suggestion
## Create a Matrix-Variate Distribution
```
(at least that's the recommendation for title case of words with hyphen in e.g. JSS)

PR #1614, opened 2022-09-12
TDist: use one instead of 1 for type stability

nsajko:


src/univariate/continuous/tdist.jl
```diff
@@ -79,7 +79,7 @@ end
 
 @_delegate_statsfuns TDist tdist ν
 
-rand(rng::AbstractRNG, d::TDist) = randn(rng) / ( isinf(d.ν) ? 1 : sqrt(rand(rng, Chisq(d.ν))/d.ν) )
+rand(rng::AbstractRNG, d::TDist) = randn(rng) / ( isinf(d.ν) ? one(d.ν) : sqrt(rand(rng, Chisq(d.ν))/d.ν) )
```
devmotion, line 82: I guess the function is type stable for most parameter types (but I assume that was the case already before, at least for `TDist{Float64}`, `TDist{Float32}`, and `TDist{Int}` that was the case - even though the expression `isinf(...) ? ... : ...` was not type stable for `Float64` and `Float32`). However, the expression `isinf(..) ? ... : ...` is actually still not type stable. For instance,
```julia

julia> f(ν) = isinf(ν) ? one(ν) : sqrt(rand(Random.default_rng(), Chisq(ν))/ν)

julia> @code_warntype f(1.0)
MethodInstance for f(::Float64)
  from f(ν) in Main at REPL[5]:1
Arguments
  #self#::Core.Const(f)
  ν::Float64
Body::Float64
1 ─ %1  = Main.isinf(ν)::Bool
└──       goto #3 if not %1
2 ─ %3  = Main.one(ν)::Core.Const(1.0)
└──       return %3
3 ─ %5  = Random.default_rng::Core.Const(Random.default_rng)
│   %6  = (%5)()::Core.Const(TaskLocalRNG())
│   %7  = Main.Chisq(ν)::Chisq{Float64}
│   %8  = Main.rand(%6, %7)::Float64
│   %9  = (%8 / ν)::Float64
│   %10 = Main.sqrt(%9)::Float64
└──       return %10


julia> @code_warntype f(1f0)
MethodInstance for f(::Float32)
  from f(ν) in Main at REPL[5]:1
Arguments
  #self#::Core.Const(f)
  ν::Float32
Body::Union{Float32, Float64}
1 ─ %1  = Main.isinf(ν)::Bool
└──       goto #3 if not %1
2 ─ %3  = Main.one(ν)::Core.Const(1.0f0)
└──       return %3
3 ─ %5  = Random.default_rng::Core.Const(Random.default_rng)
│   %6  = (%5)()::Core.Const(TaskLocalRNG())
│   %7  = Main.Chisq(ν)::Chisq{Float32}
│   %8  = Main.rand(%6, %7)::Float64
│   %9  = (%8 / ν)::Float64
│   %10 = Main.sqrt(%9)::Float64
└──       return %10
```
Maybe we should just change it to
```suggestion
function rand(rng::AbstractRNG, d::TDist)
    ν = d.ν
    z = sqrt(rand(rng, Chisq(ν)) / ν
    return randn(rng) / (isinf(ν) ? one(z) : z)
end
```
to be completely type stable in all possible cases.

devmotion (commented):
Thank you for the PR!

I think we could improve the function even a bit more. I wonder also if we could add a test that's currently failing and fixed by this PR?

src/univariate/continuous/tdist.jl
devmotion: One additional improvement would be
```suggestion
function rand(rng::AbstractRNG, d::TDist)
    ν = d.ν
    z = sqrt(rand(rng, Chisq{typeof(ν)}(ν)) / ν
    return randn(rng) / (isinf(ν) ? one(z) : z)
end
```
since we can skip the argument check (we know that `ν` is positive since it is a parameter of `TDist`).

devmotion (commented):
Thanks again for the PR! I applied an additional fix (see my comment above) and added some type inference tests (they fail on the master branch).

devmotion (approved):

PR #1619, opened 2022-09-20
Add `init` argument to `Product` logpdf

phipsgabler:
Otherwise, `product_distribution` fails with empty vectors of non-normal distributions:

```julia
julia> logpdf(product_distribution(typeof(Normal(0, 1))[]), Float64[])
-0.0

julia> logpdf(product_distribution(typeof(Beta(1, 1))[]), Float64[])
ERROR: MethodError: reducing over an empty collection is not allowed; consider supplying `init` to the reducer
Stacktrace:
  [1] reduce_empty(op::Base.MappingRF{Distributions.var"#527#528"{Product{Continuous, Beta{Float64}, Vector{Beta{Float64}}}, Vector{Float64}}, typeof(Base.add_sum)}, #unused#::Type{Int64})
    @ Base ./reduce.jl:356
  [2] reduce_empty_iter
    @ ./reduce.jl:379 [inlined]
  [3] mapreduce_empty_iter(f::Function, op::Function, itr::UnitRange{Int64}, ItrEltype::Base.HasEltype)
    @ Base ./reduce.jl:375
  [4] _mapreduce(f::Distributions.var"#527#528"{Product{Continuous, Beta{Float64}, Vector{Beta{Float64}}}, Vector{Float64}}, op::typeof(Base.add_sum), #unused#::IndexLinear, A::UnitRange{Int64})
    @ Base ./reduce.jl:427
  [5] _mapreduce_dim
    @ ./reducedim.jl:365 [inlined]
  [6] #mapreduce#765
    @ ./reducedim.jl:357 [inlined]
  [7] mapreduce
    @ ./reducedim.jl:357 [inlined]
  [8] #_sum#775
    @ ./reducedim.jl:999 [inlined]
  [9] _sum
    @ ./reducedim.jl:999 [inlined]
 [10] #sum#773
    @ ./reducedim.jl:995 [inlined]
 [11] sum
    @ ./reducedim.jl:995 [inlined]
 [12] _logpdf
    @ ~/.julia/packages/Distributions/yoypU/src/multivariate/product.jl:41 [inlined]
 [13] logpdf(d::Product{Continuous, Beta{Float64}, Vector{Beta{Float64}}}, x::Vector{Float64})
    @ Distributions ~/.julia/packages/Distributions/yoypU/src/common.jl:250
 [14] top-level scope
    @ REPL[45]:1
```

Or should it even be negative zero, for consistency?


devmotion:
I think it should not even be allowed to construct such a product distribution as it is not an actual distribution.

Did you check what happens for `ProductDistribution`s, the successor of the to-be-removed `Product` (it's only kept to avoid breaking changes for now and `product_distribution` will create a `ProductDistribution` at some point)?

phipsgabler:
> I think it should not even be allowed to construct such a product distribution as it is not an actual distribution.

How is it not? I interpret it as just the trivial distribution over length zero arrays, which I find very convenient for cases where I have 0 to many observations. (Or equivalently, the identity element of distributions under product formation).

> Did you check what happens for `ProductDistribution`s, the successor of the to-be-removed `Product` (it's only kept to avoid breaking changes for now and `product_distribution` will create a `ProductDistribution` at some point)?

No, I didn't know such a thing existed! Interestingly, its documentation says to use `product_distribution` to construct one, but the latter still produces a `Product`.

Anyway I fear `ProductDistribution` will suffer from the same issue, as it also uses `sum` over logpdfs without `init`.

src/multivariate/product.jl
```diff
@@ -39,7 +39,7 @@ end
 _rand!(rng::AbstractRNG, d::Product, x::AbstractVector{<:Real}) =
     map!(Base.Fix1(rand, rng), x, d.v)
 _logpdf(d::Product, x::AbstractVector{<:Real}) =
-    sum(n->logpdf(d.v[n], x[n]), 1:length(d))
+    sum(n->logpdf(d.v[n], x[n]), 1:length(d); init=0.0)
```
devmotion, line 42: Probably we rather want something like
```suggestion
function _logpdf(d::Product, x::AbstractVector{<:Real})
    if isempty(dists)
        return sum(map(logpdf, d.v, x))
    end
    dists = d.v
    return sum(n -> logpdf(dists[n], x[n]), 1:length(d))
end
```

devmotion:
I think we can support this in `Product`. I still think it's a bit weird case, so not sure if we should support it in `ProductDistribution`, but since `Product`s can be constructed with empty arrays of distributions we might as well fix the example above.

src/multivariate/product.jl
```diff
@@ -38,8 +38,13 @@ end
 
 _rand!(rng::AbstractRNG, d::Product, x::AbstractVector{<:Real}) =
     map!(Base.Fix1(rand, rng), x, d.v)
-_logpdf(d::Product, x::AbstractVector{<:Real}) =
-    sum(n->logpdf(d.v[n], x[n]), 1:length(d); init=0.0)
+function _logpdf(d::Product, x::AbstractVector{<:Real})
+    if isempty(dists)
```
phipsgabler, line 43: @devmotion are you sure that works? `dists` is only assigned below...

devmotion: It didn't, it was a mix of two versions I had played around with locally 😅 But I fixed it in a subsequent commit in this PR.

PR #1626, opened 2022-10-10
Add missing univariate dists to documentation

frankier:
This PR improves the SkewNormal docs a bit and adds it, along with the Soliton distribution to the documentation. Hopefully this will make them a bit more visible to users of the library.

### DIFF
--- a/src/univariate/continuous/skewnormal.jl
+++ b/src/univariate/continuous/skewnormal.jl
@@ -7,9 +7,9 @@ location `ξ`, scale `ω`, and shape `α`, it has the probability density functi
 
 ```math
 f(x; \\xi, \\omega, \\alpha) =
-\\frac{2}{\\omega \\sqrt{2 \\pi}} e^{-\\frac{(x-\\xi)^2}{2\\omega^2}}
+\\frac{2}{\\omega \\sqrt{2 \\pi}} \\exp{\\bigg(-\\frac{(x-\\xi)^2}{2\\omega^2}\\bigg)}
 \\int_{-\\infty}^{\\alpha\\left(\\frac{x-\\xi}{\\omega}\\right)}
-\\frac{1}{\\sqrt{2 \\pi}}  e^{-\\frac{t^2}{2}}\\ dt
+\\frac{1}{\\sqrt{2 \\pi}}  \\exp{\\bigg(-\\frac{t^2}{2}\\bigg)} \\, \\mathrm{d}t
 ```
 
 External links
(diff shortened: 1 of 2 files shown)


frankier:
The documentation CI seems to be failing, but I don't really understand why. It looks like it's trying to push/desploy the documentation, but this is probably not the correct thing to do for a pull request?

devmotion:
That's for showing previews. I am a bit surprised though that Documenter tries to deploy from your PR as the feature in Documenter should be disabled automatically for PRs from forks.

devmotion:
Seems to be a recent issue and hopefully fixed soon, see e.g. https://github.com/JuliaDocs/Documenter.jl/pull/1966.

devmotion:
Did you try building the docs locally? And if yes, could you add a screenshot of the changes to this PR?

src/univariate/continuous/skewnormal.jl
```diff
@@ -1,11 +1,23 @@
...
-    that generalises the normal distribution to allow for non-zero skewness.
+    SkewNormal(ξ, ω, α)
+
+The *skew normal distribution* is a continuous probability distribution that
+generalises the normal distribution to allow for non-zero skewness. Given a
+location `ξ`, scale `ω`, and shape `α`, it has the probability density function
+
+```math
+f(x; \\xi, \\omega, \\alpha) =
+\\frac{2}{\\omega \\sqrt{2 \\pi}} e^{-\\frac{(x-\\xi)^2}{2\\omega^2}}
```
devmotion, line 10: I think usually we use `exp`:
```suggestion
\\frac{2}{\\omega \\sqrt{2 \\pi}} \\exp{\\bigg(-\\frac{(x-\\xi)^2}{2\\omega^2}\\bigg)}
```

src/univariate/continuous/skewnormal.jl
```diff
@@ -1,11 +1,23 @@
...
+\\int_{-\\infty}^{\\alpha\\left(\\frac{x-\\xi}{\\omega}\\right)}
+\\frac{1}{\\sqrt{2 \\pi}}  e^{-\\frac{t^2}{2}}\\ dt
```
devmotion, line 12:
```suggestion
\\frac{1}{\\sqrt{2 \\pi}}  \\exp{\\bigg(-\\frac{t^2}{2}\\bigg)} \\, \\mathrm{d}t
```

frankier:
Yes I did. Here:

2022-10-10T13:51:53,433354582+03:00
2022-10-10T13:47:21,971389802+03:00

PR #1631, opened 2022-10-26
Folded Distributions

Potatoasad:
# Folded Distributions
Here's a Generic implementation for `Folded` continuous univariate distributions. I worked on this a while ago and only recently cleaned it up. This has been useful to me in my research and hopefully is helpful to others as well.

### Definition

A folded distribution $F_{c}(D)$ of a distribution $D$ at crease $c \in \mathbb{R}$ is the reflection of the distribution below $c$ onto the the distribution above $c$. 

The pdf of such a distribution is given by:

$$
p(x | F_{c}(D)) = p(x | D) + p(x' | D)
$$

where $x' = 2c - x$

### Interface
The interface is very very similar to the `truncated` interface. 
You can take any continuous univariate distribution and make it into a folded distribution like so:
```Julia
julia> normal_absolute = folded(Normal(0.2,0.1), 0.0)
# Folded(Normal{Float64}(μ=0.2, σ=0.1); crease = 0.0 | fold left onto right)
julia> plot(x -> pdf(normal_absolute, x))
```
This gives the result:
folded

The `folded` function creates a _folded distribution_ from the original distribution `d` about the value `crease`.
This function defaults to folding the left side onto the right, but by using `keep_right=false` one can reflect the right side onto the left.

This can be implemented for any univariate continuous distribution by using the following method:
```Julia
folded(d::ContinuousUnivariateDistribution, crease::Real)
```
If one wants to reflect points _above_ the crease $c$ onto points _below_ $c$ (such that the resultant distribution lives on points below $c$) one can do that using the `keep_right=false` tag:
```Julia
folded(d::ContinuousUnivariateDistribution, crease::Real, keep_right=false)
```
A very useful and oft-occuring example of this is the action of the absolute value function ($|\\cdot|$) on random variables. 
For example if we have a random variable obeying the normal distribution:
$$\hat{X} \sim \mathcal{N}(\alpha,\beta)$$

Taking the absolute value of this random variable always follows a _folded normal distribution_ with a crease at $0$

$$|\hat{X}| \sim F_{c=0}(\mathcal{N}(\alpha,\beta))$$
In julia that looks like:

```julia
folded_normal = folded(Normal(μ,σ), 0.0)
```
and then we can ask for many things like the pdf, or sample the distribution. The samples of course will then be positive:

```julia
all(rand(folded_normal, 1000) .≥ 0) # True
```

In general, one can take __any__ univariate continous distribution and fold it using this function.

##### Example
We can create the absolute value of the laplace distribution by:
```julia
folded_laplace = folded(Laplace(μ,b), 0.0)
```
as a quick check we can see if the 1000 samples generated from this are all greater than 0:
```julia
all(rand(folded(Lapace(μ,b),0),1000) .≥ 0) # -> true
```

For more one can check the docs in the PR. 

### Conclusion & Why would this be useful?
With this addition one should be able to define stuff like the distribution of the absolute value of any univariate random variable. 

I've added tests and docs, and wanted to see if this is an appropriate inclusion to the package. 

I saw previously people wanting an implementation of the `HalfNormal` distribution, and I think that is as simple as 

```Julia
half_normal(σ) = folded(Normal(zero(σ), σ), zero(σ))
```
Let me know if this is would be welcomed by this package, and if there's anything I should add to the tests or the docs, or change the code style. 

Thank you for all the work you all do on this amazing package!


azev77:
This is a brilliant incredibly useful idea!!!

Potatoasad:
> This is a brilliant incredibly useful idea!!!

Thank you so much! Hoping this can get merged soon 🤞

Potatoasad:
Hi! Just wanted to get an update on this if this is good enough to merge - I think it's ready for review from my end. 
I understand the maintainers have a limited time so no worries at all, just wanted to bump this.

sethaxen:
Hey @Potatoasad, I had just hacked together a `Folded` distribution for my own research and just now saw your PR. I'd be happy to review it and will try to do so this weekend.

src/fold.jl
```diff
@@ -0,0 +1,320 @@
...
+    Folded(d, included, excluded, crease, keep_right)
+end
+
+############################################################################
+# This section is there so that typeof(d)(params(d)...) == d works:
+############################################################################
+
+# Define params as the unfolded original parameters, just appended with the crease and keep_right variables
+params(d::Folded) = tuple(params(d.original)..., d.crease, d.keep_right)
+partype(d::Folded) = partype(d.original)
```
ParadaCarleton, line 115: I'm not sure if this is correct, but `partype` doesn't have a docstring so I'm not sure what it promises. @devmotion ?

src/fold.jl
```diff
@@ -0,0 +1,320 @@
...
+        excluded = left_distribution
+    else
+        included = left_distribution
+        excluded = right_distribution
+    end
+    Folded(d, included, excluded, crease, keep_right)
+end
+
+############################################################################
+# This section is there so that typeof(d)(params(d)...) == d works:
```
ParadaCarleton, line 110: Could you clarify? I'm not sure what you mean here.

src/fold.jl
```diff
@@ -0,0 +1,320 @@
...
+minimum(d::Folded) = minimum(d.included)
+
+
+
+"""
+    maximum(d::Folded)
+
+Returns the highest value of the domain of the distribution
+"""
+maximum(d::Folded) = maximum(d.included)
```
ParadaCarleton, line 199: Same as above.

src/fold.jl
```diff
@@ -0,0 +1,320 @@
...
+```julia
+pdf(folded(Normal(μ,σ),0), -3) # -> 0.0
+pdf(folded(Normal(1.0,1.0),0), 3) # -> 0.05412479673895295
+pdf(folded(Normal(3,4),0.0),0.1) ≈ pdf(Normal(3,4),-0.1) + pdf(Normal(3,4),0.1) # -> true
+```
+"""
+function pdf(d::Folded, x::Real)
+    !insupport(d,x) && return zero(x)
+    #return pdf(d.original,x) + pdf(d.original,unfold_value(x,d)) -- removed for numerical stability
+    return exp(logpdf(d,x))
```
ParadaCarleton, line 244: If `logpdf` is already included, I think `pdf` isn't necessary, because `pdf` already has a method defined as `exp(logpdf)`.

src/fold.jl
```diff
@@ -0,0 +1,320 @@
...
+```
+
+Note that if the mean for the truncated distribution is not implemented, then this will give a Method 
+Error:
+```
+mean(folded(Beta(1,2)),0) # Errors out because mean(truncated(Beta(1,2),..)) is not implemented
+# >> MethodError: no method matching iterate(::Truncated{Beta{...}...})
+```
+"""
+mean(d::Folded) = mean(d.included)*d.included.tp - mean(d.excluded)*d.excluded.tp + 2*d.crease
```
ParadaCarleton, line 291: Can you add comments or expand this out? It's not immediately obvious to me why this is correct.

test/fold.jl
```diff
@@ -0,0 +1,51 @@
...
+	@test fold_value( 4, folded(Normal(2,3),10, keep_right=false)) == 4
+	@test fold_value( 16, folded(Normal(2,3),10, keep_right=false)) == 4
+	@test unfold_value(10, folded(Normal(3,4),0)) == -10
+	@test unfold_value( 5 , folded(Normal(3,4),0)) == -5
+	@test unfold_value( 16, folded(Normal(2,3),10, keep_right=false)) == 16
+	@test unfold_value( 4, folded(Normal(2,3),10, keep_right=false)) == 16
+end 
+
+# Testing if the insupport function works
+@testset "Folded insupport, minimum and maximum functions" begin
```
ParadaCarleton, line 24: Can you add more tests? We should be covering at least `mean`, `var`, `median`, and at least two different quantiles.

test/fold.jl
```diff
@@ -0,0 +1,51 @@
...
+	@test insupport(folded(Normal(2,1),0),-10) == false
+	# Check that the crease is always the minimum of a folded distrbution
+	@test all(((x->Distributions.minimum(folded(Normal(2,1),x))).(-10:0.1:10)) .≈ collect(-10:0.1:10))
+	# Check that the crease is always the maximum of a distrbution folded to the right
+	@test all(((x->Distributions.maximum(folded(Normal(2,1),x,keep_right=false))).(-10:0.1:10)) .≈ collect(-10:0.1:10))
+end
+
+@testset "Folded Rand functions" begin
```
ParadaCarleton, line 32: Same as for the above--we need tests that check if `rand` is working correctly, e.g. testing if a big sample from a folded normal distribution has the right distribution.

test/fold.jl
```diff
@@ -0,0 +1,51 @@
...
+	# Check that a folded distribution with a crease at 0 always generates positive numbers
+	@test all(rand(folded(Normal(1.0,1.0),0.0),100) .> 0.0)
+end
+
+@testset "Folded pdf, logpdf, cdf and logcdf functions" begin
+	@test pdf(truncated(Normal(3,4),-1,1),-0.1) + pdf(truncated(Normal(3,4),-1,1),0.1) ≈ pdf(folded(truncated(Normal(3,4),-1,1),0.0),0.1)
+	@test pdf(folded(truncated(Normal(3,4),-1,1),0.0),-0.1) == 0
+end
```
ParadaCarleton, line 40:
```suggestion
@testset "Folded pdf, logpdf, cdf and logcdf functions" begin
    trunc_dist = truncated(Normal(3, 4), -1, 1)
    fold_dist = folded(trunc_dist, 0.0)
	@test pdf(truncated(trunc_dist, -0.1) + pdf(trunc_dist, 0.1) ≈ pdf(fold_dist, 0.1)
	@test pdf(fold_dist, -0.1) ≈ 0
end
```

test/fold.jl
```diff
@@ -0,0 +1,51 @@
...
+	# Check that the crease is always the maximum of a distrbution folded to the right
+	@test all(((x->Distributions.maximum(folded(Normal(2,1),x,keep_right=false))).(-10:0.1:10)) .≈ collect(-10:0.1:10))
+end
+
+@testset "Folded Rand functions" begin
+	# Check that a folded distribution with a crease at 0 always generates positive numbers
+	@test all(rand(folded(Normal(1.0,1.0),0.0),100) .> 0.0)
+end
+
+@testset "Folded pdf, logpdf, cdf and logcdf functions" begin
```
ParadaCarleton, line 37: Can you add tests that deal with cases other than folding at 0?

ParadaCarleton (changes requested):

src/fold.jl
```diff
@@ -0,0 +1,320 @@
...
+cdf(folded(Beta(α,β),0.5), 3) # -> 1.0
+cdf(folded(Normal(1.0,1.0),0), 1.0) # -> 0.4772498680518208
+```
+"""
+function cdf(d::Folded, x::Real) 
+    if !insupport(d,x)
+        (x ≤ minimum(d)) && return zero(x) # Is below the support
+        (x ≥ maximum(d)) && return one(x) # Is above the support
+    end
+    d.included.tp*cdf(d.included, x) + d.excluded.tp*ccdf(d.excluded, unfold_value(x,d))
```
devmotion, line 261: d.included.tp does not exist for general truncated distributions. Could we use something like
```suggestion
cdf(d::Folded, x::Real) = exp(logcdf(d, x)) # not sure, maybe the default definition already?
function logcdf(d::Folded, x::Real)
    _x = clamp(x, extrema(d.included)...)
    y = unfolded_value(d, _x)
    a, b = d.keep_right ? (y, _x) : (_x, y)
    return logdiffcdf(d.original, b, a)
end
```

src/fold.jl
```diff
@@ -0,0 +1,320 @@
...
+end
+
+
```
devmotion, line 264: ccdf and logccdf are not defined, ie onky fallback definitions will be used?

src/fold.jl
```diff
@@ -0,0 +1,320 @@
...
+# Quantiles will be calculated by the newton method:
+@quantile_newton(Folded)
+
+
+######################################################
+#         Statistics extended methods
+######################################################
+
+"""
+    mean(d::Folded)
```
devmotion, line 274: Please remove.

src/fold.jl
devmotion, line 291: Again tp does not exist in general.

devmotion (commented):
Thank you for the PR! I added some initial comments.

Potatoasad:
Thank you so much for the detailed review! I'll pour over these and make the changes as needed thank you so much for the feedback and explanations.

src/fold.jl
```diff
@@ -0,0 +1,320 @@
...
+"""
+function folded end
+
+
+#the folded function with no arguments will just return the function itself
+folded(d::ContinuousUnivariateDistribution, ::Nothing, ::Nothing) = d
+
+
+##### Defining the function and the initialization functions #####
+"""
```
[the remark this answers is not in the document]
devmotion: No. JuliaStats does not use Blue style. One should just be consistent with the existing style until the org decides on a specific style. It does not matter anyway as all these docstrings should be removed (see my review).

azev77:
> Thank you so much for the detailed review! I'll pour over these and make the changes as needed thank you so much for the feedback and explanations.

@Potatoasad have you had any luck with this?

Potatoasad:
> > Thank you so much for the detailed review! I'll pour over these and make the changes as needed thank you so much for the feedback and explanations.
> 
> @Potatoasad have you had any luck with this?

Hey, sorry. Yes I did make some slight progress, but got sidetracked into a paper I was working on. I think I can finish this up in the next two weeks. Apologies for the very long delay on this.

src/fold.jl
Potatoasad: My understanding, by looking at the source code was that the Truncated object has a field `tp` in its definition, and having this would prevent recalculation each time. 

Is there a way to infer from the types whether a truncated distribution has the `tp` field? So if it doesn't exist for a particular type, we could dispatch to the code you provided? 

https://github.com/JuliaStats/Distributions.jl/blob/fa8c30dedf058cb9891ef141c14cc857576221ac/src/truncate.jl#L93

ParadaCarleton:
> > > Thank you so much for the detailed review! I'll pour over these and make the changes as needed thank you so much for the feedback and explanations.
> > 
> > 
> > @Potatoasad have you had any luck with this?
> 
> Hey, sorry. Yes I did make some slight progress, but got sidetracked into a paper I was working on. I think I can finish this up in the next two weeks. Apologies for the very long delay on this.

Hi, just wanted to know if you're still interested in working on this :)

Ph0non:
Any updates on this issue?

PR #1635, opened 2022-11-03
Fix argument checks for BetaBinomial

damonbayer:
Letting α = 0 or β=0 results in many methods (pdf, quantile, rand) failing. Per Wikipedia, α and β should be positive.  https://en.wikipedia.org/wiki/Beta-binomial_distribution.


devmotion:
Seems reasonable and matches the checks of `Beta`.

devmotion:
I added some tests.

PR #1636, opened 2022-11-07
Fix of PGeneralizedGaussian description

slamajakub:
Fixed description of PGeneralizedGaussian to be compliant with mathematical formula and implementation (mismatch between μ, α).

The order of function parameters is given on lines 31-33, where it is compliant with math formula in the docs. However, the documentation for this function had mismatch between μ, α symbols, so this fix is fixing it such that the docs comply with math formula and real implementation.


devmotion:
Thank you!

Issue #1638, opened 2022-11-16
Weibull logpdf yields NaN at Inf


oschulz:
We currently have

```julia
julia> logpdf(Weibull(), Inf)
NaN
```

but we should get `-Inf`, of course. `pdf` is zero at `Inf`, as it should be:

```julia
julia> log(pdf(Weibull(), Inf))
-Inf
```

oschulz:
Should we just do

´´´julia
function mylogpdf(d::Weibull{T}, x::Real) where T<:Real
    if x >= 0 && isfinite(x)
        α, θ = params(d)
        z = x / θ
        log(α / θ) + (α - 1) * log(z) - z^α
    else
        -T(Inf)
    end
end
```

?

**Comment by devmotion** [comment]
There's also a problem/inconsistency with `logpdf` at `x = 0`:
```julia
julia> logpdf(Weibull(), 0)
NaN

julia> pdf(Weibull(), 0)
1.0
```

Regarding the suggested fix, relying on `T` is wrong (but also part of the current code) and `NaN` inputs should yield `NaN` outputs, and I assume we might want to use `xlogy`. So I suggest
```julia
function logpdf(d::Weibull, x::Real)
    α, θ = params(d)
    z = abs(x) / θ
    res = log(α / θ) + xlogy(α - 1, z) - z^α
    x < 0 || isinf(x) ? oftype(res, -Inf) : res
end
```
Then I get
```julia
julia> logpdf(Weibull(), -3)
-Inf

julia> logpdf(Weibull(), 0)
0.0

julia> logpdf(Weibull(), Inf)
-Inf

julia> logpdf(Weibull(), -Inf)
-Inf

julia> logpdf(Weibull(), NaN)
NaN

julia> logpdf(Weibull(), -NaN)
NaN
```
and also the type instability is fixed: Before
```julia
julia> @inferred logpdf(Weibull(1f0), 3.0)
ERROR: return type Float64 does not match inferred return type Union{Float32, Float64}
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] top-level scope
   @ REPL[34]:1
```
and with the suggestion:
```julia
julia> @inferred _logpdf(Weibull(1f0), 3.0)
-3.0
```

oschulz:
I like it! :-)

PR #1639, opened 2022-11-16
Fix `logpdf` and `pdf` of `Weibull`

devmotion:
Implements the suggestions in and fixes https://github.com/JuliaStats/Distributions.jl/issues/1638.

Will add tests for all the special cases.


oschulz:
Thanks @devmotion !

PR #1641, opened 2022-11-18
added DistributionsCore so other packages don't have to depend on Distributions

oscardssmith:
This won't work quite yet because we need to tag `DistributionsCore` as a package so we can depend on it, but it means that (for example) DiffEQBase can rely on `DistributionsCore` instead of `DistributionsCore` which will save almost 700ms of the current 3 second include time.


lib/DistributionsCore/src/DistributionsCore.jl
```diff
@@ -0,0 +1,521 @@
+module DistributionsCore
+import Base: maximum, minimum, extrema, ==
```
devmotion, line 2: Maybe we move away from using `import` and instead qualify functions explicitly when we extend them? At least I prefer that over `import`.

lib/DistributionsCore/src/DistributionsCore.jl
```diff
@@ -0,0 +1,521 @@
...
+## sample space/domain
+
+"""
+`F <: VariateForm` specifies the form or shape of the variate or a sample.
+"""
```
devmotion, line 7: What did you include in this file? Maybe keep the same file names as in Distributions to make it clearer? Maybe we want to reorganize it at some point but it might be clearer for the transition.

devmotion (commented):
I think it's a good idea. The main question and challenge might only be to discuss what actually the core parts are and how much should be included.

I also wonder if there are any disadvantages that we should consider before doing that. For instance, there are many PRs (one I have to revive at some point at least) and long discussions about `eltype` and how its behaviour should be changed, how `rand` should follow more closely the syntax in Random etc. If we want to make these changes to the design at some point, will it be more difficult if the package is split into multiple packages?

lib/DistributionsCore/src/DistributionsCore.jl
oscardssmith: This file is a copy and paste of common.jl.

devmotion: Just keep it in a separate common.jl file? I assume there might be also other things that we would like to add, and it seems cleaner to move the files than to merge them, doesn't it?

oscardssmith: makes sense. Will do.

oscardssmith:
This PR will make it a little bit harder to make breaking changes, but IMO the parts that I've split out are a pretty good starting place that will let packages define dispatches on abstract distribution types without bringing more than necessary in.

oscardssmith:
@devmotion how would you feel about merging this as is? I think it is a reasonable division, and registering DistributionsCore will take 3 days, so I would prefer to get an initial version tagged and work from there.

oscardssmith:
bump.

devmotion:
I think such a major change should be approved by multiple somewhat active maintainers.

oscardssmith:
Is there anyone specific you want to here from? Should they be pinged?

andreasnoack:
Yeah. This is a pretty substantial change so let's not rush it. Do we understand the components of the loading time here?

oscardssmith:
```
julia> @time @time_imports using Distributions
      0.8 ms  DataAPI
      0.1 ms  Compat
     50.3 ms  DataStructures
      0.2 ms  SortingAlgorithms
      6.6 ms  Missings
      4.0 ms  DocStringExtensions 84.46% compilation time
     56.5 ms  ChainRulesCore
      0.7 ms  ChangesOfVariables
      0.7 ms  InverseFunctions
      3.3 ms  IrrationalConstants
      0.4 ms  LogExpFunctions
      0.2 ms  StatsAPI
     19.5 ms  StatsBase
      0.1 ms  SuiteSparse
     15.7 ms  PDMats
      0.1 ms  Reexport
      0.1 ms  OpenLibm_jll
     18.4 ms  Preferences
      0.2 ms  JLLWrappers
      4.4 ms  OpenSpecFun_jll 90.54% compilation time
     12.3 ms  SpecialFunctions
      0.3 ms  Rmath_jll
     62.3 ms  Rmath 111.15% compilation time
      0.2 ms  NaNMath
      1.2 ms  Calculus
     11.1 ms  DualNumbers
      0.7 ms  HypergeometricFunctions
      3.7 ms  StatsFuns
      2.1 ms  QuadGK
    137.1 ms  FillArrays
      1.0 ms  DensityInterface
    172.0 ms  Distributions
  0.667988 seconds (1.46 M allocations: 93.295 MiB, 5.31% gc time, 13.08% compilation time)
```
So in short, the big time hits are `DataStructures`, `ChainRules`, `RMath`, `FillArrays` and all the code in `Distributions` itself. The fundamental problem is that `Distributions` defines a lot of code and brings in a lot of dependencies which causes problems for packages that want to interoperate with `Distributions` but don't depend on it. I believe with this PR, I've pulled out a minimal set of functionality that allows interop for these packages, and takes `8ms`  to load (and won't invalidate anything).

oscardssmith:
I think 1.9's weakdeps might make this irrelevant. Closing for now.

PR #1645, opened 2022-12-08
Add the Johnson's S_U-distribution

Yuan-Ru-Lin:
This PR adds the implementation of the Johnson's S_U-distribution, which is a transformation of the normal distribution. 

See also [the relevant page on Wikipedia](https://en.wikipedia.org/wiki/Johnson%27s_SU-distribution).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.25.85"
+version = "0.25.86"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"


src/univariate/continuous/johnson.jl
```diff
@@ -0,0 +1,141 @@
+"""
+    Johnson(ξ, λ, γ, δ)
+
+The Johnson's ``S_U``-distribution with parameters ξ, λ, γ and δ is a transformation of the normal distribution:
+
+```math
+z \\sim \\gamma + \\delta \\sinh^{-1}\\Big(\\frac{x-\\xi}{\\lambda}\\Big),
+```
+
+where ``z \\sim \\mathcal{N}(0,1)``.
```
devmotion, line 10: It seems this is copied from Wikipedia (apart from a `\sim` instead of a `=` in the first equation which should be changed) but IMO this needs more explanations. For instance, you could add something like: if `X` is defined according to .... where `Z ~ N(0, 1)`, then `X ~ Johnson(....)`.

src/univariate/continuous/johnson.jl
```diff
@@ -0,0 +1,141 @@
...
+end
+
+function Johnson(ξ::T, λ::T, γ::T, δ::T; check_args::Bool=true) where {T<:Real}
+    @check_args Johnson (λ, λ ≥ zero(λ)) (δ, δ ≥ zero(δ))
+	return Johnson{T}(ξ, λ, γ, δ)
+end
+
+Johnson() = Johnson{Float64}(0.0, 1.0, 0.0, 1.0)
+Johnson(ξ::Real, λ::Real, γ::Real, δ::Real; check_args::Bool=true) = Johnson(promote(ξ, λ, γ, δ)...; check_args=check_args)
+Johnson(ξ::Integer, λ::Integer, γ::Integer, δ::Integer; check_args::Bool=true) = Johnson(float(ξ), float(λ), float(γ), float(δ); check_args=check_args)
```
devmotion, line 40: Is this actually needed?
```suggestion
```

src/univariate/continuous/johnson.jl
```diff
@@ -0,0 +1,141 @@
...
+	ξ::T
+	λ::T
+	γ::T
+	δ::T
+	Johnson{T}(ξ, λ, γ, δ) where {T} = new{T}(ξ, λ, γ, δ)
+end
+
+function Johnson(ξ::T, λ::T, γ::T, δ::T; check_args::Bool=true) where {T<:Real}
+    @check_args Johnson (λ, λ ≥ zero(λ)) (δ, δ ≥ zero(δ))
+	return Johnson{T}(ξ, λ, γ, δ)
```
devmotion, line 35:
```suggestion
    return Johnson{T}(ξ, λ, γ, δ)
```

src/univariate/continuous/johnson.jl
```diff
@@ -0,0 +1,141 @@
...
+External links
+
+* [Johnson's ``S_U``-distribution on Wikipedia](http://en.wikipedia.org/wiki/Johnson%27s_SU-distribution)
+"""
+struct Johnson{T<:Real} <: ContinuousUnivariateDistribution
+	ξ::T
+	λ::T
+	γ::T
+	δ::T
+	Johnson{T}(ξ, λ, γ, δ) where {T} = new{T}(ξ, λ, γ, δ)
```
devmotion, line 30: The indentation seems wrong? Did you use tabs? Please change it to four spaces throughout the files.
```suggestion
    ξ::T
    λ::T
    γ::T
    δ::T
    Johnson{T}(ξ::T, λ::T, γ::T, δ::T) where {T<:Real} = new{T}(ξ, λ, γ, δ)
```

src/univariate/continuous/johnson.jl
```diff
@@ -0,0 +1,141 @@
...
+end
+
+function Johnson(ξ::T, λ::T, γ::T, δ::T; check_args::Bool=true) where {T<:Real}
+    @check_args Johnson (λ, λ ≥ zero(λ)) (δ, δ ≥ zero(δ))
+	return Johnson{T}(ξ, λ, γ, δ)
+end
+
+Johnson() = Johnson{Float64}(0.0, 1.0, 0.0, 1.0)
```
devmotion, line 38: Just
```suggestion
Johnson() = Johnson{Int}(0, 1, 0, 1)
```
?

src/univariate/continuous/johnson.jl
```diff
@@ -0,0 +1,141 @@
...
+Johnson(ξ::Real, λ::Real, γ::Real, δ::Real; check_args::Bool=true) = Johnson(promote(ξ, λ, γ, δ)...; check_args=check_args)
+Johnson(ξ::Integer, λ::Integer, γ::Integer, δ::Integer; check_args::Bool=true) = Johnson(float(ξ), float(λ), float(γ), float(δ); check_args=check_args)
+
+@distr_support Johnson -Inf Inf
+
+#### Conversions
+
+convert(::Type{Johnson{T}}, ξ::S, λ::S, γ::S, δ::S) where {T<:Real, S<:Real} = Johnson(T(ξ), T(λ), T(γ), T(δ))
```
devmotion, line 46: I don't think this method (and similar ones for other distributions) is useful at all.
```suggestion
```

src/univariate/continuous/johnson.jl
```diff
@@ -0,0 +1,141 @@
...
+
+params(d::Johnson) = (d.ξ, d.λ, d.γ, d.δ)
+partype(d::Johnson{T}) where {T<:Real} = T
+
+#### Statistics
+
+function mean(d::Johnson)
+    a = exp(d.δ^-2)
+    r = d.γ/d.δ
+    d.ξ - d.λ * sqrt(a) * sinh(r)
```
devmotion, line 63:
```suggestion
    a = exp(1/(2*d.δ^2))
    r = d.γ/d.δ
    d.ξ - d.λ * a * sinh(r)
```

src/univariate/continuous/johnson.jl
```diff
@@ -0,0 +1,141 @@
...
+end
+function var(d::Johnson)
+    a = exp(d.δ^-2)
+    r = d.γ/d.δ
+    d.λ^2/2 * (a-1) * (a*cosh(2r)+1)
+end
+function skewness(d::Johnson)
+    a = exp(d.δ^-2)
+    r = d.γ/d.δ
+    - (d.λ^3 * sqrt(a) * (a-1)^2 * (a*(a+2)*sinh(3r)+3sinh(2r))) / 4var(d)^1.5
```
devmotion, line 77: Please avoid floating literal floating point numbers as often they will cause undesired promotions.

src/univariate/continuous/johnson.jl
```diff
@@ -0,0 +1,141 @@
...
+function logccdf(d::Johnson, x::Real)
+	y = (x - d.ξ) / d.λ
+	z = d.γ + d.δ * asinh(y)
+	logccdf(Normal(), z)
+end
+
+quantile(d::Johnson, q::Real) = d.λ * sinh((quantile(Normal(), q) - d.γ) / d.δ) + d.ξ
+cquantile(d::Johnson, q::Real) = d.λ * sinh((cquantile(Normal(), q) - d.γ) / d.δ) + d.ξ
+invlogcdf(d::Johnson, lq::Real) = d.λ * sinh((invlogcdf(Normal(), lq) - d.γ) / d.δ) + d.ξ
+invlogccdf(d::Johnson, lq::Real) = d.λ * sinh((invlogccdf(Normal(), lq) - d.γ) / d.δ) + d.ξ
```
devmotion, line 129: This can be simplified by extracting the transformations. Also please don't use literal floats here and don't use `Normal`, as both cause possibly undesired promotions.
```suggestion
zval(d::Johnson, x::Real) = d.γ + d.δ * asinh((x - d.ξ) / d.λ)
xval(d::Johnson, z::Real) = d.λ * sinh(x - d.γ) / d.δ) + d.ξ

pdf(d::Johnson, x::Real) = normpdf(zval(d, x))
logpdf(d::Johnson, x::Real) = normlogpdf(zval(d, x))
cdf(d::Johnson, x::Real) = normcdf(zval(d, x))
ccdf(d::Johnson, x::Real) = normccdf(zval(d, x))
logcdf(d::Johnson, x::Real) = normlogcdf(zval(d, x))
logccdf(d::Johnson, x::Real) = normlogccdf(zval(d, x))

quantile(d::Johnson, p::Real) = xval(d, norminvcdf(p))
cquantile(d::Johnson, p::Real) = xval(d, norminvccdf(p))
invlogcdf(d::Johnson, lp::Real) = xval(d, norminvlogcdf(lp))
invlogccdf(d::Johnson, lp::Real) = xval(d, norminvlogccdf(lp))
```

src/univariate/continuous/johnson.jl
```diff
@@ -0,0 +1,141 @@
...
+
+# entropy(d::Johnson)
+# mgf(d::Johnson)
+# cf(d::Johnson)
+
+#### Sampling
+
+rand(rng::AbstractRNG, d::Johnson) = d.λ * sinh((rand(rng) - d.γ) / d.δ) + d.ξ
```
devmotion, line 137: This should be `randn` instead of `rand`, I assume. With the function above it can be simplified to:
```suggestion
rand(rng::AbstractRNG, d::Johnson) = xval(d, randn(rng))
```

devmotion (commented):
Thank you for the PR! I added some comments.

src/univariate/continuous/johnson.jl
Yuan-Ru-Lin: Yes the indentation were wrong. Now changed to spaces.

src/univariate/continuous/johnson.jl
Yuan-Ru-Lin: Yes, you're right!

Yuan-Ru-Lin: Yes, you're right!

src/univariate/continuous/johnson.jl
Yuan-Ru-Lin: Is it because functions shouldn't expect the distributions to be of type other than `Real`?

src/univariate/continuous/johnson.jl
Yuan-Ru-Lin: Nice to know those `norm*` helper functions! I agree that they should be reformatted as above.

However, the pdf is not just `normpdf(d, x)` but `d.δ / d.λ / sqrt(1 + (x - d.ξ) / d.λ)^2) * normpdf(zval(d, x))`.

I'll refactor out another helper function `yval(d::Johnson, x::Real) = (x - d.ξ) / d.λ` so that it's just `d.δ / d.λ / sqrt(1 + yval(d, x)^2) * normpdf(zval(d, x))`

src/univariate/continuous/johnson.jl
Yuan-Ru-Lin: I'll delete it anyway.

src/univariate/continuous/johnson.jl
Yuan-Ru-Lin: Yeah, let's not force users to use `float`.

src/univariate/continuous/johnson.jl
Yuan-Ru-Lin: Agreed. `Int` is fine.

src/univariate/continuous/johnson.jl
Yuan-Ru-Lin: Agreed.

src/univariate/continuous/johnson.jl
Yuan-Ru-Lin: I suppose `sqrt(var(d)^3)` can avoid that?

src/univariate/continuous/johnson.jl
Yuan-Ru-Lin: Thanks for pointing out the `\sim` typo. I'll fix that.

I'll also add

```
i.e. if a random variable ``X`` is defined to be

```math
X = \\lambda\\sinh\\Bigg( \\frac{Z - \\gamma}{\\delta} \\Bigg) + \\xi,
```.

where ``Z \\sim \\mathcal{N}(0,1)``,  then ``X \\sim {\\rm Johnson}(\\xi, \\lambda, \\gamma, \\delta)``.
```

The dot after the enclosing math code block fence is only added here to avoid GitHub from closing the code block there.

src/univariate/continuous/johnson.jl
Yuan-Ru-Lin: Turned tabs to spaces.

src/univariate/continuous/johnson.jl
devmotion: No, it's because I don't think there should be (automatic) conversions of four real numbers to `Johnson` - the constructor exists for this purpose.

devmotion (changes requested):
Sorry for the slow response, somehow this PR got lost. I think the PR looks quite good, the main thing that seems missing are correctness checks of the implementation. Can you add reference values and code such that we can compare it with libraries in R?

src/univariate/continuous/johnson.jl
ararslan: You could use `3//2` rather than `1.5`

ararslan:
Something potentially worth noting is that there are other Johnson family distributions, e.g. S<sub>B</sub> as mentioned in the Wikipedia article, so it may not be immediately clear to a user that `Johnson` refers specifically to S<sub>U</sub>. If we were to also add S<sub>B</sub> to Distributions, what would it be called?

devmotion:
> Something like `@test Johnson(0, 1, 0, 1) ≈ 0.399` in the test file?

No, something like the changes in the test/ref folder in https://github.com/JuliaStats/Distributions.jl/pull/1678.

Yuan-Ru-Lin:
Was the json part generated by R? I am wondering if I should edit myself.

src/univariate/continuous/johnson.jl
heltonmc: You could use `sqrt(x) * x` which will be much faster and slightly more accurate than `x^(3/2)`.

devmotion:
Yes, by running the gendref.R script: https://github.com/JuliaStats/Distributions.jl/tree/master/test/ref#usage

Yuan-Ru-Lin:
I'm trying to run `Rscript gendref.R` to learn how to contribute to that part. However, I got the following error message when I tried to `install.packages('LindleyR', repo=xxx)`.

```
Warning message:
package ‘LindleyR’ is not available for this version of R
```

I tried several repos, as suggested in some blog posts, none of which worked.

May I ask how to install that package?

Yuan-Ru-Lin:
Also, there is another error after I commented code relevant to `Lindley`:

```
On Rician(10.0, 1.0) --> Rician: nu=10 sigma=1 
On StudentizedRange(2.0, 2.0) --> StudentizedRange: nu=2 k=2 
Error in distr$pdf(xs) : attempt to apply non-function
Calls: do.main -> eval.points
3: (function () 
   {
       traceback(2)
       q()
   })()
2: eval.points(distr)
1: do.main("continuous_test")
```

Not sure if I should report it.

Yuan-Ru-Lin:
I changed the name of the branch under development which closed this PR. The new PR is at  #1690

Yuan-Ru-Lin:
> Also, there is another error after I commented code relevant to `Lindley`:
> 
> ```
> ...
> On Rician(10.0, 1.0) --> Rician: nu=10 sigma=1 
> On StudentizedRange(2.0, 2.0) --> StudentizedRange: nu=2 k=2 
> Error in distr$pdf(xs) : attempt to apply non-function
> Calls: do.main -> eval.points
> 3: (function () 
>    {
>        traceback(2)
>        q()
>    })()
> 2: eval.points(distr)
> 1: do.main("continuous_test")
> ```
> 
> Not sure if I should report it.

Nevermind, realized how to with these 2 links:

1. https://www.displayr.com/installing-r-packages-from-github/
1. https://github.com/cran/LindleyR/blob/master/man/Lindley.Rd

Issue #1463, opened 2021-12-24
`pdf` is ambiguous for `UnivariateGMM`


ForceBru:
## Code

```julia
julia> distr = UnivariateGMM([-1, 1], [.1, .2], Categorical([.4, .6]))
UnivariateGMM{Vector{Int64}, Vector{Float64}, Categorical{Float64, Vector{Float64}}}(
K: 2
means: [-1, 1]
stds: [0.1, 0.2]
prior: Categorical{Float64, Vector{Float64}}(support=Base.OneTo(2), p=[0.4, 0.6])
)


julia> pdf(distr, collect(-1:0.1:1))
ERROR: MethodError: pdf(::UnivariateGMM{Vector{Int64}, Vector{Float64}, Categorical{Float64, Vector{Float64}}}, ::Vector{Float64}) is ambiguous. Candidates:
  pdf(d::Distribution{ArrayLikeVariate{N}}, x::AbstractArray{<:Real, M}) where {N, M} in Distributions at ~/.julia/packages/Distributions/vR2pk/src/common.jl:249
  pdf(d::UnivariateDistribution, X::AbstractArray) in Distributions at deprecated.jl:70
Possible fix, define
  pdf(::UnivariateDistribution, ::AbstractArray{T, M} where T<:Real) where M
Stacktrace:
 [1] top-level scope
   @ REPL[100]:1
```

## Versions
- Julia 1.7.1
- Distributions v0.25.37

devmotion:
For univariate distributions, calling `pdf` on an array of values is deprecated. Nevertheless it should not throw an error but only display a deprecation warning (if they are enabled).

PR #1651, opened 2022-12-17
Fix ambiguity in univariate pdf for 0-dim arrays

### LINKED ISSUE #1650: Ambiguities with 0-dimensional arrays and univariate distributions
I found these two methods that I was hoping to use in my code:
https://github.com/JuliaStats/Distributions.jl/blob/cd45ecc6ab9ed186ad741de41a64b24c5336e4cc/src/univariates.jl#L314
https://github.com/JuliaStats/Distributions.jl/blob/cd45ecc6ab9ed186ad741de41a64b24c5336e4cc/src/univariates.jl#L326

But when I try to use them with the exported methods it does not work:

```julia
julia> using Distributions

julia> x = fill(0, ())
0-dimensional Array{Int64, 0}:
0

julia> logpdf(Normal(), x)
ERROR: MethodError: logpdf(::Normal{Float64}, ::Array{Int64, 0}) is ambiguous. Candidates:
  logpdf(d::Distribution{ArrayLikeVariate{N}}, x::AbstractArray{<:Real, N}) where N in Distributions at C:\Users\mosullivan\.julia\packages\Distributions\gggmX\src\common.jl:243
  logpdf(d::Distribution{ArrayLikeVariate{N}}, x::AbstractArray{<:Real, M}) where {N, M} in Distributions at C:\Users\mosullivan\.julia\packages\Distributions\gggmX\src\common.jl:308
  logpdf(d::UnivariateDistribution, X::AbstractArray) in Distributions at deprecated.jl:70
Possible fix, define
  logpdf(::UnivariateDistribution, ::AbstractArray{T, 0} where T<:Real)
Stacktrace:
 [1] top-level scope
   @ REPL[3]:1

julia> Distributions._logpdf(Normal(), x)
-0.9189385332046728
```

Is this the correct behaviour?

---

Julia: v1.8.1
Distributions: v0.25.79

mosullivan93:
The internal methods (`_pdf` & `_logpdf`) have implementations that support zero dimensional arrays, which are not accessible through the normal API. This PR changes this by having the behaviour defined directly on the exported methods (`pdf` and `logpdf`).

Closes #1650

### DIFF
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -311,7 +311,7 @@ See also: [`logpdf`](@ref).
 pdf(d::UnivariateDistribution, x::Real) = exp(logpdf(d, x))
 
 # extract value from array of zero dimension
-_pdf(d::UnivariateDistribution, x::AbstractArray{<:Real,0}) = pdf(d, first(x))
+pdf(d::UnivariateDistribution, x::AbstractArray{<:Real,0}) = pdf(d, first(x))
 
 """
     logpdf(d::UnivariateDistribution, x::Real)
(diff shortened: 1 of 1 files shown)


devmotion:
I think #1650 is a duplicate of https://github.com/JuliaStats/Distributions.jl/issues/1463, and hence maybe it would be better to fix the ambiguity issue of `pdf(::UnivariateDistribution, ::AbstractArray)`  instead of replacing `_pdf` and `_logpdf` with `pdf` and `logpdf`.

mosullivan93:
Yea, that makes sense. The fix for the existing issue should solve this, too. I'll close this PR and the issue.

PR #1654, opened 2022-12-27
add alpha distribution

pri1311:
Hi team, this PR aims to add the implementation for alpha distribution as mentioned in `Continuous Univariate Distributions, Volumes I and II` Book. I haven't added tests yet, but once the implementation is regarded as correct, will add the tests too.

Thank you!

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -65,6 +65,7 @@ export
     MultivariateMixture,
 
     # distribution types
+    Alpha,
     Arcsine,
     Bernoulli,
     BernoulliLogit,
--- a/src/univariate/continuous/alpha.jl
+++ b/src/univariate/continuous/alpha.jl
@@ -1,0 +1,78 @@
+"""
+    Alpha(α, β)
+
+The *Alpha distribution* with shape parameter `α` and scale parameter `β` has probability density function
+
+```math
+f(x, \\alpha) = \\frac{\\beta}{x^2 \\Phi(\\alpha) \\sqrt{2\\pi}} *
+\\exp(-\\frac{1}{2} (\\alpha-\\beta/x)^2)
+```
+where :math:`\\Phi` is the normal CDF.
+
+
+```julia
+Alpha()        # equivalent to Alpha distribution with unit shape and unit scale i.e. Alpha(1, 1)
+Alpha(α)       # equivalent to Alpha distribution with shape α and unit scale, i.e. Alpha(α, 1)
+Alpha(α, β)    # equivalent to Alpha distribution with shape α and scale β i.e. Alpha(α, β)
+
+params(d)        # Get the parameters, i.e. (α, β)
+shape(d)         # Get the shape parameter, i.e. α
+scale(d)         # Get the scale parameter, i.e. β
+```
+
+External links
+
+* [Continuous Univariate Distributions, Volumes I and II](https://www.wiley.com/en-us/Continuous+Univariate+Distributions%2C+Volume+1%2C+2nd+Edition-p-9780471584957)
+* [scipy.stats.alpha](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.alpha.html)
+
+"""
+struct Alpha{T<:Real} <: ContinuousUnivariateDistribution
+    α::T
+    β::T
+    Alpha{T}(α, β) where {T} = new{T}(α, β)
+end
+
+# as per NIST. ALPPDF. Online: https://www.itl.nist.gov/div898/software/dataplot/refman2/auxillar/alppdf.htm
+function Alpha(α::T; check_args::Bool=true) where {T <: Real}
+    @check_args Alpha (α, α > zero(α))
+    return Alpha{T}(α, 1)
+end
+
+function Alpha(α::T, β::T; check_args::Bool=true) where {T <: Real}
+    @check_args Alpha (α, α > zero(α)) (β, β > zero(β))
+    return Alpha{T}(α, β)
+end
+
+Alpha(α::Real, β::Real; check_args::Bool=true) = Alpha(promote(α, β)...; check_args=check_args)
+Alpha(α::Integer, β::Integer; check_args::Bool=true) = Alpha(float(α), float(β); check_args=check_args)
+Alpha(α::Real; check_args::Bool=true) = Alpha(α, one(α); check_args=check_args)
+Alpha() = Alpha{Float64}(1.0, 1.0)
+
+@distr_support Alpha 0.0 Inf
+
+#### Conversions
+convert(::Type{Alpha{T}}, α::S, β::S) where {T <: Real, S <: Real} = Alpha(T(α), T(β))
+Base.convert(::Type{Alpha{T}}, d::Alpha) where {T<:Real} = Alpha{T}(T(d.α), T(d.β))
+Base.convert(::Type{Alpha{T}}, d::Alpha{T}) where {T<:Real} = d
+
+#### Parameters
+
+shape(d::Alpha) = d.α
+scale(d::Alpha) = d.β
+
+params(d::Alpha) = (d.α, d.β)
+partype(::Alpha{T}) where {T} = T
+
+#### Statistics
+
+mode(d::Alpha) = ((sqrt(d.α^2 + 8) - d.α) * d.β) / 4
+
+function pdf(d::Alpha{T}, x::Real) where T<:Real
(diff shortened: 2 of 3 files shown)


src/univariate/continuous/alpha.jl
```diff
@@ -0,0 +1,78 @@
...
+External links
+
+* [Continuous Univariate Distributions, Volumes I and II](https://www.wiley.com/en-us/Continuous+Univariate+Distributions%2C+Volume+1%2C+2nd+Edition-p-9780471584957)
+* [scipy.stats.alpha](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.alpha.html)
+
+"""
+struct Alpha{T<:Real} <: ContinuousUnivariateDistribution
+    α::T
+    β::T
+    Alpha{T}(α, β) where {T} = new{T}(α, β)
```
devmotion, line 32:
```suggestion
    Alpha{T}(α::T, β::T) where {T} = new{T}(α, β)
```

src/univariate/continuous/alpha.jl
```diff
@@ -0,0 +1,78 @@
...
+end
+
+function Alpha(α::T, β::T; check_args::Bool=true) where {T <: Real}
+    @check_args Alpha (α, α > zero(α)) (β, β > zero(β))
+    return Alpha{T}(α, β)
+end
+
+Alpha(α::Real, β::Real; check_args::Bool=true) = Alpha(promote(α, β)...; check_args=check_args)
+Alpha(α::Integer, β::Integer; check_args::Bool=true) = Alpha(float(α), float(β); check_args=check_args)
+Alpha(α::Real; check_args::Bool=true) = Alpha(α, one(α); check_args=check_args)
```
devmotion, line 48: A bit simpler:
```suggestion
function Alpha(α::Real; check_args::Bool=true)
    @check_args Alpha (α, α > zero(α))
    return Alpha{typeof(α)}(α, oneunit(α))
end
function Alpha(α::Real, β::Real; check_args::Bool=true)
    @check_args Alpha (α, α > zero(α)) (β, β > zero(β))
    a, b = promote(α, β)
    return Alpha{typeof(a)}(a, b)
end
```

src/univariate/continuous/alpha.jl
```diff
@@ -0,0 +1,78 @@
...
+Alpha() = Alpha{Float64}(1.0, 1.0)
+
+@distr_support Alpha 0.0 Inf
+
+#### Conversions
+convert(::Type{Alpha{T}}, α::S, β::S) where {T <: Real, S <: Real} = Alpha(T(α), T(β))
```
devmotion, line 54:
```suggestion
```

src/univariate/continuous/alpha.jl
```diff
@@ -0,0 +1,78 @@
...
+
+function pdf(d::Alpha{T}, x::Real) where T<:Real
+    nd = Normal()
+    return pdf(nd, d.α-d.β/x) / ((x^2) * cdf(nd, d.α))
+end
+
+function cdf(d::Alpha{T}, x::Real) where T<:Real
+    nd = Normal()
+    return cdf(nd, d.α-d.β/x) / cdf(nd, d.α)
+end
```
devmotion, line 78:
```suggestion
function pdf(d::Alpha, x::Real)
    return normpdf(d.α - d.β/x) / (x^2 * normcdf(d.α))
end

function cdf(d::Alpha, x::Real)
    return normcdf(d.α - d.β/x) / normcdf(d.α)
end
```

devmotion, line 78:
```suggestion
function pdf(d::Alpha, x::Real)
    return normpdf(d.α - d.β/x) / (x^2 * normcdf(d.α))
end

function cdf(d::Alpha, x::Real)
    return normcdf(d.α - d.β/x) / normcdf(d.α)
end
```

src/univariate/continuous/alpha.jl
```diff
@@ -0,0 +1,78 @@
+"""
+    Alpha(α, β)
+
+The *Alpha distribution* with shape parameter `α` and scale parameter `β` has probability density function
+
+```math
+f(x, \\alpha, \\beta) = \\frac{\\beta}{x^2 \\Phi(\\alpha) \\sqrt{2\\pi}} *
+\\exp(-\\frac{1}{2} (\\alpha-\\beta/x)^2)
+```
+where :math:`\\Phi` is the normal CDF.
```
devmotion, line 10:
```suggestion
where ``\\Phi`` is the normal CDF.
```

src/univariate/continuous/alpha.jl
```diff
@@ -0,0 +1,78 @@
+"""
+    Alpha(α, β)
+
+The *Alpha distribution* with shape parameter `α` and scale parameter `β` has probability density function
+
+```math
+f(x, \\alpha, \\beta) = \\frac{\\beta}{x^2 \\Phi(\\alpha) \\sqrt{2\\pi}} *
+\\exp(-\\frac{1}{2} (\\alpha-\\beta/x)^2)
```
devmotion, line 8:
```suggestion
f(x; \\alpha, \\beta) = \\frac{\\beta}{\\sqrt{2\\pi} x^2 \\Phi(\\alpha)}
\\exp{\\bigg(-\\frac{(\\alpha-\\beta/x)^2}{2}\\bigg)}
```

devmotion (commented):
Thank you for the PR! I made some comments.

I think generally maybe it could be useful to precompute `normcdf(alpha)` and/or the log cdf and store it in the struct, similar to what we do for truncated distributions.
Could you also add more functions of the interface such as mean, std, variance, entropy, logpdf, logpdf, quantile etc?

pri1311:
Mean does not exist for Alpha distribution - [Ref](https://www.statisticshowto.com/alpha-distribution/#:~:text=Mean%3A%20Does%20not%20exist.)

I added the other functions as found in the paper - [Reliability Application of the Alpha Distribution](https://ieeexplore.ieee.org/abstract/document/5222136)

src/Distributions.jl
```diff
@@ -260,6 +261,7 @@ export
     suffstats,          # compute sufficient statistics
     succprob,           # the success probability
     support,            # the support of a distribution (or a distribution type)
+    survivor_function,  # complementary cumulative distribution function
```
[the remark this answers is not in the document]
pri1311: Aah yes!

src/univariate/continuous/alpha.jl
```diff
@@ -0,0 +1,86 @@
...
+
+params(d::Alpha) = (d.α, d.β)
+partype(::Alpha{T}) where {T} = T
+
+#### Statistics
+
+mode(d::Alpha) = ((sqrt(d.α^2 + 8) - d.α) * d.β) / 4
+
+function pdf(d::Alpha, x::Real)
+    return d.β * normpdf(d.α - d.β/x) / (x^2 * normcdf(d.α))
```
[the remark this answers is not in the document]
pri1311: but since I've added `@distr_support Alpha 0.0 Inf` that we only support from 0, is it necessary to check here?

src/univariate/continuous/alpha.jl
```diff
@@ -0,0 +1,86 @@
...
+end
+
+function cdf(d::Alpha, x::Real)
+    return normcdf(d.α - d.β/x) / normcdf(d.α)
+end
+
+logpdf(d::Alpha, x::Real) = d.β - (2 * log(x)) + normlogpdf(d.α - d.β/x) - normlogcdf(d.α)
+logcdf(d::Alpha, x::Real) = log(cdf(d, x))
```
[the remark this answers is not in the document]
pri1311: I am thrown off a little here, I don't quite understand the reason behind doing it in this way, especially the
> v = normlogcdf(d, max(d.α - d.β/x, d.α))

Could you explain, TIA.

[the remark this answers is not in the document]
devmotion: The general structure is the same also used in e.g. `logdiffcdf`: It is numerically better to use `u + log1mexp(v - u)` instead of `log(exp(u) - exp(v))` where `u >= v`.

Regarding the definition of `v`: Again we have to support arbitrary `x`, it's not possible to assume that `x >= 0`. By defining `v` in that way, the result should be correct also for `x < 0` (if I'm not mistaken). In that case we will just have `v = u`, and hence `u + log1mexp(v - u) = -Inf` (of possibly different type) as desired.

PR #1655, opened 2022-12-28
New distribution: Maximum

Vilin97:
A bare bones implementation of the distribution of maximum of n iid random variables.


sethaxen:
I like the idea of having such a distribution. I imagine though that users would find an analogous `Minimum`  or `Median` useful. Which makes me wonder if it would be better to add a general `OrderStatistic(dist, n, i)` distribution that takes a parameter `n` (sample size) and a parameter `i` (order statistic). `Maximum` would correspond to `OrderStatistic(dist, n, n)`, while `Minimum` would be `OrderStatistic(dist, n, 1)`, and `Median` would be `OrderStatistic(dist, n, fld(n, 2))`. One could even have a `Quantile(dist, n, p)` alias that is just `OrderStatistic(dist, n, floor(Int, p*n))`. Though I'm in general not a fan of things that look like constructors returning other objects.

Vilin97:
Thank you for your interest, @sethaxen! I agree that `OrderStatistic` would be a good generalization but the [formulas](https://en.wikipedia.org/wiki/Order_statistic#Cumulative_distribution_function_of_order_statistics) for its cdf and pdf are considerably more complicated. AFAIK, there is no closed form formula for the inverse of the cdf. Therefore, I don't think it makes sense to implement `OrderStatistic`.

sethaxen:
> Thank you for your interest, @sethaxen! I agree that `OrderStatistic` would be a good generalization but the [formulas](https://en.wikipedia.org/wiki/Order_statistic#Cumulative_distribution_function_of_order_statistics) for its cdf and pdf are considerably more complicated.

They seem complicated, but they're much simpler when written as a function of the PDF and CDF of the binomial distribution:

```julia
function orderstat_pdf(dist, n, r, x)
    d = Binomial(n - 1, cdf(dist, x))
    return n * pdf(d, r - 1) * pdf(dist, x)
end

function orderstat_cdf(dist, n, r, x)
    d = Binomial(n, cdf(dist, x))
    return cdf(d, n) - cdf(d, r - 1)
end
```

> AFAIK, there is no closed form formula for the inverse of the cdf. Therefore, I don't think it makes sense to implement `OrderStatistic`.

Several other distributions in this package don't have closed-form expressions for the quantile function e.g.
https://github.com/JuliaStats/Distributions.jl/blob/432a7f93c0cb096b1852b01ef0abd678eba1bc60/src/mixtures/mixturemodel.jl#L444-L448
There are several quantile algorithms in https://github.com/JuliaStats/Distributions.jl/blob/master/src/quantilealgs.jl that can be used here. There's likely a way to set absolute lower and upper bounds for the quantile for `quantile_bisect`.

sethaxen:
Hi @Vilin97 are you interested in tackling this more general `OrderStatistics` distribution?

Vilin97:
> Hi @Vilin97 are you interested in tackling this more general `OrderStatistics` distribution?

I'm not, sorry! I understand the desire for more generality but in this case it will come at the expense of less certainty, e.g. with quantiles. I will leave the implementation of orderStatistics to someone else.

src/univariate/continuous/maximum.jl
```diff
@@ -0,0 +1,25 @@
...
+
+#### Evaluation
+
+cdf(d::Maximum{D}, x::Real) where {D} = cdf(d.dist, x)^d.n
+pdf(d::Maximum{D}, x::Real) where {D} = d.n*pdf(d.dist, x)*cdf(d.dist, x)^(d.n-1)
+logpdf(d::Maximum{D}, x::Real) where {D} = log(d.n)+logpdf(d.dist, x)+(d.n-1)*logcdf(d.dist, x)
+quantile(d::Maximum{D}, q::Real) where {D} = quantile(d.dist, q^(1/d.n))
+minimum(d::Maximum{D}) where {D} = minimum(d.dist)
+maximum(d::Maximum{D}) where {D} = maximum(d.dist)
+insupport(d::Maximum{D}, x::Real) where {D} = insupport(d.dist, x)
```
sethaxen, line 25: No need for `D`:
```suggestion
minimum(d::Maximum) = minimum(d.dist)
maximum(d::Maximum) = maximum(d.dist)
insupport(d::Maximum, x::Real) = insupport(d.dist, x)
```

src/univariate/continuous/maximum.jl
```diff
@@ -0,0 +1,25 @@
...
+end
+
+rand(rng::AbstractRNG, d::Maximum{D}) where {D} = maximum([rand(rng, d.dist) for _ in 1:d.n])
+
+#### Evaluation
+
+cdf(d::Maximum{D}, x::Real) where {D} = cdf(d.dist, x)^d.n
+pdf(d::Maximum{D}, x::Real) where {D} = d.n*pdf(d.dist, x)*cdf(d.dist, x)^(d.n-1)
+logpdf(d::Maximum{D}, x::Real) where {D} = log(d.n)+logpdf(d.dist, x)+(d.n-1)*logcdf(d.dist, x)
+quantile(d::Maximum{D}, q::Real) where {D} = quantile(d.dist, q^(1/d.n))
```
sethaxen, line 22: No need for `D`, and we can use `//` to avoid unnecessarily promoting to a `Float64`. Also, I'd have to think more about how if this needs to be changed to handle the discrete case.
```suggestion
quantile(d::Maximum, q::Real) = quantile(d.dist, q^(1//d.n))
```

sethaxen (changes requested):
On further thought, even if we have an `OrderStatistic` distribution, it makes sense to have independent `Maximum` and `Minimum` distributions. I'll tackle the other two once this is finished, but I have some suggestions:

src/univariate/continuous/maximum.jl
```diff
@@ -0,0 +1,25 @@
...
+    n::Int
+    Maximum{D}(dist, n) where {D<:ContinuousUnivariateDistribution} = new{D}(dist, n)
+end
+
+function Maximum(dist::D, n::Integer; check_args::Bool=true) where {D <: ContinuousUnivariateDistribution}
+    @check_args Maximum (n, n >= one(n))
+    return Maximum{D}(dist, n)
+end
+
+rand(rng::AbstractRNG, d::Maximum{D}) where {D} = maximum([rand(rng, d.dist) for _ in 1:d.n])
```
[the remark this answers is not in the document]
sethaxen: Even better, since `rand(Beta(n, 1)` is equivalent to `rand()^(1//n)`
```suggestion
rand(rng::AbstractRNG, d::Maximum) = maximum(rand(rng, d.dist) for _ in 1:d.n)
function rand(rng::AbstractRNG, d::Maximum{<:ContinuousUnivariateDistribution})
    return quantile(d.dist, rand(rng)^(1//d.n))
end
```

sethaxen:
> ...in this case it will come at the expense of less certainty, e.g. with quantiles.

Interestingly, https://epubs.siam.org/doi/10.1137/1.9780898719062.ch4, they point out that the `p`th quantile of the order statistic `i` for sample size `n` can be computed as `quantile(d.dist, quantile(Beta(i, n-i+1), p))`.

src/univariate/continuous/maximum.jl
sethaxen: It seems nothing special needs to be done, see https://github.com/JuliaStats/Distributions.jl/pull/1655#issuecomment-1407654229.

Vilin97:
In light of https://github.com/JuliaStats/Distributions.jl/pull/1668 is this now obsolete, @sethaxen ? I have not had the time to address your comments but if it's not obsolete, I will address your comments within a week to make this merge-ready.

sethaxen:
> In light of #1668 is this now obsolete, @sethaxen ?

I'm not certain. I just benchmarked, and `pdf`, `cdf`, and `logpdf` in this PR are all 1.5-3x faster than representing the distribution of the maximum using `OrderStatistic`, while `quantile` is ~20x faster. I don't know whether it's better to have a custom type for `Maximum` or just to special-case the maximum (and minimum) in these functions in `OrderStatistic`.

PR #1664, opened 2023-01-21
SkewedExponentialPower, PGeneralizedGaussian: reproducible rand

### LINKED ISSUE #3: Should the Link type be part of Distributions?
Link functions for generalized linear models are, to some extent, connected with distributions.  In particular, distributions in the exponential family have canonical link functions.  In R a object in the "family" class is generated from a distribution and, optionally, the name of a link.

``` bash
> class(binomial())
[1] "family"
> binomial()

Family: binomial 
Link function: logit 

> binomial("probit")

Family: binomial 
Link function: probit 

> poisson()

Family: poisson 
Link function: log 

> gaussian()

Family: gaussian 
Link function: identity 

> Gamma()

Family: Gamma 
Link function: inverse 
```

Currently I have the `Link` types in the Glm package (not yet a package but it will be) but that leaves methods for `canonicallink(d::Distribution)` being defined in `Glm.jl` whereas these really are properties of the distribution.  Would it be too confusing to incorporate the `Link` abstract type and the concrete Link types in the `Distributions` module?

I have the feeling that we may end up with too many interlinking parts if we create fine-grained packages and we may be better off using fewer but more coarse-grained packages, which is why I would like to move many things that may be associated with probability distributions into this package.

### LINKED ISSUE #1663: Irreproducible sampling results for PGeneralizedGaussian and SkewedExponentialPower
The samples drawn from the above distributions are not reproducible, even when the same random number generator is used:

```julia
julia> using Random, Distributions

julia> hcat(rand(Xoshiro(42), PGeneralizedGaussian(), (10,1)), rand(Xoshiro(42), PGeneralizedGaussian(), (10,1)))
10×2 Matrix{Float64}:
  0.725982  -0.725982
  0.58417    0.58417
  0.054883  -0.054883
  0.360157  -0.360157
  0.363709  -0.363709
  0.724104   0.724104
 -1.13497   -1.13497
  0.213059   0.213059
  0.122888  -0.122888
  0.855553  -0.855553
```

and likewise

```julia
julia> hcat(rand(Xoshiro(42), SkewedExponentialPower(), (10,1)), rand(Xoshiro(42), SkewedExponentialPower(), (10,1)))
10×2 Matrix{Float64}:
  2.19035     0.691078
 -1.43392    -0.99833
 -0.853063   -0.379221
  1.31484     0.514229
  0.663437    0.402819
 -0.701881   -1.01172
  0.0537253   0.167032
  0.555474    0.435179
 -0.940844   -0.172264
 -0.951032   -1.85481
```

It happens because the rng is not passed to some of the internal rand calls (like the one responsible for the sign in the first case above). I will make a PR that fixes this.

philbit:
Fixes #1663


philbit:
After this change, the behavior mentioned in #1663 changes to the expected outcome:

```julia
julia> hcat(rand(Xoshiro(42), SkewedExponentialPower(), (10,1)), rand(Xoshiro(42), SkewedExponentialPower(), (10,1)))
10×2 Matrix{Float64}:
  0.327906   0.327906
  0.649174   0.649174
 -0.60571   -0.60571
 -1.32062   -1.32062
 -1.02497   -1.02497
 -0.358321  -0.358321
 -0.931176  -0.931176
 -0.566247  -0.566247
  1.06089    1.06089
  1.41578    1.41578

julia> hcat(rand(Xoshiro(42), PGeneralizedGaussian(), (10,1)), rand(Xoshiro(42), PGeneralizedGaussian(), (10,1)))
10×2 Matrix{Float64}:
 -0.194974  -0.194974
 -0.386001  -0.386001
  0.360157   0.360157
  0.785246   0.785246
  0.609452   0.609452
  0.213059   0.213059
  0.55368    0.55368
  0.336693   0.336693
 -0.630806  -0.630806
 -0.841828  -0.841828
```

src/univariate/continuous/pgeneralizedgaussian.jl
```diff
@@ -127,7 +127,7 @@ function rand(rng::AbstractRNG, d::PGeneralizedGaussian)
     g = Gamma(inv(d.p), 1)
 
     # random variable with value -1 or 1 with probability (1/2).
-    b = 2.0 * rand(Bernoulli()) -1
+    b = 2.0 * rand(rng, Bernoulli()) -1
```
devmotion, line 130: Usually, `Float64` literals should be avoided. Not sure if it's relevant here but no harm either to follow this guideline:
```suggestion
    b = 2 * rand(rng, Bernoulli()) -1
```

devmotion (commented):
Thank you! Could you add some tests for these changes as well such that the bug does not reappear accidentally?

src/univariate/continuous/pgeneralizedgaussian.jl
philbit: Sure, why not. I didn't change anything other than what pertains to the specific issue so far.

philbit:
> Thank you! Could you add some tests for these changes as well such that the bug does not reappear accidentally?

Sure, that's a good idea. However, I looked around and couldn't find any other distribution that does this (maybe I didn't look hard enough though). Since this problem potentially applies to every single distribution, I am wondering what to do.

1. Add this kind of test for these two distributions. But does this make sense simply because these two happened to have this problem?
2. Add a function to `test/testutils.jl` that does this kind of reproducibility check for any distribution, but only add it to the test files for these two distributions and deal with other distributions later (where the new function could then be used).
3. Find a function that many tests of other distributions use already anyway (if it exists) and add the check to that function, so that it is automatically applied to a broad range of distributions. Does such a function exist, or, in other words, which one would you recommend?

There is a fourth option which is probably the "right" thing to do, which is to make sure that sample reproducibility is tested for *all* distributions. But to be honest, I feel like a small bug fix pull request is not the right place for that. Or what do you think?

devmotion:
I added a reproducible check to `test_samples` in testutils.jl. This function is already used to check sampling of all univariate distributions that are compared with R in test/univariates.jl (and some other distributions, I think). This function also checks if the samples seem to be distributed according to the desired distribution - and adding these checks to the tests of `PGeneralizedGaussian` and `SkewedExponentialPower` revealed that `rand` for `PGeneralizedGaussian` is implemented incorrectly. I fixed this and some other things in `PGeneralizedGaussian`, added comparisons with R, and added more (IMO better) tests. Probably the tests for `SkewedExponentialPower` should be updated in a similar way but since `rand` seems to be correct for that distribution I did  not do that in this PR.

philbit:
Thanks a lot!

Issue #1643, opened 2022-12-01
New distribution: maximum


Vilin97:
Would `Distributions.jl` be the right place for the maximum of n iid random variables? The cdf and pdf are easy to implement, as well as `rand`, and others listed [here](https://juliastats.org/Distributions.jl/stable/extends/#Univariate-Distribution). 

To be precise, given `n` iid random variables `X_1`, ..., `X_n`, I am talking about the distribution of the random variable `M_n = maximum([X_1,...,X_n])`.

ParadaCarleton:
Yes, this would be great! Feel free to make a pull request implementing this!

PR #1668, opened 2023-01-30
Add OrderStatistic and JointOrderStatistics distributions

### LINKED ISSUE #1284: Adding order statistics sampling
Hello,

I've implemented sampling from the order statistics of an arbitrary sampleable. Are you interested in merging this kind of functionality into Distributions.jl? 

The `k`-th order statistic is the `k`-th smallest value out of `n` (`n >= k`) variables. I'm using order statistics to model latency in distributed computing (the overall latency is the `n`-th order statistic of the latency of the `n` workers). Finding the distribution of the order statistics is often non-trivial. An implementation like this is a useful fallback.

I'm attaching a prototype implementation.

```julia
struct OrderStatistic{S<:Union{Discrete,Continuous},Spl<:Sampleable{Univariate,S},T} <: Sampleable{Univariate,S}
    spl::Spl
    k::Int
    buffer::Vector{T}
end

Base.show(io::IO, s::OrderStatistic) = print(io, "OrderStatistic($(s.spl), k=$(s.k), n=$(length(s.buffer)))")

OrderStatistic(s::Sampleable, k::Integer, n::Integer) = OrderStatistic(s, k, Vector{eltype(s)}(undef, n))

function rand(rng::AbstractRNG, s::OrderStatistic)
    Distributions.rand!(rng, s.spl, s.buffer)
    partialsort!(s.buffer, s.k)
    s.buffer[s.k]
end
```

sethaxen:
This PR will add two new distributions:
- `OrderStatistic`: a univariate distribution representing the distribution of the rank `i`th draw in an IID sample of length `n` from some (continuous or discrete) univariate distribution `dist`.
- `JointOrderStatistics`: a continuous multivariate distribution representing the joint distribution of the rank `r` draws in the same sample.

These have a number of uses. For example, they can represent transformations that have been performed to data (e.g. outlier removal and calculation of summary statistics). They also can be used to draw confidence intervals, e.g. of ECDF plots. For `JointOrderStatistics`, in the extreme case where `r=1:n`, i.e. when the distribution is over all ranks, it can be used in a PPL like Turing to restrict an IID array of parameters to the subset of arrays that are ordered.

Relates #1643 #1655, closes #1284


sethaxen:
I've only seen the expression of the density of joint order statistics `1:n` or `[i,j]`, so for completeness, here's the derivation for the more general case used in `JointOrderStatistics`:

<details>

<summary>
<b>Derivation</b>
</summary>

Given the density function $f_X$ of the sample distribution, with CDF $F_X$, the joint density for all order statistics is
$$f_{1\ldots n:n}(x_1, \ldots, x_n) = n! \prod_{k=1}^n f_X(x_k)$$

Suppose now we marginalize all $x_k$ for $i < k < j$, where $i$ and $j$ are integers in $1\ldots n$. The resulting density is
$$f_{1\ldots i,j\ldots n:n}(x_1, \ldots, x_i, x_j, \ldots x_n) = n! \left(\prod_{k=1}^i f_X(x_k)\right) \left(\prod_{k=j}^n f_X(x_k)\right) C_{ij}(x_i, x_j),$$
where
$$C_{ij}(x_i, x_j) = \int_{x_i}^{x_j} \cdots \int_{x_i}^{x_{i+3}}\int_{x_i}^{x_{i+2}} f_X(x_{i+1}) f_X(x_{i+2}) \cdots f_X(x_{j-1}) \mathrm{d}x_{i+1} \mathrm{d}x_{i+2} \cdots \mathrm{d}x_{j-1}.$$

Let $u_k = F_X(x_k) - F_X(x_i)$. Then $\mathrm{d}u_k = f_X(x_k) \mathrm{d}x_k$, and
$$C_{ij}(x_i, x_j) = \int_{0}^{u_j} \cdots \int_{0}^{u_{i+3}}\int_{0}^{u_{i+2}} \mathrm{d}u_{i+1} \mathrm{d}u_{i+2} \cdots \mathrm{d}u_{j-1} = \frac{u_j^{j-i-1}}{(j-i-1)!} = \frac{(F_X(x_j) - F_X(x_i))^{j-i-1}}{(j-i-1)!}$$

We can repeat this to get the joint density of any subset of order statistics.

</details>

src/multivariate/jointorderstatistics.jl
```diff
@@ -0,0 +1,132 @@
...
+For a single order statistic, use [`OrderStatistic`](@ref) instead.
+
+## Examples
+
+```julia
+JointOrderStatistics(Normal(), 10)           # Product(fill(Normal(), 10)) restricted to ordered vectors
+JointOrderStatistics(Cauchy(), 10, [1, 10])  # joint distribution of the extrema
+```
+"""
+struct JointOrderStatistics{D<:ContinuousUnivariateDistribution,R<:AbstractVector{Int}} <:
```
sethaxen, line 34: For the continuous case, we can write down the PDF and not much else (except the mean and covariance for a handful of `dist`s, for a future PR).

For the discrete case, the PDF is much more complicated and probably only tractable when `length(r) <= 2`.  For a discrete `dist` whose support is a finite set, we can define a distribution of the order statistics for a sample _without_ replacement, at least in the bivariate case. The PMF, mean, and covariance are known. I wonder if that should be a special case of this distribution.

sethaxen (commented):
I haven't added them to the docs yet. Univariate distributions are currently separated into Continuous and Discrete categories, but `OrderStatistic` can be either. Plus, `JointOrderStatistics` and `OrderStatistic` are so closely related, it might make sense to give them their own docs page. Could even be a page devoted to distributions of statistics of samples, in case in the future someone adds asymptotic distributions of statistics like quantiles, etc.

sethaxen:
@devmotion or whoever, this is now ready for initial review.

sethaxen:
Bump @devmotion or maybe @andreasnoack

devmotion:
Ah sorry, missed your last ping. I'll have a look at the PR this weekend.

devmotion:
Ah missed it again :see_no_evil: Well, hopefully in the next few days :crossed_fingers:

sethaxen:
> Ah missed it again see_no_evil Well, hopefully in the next few days crossed_fingers

No worries! I start traveling tomorrow, so it's unlikely I will address any review for several weeks, so whenever is convenient for you.

sethaxen:
@devmotion I would have time for the next couple of weeks to finish this up, in case you have time for a review.

src/multivariate/jointorderstatistics.jl
sethaxen: > I wonder if that should be a special case of this distribution.

Nah, seems complicated and not all that useful.

sethaxen:
Bump @devmotion

src/multivariate/jointorderstatistics.jl
```diff
@@ -0,0 +1,132 @@
...
+    length(r) == n && return lp
+    i = 0
+    xᵢ = oftype(float(first(x)), -Inf)
+    for (j, xⱼ) in zip(r, x)
+        lp += _marginalize_range(d.dist, n, i, j, xᵢ, xⱼ, T)
+        i = j
+        xᵢ = xⱼ
+    end
+    j = n + 1
+    xⱼ = oftype(xᵢ, Inf)
```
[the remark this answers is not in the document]
devmotion: Again, maybe just handle this separately, ie., not with `_marginalize_range` that takes `x_i` and `x_j`.

src/multivariate/jointorderstatistics.jl
```diff
@@ -0,0 +1,132 @@
...
+    else
+        T(logdiffcdf(dist, xⱼ, xᵢ))
+    end
+    return k * lpdiff - loggamma(T(k + 1))
+end
+
+function _rand!(rng::AbstractRNG, d::JointOrderStatistics, x::AbstractVector{<:Real})
+    n = d.n
+    if n == length(d.r)  # r == 1:n
+        # direct method, slower than inversion method for large `n` and distributions with
```
devmotion, line 106: Seems we should maybe use some heuristic for choosing one of the two algorithms, depending on `n` and possibly the type of distribution?

src/multivariate/jointorderstatistics.jl
```diff
@@ -0,0 +1,132 @@
...
+        sort!(x)
+    else
+        # use exponential generation method with inversion, where for gaps in the ranks, we
+        # use the fact that the sum Y of k IID variables xₘ ~ Exp(1) is Y ~ Gamma(k, 1).
+        # Lurie, D., and H. O. Hartley. "Machine-generation of order statistics for Monte
+        # Carlo computations." The American Statistician 26.1 (1972): 26-27.
+        # this is slow if length(d.r) is close to n and quantile for d.dist is expensive,
+        # but this branch is probably taken when length(d.r) is small or much smaller than n.
+        T = typeof(one(eltype(x)))
+        s = zero(one(T))
```
devmotion, line 118: That is, this is
```suggestion
        s = zero(eltype(x))
```
?

src/multivariate/jointorderstatistics.jl
```diff
@@ -0,0 +1,132 @@
...
+        i = 0
+        for (m, j) in zip(eachindex(x), d.r)
+            k = j - i
+            s += k > 1 ? rand(rng, Gamma(k, one(T); check_args=false)) : randexp(rng, T)
```
devmotion, line 122: Is this type stable? `eltype`, ie. element type of `rand`, is supposed to be not (necessarily) dictated by the type of the parameters.

src/univariate/orderstatistic.jl
```diff
@@ -0,0 +1,106 @@
...
+OrderStatistic(Cauchy(), 10, 1)              # distribution of the sample minimum
+OrderStatistic(DiscreteUniform(10), 10, 10)  # distribution of the sample maximum
+OrderStatistic(Gamma(1, 1), 11, 5)           # distribution of the sample median
+```
+"""
+struct OrderStatistic{D<:UnivariateDistribution,S<:ValueSupport} <:
+       UnivariateDistribution{S}
+    dist::D
+    n::Int
+    i::Int
```
devmotion, line 45: Not sure if there are more descriptive alternatives to the name `i`.

src/univariate/orderstatistic.jl
```diff
@@ -0,0 +1,106 @@
...
+    end
+end
+
+minimum(d::OrderStatistic) = minimum(d.dist)
+maximum(d::OrderStatistic) = maximum(d.dist)
+insupport(d::OrderStatistic, x::Real) = insupport(d.dist, x)
+
+params(d::OrderStatistic) = tuple(params(d.dist)..., d.n, d.i)
+partype(d::OrderStatistic) = partype(d.dist)
+Base.eltype(::Type{<:OrderStatistic{D}}) where {D} = Base.eltype(D)
```
devmotion, line 60: Same here, maybe extend it to
```suggestion
Base.eltype(::Type{<:OrderStatistic{D}}) where {D} = Base.eltype(D)
Base.eltype(d::OrderStatistic) = eltype(d.dist)
```

src/univariate/orderstatistic.jl
```diff
@@ -0,0 +1,106 @@
+# Implementation based on chapters 2-4 of
+# Arnold, Barry C., Narayanaswamy Balakrishnan, and Haikady Navada Nagaraja.
+# A first course in order statistics. Society for Industrial and Applied Mathematics, 2008.
+
+"""
+    OrderStatistic{D<:UnivariateDistribution,S<:ValueSupport} <: UnivariateDistribution{S}
+
+The distribution of an order statistic from IID samples from a univariate distribution
```
devmotion, line 8:
```suggestion
The distribution of an order statistic from IID samples from a univariate distribution.
```

src/multivariate/jointorderstatistics.jl
```diff
@@ -0,0 +1,132 @@
...
+        n::Int,
+        r::AbstractVector{Int}=1:n;
+        check_args::Bool=true,
+    )
+        @check_args(
+            JointOrderStatistics,
+            (n, n ≥ 1, "`n` must be a positive integer."),
+            (
+                r,
+                1 ≤ first(r) && last(r) ≤ n && issorted(r) && allunique(r),
```
devmotion, line 50: This could be checked more efficiently by just iterating through `r` once.

devmotion (commented):
Sorry for the embarassingly long delay, I had completely forgotten this PR (again).

I made a few comments. I also wonder if we could add a comparison with existing R implementations, to exploit additionallu the existing test infrastructure and run standardized tests for evaluations and sampling?

> Plus, JointOrderStatistics and OrderStatistic are so closely related, it might make sense to give them their own docs page

I agree, I think a separate page in the docs could be useful.

src/univariate/orderstatistic.jl
sethaxen: We could go with `rank` instead.

src/multivariate/jointorderstatistics.jl
sethaxen: I'm not certain what heuristic we could use here, since which strategy is faster strongly depends on the relative speed of the quantile function and the `rand` method of the wrapped distribution. Also `n`, since the direct method has an `O(nlog(n))` sort step that for `n` large enough will make it slower than the exponential method, which is always `O(n)`, though potentially with a large constant factor.

src/multivariate/jointorderstatistics.jl
sethaxen: Good point, and it turns out that `issorted` can handle this if we just provide a function to `lt` that is also adds an equality check; this is much more efficient than calling `allunique`.

src/univariate/orderstatistic.jl
devmotion: I guess similarly we could even rename the field `r` in `JointOrderStatistics` to `ranks`?

sethaxen:
> Sorry for the embarassingly long delay, I had completely forgotten this PR (again).

No problem!

> I also wonder if we could add a comparison with existing R implementations, to exploit additionallu the existing test infrastructure and run standardized tests for evaluations and sampling?

Unfortunately I don't think there are any standard R implementations of distributions of order statistics against which we can compare.

> I agree, I think a separate page in the docs could be useful.

Done!

src/multivariate/jointorderstatistics.jl
```diff
@@ -0,0 +1,156 @@
...
+end
+function _are_ranks_valid(ranks::AbstractRange, n)
+    first(ranks) ≥ 1 || return false
+    last(ranks) ≤ n || return false
+    return issorted(ranks) && allunique(ranks)
+end
+
+length(d::JointOrderStatistics) = length(d.ranks)
+function insupport(d::JointOrderStatistics, x::AbstractVector)
+    return length(d) == length(x) && issorted(x) && all(Base.Fix1(insupport, d.dist), x)
```
devmotion, line 79: Also here, I think one loop over `x` would be sufficient, e.g.:
```suggestion
    length(d) == length(x) || return false
    xi, state = iterate(x) # at least one element!
    dist = d.dist
    insupport(dist, xi) || return false
    while (xj_state = iterate(x, state)) !== nothing
        xj, state = xj_state
        xj ≥ xi && insupport(dist, xj) || return false
        xi = xj
    end
    return true       
```

src/multivariate/jointorderstatistics.jl
devmotion: Hmm, no I'm not aware of such an example. Might be good enough for now.

src/multivariate/jointorderstatistics.jl
```diff
@@ -0,0 +1,156 @@
...
+        # Carlo computations." The American Statistician 26.1 (1972): 26-27.
+        # this is slow if length(d.ranks) is close to n and quantile for d.dist is expensive,
+        # but this branch is probably taken when length(d.ranks) is small or much smaller than n.
+        T = typeof(one(eltype(x)))
+        s = zero(eltype(x))
+        i = 0
+        for (m, j) in zip(eachindex(x), d.ranks)
+            k = j - i
+            if k > 1
+                s += T(rand(rng, GammaMTSampler(Gamma{T}(T(k), T(1)))))
```
devmotion, line 139: Is this the method that would be picked by calling `rand(Gamma(...))` directly?

sethaxen: It is, but only after checking the shape `k`, which we've already done:
https://github.com/JuliaStats/Distributions.jl/blob/ef42afb215cb08304dd64e309da8002c333caef0/src/univariate/continuous/gamma.jl#L103-L113
Plus, I benchmarked all samplers that apply for `Gamma(k, 1)` with `k>1`, and this one is the fastest.

devmotion: Maybe add a short comment in the code?

devmotion (approved):
Great PR, looks good to me @sethaxen!

I spotted only a few final things that, IMO, should be addressed before merging the PR. But I already approve now to indicate that it seems fine to me otherwise.

sethaxen:
I realized that when we evaluated `pdf(::JointOrderStatistics, x)` for unsorted `x`, an error would be raised by `logdiffcdf`, so I fixed that. This should now be ready to merge.

devmotion:
Great! I'll merge it within the next days, if there are no objections.

Issue #1574, opened 2022-06-29
fit_mle(Type, y) vs fit_mle(Distribution, y)


dmetivie:
I opened [this issue](https://discourse.julialang.org/t/distributions-fit-mle-of-distribution-type-vs-distributions-with-parameters/75228) a while ago but was not satisfied with the answer. I tried workarounds but it seems just not right.

Basically, I don't understand the benefits of having 1)`fit_mle(Normal, y)` work and not 2) `fit_mle(Normal(0,1), y)`. Why not allow both syntaxes?

Sure in case 2) parameters are useless. However, in more complexes `mle` estimations 1) is limited while 2) is not.
For example, in the package [ExpectationMaximization.jl](https://github.com/dmetivie/ExpectationMaximization.jl) I implemented `fit_mle` for `MixtureModels` and for `Product` distributions. 
The `fit_mle` of product distribution is simply the product of the `mle`. 
Say you have
`dist = product_distribution([Normal(1), Exponential(1)])`.
Having
`fit_mle(typeof(dist), y)` does not say which distributions are the elements of the product (`Normal`, `Exponential` etc.). 
While having `fit_mle(dist, y)` makes it possible.

Similarly `fit_mle(dist::MixtureModel,y)` needs the element of the mixture and the type `MixtureModel` is not enough as an information. 
Plus in that case, the actual parameters of `dist` are not useless since they can be used as starting point of the EM algorithm used to find the `mle`.

Note that in this package I need (for now) to add 
```
function fit_mle(g::Distribution, args...)
    fit_mle(typeof(g), args...)
end
```
for my sintaxte to work with `Disrtibutions.fit_mle`.

timholy:
Related: when fitting, e.g., `MvNormal`, sometimes it might be desirable to use [shrinkage](https://github.com/mateuszbaran/CovarianceEstimation.jl). Consequently it might be nice to have an interface accepting a `CovarianceEstimator`.

That said, there are counterarguments:
- if you're using shrinkage, are you really doing MLE? Technically not, I think.
- I also agree with the perspective in https://github.com/JuliaStats/Distributions.jl/pull/1670#discussion_r1092571161

ForceBru:
> I think if desirable/necessary such estimation/optimization settings should be specified explicitly, e.g. as (mandatory?) keyword arguments to `fit_mle`. Specifying them via a distribution instance feels too implicit and magical to me.

I think this is as explicit as possible:

[...]

PR #1670, opened 2023-01-31
Add fit_mle methods

dmetivie:
Hello,
I open this PR after asking a [question on discourse](https://discourse.julialang.org/t/distributions-fit-mle-of-distribution-type-vs-distributions-with-parameters/75228) and an [issue](https://github.com/JuliaStats/Distributions.jl/issues/1574) (more like feature enhancement) here. I was not too convinced by the answer on Discourse, so I decided to show you directly what I meant in this PR.

Here are the additions of this PR

1. **Main one** add `fit_mle(dist::Distribution, x)` instead of just `fit_mle(::Type{<:Distribution}, x)`.
    Why?
    - Why not: From the tests I ran, it does not break anything. If `fit_mle(::Type{<:Distribution}, x)` is available, then it will use it.
    - It allows to fully use multiple dispatch of `fit_mle` as all “structure parameters” are now stored in `dist` and not as extra parameters. E.g. `fit_mle(Binomial(n, 0.1), x)` is `fit_mle(Binomial, ntrials(Binomial(n, 0.1)), x)`. Of course, the 0.1 is not used and is kind of useless here.
    - This `dist` can be used when needed as a starting point for a maximum likelihood estimation.
    - This could maybe benefit many downstream pkg using `fit_mle` for complex distributions like [Copulas.jl](https://github.com/lrnv/Copulas.jl/blob/08aca27dc0a28e4932e5e2c0dd04482bb3b04f48/test/some_tests.jl#L8), [ExpectationMaximization.jl](https://github.com/dmetivie/ExpectationMaximization.jl) or [HMMBase.jl](https://github.com/maxmouchet/HMMBase.jl).
2. Point 1. allows adding very easily a `fit_mle` method for Product distributions.
    `fit_mle(product_distribution([Exponential(0.5), Normal(11.3, 3.2)]),x)` is simply `product_distribution([fit_mle(Exponential, x[1,:]), fit_mle(Normal, x[2, :])])`
    I am not too sure how it would be done only with types. Thought this seems to be done in [Copulas.jl](https://github.com/lrnv/Copulas.jl/blob/08aca27dc0a28e4932e5e2c0dd04482bb3b04f48/test/some_tests.jl#L8) (notation becomes quite heavy), but again not sure how one would infer the components of a mixtures distribution or a product distribution.
3. Add `fit_mle(Laplace, x, w)` using the weighted `median`
4. Add `fit_mle(Dirac, x[,w])` (why not).
5. I added test for all the new things and doc.
6. I added in [the estimation of mixture section](https://juliastats.org/Distributions.jl/stable/mixture/#Estimation) a ref to my pkg [ExpectationMaximization.jl](https://github.com/dmetivie/ExpectationMaximization.jl) which fits mixture models (and thanks to `fit_mle(dist::Distribution, x)` is able to use EM algorithm on very generic distribution e.g., mixture of mixtures).

I tried to follow as much as I understood the contributions guidelines (my VSCode formatter changed quite a bit of things).

### DIFF
--- a/docs/src/fit.md
+++ b/docs/src/fit.md
@@ -61,6 +61,7 @@ The `fit_mle` method has been implemented for the following distributions:
 - [`Multinomial`](@ref)
 - [`MvNormal`](@ref)
 - [`Dirichlet`](@ref)
+- [`Product`](@ref)
 
 For most of these distributions, the usage is as described above. For a few special distributions that require additional information for estimation, we have to use a modified interface:
 
--- a/docs/src/fit.md
+++ b/docs/src/fit.md
@@ -42,6 +42,7 @@ The `fit_mle` method has been implemented for the following distributions:
 - [`Binomial`](@ref)
 - [`Categorical`](@ref)
 - [`DiscreteUniform`](@ref)
+- [`Dirac`](@ref)
 - [`Exponential`](@ref)
 - [`LogNormal`](@ref)
 - [`Normal`](@ref)
@@ -60,6 +61,7 @@ The `fit_mle` method has been implemented for the following distributions:
 - [`Multinomial`](@ref)
 - [`MvNormal`](@ref)
 - [`Dirichlet`](@ref)
+- [`Product`](@ref)
 
 For most of these distributions, the usage is as described above. For a few special distributions that require additional information for estimation, we have to use a modified interface:
 
@@ -73,6 +75,21 @@ fit_mle(Categorical, k, x, w)
 fit_mle(Categorical, x)        # equivalent to fit_mle(Categorical, max(x), x)
 fit_mle(Categorical, x, w)
 ```
+
+It is also possible to directly input a distribution `fit_mle(d::Distribution, x[, w])`. This form avoids the extra arguments:
+```julia
+fit_mle(Binomial(n, 0.1), x) 
+# equivalent to fit_mle(Binomial, ntrials(Binomial(n, 0.1)), x), here the parameter 0.1 is not used
+
+fit_mle(Categorical(p), x) 
+# equivalent to fit_mle(Categorical, ncategories(Categorical(p)), x), here the only the length of p is used not its values
+
+d = product_distribution([Exponential(0.5), Normal(11.3, 3.2)])
+fit_mle(d, x) 
+# equivalent to product_distribution([fit_mle(Exponential, x[1,:]), fit_mle(Normal, x[2, :])]). Again parameters of d are not used.
+```
+Note that for standard distributions, the values of the distribution parameters `d` are not used in `fit_mle` only the “structure” of `d` is passed into `fit_mle`.
+However, for complex Maximum Likelihood estimation requiring optimization, e.g., EM algorithm, one could use `D` as an initial guess.
 
 ## Sufficient Statistics
 
--- a/src/genericfit.jl
+++ b/src/genericfit.jl
@@ -1,6 +1,6 @@
 # generic functions for distribution fitting
 
-function suffstats(dt::Type{D}, xs...) where D<:Distribution
+function suffstats(dt::Type{D}, xs...) where {D<:Distribution}
     argtypes = tuple(D, map(typeof, xs)...)
     error("suffstats is not implemented for $argtypes.")
 end
@@ -29,6 +29,8 @@ fit_mle(dt::Type{D}, x::AbstractArray, w::AbstractArray) where {D<:UnivariateDis
 
 fit_mle(dt::Type{D}, x::AbstractMatrix) where {D<:MultivariateDistribution} = fit_mle(D, suffstats(D, x))
 fit_mle(dt::Type{D}, x::AbstractMatrix, w::AbstractArray) where {D<:MultivariateDistribution} = fit_mle(D, suffstats(D, x, w))
+
+fit_mle(dt::D, args...) where {D<:Distribution} = fit_mle(typeof(dt), args...)
 
 fit(dt::Type{D}, x) where {D<:Distribution} = fit_mle(D, x)
 fit(dt::Type{D}, args...) where {D<:Distribution} = fit_mle(D, args...)
--- a/src/multivariate/product.jl
+++ b/src/multivariate/product.jl
@@ -27,14 +27,16 @@ function Product(v::V) where {S<:ValueSupport,T<:UnivariateDistribution{S},V<:Ab
         "`Product(v)` is deprecated, please use `product_distribution(v)`",
         :Product,
     )
-    return Product{S, T, V}(v)
+    return Product{S,T,V}(v)
 end
 
 length(d::Product) = length(d.v)
 function Base.eltype(::Type{<:Product{S,T}}) where {S<:ValueSupport,
-                                                    T<:UnivariateDistribution{S}}
+    T<:UnivariateDistribution{S}}
     return eltype(T)
 end
+
+params(g::Product) = params.(g.v)
 
 _rand!(rng::AbstractRNG, d::Product, x::AbstractVector{<:Real}) =
     map!(Base.Fix1(rand, rng), x, d.v)
@@ -59,4 +61,18 @@ maximum(d::Product) = map(maximum, d.v)
 # higher-dimensional arrays and distributions
 function product_distribution(dists::V) where {S<:ValueSupport,T<:UnivariateDistribution{S},V<:AbstractVector{T}}
     return Product{S,T,V}(dists)
+end
+
+#### Fitting
+
+"""
+    fit_mle(g::Product, x::AbstractMatrix)
+    fit_mle(g::Product, x::AbstractMatrix, γ::AbstractVector)
+
+The `fit_mle` for a multivariate Product distributions `g` is the `product_distribution` of `fit_mle` of each components of `g`.
+"""
+function fit_mle(g::Product, x::AbstractMatrix, args...)
+    d = size(x, 1)
+    length(g) == d || throw(DimensionMismatch("The dimensions of g and x are inconsistent."))
+    return product_distribution([fit_mle(g.v[s], y, args...) for (s, y) in enumerate(eachrow(x))])
 end
--- a/src/univariate/continuous/laplace.jl
+++ b/src/univariate/continuous/laplace.jl
@@ -136,3 +136,15 @@ function fit_mle(::Type{<:Laplace}, x::AbstractArray{<:Real})
     xc .= abs.(x .- m)
     return Laplace(m, mean(xc))
 end
+
+function fit_mle(::Type{<:Laplace}, x::AbstractArray{<:Real}, w::AbstractArray{<:Real})
+    n = length(x)
+    if n != length(w)
+        throw(DimensionMismatch("Inconsistent array lengths."))
+    end
+    xc = similar(x)
+    copyto!(xc, x)
+    m = median(xc, weights(w)) # https://github.com/JuliaStats/StatsBase.jl/blob/0b64a9c8a16355da16469d0fe5016e0fdf099af5/src/weights.jl#L788
+    xc .= abs.(x .- m)
+    return Laplace(m, mean(xc, weights(w)))
+end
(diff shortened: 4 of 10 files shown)


devmotion (commented):
Hej! Thank you for taking the time and putting together a PR. Before diving into more detailed discussions, I think it would be good if you could revert the formatting changes. Additionally, I think it would be good to move a few parts to separate PRs as they do not involve any design questions (see my comments below).

src/univariate/discrete/categorical.jl
```diff
@@ -26,7 +26,7 @@ External links:
 """
 const Categorical{P<:Real,Ps<:AbstractVector{P}} = DiscreteNonParametric{Int,P,Base.OneTo{Int},Ps}
 
-function Categorical{P,Ps}(p::Ps; check_args::Bool=true) where {P<:Real, Ps<:AbstractVector{P}}
+function Categorical{P,Ps}(p::Ps; check_args::Bool=true) where {P<:Real,Ps<:AbstractVector{P}}
```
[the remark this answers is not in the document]
dmetivie: Sure

src/univariate/discrete/dirac.jl
```diff
@@ -50,9 +50,21 @@ logccdf(d::Dirac, x::Real) = x < d.value ? 0.0 : isnan(x) ? NaN : -Inf
...
+
+fit_mle(::Type{<:Dirac}, x::AbstractArray{T}) where {T<:Real} = length(unique(x)) == 1 ? Dirac(first(x)) : Dirac(NaN)
+
+function fit_mle(::Type{<:Dirac}, x::AbstractArray{T}, w::AbstractArray{Float64}) where {T<:Real}
+    n = length(x)
+    if n != length(w)
+        throw(DimensionMismatch("Inconsistent array lengths."))
+    end
+    return length(unique(x[findall(!iszero, w)])) == 1 ? Dirac(first(x)) : Dirac(NaN)
+end
```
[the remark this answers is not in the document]
dmetivie: Ok. For each new fit_mle method? I have never work with more than one branch but I'll try

src/univariate/continuous/laplace.jl
```diff
@@ -136,3 +136,15 @@ function fit_mle(::Type{<:Laplace}, x::AbstractArray{<:Real})
     xc .= abs.(x .- m)
     return Laplace(m, mean(xc))
 end
+
+function fit_mle(::Type{<:Laplace}, x::AbstractArray{<:Real}, w::AbstractArray{<:Real})
```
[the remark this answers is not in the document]
dmetivie: ok

src/multivariate/product.jl
```diff
@@ -60,3 +62,17 @@ maximum(d::Product) = map(maximum, d.v)
...
+    fit_mle(g::Product, x::AbstractMatrix)
+    fit_mle(g::Product, x::AbstractMatrix, γ::AbstractVector)
+
+The `fit_mle` for a multivariate Product distributions `g` is the `product_distribution` of `fit_mle` of each components of `g`.
+"""
+function fit_mle(g::Product, x::AbstractMatrix, args...)
+    d = size(x, 1)
+    length(g) == d || throw(DimensionMismatch("The dimensions of g and x are inconsistent."))
+    return product_distribution([fit_mle(g.v[s], y, args...) for (s, y) in enumerate(eachrow(x))])
+end
```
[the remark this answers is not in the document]
dmetivie: My bad I did not see the new product framework. 
I believe this can be adapted to `ProductDistribution`. I'll do that.

src/univariate/discrete/binomial.jl
```diff
@@ -196,6 +196,7 @@ suffstats(::Type{T}, data::BinomData, w::AbstractArray{<:Real}) where {T<:Binomi
 
 fit_mle(::Type{T}, ss::BinomialStats) where {T<:Binomial} = T(ss.n, ss.ns / (ss.ne * ss.n))
 
+fit_mle(d::T, x::AbstractArray{<:Integer}) where {T<:Binomial} = fit_mle(T, suffstats(T, ntrials(d), x))
```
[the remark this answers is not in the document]
dmetivie: The arguments are the ones I made before.
Mostly, it allows using `fit_mle(dist, x)` in a very generic code without having to care about extra arguments for some distributions like `ProductDistribution`, `Bernoulli` and `Categorical`.

Docs should be clear on the difference between the instance and types difference. I tried typing something in that sense.

src/multivariate/product.jl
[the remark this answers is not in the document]
dmetivie: For now, I keep `Product` because it seems that `d = product_distribution([Exponential(4), Normal(1)])` `isa Product` but not a `Distributions.ArrayOfUnivariateDistribution`.
I guess when `Product` are removed, `d` will be an `DistributionsVectorOfUnivariateDistribution` and the code I added will work.

docs/src/fit.md
```diff
@@ -60,6 +60,7 @@ The `fit_mle` method has been implemented for the following distributions:
 - [`Multinomial`](@ref)
 - [`MvNormal`](@ref)
 - [`Dirichlet`](@ref)
+- [`ProductDistribution`](@ref)
```
devmotion, line 63: The type is an internal implementation detail and not exported, the official API is `product_distribution`.
```suggestion
```

docs/src/fit.md
```diff
@@ -74,6 +75,23 @@ fit_mle(Categorical, x)        # equivalent to fit_mle(Categorical, max(x), x)
...
+fit_mle(Categorical(p), x) 
+# equivalent to fit_mle(Categorical, ncategories(Categorical(p)), x), here the only the length of p is used not its values
+
+d = product_distribution([Exponential(0.5), Normal(11.3, 3.2)])
+fit_mle(d, x) 
+# equivalent to product_distribution([fit_mle(Exponential, x[1,:]), fit_mle(Normal, x[2, :])]). Again parameters of d are not used.
+```
+
+Note that for standard distributions, the values of the distribution parameters `d` are not used in `fit_mle` only the “structure” of `d` is passed into `fit_mle`.
+However, for complex Maximum Likelihood estimation requiring optimization, e.g., EM algorithm, one could use `D` as an initial guess.
```
devmotion, line 93: This is yet another behaviour and IMO adds to the confusion.

src/genericfit.jl
```diff
@@ -30,5 +30,7 @@ fit_mle(dt::Type{D}, x::AbstractArray, w::AbstractArray) where {D<:UnivariateDis
 fit_mle(dt::Type{D}, x::AbstractMatrix) where {D<:MultivariateDistribution} = fit_mle(D, suffstats(D, x))
 fit_mle(dt::Type{D}, x::AbstractMatrix, w::AbstractArray) where {D<:MultivariateDistribution} = fit_mle(D, suffstats(D, x, w))
 
+fit_mle(dt::D, args...) where {D<:Distribution} = fit_mle(typeof(dt), args...)
```
devmotion, line 33: `D` is not needed:
```suggestion
fit_mle(dist::Distribution, args...) = fit_mle(typeof(dist), args...)
```
Just adding this definition would not cause any confusion. But I also don't see a completely clear benefit of it as one could already use `typeof(dist)`.

src/multivariate/product.jl
```diff
@@ -60,3 +62,17 @@ maximum(d::Product) = map(maximum, d.v)
...
+"""
+    fit_mle(g::Product, x::AbstractMatrix)
+    fit_mle(g::Product, x::AbstractMatrix, γ::AbstractVector)
+
+The `fit_mle` for a multivariate Product distributions `g` is the `product_distribution` of `fit_mle` of each components of `g`.
+"""
+function fit_mle(g::Product, x::AbstractMatrix, args...)
+    d = size(x, 1)
+    length(g) == d || throw(DimensionMismatch("The dimensions of g and x are inconsistent."))
+    return product_distribution([fit_mle(g.v[s], y, args...) for (s, y) in enumerate(eachrow(x))])
```
devmotion, line 77: In addition to what I mentioned above, I think there's another issue here: It is not possible to pass different arguments/options to the fits of the different components, which makes this approach much less flexible than just fitting each component manually.

src/multivariate/product.jl
```diff
@@ -36,6 +36,8 @@ function Base.eltype(::Type{<:Product{S,T}}) where {S<:ValueSupport,
     return eltype(T)
 end
 
+params(g::Product) = params.(g.v)
+
```
devmotion, line 40: This seems unrelated:
```suggestion
```

src/product.jl
```diff
@@ -74,6 +74,8 @@ end
 
 size(d::ProductDistribution) = d.size
 
+params(d::ArrayOfUnivariateDistribution) = params.(d.dists)
+
```
devmotion, line 78:
```suggestion
```

devmotion (commented):
I added some more comments to explain my reservations. Currently, my opinion is that the suggestions in this PR would lead to confusing behaviour that is difficult to predict, without a clear benefit. I'm also not completely convinced by the definitions for product distributions as they are not very flexible and make some seemingly strong assumptions such as e.g. that the same arguments are provided to all fits.

Therefore currently I would suggest not merging this PR.

src/genericfit.jl
dmetivie: If `fit_mle(dist::Distribution, x)` does not exist for some distribution, this line will try `fit_mle(::Type, x)`. 
This ensures that `fit_mle(dist::Distribution, x)` will call currently implemented `fit_mle` methods that use type and not instance. 
If this is added for me, it is already something.

src/multivariate/product.jl
dmetivie: I wondered about that: is there a way to add `kwargs = [kwargs[i] for i in 1:d]` and apply them to each distribution?
About the arguments, I believe in most situation you have a sample and fit the whole `ProductDistribution` at once.
Hence, IMO the code above would correspond to the most typical situation. If you think about a situation where some distributions have weighted samples and some other not, one could either try to define a method for this case or as you said do it manually.

docs/src/fit.md
```diff
@@ -74,6 +75,23 @@ fit_mle(Categorical, x)        # equivalent to fit_mle(Categorical, max(x), x)
...
+```julia
+fit_mle(Binomial(n, 0.1), x) 
+# equivalent to fit_mle(Binomial, ntrials(Binomial(n, 0.1)), x), here the parameter 0.1 is not used
+
+fit_mle(Categorical(p), x) 
+# equivalent to fit_mle(Categorical, ncategories(Categorical(p)), x), here the only the length of p is used not its values
+
+d = product_distribution([Exponential(0.5), Normal(11.3, 3.2)])
+fit_mle(d, x) 
+# equivalent to product_distribution([fit_mle(Exponential, x[1,:]), fit_mle(Normal, x[2, :])]). Again parameters of d are not used.
```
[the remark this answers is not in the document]
dmetivie: For `Categorical` the current `fit_mle(Categorical, x) = fit_mle(Categorical, max(x), x)` could also lead to confusion (and error in really unlucky case where there are 0 samples of the last category) and not so different of `fit_mle(Categorical, ncategories(Categorical(p)), x)`.

It is not about code length. See my comment below.

dmetivie:
Thanks for giving thoughts and time.
The main benefits of this PR are not about shorter code. I find that the instance version of `fit_mle` convey more information than the type version. In `Distributions.jl` advantages are not super obvious (except for `ProductDistributions`) since for simple univariate of multivariate normal MLE is known exactly.
The instance version allows writing the EM algorithm very simply for all kind of distributions see [here](https://github.com/dmetivie/ExpectationMaximization.jl/blob/2c6f6b1f5f88727c189ca93d4aaba4003e2019a9/src/fit_em.jl#L65) sure I am very bias about that.
I feel that with types only we are missing something:
- How would one code `fit_mle` for product distributions with types only? IMO `fit_mle` for product distribution is a very neat feature (even if it is not hard to code). 
- How would you extract the types of a mixture with only the type version of `fit_mle` (or another similar function)

lrnv:
Although i really have no oppinon on the confusion argument, i think that a better solution would be to parametrize mixtures distributions: 

`MixtureModel{Gamma,Normal,Binomial}` would be the type, and not only `MixtureModel`. This would be coherent with the current behavior of `ProductDistribution`, which is parametrized by types. But I agree that this is probably a more dificult change to implement (might even be considered breaking ?).

ParadaCarleton:
> * How would you extract the types of a mixture with only the type version of `fit_mle` (or another similar function)

Hmm, maybe the types of a mixture should also include how many of each component is included? e.g. 
```julia
Mixture{MixtureComponent{Normal, 2}, MixtureComponent{Gamma, 3}}
```

PR #1674, opened 2023-02-02
fix cdf & pdf for TriangularDist

jwright11:
fixes issue 1673

### DIFF
--- a/src/univariate/continuous/triangular.jl
+++ b/src/univariate/continuous/triangular.jl
@@ -35,7 +35,7 @@ struct TriangularDist{T<:Real} <: ContinuousUnivariateDistribution
 end
 
 function TriangularDist(a::T, b::T, c::T; check_args::Bool=true) where {T <: Real}
-    @check_args TriangularDist (a <= c <= b)
+    @check_args(TriangularDist, (a <= c <= b, "must have a <= c <= b"), (a < b, "must have a < b"))
     return TriangularDist{T}(a, b, c)
 end
 
--- a/src/univariate/continuous/triangular.jl
+++ b/src/univariate/continuous/triangular.jl
@@ -35,7 +35,7 @@ struct TriangularDist{T<:Real} <: ContinuousUnivariateDistribution
 end
 
 function TriangularDist(a::T, b::T, c::T; check_args::Bool=true) where {T <: Real}
-    @check_args TriangularDist (a <= c <= b) (a < b)
+    @check_args TriangularDist (a <= c <= b)
     return TriangularDist{T}(a, b, c)
 end
 
@@ -93,30 +93,26 @@ entropy(d::TriangularDist{T}) where {T<:Real} = one(T)/2 + log((d.b - d.a) / 2)
 
 function pdf(d::TriangularDist, x::Real)
     a, b, c = params(d)
-    T = typeof(one(x) / one(a))
-    if x < a || b < x
-        return T(0)
-    elseif a <= x < c
-        return 2 * (x - a) / ((b - a) * (c - a))
-    elseif x == c
-        return T(2) / (b - a)
+    res = if x < c
+        2 * (x - a) / ((b - a) * (c - a))
+    elseif x > c
+        2 * (b - x) / ((b - a) * (b - c))
     else
-        return 2 * (b - x) / ((b - a) * (b - c))
+        # Handle x == c separately to avoid `NaN` if `c == a` or `c == b`
+        oftype(x - a, 2) / (b - a)
     end
+    return insupport(d, x) ? res : zero(res)
 end
 logpdf(d::TriangularDist, x::Real) = log(pdf(d, x))
 
 function cdf(d::TriangularDist, x::Real)
     a, b, c = params(d)
-    T = typeof(one(x) / one(a))
-    if x <= a
-        return T(0)
-    elseif x >= b
-        return T(1)
-    elseif a < x <= c
-        return (x - a) ^ 2 / ((b - a) * (c - a))
-    else c < x < b
-        return 1 - (b - x) ^ 2 / ((b - a) * (b - c))
+    if x < c
+        res = (x - a)^2 / ((b - a) * (c - a))
+        return x < a ? zero(res) : res
+    else
+        res = 1 - (b - x)^2 / ((b - a) * (b - c))
+        return x ≥ b ? one(res) : res
     end
 end
 
@@ -129,20 +125,73 @@ function quantile(d::TriangularDist, p::Real)
               b - sqrt(b_m_a * (b - c) * (1 - p))
 end
 
+"""
+    _phi2(x::Real)
+
+Compute
+```math
+2 (exp(x) - 1 - x) / x^2
+```
+with the correct limit at ``x = 0``.
+"""
+function _phi2(x::Real)
+    res = 2 * (expm1(x) - x) / x^2
+    return iszero(x) ? one(res) : res
+end
 function mgf(d::TriangularDist, t::Real)
     a, b, c = params(d)
-    u = (b - c) * exp(a * t) - (b - a) * exp(c * t) + (c - a) * exp(b * t)
-    iszero(t) && return one(u)
-    v = (b - a) * (c - a) * (b - c) * t^2
-    return 2u / v
+    # In principle, only two branches (degenerate + non-degenerate case) are needed
+    # But writing out all four cases will avoid unnecessary computations
+    if a < c
+        if c < b
+            # Case: a < c < b
+            return exp(c * t) * ((c - a) * _phi2((a - c) * t) + (b - c) * _phi2((b - c) * t)) / (b - a)
+        else
+            # Case: a < c = b
+            return exp(c * t) * _phi2((a - c) * t)
+        end
+    elseif c < b
+        # Case: a = c < b
+        return exp(c * t) * _phi2((b - c) * t)
+    else
+        # Case: a = c = b
+        return exp(c * t)
+    end
 end
 
+"""
+    _cisphi2(x::Real)
+
+Compute
+```math
+- 2 (exp(x im) - 1 - x im) / x^2
+```
+with the correct limit at ``x = 0``.
+"""
+function _cisphi2(x::Real)
+    z = x * im
+    res = -2 * (expm1(z) - z) / x^2
+    return iszero(x) ? one(res) : res
+end
 function cf(d::TriangularDist, t::Real)
     a, b, c = params(d)
-    u = (b - c) * cis(a * t) - (b - a) * cis(c * t) + (c - a) * cis(b * t)
-    iszero(t) && return one(u)
-    v = (b - a) * (c - a) * (b - c) * t^2
-    return -2u / v
+    # In principle, only two branches (degenerate + non-degenerate case) are needed
+    # But writing out all four cases will avoid unnecessary computations
+    if a < c
+        if c < b
+            # Case: a < c < b
+            return cis(c * t) * ((c - a) * _cisphi2((a - c) * t) + (b - c) * _cisphi2((b - c) * t)) / (b - a)
+        else
+            # Case: a < c = b
+            return cis(c * t) * _cisphi2((a - c) * t)
+        end
+    elseif c < b
+        # Case: a = c < b
+        return cis(c * t) * _cisphi2((b - c) * t)
+    else
+        # Case: a = c = b
+        return cis(c * t)
+    end
 end
 
 
--- a/src/univariate/continuous/triangular.jl
+++ b/src/univariate/continuous/triangular.jl
@@ -125,6 +125,12 @@ function quantile(d::TriangularDist, p::Real)
               b - sqrt(b_m_a * (b - c) * (1 - p))
 end
 
+_expm1(x::Number) = expm1(x)
+if VERSION < v"1.7.0-DEV.1172"
+    # expm1(::Float16) is not defined in older Julia versions
+    _expm1(x::Float16) = Float16(expm1(Float32(x)))
+end
+
 """
     _phi2(x::Real)
 
@@ -135,7 +141,7 @@ Compute
 with the correct limit at ``x = 0``.
 """
 function _phi2(x::Real)
-    res = 2 * (expm1(x) - x) / x^2
+    res = 2 * (_expm1(x) - x) / x^2
     return iszero(x) ? one(res) : res
 end
 function mgf(d::TriangularDist, t::Real)
@@ -170,7 +176,7 @@ with the correct limit at ``x = 0``.
 """
 function _cisphi2(x::Real)
     z = x * im
-    res = -2 * (expm1(z) - z) / x^2
+    res = -2 * (_expm1(z) - z) / x^2
     return iszero(x) ? one(res) : res
 end
 function cf(d::TriangularDist, t::Real)
--- a/src/univariate/continuous/triangular.jl
+++ b/src/univariate/continuous/triangular.jl
@@ -122,27 +122,22 @@ function quantile(d::TriangularDist, p::Real)
               b - sqrt(b_m_a * (b - c) * (1 - p))
 end
 
-function mgf(d::TriangularDist{T}, t::Real) where T<:Real
-    if t == zero(t)
-        return one(T)
-    else
-        (a, b, c) = params(d)
+function mgf(d::TriangularDist, t::Real)
+    a, b, c = params(d)
+    T = typeof(t / a)
+    t == zero(T) && return one(T)
         u = (b - c) * exp(a * t) - (b - a) * exp(c * t) + (c - a) * exp(b * t)
         v = (b - a) * (c - a) * (b - c) * t^2
         return 2u / v
-    end
 end
 
-function cf(d::TriangularDist{T}, t::Real) where T<:Real
-    # Is this correct?
-    if t == zero(t)
-        return one(Complex{T})
-    else
-        (a, b, c) = params(d)
+function cf(d::TriangularDist, t::Real)
+    a, b, c = params(d)
+    T = typeof(t / a)
+    t == zero(T) && return complex(one(T))
         u = (b - c) * cis(a * t) - (b - a) * cis(c * t) + (c - a) * cis(b * t)
         v = (b - a) * (c - a) * (b - c) * t^2
         return -2u / v
-    end
 end
 
 
--- a/src/univariate/continuous/triangular.jl
+++ b/src/univariate/continuous/triangular.jl
@@ -93,20 +93,28 @@ entropy(d::TriangularDist{T}) where {T<:Real} = one(T)/2 + log((d.b - d.a) / 2)
 
 function pdf(d::TriangularDist, x::Real)
     a, b, c = params(d)
-    return if x < c
-        2 * max(x - a, 0) / ((b - a) * (c - a))
-    else
-        2 * max(b - x, 0) / ((b - a) * (b - c))
+    if x < a || b < x
+        return 0
+    elseif a <= x < c
+        return 2 * (x - a) / ((b - a) * (c - a))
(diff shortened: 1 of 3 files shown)


src/univariate/continuous/triangular.jl
```diff
@@ -93,20 +93,28 @@ entropy(d::TriangularDist{T}) where {T<:Real} = one(T)/2 + log((d.b - d.a) / 2)
...
-    else
-        2 * max(b - x, 0) / ((b - a) * (b - c))
+    if x < a || b < x
+        return 0
+    elseif a <= x < c
+        return 2 * (x - a) / ((b - a) * (c - a))
+    elseif x == c
+        return 2 / (b - a)
+    elseif c < x <= b
+        return 2 * (b - x) / ((b - a) * (b - c))
```
devmotion, line 103: I'm pretty sure that was (roughly) the original code before someone (I?) tried to make it type stable and apparently introduced bugs which were not covered by the tests (even though we compare it with R, IIRC). Unfortunately, this PR has the same type stability issues as the original versions which causes the test errors.

devmotion: It seems the current version is correct apart from degenerate cases where `a == c` or `b  == c` since in these cases we would divide by zero.

jwright11: Ok, thanks. Can you point me to where the test for type stability is performed? I can try to modify for type stability, would be helpful to be able to test it directly.

devmotion: It's part of the large test suite for univariate distributions, there's no simple TriangularDist-specific test. For instance, type stability checks are performed in https://github.com/JuliaStats/Distributions.jl/blob/c431d20d3a0f00e0cf22f9fd959b9bb11d978517/test/testutils.jl#L487, which is part of https://github.com/JuliaStats/Distributions.jl/blob/2dc764e450a5b33f1506669e3649f37a1dbda4f1/test/testutils.jl#L69, which is called by https://github.com/JuliaStats/Distributions.jl/blob/c431d20d3a0f00e0cf22f9fd959b9bb11d978517/test/univariates.jl#L51.

But locally something like
```julia
using Test
@inferred(pdf(TriangularDist(0f0, 2f0, 1f0), 1.0))
```
or just
```julia
@code_warntype pdf(TriangularDist(0f0, 2f0, 1f0), 1.0)
```
should be sufficient to reproduce the errors with the `pdf` code in this PR: In one branch the function would return `Int`, in one branch `Float32`, and in the other branches `Float64`.

jwright11: Thanks for the help. Should be a very educational exercise for me 🙂

jwright11: I have updated to be type stable now. All the tests passed on my local. I think this should be good to go now but let me know if anything else is required.

devmotion: Unfortunately, it's still not type stable.

src/univariate/continuous/triangular.jl
```diff
@@ -93,20 +93,34 @@ entropy(d::TriangularDist{T}) where {T<:Real} = one(T)/2 + log((d.b - d.a) / 2)
 
 function pdf(d::TriangularDist, x::Real)
     a, b, c = params(d)
-    return if x < c
-        2 * max(x - a, 0) / ((b - a) * (c - a))
+    t = promote(x, a)[1]
```
devmotion, line 96: The problem with heuristics in general is that it is easy to miss some cases. Therefore usually in Distributions the typical approach is to perform computations as much as possible (possibly by clamping values to the support) and handle special cases (such as out of support) only afterwards, based on the type of these computations. In that way we can avoid having to implement any heuristics.

One example where the current heuristic fails:
```julia
julia> @inferred pdf(TriangularDist{Int}(0, 2, 1), 1)
ERROR: return type Int64 does not match inferred return type Union{Float64, Int64}
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] top-level scope
   @ REPL[5]:1
```

jwright11: Ok, I've given it another shot.  I went back to the form of the functions that were previously there and modified the cdf to accurately handle the case when b=c, the Inf test, and the NaN test.  I tried to follow the "spirit" you've mentioned above.  All tests passed on my local.  I have my fingers crossed.

test/univariate/continuous/triangular.jl
```diff
@@ -0,0 +1,21 @@
+using Distributions
+using Test
+
+@testset "triangular" begin
+    a1 = 0.0
+    b1 = 1.0
+    c1 = 1.0
+    x1 = 0.0
+    x2 = 0.5
+    x3 = 1.0
```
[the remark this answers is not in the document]
jwright11: Yeah ok. There was no testing before I started and the only problem I had noticed was with the cdf function. I'll take a look at pdf and add more testing scenarios.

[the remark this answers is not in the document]
jwright11: When this is all over I'll be the triangular distribution champion, lol

jwright11:
Alright, I learned a good lesson about the importance of setting up robust testing.  I thought my last version of the cdf function was good to go but robust testing proved otherwise.  I went back to the heuristic approach, which I think is necessary for this distribution.  I could not see how to handle the various scenarios/outliers using a clamp approach.  I believe these functions are now type stable and robustly tested.

Using the same sort of "@inferred" tests on the mgf & cf functions, it was clear that the prior versions were not type stable so I've modified them to be type stable.

Lastly, I added "a<b" to check_args.  Wikipedia has this as a parameter constraint "a<b & a<=c<=b".  Having it in there disallows declaring a trivial distribution like TriangularDist(0,0,0).

Hopefully this is it but let me know if there's anything else I'm missing.

src/univariate/continuous/triangular.jl
```diff
@@ -93,20 +93,34 @@ entropy(d::TriangularDist{T}) where {T<:Real} = one(T)/2 + log((d.b - d.a) / 2)
 
 function pdf(d::TriangularDist, x::Real)
     a, b, c = params(d)
-    return if x < c
-        2 * max(x - a, 0) / ((b - a) * (c - a))
+    T = typeof(x / a)
```
devmotion, line 96: This line will error e.g. in the following case:
```julia
julia> pdf(TriangularDist{Int}(0, 2, 1), 0//1)
ERROR: DivideError: integer division error
Stacktrace:
 [1] div
   @ ./int.jl:288 [inlined]
 [2] divgcd(x::Int64, y::Int64)
   @ Base ./rational.jl:45
 [3] //
   @ ./rational.jl:65 [inlined]
 [4] /
   @ ./rational.jl:346 [inlined]
 [5] pdf(d::TriangularDist{Int64}, x::Rational{Int64})
   @ Main ./REPL[4]:3
 [6] top-level scope
   @ REPL[13]:1
```

src/univariate/continuous/triangular.jl
```diff
@@ -35,7 +35,7 @@ struct TriangularDist{T<:Real} <: ContinuousUnivariateDistribution
 end
 
 function TriangularDist(a::T, b::T, c::T; check_args::Bool=true) where {T <: Real}
-    @check_args TriangularDist (a <= c <= b)
+    @check_args(TriangularDist, (a <= c <= b, "must have a <= c <= b"), (a < b, "must have a < b"))
```
devmotion, line 38: Is this actually needed? A few distributions also support the degenerate case (e.g., `Normal(m, 0)`), so I would not want to add this restrictions if it is not required.

I would also suggest using the default error messages, IMO the manual ones do not add more explanations:
```suggestion
    @check_args TriangularDist (a <= c <= b) (a < b)
```

src/univariate/continuous/triangular.jl
```diff
@@ -93,20 +93,34 @@ entropy(d::TriangularDist{T}) where {T<:Real} = one(T)/2 + log((d.b - d.a) / 2)
...
+    elseif a <= x < c
+        return 2 * (x - a) / ((b - a) * (c - a))
+    elseif x == c
+        return T(2) / (b - a)
+    elseif c < x <= b
+        return 2 * (b - x) / ((b - a) * (b - c))
     else
-        2 * max(b - x, 0) / ((b - a) * (b - c))
+        return T(NaN)
     end
```
devmotion, line 107: This branch can't be reached, can it? We also would never want to return `NaN` - inside of the support we would like a finite value, and outside of it 0.
It would also error in cases such as `pdf(::TriangularDist{Int}, ::Rational{Int})`:
```julia
julia> Rational{Int}(NaN)
ERROR: InexactError: Int64(NaN)
```

src/univariate/continuous/triangular.jl
```diff
@@ -93,20 +93,34 @@ entropy(d::TriangularDist{T}) where {T<:Real} = one(T)/2 + log((d.b - d.a) / 2)
...
 end
 logpdf(d::TriangularDist, x::Real) = log(pdf(d, x))
 
 function cdf(d::TriangularDist, x::Real)
     a, b, c = params(d)
-    return if x < c
-        max(x - a, 0)^2 / ((b - a) * (c - a))
+    T = typeof(x / a)
```
devmotion, line 113: Same here, this can lead to errors.

src/univariate/continuous/triangular.jl
```diff
@@ -93,20 +93,34 @@ entropy(d::TriangularDist{T}) where {T<:Real} = one(T)/2 + log((d.b - d.a) / 2)
...
+    elseif a < x <= c
+        return (x - a) ^ 2 / ((b - a) * (c - a))
+    elseif c < x < b
+        return 1 - (b - x) ^ 2 / ((b - a) * (b - c))
+    elseif b <= x
+        return T(1)
     else
-        1 - max(b - x, 0)^2 / ((b - a) * (b - c))
+        return T(NaN)
     end
```
devmotion, line 124: Also this branch should not be needed.

src/univariate/continuous/triangular.jl
```diff
@@ -119,27 +133,22 @@ function quantile(d::TriangularDist, p::Real)
...
-        return one(T)
-    else
-        (a, b, c) = params(d)
-        u = (b - c) * exp(a * t) - (b - a) * exp(c * t) + (c - a) * exp(b * t)
-        v = (b - a) * (c - a) * (b - c) * t^2
-        return 2u / v
-    end
+function mgf(d::TriangularDist, t::Real)
+    a, b, c = params(d)
+    T = typeof(t / a)
```
devmotion, line 138: Same problems as above here.

src/univariate/continuous/triangular.jl
```diff
@@ -119,27 +133,22 @@ function quantile(d::TriangularDist, p::Real)
...
+    t == zero(T) && return one(T)
```
devmotion, line 139: This can be written simpler as
```suggestion
    iszero(t) && return one(T)
```

src/univariate/continuous/triangular.jl
```diff
@@ -119,27 +133,22 @@ function quantile(d::TriangularDist, p::Real)
...
+    u = (b - c) * exp(a * t) - (b - a) * exp(c * t) + (c - a) * exp(b * t)
+    v = (b - a) * (c - a) * (b - c) * t^2
+    return 2u / v
```
devmotion, line 142: The function is not type stable since sometimes `T` is not a good heuristic for the type of the result here.
```julia
julia> @inferred mgf(TriangularDist{Int}(0, 2, 1), 1//1)
ERROR: return type Float64 does not match inferred return type Union{Rational{Int64}, Float64}
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] top-level scope
   @ REPL[22]:1
```

src/univariate/continuous/triangular.jl
```diff
@@ -119,27 +133,22 @@ function quantile(d::TriangularDist, p::Real)
...
-    # Is this correct?
-    if t == zero(t)
-        return one(Complex{T})
-    else
-        (a, b, c) = params(d)
-        u = (b - c) * cis(a * t) - (b - a) * cis(c * t) + (c - a) * cis(b * t)
-        v = (b - a) * (c - a) * (b - c) * t^2
-        return -2u / v
-    end
+function cf(d::TriangularDist, t::Real)
```
devmotion, line 145: Same comments as for `mgf`.

src/univariate/continuous/triangular.jl
jwright11: I'm good with default error messages and have accepted your suggested change.  To me, there's no point in allowing someone to define a trivial/degenerate distribution.  I can't think of a reason to intentionally define it.

src/univariate/continuous/triangular.jl
jwright11: I've updated, which I believe solves the issue more generally as there won't be a div0 error anymore to interfere with getting the type.

src/univariate/continuous/triangular.jl
jwright11: I believe it can with:
pdf(TriangularDist{Int}(0, 2, 1), NaN)

I added this in because of the check you pointed me to in test_nonfinite():
    for f in (cdf, ccdf, logcdf, logccdf)
        if distr isa NoncentralT
            # broken in StatsFuns/R
            @test_broken isnan(f(distr, NaN))
        else
            @test isnan(f(distr, NaN))
        end
    end

    Having this last branch for NaN allows me to pass this NaN test.

src/univariate/continuous/triangular.jl
jwright11: same as above, believe this is fixed now

src/univariate/continuous/triangular.jl
jwright11: see comments above on testing for NaN

src/univariate/continuous/triangular.jl
jwright11: believe this is fixed now

src/univariate/continuous/triangular.jl
jwright11: ok, I took your suggestion

src/univariate/continuous/triangular.jl
jwright11: updated

src/univariate/continuous/triangular.jl
jwright11: updated, should be now

jwright11:
I plan on adding tests regarding type stability over the weekend

jwright11:
So what do you think? Is this good to go?

test/univariate/continuous/triangular.jl
```diff
@@ -0,0 +1,473 @@
...
+n = length(types)^2
+typeslist = Array{String}(undef, n, 2)
+j = 1
+for t in eachindex(types)
+    for u in eachindex(types)
+        typeslist[j, 1] = types[t]
+        typeslist[j, 2] = types[u]
+        global j += 1
+    end
+end
```
devmotion, line 40: Can we avoid dealing with strings alltogether? It would be much simpler to just loop over an array/tuple of types, such as `Ts = (Int, Float64, Float32, Rational{Int})`.

src/univariate/continuous/triangular.jl
```diff
@@ -93,20 +93,34 @@ entropy(d::TriangularDist{T}) where {T<:Real} = one(T)/2 + log((d.b - d.a) / 2)
...
+        return T(0)
+    elseif a <= x < c
+        return 2 * (x - a) / ((b - a) * (c - a))
+    elseif x == c
+        return T(2) / (b - a)
+    elseif c < x <= b
+        return 2 * (b - x) / ((b - a) * (b - c))
     else
-        2 * max(b - x, 0) / ((b - a) * (b - c))
+        return T(NaN)
```
devmotion, line 106: Can you remove this branch? I don't think https://github.com/JuliaStats/Distributions.jl/pull/1674/files#r1101068580 is resolved - IMO the tests should not be fixed by adding a `NaN` branch but by changing the last `elseif` to an `else`.

src/univariate/continuous/triangular.jl
```diff
@@ -93,20 +93,34 @@ entropy(d::TriangularDist{T}) where {T<:Real} = one(T)/2 + log((d.b - d.a) / 2)
...
+        return T(0)
+    elseif a < x <= c
+        return (x - a) ^ 2 / ((b - a) * (c - a))
+    elseif c < x < b
+        return 1 - (b - x) ^ 2 / ((b - a) * (b - c))
+    elseif b <= x
+        return T(1)
     else
-        1 - max(b - x, 0)^2 / ((b - a) * (b - c))
+        return T(NaN)
```
devmotion, line 123: Same here, this should be removed.

src/univariate/continuous/triangular.jl
```diff
@@ -119,27 +133,20 @@ function quantile(d::TriangularDist, p::Real)
...
-    else
-        (a, b, c) = params(d)
-        u = (b - c) * cis(a * t) - (b - a) * cis(c * t) + (c - a) * cis(b * t)
-        v = (b - a) * (c - a) * (b - c) * t^2
-        return -2u / v
-    end
+function cf(d::TriangularDist, t::Real)
+    a, b, c = params(d)
+    u = (b - c) * cis(a * t) - (b - a) * cis(c * t) + (c - a) * cis(b * t)
+    iszero(t) && return complex(one(u))
```
devmotion, line 147: I think this could be simplified to
```suggestion
    iszero(t) && return one(u)
```

test/univariate/continuous/triangular.jl
```diff
@@ -0,0 +1,473 @@
...
+        @test median(d) == (c >= middle(a, b) ?
+            a + sqrt((b - a) * (c - a) / 2) :
+            b - sqrt((b - a) * (b - c) / 2))
+        @test var(d) == (a^2 + b^2 + c^2 - a * b - a * c - b * c) / 18
+        @test skewness(d) == sqrt(2) * (a + b - 2c) * (2a - b - c) * (a - 2b + c) /
+            (5 * (a^2 + b^2 + c^2 - a * b - a * c - b * c) ^ (3 / 2))
+        @test kurtosis(d) == -3 / 5
+        @test entropy(d) == 1 / 2 + log((b - a) / 2)
+        @test pdf(d, xa) == 2 * (xa - a) / ((b - a) * (c - a))
+        @test pdf(d, xb) == 2 * (b - xb) / ((b - a) * (b - c))
```
devmotion, line 103: Can these tests be unified? It seems a bit redundant to test every scenario completely separately.

src/univariate/continuous/triangular.jl
jwright11: updated

src/univariate/continuous/triangular.jl
jwright11: removed

jwright11:
These 3 checks have failed because we've removed the NaN branches on the pdf & cdf functions.  If we add those back these will pass.  I think we should add them back and then put a bow on this thing.

devmotion:
No, they really shouldn't be there. It's only a problem currently because the fall-through branch in `cdf` returns a constant, independent of the values of `x` and `a`, `b`, `c`. Let me clean up the PR so we can speed up the merge process.

jwright11:
> No, they really shouldn't be there. It's only a problem currently because the fall-through branch in `cdf` returns a constant, independent of the values of `x` and `a`, `b`, `c`. Let me clean up the PR so we can speed up the merge process.

Thanks for pointing this out. I made one last edit to the cdf function to change this and all checks are passing now.

devmotion:
I pushed some changes, the tests seemed a bit too complicated to me. Since they used the same code for the tests that was used in the implementation they also didn't catch some of the bugs in the `cf` and `mgf` functions: For all special cases they always returned `NaN`. This should be fixed by my commit.

PR #1676, opened 2023-02-07
Add fit mle weighted laplace

dmetivie:
I added `fit_mle(Laplace, x, w)` method using the weighted median plus a test in the test section of Laplace fit.

### DIFF
--- a/src/univariate/continuous/laplace.jl
+++ b/src/univariate/continuous/laplace.jl
@@ -136,3 +136,15 @@ function fit_mle(::Type{<:Laplace}, x::AbstractArray{<:Real})
     xc .= abs.(x .- m)
     return Laplace(m, mean(xc))
 end
+
+function fit_mle(::Type{<:Laplace}, x::AbstractArray{<:Real}, w::AbstractArray{<:Real})
+    n = length(x)
+    if n != length(w)
+        throw(DimensionMismatch("Inconsistent array lengths."))
+    end
+    xc = similar(x)
+    copyto!(xc, x)
+    m = median(xc, weights(w)) # https://github.com/JuliaStats/StatsBase.jl/blob/0b64a9c8a16355da16469d0fe5016e0fdf099af5/src/weights.jl#L788
+    xc .= abs.(x .- m)
+    return Laplace(m, mean(xc, weights(w)))
+end
--- a/test/fit.jl
+++ b/test/fit.jl
@@ -367,10 +367,18 @@ end
 
 @testset "Testing fit for Laplace" begin
     for func in funcs, dist in (Laplace, Laplace{Float64})
-        d = fit(dist, func[2](dist(5.0, 3.0), N + 1))
+        x = func[2](dist(5.0, 3.0), N + 1)
+        w = func[1](N + 1)
+
+        d = fit(dist, x)
         @test isa(d, dist)
         @test isapprox(location(d), 5.0, atol=0.02)
         @test isapprox(scale(d)   , 3.0, atol=0.03)
+
+        dw = fit(dist, x, w)
+        @test isa(dw, dist)
+        @test isapprox(location(dw), 5.0, atol=0.02)
+        @test isapprox(scale(dw), 3.0, atol=0.03)
     end
 end
 


ParadaCarleton:
> I added `fit_mle(Laplace, x, w)` method using the weighted median plus a test in the test section of Laplace fit.

Do you have a citation for this being equivalent to the MLE of a Laplace distribution?

This looks good, though! I don't see any problems.

ParadaCarleton:
Actually, I think this is just clear/well-known enough that we don't need a citation. Thanks for the PR!

src/univariate/continuous/laplace.jl
```diff
@@ -136,3 +136,15 @@ function fit_mle(::Type{<:Laplace}, x::AbstractArray{<:Real})
...
+    n = length(x)
+    if n != length(w)
+        throw(DimensionMismatch("Inconsistent array lengths."))
+    end
+    xc = similar(x)
+    copyto!(xc, x)
+    m = median(xc, weights(w)) # https://github.com/JuliaStats/StatsBase.jl/blob/0b64a9c8a16355da16469d0fe5016e0fdf099af5/src/weights.jl#L788
+    xc .= abs.(x .- m)
+    return Laplace(m, mean(xc, weights(w)))
+end
```
ParadaCarleton, line 150: Actually, I've taken the opportunity to remove an unneeded allocation; there's no need to create an extra array for the centered variables.
```suggestion
function fit_mle(::Type{<:Laplace}, data::AbstractArray{<:Real})
    med = median(data)
    # TODO: is it possible to do a single-pass median and mean absolute deviation?
    mean_abs_deviation = mean(x) do x
        return abs(x - mean_abs_deviation)
    end
    return Laplace(med, mean_abs_dev)
end

function fit_mle(::Type{<:Laplace}, x::AbstractArray{<:Real}, w::AbstractWeights)
    if length(x) != length(w)
        throw(DimensionMismatch("Inconsistent array lengths."))
    end
    # TODO: can this be done in a single-pass?
    # code for weighted median can be found at:
    # https://github.com/JuliaStats/StatsBase.jl/blob/0b64a9c8a16355da16469d0fe5016e0fdf099af5/src/weights.jl#L788
    med = median(xc, weights(w))
    norm = sum(w)
    mean_abs_deviation = mean(zip(x, weights)) / norm do x, weight
        return weight * abs(x - med)
    end
    return Laplace(med, mean_abs_deviation)
end
```

test/fit.jl
```diff
@@ -367,10 +367,18 @@ end
...
+
+        d = fit(dist, x)
         @test isa(d, dist)
         @test isapprox(location(d), 5.0, atol=0.02)
         @test isapprox(scale(d)   , 3.0, atol=0.03)
+
+        dw = fit(dist, x, w)
+        @test isa(dw, dist)
+        @test isapprox(location(dw), 5.0, atol=0.02)
+        @test isapprox(scale(dw), 3.0, atol=0.03)
```
ParadaCarleton, line 381: Can you test the weighted method gives the same answers as the unweighted method when `w` is a set of unit weights?

test/fit.jl
```diff
@@ -367,10 +367,18 @@ end
...
     for func in funcs, dist in (Laplace, Laplace{Float64})
-        d = fit(dist, func[2](dist(5.0, 3.0), N + 1))
+        x = func[2](dist(5.0, 3.0), N + 1)
+        w = func[1](N + 1)
+
+        d = fit(dist, x)
         @test isa(d, dist)
         @test isapprox(location(d), 5.0, atol=0.02)
         @test isapprox(scale(d)   , 3.0, atol=0.03)
+
```
ParadaCarleton, line 377: Could you add a test for the case where the weights don't sum to 1?

ParadaCarleton: (And ideally one where the weights sum to 0, making sure we get either a NaN or an error?)

src/univariate/continuous/laplace.jl
devmotion: The suggestion breaks the non-weighted implementation without improving its allocations (I don't think the non-weighted algorithm can be improved allocationwise).

ParadaCarleton: > The suggestion breaks the non-weighted implementation without improving its allocations

I made a mistake earlier. Not sure if it works now, but @dmetivie can double-check hopefully. (Is there an easy way to run tests on suggestions?)


> I don't think the non-weighted algorithm can be improved allocationwise

The code for both algorithms is the same. But in any case, there's definitely no need to copy an array just to center it.

dmetivie:
Thanks for the review, I'll do your suggestion and test, when I find some time.

> Do you have a citation for this being equivalent to the MLE of a Laplace distribution?
I was going to write down the proof because I don't see it written explicitly anywhere.

src/univariate/continuous/laplace.jl
devmotion: > But in any case, there's definitely no need to copy an array just to center it.

The code on master seems to be as efficient as possible: The copy is used by the median calculations and then only reused by the centering operation. Note that the code uses `median!` - if you switch to `median`, a copy is created internally and discarded afterwards. So IMO the current code should not be changed.

ParadaCarleton:
> I was going to write down the proof because I don't see it written explicitly anywhere.

Yeah, I think it's a well-known enough result that we can use it. (The proof should be identical to the one for the unweighted case).

PR #1678, opened 2023-02-13
Add the one-parameter Lindley distribution

ararslan:
~~This requires a new dependency on LambertW.jl, which is actively maintained and has no dependencies of its own. I'm also currently using `evalpoly`, which was introduced in Julia 1.4, but this package currently requires at least 1.3. IMO we should raise the minimum supported version to 1.6 anyway; I've done so in a separate commit.~~


Project.toml
```diff
@@ -28,7 +29,7 @@ QuadGK = "2"
 SpecialFunctions = "1.2, 2"
 StatsBase = "0.32, 0.33"
 StatsFuns = "0.9.15, 1"
-julia = "1.3"
+julia = "1.6"
```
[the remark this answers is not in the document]
devmotion: Just saw your comment, it's because of `evalpoly`? We can just use `@evalpoly` which uses `evalpoly` in newer Julia versions.

src/univariate/continuous/lindley.jl
```diff
@@ -0,0 +1,184 @@
...
+
+### Statistics
+
+mean(d::Lindley) = (2 + d.θ) / d.θ / (1 + d.θ)
+
+var(d::Lindley) = 2 / d.θ^2 - 1 / (1 + d.θ)^2
+
+skewness(d::Lindley) = 2 * evalpoly(d.θ, (2, 6, 6, 1)) / evalpoly(d.θ, (2, 4, 1))^(3//2)
+
+kurtosis(d::Lindley) = 3 * evalpoly(d.θ, (8, 32, 44, 24, 3)) / evalpoly(d.θ, (2, 4, 1))^2 - 3
```
devmotion, line 54:
```suggestion
skewness(d::Lindley) = 2 * @evalpoly(d.θ, 2, 6, 6, 1) / @evalpoly(d.θ, 2, 4, 1)^(3//2)

kurtosis(d::Lindley) = 3 * @evalpoly(d.θ, 8, 32, 44, 24, 3) / @evalpoly(d.θ, 2, 4, 1)^2 - 3
```

src/univariate/continuous/lindley.jl
```diff
@@ -0,0 +1,184 @@
...
+        return _oftype(d, y, NaN)
+    elseif y > 0
+        if isfinite(y)
+            return log1p(θ * (1 + y)) - log1p(θ) - θ * y
+        else
+            return _oftype(d, y, -Inf)
+        end
+    else
+        return _zero(d, y)
+    end
```
devmotion, line 151:
```suggestion
    θy = θ * y
    res = log1p(θy / (1 + θ)) - θy
    return y < 0 ? zero(res) : (y == Inf ? oftype(res, -Inf) : res)
```

src/univariate/continuous/lindley.jl
```diff
@@ -0,0 +1,184 @@
...
+end
+
+cdf(d::Lindley, y::Real) = 1 - ccdf(d, y)
+
+logcdf(d::Lindley, y::Real) = log1p(-ccdf(d, y))
```
devmotion, line 156:
```suggestion
logcdf(d::Lindley, y::Real) = log1mexp(logccdf(d, y))
```

src/univariate/continuous/lindley.jl
```diff
@@ -0,0 +1,184 @@
...
+
+# Jodrá, P. (2010). Computer generation of random variables with Lindley or
+# Poisson–Lindley distribution via the Lambert W function. Mathematics and Computers
+# in Simulation, 81(4), 851–859.
+function quantile(d::Lindley, q::Real)
+    θ = shape(d)
+    return -1 - inv(θ) - lambertw((1 + θ) * (q - 1) / exp(1 + θ), -1) / θ
```
devmotion, line 163: Could avoid one division:
```suggestion
    return -(1 + (1 + lambertw((1 + θ) * (q - 1) / exp(1 + θ), -1)) / θ)
```

src/univariate/continuous/lindley.jl
```diff
@@ -0,0 +1,184 @@
...
+
+# Ghitany, M. E., Atieh, B., & Nadarajah, S. (2008). Lindley distribution and its
+# application. Mathematics and Computers in Simulation, 78(4), 493–506.
+function rand(rng::AbstractRNG, d::Lindley{T}) where {T}
+    θ = shape(d)
+    λ = inv(θ)
+    u = rand(rng, T)
+    p = θ / (1 + θ)
+    return T(rand(rng, u <= p ? Exponential{T}(λ) : Gamma{T}(2, λ)))
+end
```
devmotion, line 176: `T` indicates the parameter type, not the type of the samples. Also `λ` might not be of the same type as `θ`, and `rand(rng, T)` won't work if e.g. `T = Rational{Int}`. So I would suggest:
```suggestion
function rand(rng::AbstractRNG, d::Lindley)
    θ = shape(d)
    λ = inv(θ)
    if (1 + θ) * rand(rng) <= θ
        return rand(rng, Exponential{typeof(λ)}(λ))
    else
        return rand(rng, Gamma{typeof(λ)}(oftype(λ, 2), λ))
    end
```

src/univariate/continuous/lindley.jl
```diff
@@ -0,0 +1,184 @@
...
+
+    Lindley{T}(θ::T) where {T} = new{T}(θ)
+end
+
+function Lindley(θ::Real; check_args::Bool=true)
+    @check_args Lindley (θ, θ > zero(θ))
+    return Lindley{typeof(θ)}(θ)
+end
+
+Lindley(θ::Integer; check_args::Bool=true) = Lindley(float(θ); check_args=check_args)
```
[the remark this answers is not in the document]
ararslan: Originally I didn't have this but one of the automated tests checks that integer input gives you a float parameter.

Project.toml
```diff
@@ -7,6 +7,7 @@ version = "0.25.81"
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 DensityInterface = "b429d917-457f-4dbc-8f4c-0cc954292b1d"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
+LambertW = "984bce1d-4616-540c-a9ee-88d1112d94c9"
```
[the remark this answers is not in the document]
ararslan: Very fair and upon further inspection of the package I agree.

Project.toml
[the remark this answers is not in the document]
ararslan: Actually, `@horner` has existed for a long time and similarly uses `evalpoly` on newer Julia versions (though via `invoke` for some reason).

src/univariate/continuous/lindley.jl
ararslan: Good point though without some kind of type conversion this won't actually be type stable since `rand` for `Exponential` always gives `Float64` whereas `rand` for `Gamma` gives you the parameter type.

src/univariate/continuous/lindley.jl
```diff
@@ -0,0 +1,184 @@
...
+
+function pdf(d::Lindley, y::Real)
+    θ = shape(d)
+    if isnan(y)
+        return _oftype(d, y, NaN)
+    elseif isfinite(y) && y > 0
+        return θ^2 / (1 + θ) * (1 + y) * exp(-θ * y)
+    else
+        return _zero(d, y)
+    end
```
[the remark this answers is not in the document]
ararslan: Ah very nice, I figured there must be a better way to express these but didn't think too hard about it. Note that the distribution's support is `y > 0` though so the conditional should be `y <= 0`.

test/univariate/continuous/lindley.jl
```diff
@@ -0,0 +1,82 @@
+using Distributions
+using FiniteDifferences
+using ForwardDiff
+using Random
+using Test
+
+using Base: Fix1
+using Distributions: expectation
+using ForwardDiff: derivative
```
devmotion, line 9: This import causes a warning it seems.

src/univariate/continuous/lindley.jl
```diff
@@ -0,0 +1,180 @@
...
+
+mean(d::Lindley) = (2 + d.θ) / d.θ / (1 + d.θ)
+
+var(d::Lindley) = 2 / d.θ^2 - 1 / (1 + d.θ)^2
+
+skewness(d::Lindley) = 2 * @evalpoly(d.θ, 2, 6, 6, 1) / @evalpoly(d.θ, 2, 4, 1)^(3//2)
+
+kurtosis(d::Lindley) = 3 * @evalpoly(d.θ, 8, 32, 44, 24, 3) / @evalpoly(d.θ, 2, 4, 1)^2 - 3
+
+mode(d::Lindley) = d.θ < 1 ? (1 - d.θ) / d.θ : zero(d.θ)
```
devmotion, line 57: If theta is e.g. an integer, this would not be type stable. Maybe safer:
```suggestion
mode(d::Lindley) = max(0, (1 - d.θ) / d.θ)
```

src/univariate/continuous/lindley.jl
```diff
@@ -0,0 +1,180 @@
...
+
+mgf(d::Lindley, t::Real) = _lindley_mgf(shape(d), t)
+
+cf(d::Lindley, t::Real) = _lindley_mgf(shape(d), t * im)
+
+cgf(d::Lindley, t::Real) = log1p(-t / (1 + d.θ)) - 2 * log1p(-t / d.θ)
+
+_zero(d::Lindley, y::Real) = zero(shape(d)) * zero(y)
+_oftype(d::Lindley, y::Real, x) = oftype(_zero(d, y), x)
+
```
devmotion, line 91: Not needed anymore, I assume?
```suggestion
```

Project.toml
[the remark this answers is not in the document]
ararslan: Went with `@evalpoly` per your suggestion, which appears to exist in 1.3.

ararslan:
I really appreciate the thorough review, @devmotion!

devmotion (approved):
Looks good to me.

ararslan:
Thanks again!

PR #1679, opened 2023-02-15
Replace `GLOBAL_RNG` with `default_rng()`

quildtide:
See https://github.com/JuliaLang/julia/issues/48683 for some motivation on this.

`GLOBAL_RNG` can result in reduced performance when compared to the newer `default_rng()` in newer versions of Julia. `default_rng()` is present as early as Juliar 1.3, so switching to `default_rng()` should not be a major issue.

I actually first ran into this issue with `GLOBAL_RNG` while testing a faster vectorized method for `rand(Normal...)`. This pull request by itself does not impact performance significantly, but makes room for others like #1680 .

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -14,7 +14,7 @@ using LinearAlgebra, Printf
 import LinearAlgebra: dot, rank
 
 using Random
-import Random: GLOBAL_RNG, rand!, SamplerRangeInt
+import Random: default_rng, rand!, SamplerRangeInt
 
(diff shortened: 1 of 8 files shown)


devmotion (approved):
Yes, we should use `default_rng`. Are these all occurrences of `GLOBAL_RNG`?

quildtide:
Yep, I believe these are all cases, although I just realized that I forgot to rerun the docs (there is one docstring which has been updated).

devmotion:
Thank you!

PR #1680, opened 2023-02-15
Add vectorized rand for Normal dist

quildtide:
In the same spirit as #1269, but requires #1679 to be effective.

Before this and #1679:
```julia
julia> @time randn(n);
  0.002643 seconds (2 allocations: 9.419 MiB)

julia> @time randn(n);
  0.002520 seconds (2 allocations: 9.419 MiB)
  
julia> @time rand(Normal(0,1), n);
  0.008471 seconds (2 allocations: 9.419 MiB)

julia> @time rand(Normal(0,1), n);
  0.008220 seconds (2 allocations: 9.419 MiB)
```

After this and #1679:
```julia
julia> @time randn(n);
  0.002543 seconds (2 allocations: 9.419 MiB)

julia> @time randn(n);
  0.002482 seconds (2 allocations: 9.419 MiB)

julia> @time rand(Normal(0,1), n);
  0.002880 seconds (2 allocations: 9.419 MiB)

julia> @time rand(Normal(0,1), n);
  0.002934 seconds (2 allocations: 9.419 MiB)
```

Like #1269, this also improves compatibility with CUDA by replacing indexing with broadcasting; this does not require #1679:

```julia
using CUDA
using Random
using Distributions

CUDA.allowscalar(false);

v = CuArray{Float64}(undef, 1234567);

randn!(v); # works

rand!(CURAND.default_rng(), Normal(0,1), v); # doesn't work currently
```

```julia
julia> @time randn!(v);
  0.000065 seconds (5 allocations: 192 bytes)

julia> @time randn!(v);
  0.000065 seconds (5 allocations: 192 bytes)

julia> @time rand!(CURAND.default_rng(), Normal(0,1), v);
  0.000111 seconds (49 allocations: 2.672 KiB)

julia> @time rand!(CURAND.default_rng(), Normal(0,1), v);
  0.000103 seconds (49 allocations: 2.672 KiB)
```

### DIFF
--- a/src/univariate/continuous/normal.jl
+++ b/src/univariate/continuous/normal.jl
@@ -107,6 +107,9 @@ Base.:*(c::Real, d::Normal) = Normal(c * d.μ, abs(c) * d.σ)
 
 rand(rng::AbstractRNG, d::Normal{T}) where {T} = d.μ + d.σ * randn(rng, float(T))
 
+rand!(rng::AbstractRNG, d::Normal, A::AbstractArray{<:Real}) = 
+    A .= randn!(rng, A) .* d.σ .+ d.μ
+
 #### Fitting
 
 struct NormalStats <: SufficientStats
--- a/src/univariate/continuous/normal.jl
+++ b/src/univariate/continuous/normal.jl
@@ -107,8 +107,7 @@ Base.:*(c::Real, d::Normal) = Normal(c * d.μ, abs(c) * d.σ)
 
 rand(rng::AbstractRNG, d::Normal{T}) where {T} = d.μ + d.σ * randn(rng, float(T))
 
-rand!(rng::AbstractRNG, d::Normal, A::AbstractArray{<:Real}) = 
-    A .= randn!(rng, A) .* d.σ .+ d.μ
+rand!(rng::AbstractRNG, d::Normal, A::AbstractArray{<:Real}) = A .= muladd.(d.σ, randn!(rng, A), d.μ)
 
 #### Fitting
 


src/univariate/continuous/normal.jl
```diff
@@ -107,6 +107,9 @@ Base.:*(c::Real, d::Normal) = Normal(c * d.μ, abs(c) * d.σ)
 
 rand(rng::AbstractRNG, d::Normal{T}) where {T} = d.μ + d.σ * randn(rng, float(T))
 
+rand!(rng::AbstractRNG, d::Normal, A::AbstractArray{<:Real}) = 
+    A .= randn!(rng, A) .* d.σ .+ d.μ
```
devmotion, line 111: I wonder if it would be even better to use `muladd` here:
```suggestion
rand!(rng::AbstractRNG, d::Normal, A::AbstractArray{<:Real}) = A .= muladd.(d.σ, randn!(rng, A), d.μ)
```

devmotion: Seems to be a tiny bit faster in my benchmarks:
```julia
julia> using BenchmarkTools, Distributions

julia> n = 1_000;

julia> @btime randn($n);
  1.632 μs (1 allocation: 7.94 KiB)

julia> @btime rand($(Normal()), $n);
  3.692 μs (1 allocation: 7.94 KiB)

julia> f(d::Normal, n::Int) = d.μ .+ d.σ .* randn(n)

julia> @btime f($(Normal()), $n);
  2.296 μs (2 allocations: 15.88 KiB)

julia> @btime f($(Normal()), $n);
  2.282 μs (2 allocations: 15.88 KiB)

julia> g(d::Normal, n::Int) = muladd.(d.σ, randn(n), d.μ)

julia> @btime g($(Normal()), $n);
  2.248 μs (2 allocations: 15.88 KiB)

julia> @btime g($(Normal()), $n);
  2.244 μs (2 allocations: 15.88 KiB)
```

quildtide:
I actually set it up originally with `muladd`, then moved back to the operators because I was concerned that a pull request with `muladd` would raise concerns because it might return marginally different results on different machines despite a consistent PRNG with a seed and identical versions.

I personally also prefer `muladd` though. Happy to see that using it isn't a problem.

PR #1681, opened 2023-02-19
fix typos

spaette:
src/matrix/lkj.jl
src/truncated/exponential.jl
src/univariate/continuous/semicircle.jl
src/univariate/continuous/rayleigh.jl
src/univariate/continuous/weibull.jl
test/functionals.jl
test/runtests.jl
test/multivariate/mvnormal.jl
test/testutils.jl
test/univariate_bounds.jl


devmotion (approved):
Thank you!

Issue #1140, opened 2020-06-18
DiscreteNonParametric == throws error


fredcallaway:
```
julia> DiscreteNonParametric(1:2) === DiscreteNonParametric(1:3)
ERROR: DimensionMismatch("arrays could not be broadcast to a common size; got a dimension with lengths 2 and 3")
```

The root cause is the broadcasting on possibly different length arrays [here](https://github.com/JuliaStats/Distributions.jl/blob/20c91d9efcc5f96913bf8e38be2e3fb14b21942b/src/univariate/discrete/discretenonparametric.jl#L63). Should be an easy fix, just check that the supports have the same size before the broadcast. I can make a pull request if desired.

PR #1683, opened 2023-02-23
Fix comparisons of `DiscreteNonParametric`

### LINKED ISSUE #1675: Broken `≈` for the Categorical distribution on Julia 1.6
`isapprox` started to fail for the `Categorical` distribution on Julia 1.6. Works fine on other versions of Julia. Our CI started to fail recently, it check that the result of some computation is equal to some predefine answer with `isapprox`. These tests were working for at least a year as far as I remember, but started to fail recently only on Julia 1.6. 

MWE:

```bash
julia +1.6
```

```julia
julia> using Distributions

julia> Categorical([ 0.5, 0.5 ]) ≈ Categorical([ 0.5, 0.5])
ERROR: ArgumentError: step cannot be zero
Stacktrace:
  [1] steprange_last(start::Int64, step::Int64, stop::Int64)
    @ Base ./range.jl:221
  [2] StepRange
    @ ./range.jl:208 [inlined]
  [3] _rangestyle
    @ ./range.jl:118 [inlined]
  [4] _range
    @ ./range.jl:116 [inlined]
  [5] #range#57
    @ ./range.jl:91 [inlined]
  [6] -
    @ ./range.jl:1107 [inlined]
  [7] isapprox(x::Base.OneTo{Int64}, y::Base.OneTo{Int64}; atol::Int64, rtol::Int64, nans::Bool, norm::typeof(LinearAlgebra.norm))
    @ LinearAlgebra /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/generic.jl:1657
  [8] isapprox
    @ /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/generic.jl:1657 [inlined]
  [9] isapprox(c1::Categorical{Float64, Vector{Float64}}, c2::Categorical{Float64, Vector{Float64}})
    @ Distributions ~/.julia/packages/Distributions/bQ6Gj/src/univariate/discrete/discretenonparametric.jl:70
 [10] top-level scope
    @ REPL[2]:1
```

```julia
julia> versioninfo()
Julia Version 1.6.7
Commit 3b76b25b64 (2022-07-19 15:11 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin21.4.0)
  CPU: Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)
```

Works fine on Julia 1.8:

```julia
julia> Categorical([ 0.5, 0.5 ]) ≈ Categorical([ 0.5, 0.5])
true
```

Related CI: https://github.com/biaslab/ReactiveMP.jl/actions/runs/4087611306 (it includes other test failures as well, but I'm not sure if these are related)

devmotion:
Fixes #1675. Fixes https://github.com/JuliaStats/Distributions.jl/issues/1140.

### DIFF
--- a/src/univariate/discrete/categorical.jl
+++ b/src/univariate/discrete/categorical.jl
@@ -52,6 +52,15 @@ ncategories(d::Categorical) = support(d).stop
 params(d::Categorical{P,Ps}) where {P<:Real, Ps<:AbstractVector{P}} = (probs(d),)
 partype(::Categorical{T}) where {T<:Real} = T
 
+function Base.isapprox(c1::C, c2::C; kwargs...) where {C<:Categorical}
+    # support are of type Base.OneTo, so comparing the cardinality of the support
+    # is sufficient
+    # we explicitly redefine the method for `DiscreteNonParametric` which also compares
+    # the support since `isapprox(::OneTo, ::OneTo)` is broken on Julia 1.6 (issue #1675)
+    return length(support(c1)) == length(support(c2)) &&
+        isapprox(probs(c1), probs(c2); kwargs...)
+end
+
 ### Statistics
 
 function median(d::Categorical{T}) where {T<:Real}
--- a/src/univariate/discrete/categorical.jl
+++ b/src/univariate/discrete/categorical.jl
@@ -52,7 +52,7 @@ ncategories(d::Categorical) = support(d).stop
 params(d::Categorical{P,Ps}) where {P<:Real, Ps<:AbstractVector{P}} = (probs(d),)
 partype(::Categorical{T}) where {T<:Real} = T
 
-function Base.isapprox(c1::C, c2::C; kwargs...) where {C<:Categorical}
+function Base.isapprox(c1::Categorical, c2::Categorical; kwargs...)
     # support are of type Base.OneTo, so comparing the cardinality of the support
     # is sufficient
     # we explicitly redefine the method for `DiscreteNonParametric` which also compares
--- a/src/univariate/discrete/discretenonparametric.jl
+++ b/src/univariate/discrete/discretenonparametric.jl
@@ -63,13 +63,13 @@ Get the vector of probabilities associated with the support of `d`.
 """
 probs(d::DiscreteNonParametric)  = d.p
 
-==(c1::D, c2::D) where D<:DiscreteNonParametric =
-    (support(c1) == support(c2) || all(support(c1) .== support(c2))) &&
-    (probs(c1) == probs(c2) || all(probs(c1) .== probs(c2)))
-
-Base.isapprox(c1::D, c2::D) where D<:DiscreteNonParametric =
-    (support(c1) ≈ support(c2) || all(support(c1) .≈ support(c2))) &&
-    (probs(c1) ≈ probs(c2) || all(probs(c1) .≈ probs(c2)))
+function Base.isapprox(c1::D, c2::D; kwargs...) where D<:DiscreteNonParametric
+    support_c1 = support(c1)
+    support_c2 = support(c2)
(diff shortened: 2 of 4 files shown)


src/univariate/discrete/categorical.jl
```diff
@@ -52,6 +52,15 @@ ncategories(d::Categorical) = support(d).stop
 params(d::Categorical{P,Ps}) where {P<:Real, Ps<:AbstractVector{P}} = (probs(d),)
 partype(::Categorical{T}) where {T<:Real} = T
 
+function Base.isapprox(c1::C, c2::C; kwargs...) where {C<:Categorical}
```
bvdmitri, line 55: Wouldn't it better to define this method as
```suggestion
function Base.isapprox(c1::Categorical, c2::Categorical; kwargs...) 
```
There is a big difference in two definitions, as the latter dispatches on
```julia
Categorical([ 1.0, 0.0 ]) ≈ Categorical([ 1f0, 0f0 ]) # Note `Float64` and `Float32`
```

I actually checked this on master and the behaviour is interesting. It means that `Float64 ≈ Float32` dispatches on something else.
```
Categorical([ 1.0, 0.0 ]) ≈ Categorical([ 1.0, 0.0 ])
```

```julia
julia> Categorical([ 1.0, 0.0 ]) ≈ Categorical([ 1f0, 0f0 ])
true

julia> Categorical([ 1.0, 0.0 ]) ≈ Categorical([ 1.0, 0.0 ])
ERROR: ArgumentError: step cannot be zero
Stacktrace:
  [1] steprange_last(start::Int64, step::Int64, stop::Int64)
    @ Base ./range.jl:221
  [2] StepRange
    @ ./range.jl:208 [inlined]
  [3] _rangestyle
    @ ./range.jl:118 [inlined]
  [4] _range
    @ ./range.jl:116 [inlined]
  [5] #range#57
    @ ./range.jl:91 [inlined]
  [6] -
    @ ./range.jl:1107 [inlined]
  [7] isapprox(x::Base.OneTo{Int64}, y::Base.OneTo{Int64}; atol::Int64, rtol::Int64, nans::Bool, norm::typeof(LinearAlgebra.norm))
    @ LinearAlgebra /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/generic.jl:1657
  [8] isapprox
    @ /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/generic.jl:1657 [inlined]
  [9] isapprox(c1::Categorical{Float64, Vector{Float64}}, c2::Categorical{Float64, Vector{Float64}})
    @ Distributions /private/tmp/Distributions.jl/src/univariate/discrete/discretenonparametric.jl:70
 [10] top-level scope
    @ REPL[4]:1
```

devmotion: Good catch!

The reason for this different behavior is that the fallback in src/common.jl supports arguments of different types (but if the `nameof` the types - `DiscreteNonParametric` here - or the field names are not identical, `false` is returned). And the default definition there compares fields with `===` first before evaluating `==` or `isapprox`, and hence in this example for different the check of `Base.OneTo` supports with `isapprox` is never performed.

Since the fallback definition supports arguments of different types, I guess it would be better to support arguments of non-identical types here as well. I changed it.

devmotion:
I'll merge as this fixes some real issues. We can always make additional improvements in follow-up PRs.

PR #1685, opened 2023-02-25
add InverseFunctions support

aplavin:
Simple implementation, adds no additional dependencies - `InverseFunctions.jl` is already a dep of `Distributions.jl`:
```julia
(Distributions) pkg> why InverseFunctions
  DensityInterface → InverseFunctions
  StatsFuns → InverseFunctions
```


devmotion:
Can you make it a weak dependency? Regardless of whether DensityInterface will make it a weak dependency (I assume so), probably we will make DensityInterface a weak dependency in which case InverseFunctions won't be an indirect dependency anymore.

devmotion:
There's already a PR for StatsFuns, just requires someone's approval.

Project.toml
```diff
@@ -19,10 +19,17 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
+[weakdeps]
+InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
+
+[extensions]
+InverseFunctionsDistributionsExt = "InverseFunctions"
```
devmotion, line 26: Following SpecialFunctions etc. it should be
```suggestion
DistributionsInverseFunctionsExt  = "InverseFunctions"
```

ext/InverseFunctionsDistributionsExt.jl
```diff
@@ -0,0 +1,17 @@
+module InverseFunctionsDistributionsExt
+
+import InverseFunctions: inverse
```
devmotion, line 3: I'd prefer if functions are qualified explicitly when they are overloaded:
```suggestion
import InverseFunctions
```

ext/InverseFunctionsDistributionsExt.jl
```diff
@@ -0,0 +1,17 @@
...
+using Distributions
+
+
+for (fcdf, fqua) in [
+    (:cdf, :quantile),
+    (:ccdf, :cquantile),
+    (:logcdf, :invlogcdf),
+    (:logccdf, :invlogccdf),
+]
```
devmotion, line 12:
```suggestion
for (fcdf, fqua) in (
    (:cdf, :quantile),
    (:ccdf, :cquantile),
    (:logcdf, :invlogcdf),
    (:logccdf, :invlogccdf),
)
```

ext/InverseFunctionsDistributionsExt.jl
```diff
@@ -0,0 +1,17 @@
...
+    @eval inverse(f::Base.Fix1{typeof($fcdf), <:ContinuousUnivariateDistribution}) = Base.Fix1($fqua, f.x)
```
devmotion, line 13: In general, that's not correct, `x -> quantile(dist, x)` does not necessarily invert the result of `x -> cdf(dist, x)`:
```julia
julia> using Distributions

julia> quantile(Gamma(), cdf(Gamma(), -10))
0.0
```

devmotion:
I'm still not completely convinced by this PR, it's a bit unclear to me when these definitions are actually useful and hence justify the increased code complexity (in particular due to the complications arising from undefined inverses of `cdf` for `p = 0` and `p = 1`). So I'd put a "pending clear need" label (as used e.g. in ChainRules) here to indicate that the PR needs a convincing use case first.

In its current state, it also does not support Julia < 1.9 which seems unfortunate. Since InverseFunctions is a lightweight interface package, maybe it would be reasonable to make it a regular dependency on Julia < 1.9.

Any thoughts @ararslan or e.g. @sethaxen?

ararslan:
I agree with @devmotion here.

> A reason why I have this in my personal utility package, and why propose to add here

Can you explain how the code you've shown is related to InverseFunctions? Admittedly I don't quite follow it.

PR #1686, opened 2023-02-25
Make ChainRulesCore and DensityInterface weak dependencies

devmotion:
This PR makes ChainRulesCore and DensityInterface weak dependencies on newer Julia versions that support extensions (see, e.g., https://pkgdocs.julialang.org/dev/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)).


devmotion:
This PR is ready for review. The logs (https://github.com/JuliaStats/Distributions.jl/actions/runs/4272559663/jobs/7482712943) show that on Julia >= 1.9 ChainRulesCore and DensityInterface are not installed automatically anymore but the extensions are precompiled and loaded successfully once they are (implicitly) loaded in the tests.

KristofferC (approved):

devmotion:
It's easy to make it a proper dependency again, if needed. But currently it's "just" an additional interface that we support and hence a good fit for an extension, I think - even if it is only a minor contribution to load time. As soon as you load DensityInterface in your package/script, you have access to all methods but not every user might want to work with DensityInterface (similar to how not everyone needs ChainRules support or not every user of LogExpFunctions works with InverseFunctions).

oschulz:
> But currently it's "just" an additional interface that we support and hence a good fit for an extension

I agree, but since we we careful to make DensityInterface and InverseFunctions ultra-lightweight, what's the advantage of making them optional? I would guess that if both Distributions and  DensityInterface are used then the extension package might even add another ms of load time.

I agree that we should use the new Pkg extensions extensively, but I wonder if they useful for very lightweight deps  intended to be community standards.

PR #1688, opened 2023-03-01
pdf(d, missing) and pdf(missing, x)

### LINKED ISSUE #1392: Implement `pdf(d, missing)`
Be nice if `pdf(d, missing) = missing`. But we don't have that:

```julia
julia> d=Distributions.Normal()
Distributions.Normal{Float64}(μ=0.0, σ=1.0)

julia> pdf(d, missing)
ERROR: MethodError: no method matching pdf(::Distributions.Normal{Float64}, ::Missing)
```

ArunSanganal:
closes #1392.


src/common.jl
```diff
@@ -225,6 +225,15 @@ function _pdf(d::Distribution{ArrayLikeVariate{N}}, x::AbstractArray{<:Real,N})
     return exp(@inbounds logpdf(d, x))
 end
 
+function pdf(
+    d::Union{Distribution{ArrayLikeVariate{N}},Missing}, x::Any
```
devmotion, line 229: I think we should not add a method with such a signature. I think we should not allow `d::Missing`, and `x::Any` seems too arbitrary.

devmotion (changes requested):
Thank you for the PR!

Unfortunately, in my opinion it should not be merged in its current form.

More generally, I think #1392 needs more thought and discussion. In my opinion, it is a design problem of `Missing` that somehow people now expect every method to support `missing` (just check how many open issues there are in Statistics stdlib about handling missing values). So I'm not excited about it, but I would be fine with adding `pdf(d::Distribution, x::Missing) = missing` irrespective of these deficiencies, given how widespread implementations with `missing` have become.

However, I don't see a compelling reason to add `pdf(d::Missing, x::Real)` or similar. `x::Missing` corresponds to missing data but in what situations would one encounter a `missing` distribution?

Another point is: Why should one only handle `pdf`? What about `logpdf`? `cdf`? etc.

---

In addition to these general concerns, I think there are some other problems with this PR:
- `.DS_Store` should not be added
- `x::Any` is too generic, IMO a method for `x::Missing` is sufficient
- The definition creates a stackoverflow if neither `d` nor `x` are `missing`
- More tests should be added

ArunSanganal:
> Thank you for the PR!
> 
> Unfortunately, in my opinion it should not be merged in its current form.
> 
> More generally, I think #1392 needs more thought and discussion. In my opinion, it is a design problem of `Missing` that somehow people now expect every method to support `missing` (just check how many open issues there are in Statistics stdlib about handling missing values). So I'm not excited about it, but I would be fine with adding `pdf(d::Distribution, x::Missing) = missing` irrespective of these deficiencies, given how widespread implementations with `missing` have become.
> 
> However, I don't see a compelling reason to add `pdf(d::Missing, x::Real)` or similar. `x::Missing` corresponds to missing data but in what situations would one encounter a `missing` distribution?
> 
> Another point is: Why should one only handle `pdf`? What about `logpdf`? `cdf`? etc.
> 
> In addition to these general concerns, I think there are some other problems with this PR:
> 
> * `.DS_Store` should not be added
> * `x::Any` is too generic, IMO a method for `x::Missing` is sufficient
> * The definition creates a stackoverflow if neither `d` nor `x` are `missing`
> * More tests should be added

Thank you for the review however for the second point, `pdf(missing, d)` was requested as well. If we just use `x::Missing`, it will again give a method error. But I agree that there's a very minimal chance of it happening. I will look into points 4 and 5.

devmotion:
>  however for the second point, pdf(missing, x) was requested as well.

Not every feature request is uncontroversial and should be implemented :slightly_smiling_face:

src/common.jl
```diff
@@ -225,6 +225,10 @@ function _pdf(d::Distribution{ArrayLikeVariate{N}}, x::AbstractArray{<:Real,N})
     return exp(@inbounds logpdf(d, x))
 end
 
+function pdf(d::Distribution{ArrayLikeVariate{N}}, x::Missing) where N
+    return missing
+end
```
devmotion, line 230: A bit shorter and also with a definition for `logpdf`:
```suggestion
pdf(::Distribution{<:ArrayLikeVariate}, ::Missing) = missing
logpdf(::Distribution{<:ArrayLikeVariate}, ::Missing) = missing
```

test/multivariate/dirichlet.jl
```diff
@@ -39,6 +39,7 @@ rng = MersenneTwister(123)
         @test iszero(pdf(d, [0, 0, 1]))
         @test pdf(d, [0.2, 0.3, 0.5]) ≈ 3.6
         @test pdf(d, [0.4, 0.5, 0.1]) ≈ 2.4
+        @test typeof(pdf(d, missing)) == Missing
```
devmotion, line 42: Simpler:
```suggestion
        @test pdf(d, missing)) === missing
```
IMO it would also be better to not add this check to all kinds of different files but to the generic test interfaces in e.g. `test/testutils.jl`.

PR #1689, opened 2023-03-03
move ChainRulesCore to a weak dependency

KristofferC:


devmotion:
Should we go with #1686 instead?

KristofferC:
Sure, missed that

Issue #1684, opened 2023-02-24
Accept `Number` types in distribution constructors.


rafaqz:
There are many `Number` typed objects that would work just fine in a distribution but currently don't. #1413 has examples where units would already work in a lot of cases by just using `Number`.

ModelParameters.jl is a similar case. It has `Number` typed parameters `Param` so that e.g. Unitful.jl units can be put in them.

But then they don't work as input to Distributions.jl objects when users try to do that:
https://github.com/rafaqz/ModelParameters.jl/issues/51

Embedding parametrised distributions in large nested models could be extremely easy with this method.

The alternative is we add a parallel `RealParam` object and add new types and methods to AbstractNumbers.jl to work on `Real`. This is a _lot_ of work and code duplication compared to simply swapping `Real` to `Number` here.

devmotion:
I thought #1413 showed that just replacing `Real` with `Number` would lead to many functions returning incorrect units?

rafaqz:
That's not the impression I got - there are some errors, but a lot of things actually just work? Like these examples https://github.com/JuliaStats/Distributions.jl/issues/1413#issuecomment-1014195306

I guess the question is do you want to guarantee that there are never and wrong outputs. And accepting `Number` means you cant. I get why that's a problem.

But it leaves this impasse where it is impossible to generically wrap numbers in another object and let them be useful in the ecosystem.

We really need a trait-based number system in base so you could call `isreal`, `iscomplex` on a number rather than being stuck in the hierarchy.

Then these could be checked in the constructor to guarantee correct results, rather than fixing the type to `Real`.

mschauer:
Relatedly, for example, https://github.com/gaurav-arya/StochasticAD.jl/blob/576d4e9026e2b4c225b2aab5910f4b6148a66666/src/discrete_randomness.jl#L110 is to circumvent that the Binomials don't allow numbers @gaurav-arya

devmotion:
Regarding the example, I think it's annoying if constructors "lie" about the type of the instance they return.

I think it would be nice to generalize the definitions. I'm just not a fan of simple search-replace approaches, I think it deserves to take one distribution, put together a PR for this specific distribution only, and discuss it intensively to figure out all problems and fix them. That would make the overall design and potential pitfalls much clearer. Afterwards one could generalize it to other distributions.

[...]

PR #1691, opened 2023-03-08
accept `Number` in `Exponential`

rafaqz:
This is a start on #1684, with `Exponential` as an example to start the discussion, I'm not personally across the problems this may cause so this PR is a request for input and feedback on that.

The idea is to loosen the type of `θ` to be `Number`, and add a check for `Complex` numbers so those errors are avoided. More or different checks may be required.

I'm not totally satisfied with the combination of `!(θ isa Complex)` and `isreal(θ)` checks, but the idea is to make sure that even wrapped numbers are internally `Real`.

### DIFF
--- a/src/univariate/continuous/exponential.jl
+++ b/src/univariate/continuous/exponential.jl
@@ -21,13 +21,18 @@ External links
 * [Exponential distribution on Wikipedia](http://en.wikipedia.org/wiki/Exponential_distribution)
 
 """
-struct Exponential{T<:Real} <: ContinuousUnivariateDistribution
+struct Exponential{T<:Number} <: ContinuousUnivariateDistribution
     θ::T        # note: scale not rate
     Exponential{T}(θ::T) where {T} = new{T}(θ)
 end
 
-function Exponential(θ::Real; check_args::Bool=true)
-    @check_args Exponential (θ, θ > zero(θ))
+function Exponential(θ::Number; check_args::Bool=true)
+    @check_args(
+        Exponential, 
+        (!(θ isa Complex), "Complex not allowed for θ"),
+        (isreal(θ), "θ must be a real number"),
+        (θ, θ > zero(θ)),
+    )
     return Exponential{typeof(θ)}(θ)
 end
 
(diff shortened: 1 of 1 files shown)


ParadaCarleton:
The problem here is related to Unitful, right? I'm happy to merge this if we need to, but first I'd like to check if it's possible to fix this in Unitful. @sostock ?

devmotion:
IMO this needs more thought as it affects much more than just `Exponential` (see https://github.com/JuliaStats/Distributions.jl/issues/1684). I also agree that the checks introduced in this PR are suboptimal.

rafaqz:
Its not related to Unitful.jl - its a generic problem with `Number`.

See here:
https://github.com/rafaqz/ModelParameters.jl/issues/51

I don't have bandwidth to make this "optimal" or think through the implications, that is clearly a problem for devs of this package.

I was just proposing a solution that would work for other users of my own packages, as well as Unitful.jl

sostock:
Another option would be to make the support for non-`Real` types opt-in, so that this package only supports `Exponential{<:Real}`, but other packages can add support for their types as package extensions.

For that, one would change the type parameter to `T<:Number` and rename the internal constructor to something like `unsafe_exponential` while providing an outer constructor only for `Exponential{<:Real}`. Other packages (like Unitful.jl) can then create a package extension in which they define their own outer constructor that calls the inner `unsafe_exponential` function.

src/univariate/continuous/exponential.jl
```diff
@@ -37,7 +42,7 @@ Exponential() = Exponential{Float64}(1.0)
 @distr_support Exponential 0.0 Inf
 
 ### Conversions
-convert(::Type{Exponential{T}}, θ::S) where {T <: Real, S <: Real} = Exponential(T(θ))
+convert(::Type{Exponential{T}}, θ::S) where {T <: Number, S <: Number} = Exponential(T(θ))
```
sostock, line 45:
```suggestion
convert(::Type{Exponential{T}}, θ::S) where {T <: Number, S <: Number} = Exponential(convert(T, θ))
```
To support Unitful quantities correctly, `T(θ)` should be changed to `convert(T, θ)`. Otherwise you could convert a unitless number to a unitful `Exponential` (i.e., `convert(Exponential{typeof(1.0m)}, 1.0)`).

PR #1694, opened 2023-03-13
added documentation of statistics functions

sushi26:
I think the negativebinomial.jl has been finished.

### DIFF
--- a/src/univariate/discrete/negativebinomial.jl
+++ b/src/univariate/discrete/negativebinomial.jl
@@ -14,7 +14,7 @@ P(X = k) = \\frac{\\Gamma(k+r)}{k! \\Gamma(r)} p^r (1 - p)^k, \\quad \\text{for 
 ```
 
 ```julia
-NegativeBinomial()        # Negative binomial distribution with r = 1 and p = 0.5
+NegativeBinomial()        # Negative binomial stribution with r = 1 and p = 0.5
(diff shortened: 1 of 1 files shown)


mschauer:
We should link the Wiki

enovu:
> Hi, can you explain the motivation for this PR? It's a bit unclear to me what's your intention here.

I’m new to open source and am trying to contribute but am starting small.

devmotion:
That's great :slightly_smiling_face: But can you explain why you put together this PR and why you think it would be an improvement?

enovu:
For this specific one the definition we use is different from the definition on wikipedia. We use the mathematica definition so I thought it would be clearer with this information. Do you have any suggestions where I can start - some low hanging fruit would be great. 

I was also testing how to make a PR so didn't want to mess something up by accident which is why it was a simple PR.

devmotion:
> For this specific one the definition we use is different from the definition on wikipedia. We use the mathematica definition so I thought it would be clearer with this information.

This is already mentioned in the docstring currently, it seems? I don't there's an improvement regarding this point in the PR, or maybe I missed it?

Regarding the statistics functions, is there a good reason to add them? The docs are already quite lengthy and I don't think the list is complete, so maybe it would be better to point out things that are not implement or do not exist when needed instead and otherwise refer to the description of the general interface for univariate distributions. (As a more specific point, also the functions cannot be called in the way that they are listed here.)

Issue #294, opened 2014-11-03
Replace Rmath RNGs


simonbyrne:
As the Rmath-based `rand` methods are now broken in 0.4 (https://github.com/JuliaLang/julia/issues/8874), this seems like as good as time as any to implement them in Julia. We need:
- [x] `Beta` (Fixed in https://github.com/JuliaStats/Distributions.jl/pull/830)
  - Can generate two `Chisq`s and `a/(a+b)`, or
  - R uses ["Generating beta variates with nonintegral shape parameters" by R.C.H. Cheng](http://dx.doi.org/10.1145/359460.359482). 
- [x] `Binomial` (Fixed in https://github.com/JuliaStats/Distributions.jl/pull/830)
- [x] `Chisq` (Fixed in https://github.com/JuliaStats/Distributions.jl/pull/830)
  - Can use `Gamma` method.
- [x] `FDist` (Fixed in https://github.com/JuliaStats/Distributions.jl/pull/830)
  - Can use ratio of `ChiSq`s.
- [x] `Gamma` (Fixed in https://github.com/JuliaStats/Distributions.jl/pull/830)
- [ ] `Hypergeometric`
  - R uses  ["Computer generation of hypergeometric random variates" by Kachitvichyanukul and Schmeiser](http://dx.doi.org/10.1080/00949658508810839), though apparently there's an error in there somewhere.
  - [Numpy](https://github.com/numpy/numpy/blob/master/numpy/random/mtrand/distributions.c) uses ratio-of-uniforms, which seems to be the [HRUA algorithm](http://www.researchgate.net/publication/256371290_Sampling_from_Poisson_binomial_and_hypergeometric_distributions_ratio_of_uniforms_as_a_simple_and_fast_alternative).
- [ ] `NegativeBinomial`
- [ ] `NoncentralChisq`
- [ ] `Poisson`
- [ ] `Skellam`
- [x] `TDist` (Fixed in https://github.com/JuliaStats/Distributions.jl/pull/830)
  - Use `Normal`/`Chisq` mixture.

Some of these have been implemented in the `src/samplers/` directory. They still need some checking and performance tuning.

[...]

PR #1698, opened 2023-03-18
Native RNGs

ArunSanganal:
Targeted at #294. Removes Rmath dependence of RNGs for Noncentral distributions.

### DIFF
--- a/src/univariate/continuous/noncentralbeta.jl
+++ b/src/univariate/continuous/noncentralbeta.jl
@@ -39,9 +39,6 @@ partype(::NoncentralBeta{T}) where {T} = T
 
 @_delegate_statsfuns NoncentralBeta nbeta α β λ
 
-# TODO: remove RFunctions dependency once NoncentralChisq has its removed
-@rand_rdist(NoncentralBeta)
-
 function rand(d::NoncentralBeta)
     β = d.β
     a = rand(NoncentralChisq(2d.α, β))
--- a/src/univariate/continuous/noncentralchisq.jl
+++ b/src/univariate/continuous/noncentralchisq.jl
@@ -79,6 +79,9 @@ end
 
 @_delegate_statsfuns NoncentralChisq nchisq ν λ
 
-# TODO: remove RFunctions dependency
-@rand_rdist(NoncentralChisq)
-rand(d::NoncentralChisq) = StatsFuns.RFunctions.nchisqrand(d.ν, d.λ)
+function rand(d::NoncentralChisq, rng::AbstractRNG=Random.GLOBAL_RNG)
+    χ² = Chisq(d.ν)
+    poi = Poisson(d.λ / 2)
+    x = rand(rng, χ²) + 2 * rand(rng, poi)
+    return x
+end
--- a/src/univariate/continuous/noncentralf.jl
+++ b/src/univariate/continuous/noncentralf.jl
@@ -56,8 +56,6 @@ function rand(rng::AbstractRNG, d::NoncentralF)
     r1 / r2
 end
 
-# TODO: remove RFunctions dependency once NoncentralChisq has its removed
-@rand_rdist(NoncentralF)
 function rand(d::NoncentralF)
     r1 = rand(NoncentralChisq(d.ν1,d.λ)) / d.ν1
     r2 = rand(Chisq(d.ν2)) / d.ν2


src/univariate/continuous/noncentralbeta.jl
```diff
@@ -39,9 +39,6 @@ partype(::NoncentralBeta{T}) where {T} = T
 
 @_delegate_statsfuns NoncentralBeta nbeta α β λ
 
-# TODO: remove RFunctions dependency once NoncentralChisq has its removed
-@rand_rdist(NoncentralBeta)
-
 function rand(d::NoncentralBeta)
     β = d.β
     a = rand(NoncentralChisq(2d.α, β))
```
devmotion, line 44: Similarly, we need
```suggestion
    a = rand(rng, NoncentralChisq(2d.α, β))
```
here.

devmotion: And similarly in the line below, but it seems I can't add a suggestion there.

src/univariate/continuous/noncentralchisq.jl
```diff
@@ -79,6 +79,9 @@ end
 
 @_delegate_statsfuns NoncentralChisq nchisq ν λ
 
-# TODO: remove RFunctions dependency
-@rand_rdist(NoncentralChisq)
-rand(d::NoncentralChisq) = StatsFuns.RFunctions.nchisqrand(d.ν, d.λ)
+function rand(d::NoncentralChisq, rng::AbstractRNG=Random.GLOBAL_RNG)
```
devmotion, line 82: This should be
```suggestion
function rand(rng::AbstractRNG, d::NoncentralChisq)
```

src/univariate/continuous/noncentralchisq.jl
```diff
@@ -79,6 +79,9 @@ end
...
+    χ² = Chisq(d.ν)
+    poi = Poisson(d.λ / 2)
+    x = rand(rng, χ²) + 2 * rand(rng, poi)
+    return x
```
devmotion, line 86: `rand` should return a value of type `eltype(d)`:
```suggestion
    return eltype(d)(x)
```

src/univariate/continuous/noncentralf.jl
```diff
@@ -56,8 +56,6 @@ function rand(rng::AbstractRNG, d::NoncentralF)
     r1 / r2
 end
 
-# TODO: remove RFunctions dependency once NoncentralChisq has its removed
-@rand_rdist(NoncentralF)
 function rand(d::NoncentralF)
```
devmotion, line 59: Similar to above, this has to be changed to
```suggestion
function rand(rng::AbstractRNG, d::NoncentralF)
```

src/univariate/continuous/noncentralf.jl
```diff
@@ -56,8 +56,6 @@ function rand(rng::AbstractRNG, d::NoncentralF)
...
     r1 = rand(NoncentralChisq(d.ν1,d.λ)) / d.ν1
```
devmotion, line 60:
```suggestion
    r1 = rand(rng, NoncentralChisq(d.ν1,d.λ)) / d.ν1
```

src/univariate/continuous/noncentralf.jl
```diff
@@ -56,8 +56,6 @@ function rand(rng::AbstractRNG, d::NoncentralF)
...
     r2 = rand(Chisq(d.ν2)) / d.ν2
```
devmotion, line 61:
```suggestion
    r2 = rand(rng, Chisq(d.ν2)) / d.ν2
```

devmotion (commented):
Thank you for the PR! I made some suggestions below. We should also make sure that everything is covered by the tests. In principle, the standard tests for univariate distributions include tests for sampling but maybe the distributions with Rmath-based sampling were explicitly excluded.

src/univariate/continuous/noncentralf.jl
ArunSanganal: Don't we already have [this](https://github.com/JuliaStats/Distributions.jl/blob/master/src/univariate/continuous/noncentralf.jl#L53) ?

devmotion: Ah I hadn't noticed this since these lines were hidden in the diff. If methods with signature `rand(::AbstractRNG, ::Distribution)` already exist, we should just remove `rand(::Distribution)` completely (and check that the existing method with RNG is correct).

ArunSanganal:
> Thank you for the PR! I made some suggestions below. We should also make sure that everything is covered by the tests. In principle, the standard tests for univariate distributions include tests for sampling but maybe the distributions with Rmath-based sampling were explicitly excluded.

Not really sure on how to test the RNGs 😓

ArunSanganal:
We cannot add tests for `NoncentralBeta` till #1700 gets merged.

test/univariate/continuous/noncentralchisq.jl
```diff
@@ -1 +1,12 @@
...
+using Test
+
+@testset "Native RNG" begin
+    rng = MersenneTwister(1234)
+    n = 5000
+    d = NoncentralChisq(3, 2)
+    samples = rand(rng, d, n)
+    @test mean(samples) ≈ mean(d) atol=0.05*mean(d)
+    @test var(samples) ≈ var(d) atol=0.1*var(d)
+end
```
devmotion, line 12: We should just use the existing test functionality, by making sure these distributions are included in the tests (also with `rand`) run by `test/univariates.jl`.

devmotion (commented):
Testing `rand` should be independent of the other PR - we don't need to know `mean` and `var`, being able to compute the cdf/quantile should be sufficient. As commented below, we should also use the existing test machinery instead of adding custom tests.

test/univariate/continuous/noncentralchisq.jl
ArunSanganal: Would something like this work? It more or less tests the "quality" of RNGs.
        ``samples = rand(rng, d, n)
        cdf_v = [cdf(d, x) for x in samples]
        quantiles = [quantile(d, p) for p in cdf_v]
        @test all(quantiles .≈ samples)`` where d is ContinuousUnivariateDistribution

devmotion: No, my point is that we should not come up with such new tests. We should just use the setup in test/univariates.jl, or if that's not possible (eg if mean and variance are not defined) we should at least use https://github.com/JuliaStats/Distributions.jl/blob/ec68da3a8d4a4776367f2d7ca5ec2d4666e29c78/test/testutils.jl#L99.

ArunSanganal: Ah ok. Similar to `test_cgf`.

ArunSanganal: > We should just use the existing test functionality, by making sure these distributions are included in the tests (also with `rand`) run by `test/univariates.jl`.

univariates.jl does not have any method with the `rng` parameter.

ArunSanganal: > No, my point is that we should not come up with such new tests. We should just use the setup in test/univariates.jl, or if that's not possible (eg if mean and variance are not defined) we should at least use
> 
> https://github.com/JuliaStats/Distributions.jl/blob/ec68da3a8d4a4776367f2d7ca5ec2d4666e29c78/test/testutils.jl#L99
> 
> .

Ah, like the `test_cgf` added in some test files?

ayushpatnaikgit:
What's left in the PR @ArunSanganal?

ArunSanganal:
> What's left in the PR @ArunSanganal?

It should be good to merge if the tests added are satisfactory.

test/univariate/continuous/noncentralbeta.jl
```diff
@@ -0,0 +1 @@
+test_samples(NoncentralBeta(3,2,1), 5000; rng=MersenneTwister(1234), nbins=50)
```
devmotion: This is already tested (in a better way) in https://github.com/JuliaStats/Distributions.jl/blob/35125579a77b938f8d4a8bd2be23cbf1f2ddf225/test/univariates.jl#L152

test/univariate/continuous/noncentralchisq.jl
```diff
@@ -1 +1,15 @@
...
+test_samples(NoncentralF(5, 10,5), 5000; rng=MersenneTwister(1234), nbins=50)
+
+@testset "Native RNG" begin
+    rng = MersenneTwister(1234)
+    n = 5000
+    d = NoncentralChisq(3, 2)
+    samples = rand(rng, d, n)
+    @test mean(samples) ≈ mean(d) atol=0.05*mean(d)
+    @test var(samples) ≈ var(d) atol=0.1*var(d)
+end
```
devmotion: Same here, this is already tested in a better way in https://github.com/JuliaStats/Distributions.jl/blob/35125579a77b938f8d4a8bd2be23cbf1f2ddf225/test/univariates.jl#L152

devmotion:
> It should be good to merge if the tests added are satisfactory.

I improved the tests. They were a bit redundant and even a bit worse than some existing tests, it seems. I increased the number of samples used in the standard tests as well.

ArunSanganal:
It says, `counts are out of Confidence Interval`. Where are we getting the values of Distribution parameters from? I am not able to reproduce the same for custom values.

ParadaCarleton:
> It says, `counts are out of Confidence Interval`. Where are we getting the values of Distribution parameters from? I am not able to reproduce the same for custom values.

The RNG isn't working, in that case. You should check your implementation carefully for any bugs, e.g. using the wrong parameters or parameter name.

In addition, you can try increasing the number of samples and changing the seed for the RNG to see if it might just be bad luck.

oscardssmith:
This should be closed in favor of https://github.com/JuliaStats/Distributions.jl/pull/2027

PR #1699, opened 2023-03-19
Add the Kumaraswamy distribution

ararslan:


ararslan:
Want to take a look at this, @devmotion?

src/Distributions.jl
```diff
@@ -25,6 +25,7 @@ import StatsBase: kurtosis, skewness, entropy, mode, modes,
 import PDMats: dim, PDMat, invquad
 
 using SpecialFunctions
+using Base.MathConstants
```
devmotion, line 28: Can we qualify `eulergamma` instead? I guess it could be slightly dangerous/surprising if we load generic symbols such as greek letters gamma and phi in the Distributions module (https://github.com/JuliaLang/julia/blob/8e630552924eac54c809aa7bc30871c7df1582d3/base/mathconstants.jl#L11).
```suggestion
```

src/univariate/continuous/kumaraswamy.jl
```diff
@@ -0,0 +1,139 @@
...
+logcdf(d::Kumaraswamy, x::Real) = log1mexp(logccdf(d, x))
+
+function quantile(d::Kumaraswamy, q::Real)
+    a, b = params(d)
+    return (1 - (1 - q)^inv(b))^inv(a)
+end
+
+function entropy(d::Kumaraswamy)
+    a, b = params(d)
+    H = digamma(b + 1) + eulergamma
```
devmotion, line 86:
```suggestion
    H = digamma(b + 1) + Base.MathConstants.eulergamma
```

src/univariate/continuous/kumaraswamy.jl
```diff
@@ -0,0 +1,139 @@
...
+    b::T
+end
+
+function Kumaraswamy(a::Real, b::Real; check_args::Bool=true)
+    @check_args Kumaraswamy (a, a > zero(a)) (b, b > zero(b))
+    a′, b′ = promote(a, b)
+    return Kumaraswamy{typeof(a′)}(a′, b′)
+end
+
+Kumaraswamy() = Kumaraswamy{Float64}(1.0, 1.0)
```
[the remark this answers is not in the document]
ararslan: This was for consistency with other distribution types, which default to using `Float64` as the parameter type when using the 0-argument constructor.

src/univariate/continuous/kumaraswamy.jl
```diff
@@ -0,0 +1,139 @@
...
+    a, b = params(d)
+    y = a * b * x^(a - 1) * (1 - x^a)^(b - 1)
+    return 0 < x < 1 ? y : (isnan(x) ? oftype(y, NaN) : zero(y))
+end
+
+function logpdf(d::Kumaraswamy, x::Real)
+    a, b = params(d)
+    _x = clamp(x, 0, 1)  # Ensures we can still get a value when outside the support
+    y = log(a) + log(b) + (a - 1) * log(_x) + (b - 1) * log1p(-_x^a)
+    return 0 < x < 1 ? y : (isnan(x) ? oftype(y, NaN) : oftype(y, -Inf))
```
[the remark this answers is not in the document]
ararslan: Yeah it appears this does disagree with `Beta` at the boundaries in some cases, good catch. This formulation doesn't handle `NaN` though, that will still need to be checked explicitly.

src/univariate/continuous/kumaraswamy.jl
```diff
@@ -0,0 +1,138 @@
...
+
+### Evaluation
+
+# `pdf`: Uses fallback `exp(logpdf(_))` method
+
+function logpdf(d::Kumaraswamy, x::Real)
+    a, b = params(d)
+    _x = clamp(x, 0, 1)  # Ensures we can still get a value when outside the support
+    y = log(a) + log(b) + xlogy(a - 1, _x) + xlog1py(b - 1, -_x^a)
+    return insupport(d, x) ? y : (isnan(x) ? oftype(y, NaN) : oftype(y, -Inf))
```
devmotion, line 56: Sorry, my previous suggestion was wrong - the following condition should allow to remove the `NaN` special case:
```suggestion
    return x < 0 || x > 1 ? oftype(y, -Inf) : y
```

src/univariate/continuous/kumaraswamy.jl
```diff
@@ -0,0 +1,138 @@
...
+    H = digamma(b + 1) + eulergamma
+    return (1 - inv(b)) + (1 - inv(a)) * H - log(a) - log(b)
+end
+
+function gradlogpdf(d::Kumaraswamy, x::Real)
+    a, b = params(d)
+    _x = clamp(x, 0, 1)
+    _xᵃ = _x^a
+    y = (a * (b * _xᵃ - 1) + (1 - _xᵃ)) / (_x * (_xᵃ - 1))
+    return insupport(d, x) ? y : (isnan(x) ? oftype(y, NaN) : zero(y))
```
devmotion, line 91: Similarly:
```suggestion
    return x < 0 || x > 1 ? zero(y) : y
```

src/univariate/continuous/kumaraswamy.jl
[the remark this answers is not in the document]
devmotion: I see. I think this should maybe changed generally but it's fine with me to be consistent here.

ararslan:
The failing integration tests with Tracker and ReverseDiff seem unrelated to this PR.

devmotion:
Yeah, I'm not sure yet what's going on there. They failed on master as well but not in the last PR that was merged, and CI in DistributionsAD ran successfully just yesterday. So it might not even be related to a change in Distributions.

devmotion (approved):

devmotion:
FYI the downstream tests fail due to FillArrays 1.0.1: https://github.com/JuliaArrays/FillArrays.jl/issues/252

PR #1700, opened 2023-03-21
Added mean and variance for NoncentralBeta

ArunS-tack:
Working on #1698, I discovered that NoncentraBeta Distribution does not have mean and variance functions. Opening a PR to tackle that.

### DIFF
--- a/test/univariate/continuous/noncentralbeta.jl
+++ b/test/univariate/continuous/noncentralbeta.jl
@@ -3,10 +3,10 @@ using Test
 
 
 @testset "NoncentralBeta" begin 
-    α = 11.54 ; β = 8.89; λ = 5.0
+    α = 11.54 ; β = 8.89; λ = 5/4
     d = NoncentralBeta(α, β, λ)
-    @test isapprox(mean(d), 0.6105, atol=1e-4)
-    @test isapprox(var(d), 0.0106, atol=1e-4)
+    @test isapprox(mean(d), 0.5772, atol=1e-4)
+    @test isapprox(var(d), 0.0113, atol=1e-4)
     λ = 0.0
     d = NoncentralBeta(α, β, λ)
     d1 = Beta(α, β)


ararslan (commented):
Thanks for the contribution! Just some comments and questions.

src/Distributions.jl
```diff
@@ -8,6 +8,9 @@ import Base: size, length, convert, show, getindex, rand, vec, inv
 import Base: sum, maximum, minimum, extrema, +, -, *, ==
 import Base.Math: @horner
 
+using HypergeometricFunctions
+import HypergeometricFunctions: pFq
```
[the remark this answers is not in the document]
ArunSanganal: Oh, alright. It's my first time adding a dependency to a package, I will keep that in mind.

src/univariate/continuous/noncentralbeta.jl
```diff
@@ -35,11 +35,16 @@ partype(::NoncentralBeta{T}) where {T} = T
 
 ### Evaluation & Sampling
 
-# TODO: add mean and var
+mean(d::NoncentralBeta) = (d.α / (d.α + d.β)) * exp(-d.λ / 2) * HypergeometricFunctions.pFq([1 + d.α, d.α + d.β],[d.α, d.α + d.β + 1], d.λ / 2)
```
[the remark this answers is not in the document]
ArunSanganal: We could but I actually followed [this](https://reference.wolfram.com/language/ref/NoncentralBetaDistribution.html) assuming they give the same result.

ArunSanganal:
Thank you for the review. I have committed the suggestions and will take care in future to make the code as clean as possible.

ArunSanganal:
Oops. Sorry for so many review requests, I thought the it was bugged and clicked it way too many times.😓

test/univariate/continuous/noncentralbeta.jl
```diff
@@ -0,0 +1,15 @@
+using Distributions
+using Test
+
+
+@testset "NoncentralBeta" begin 
+    α = 2.0; β = 3.0; λ = 1.0
```
devmotion, line 6: Could we actually test more random/less nice values?  Or directly use the R reference setup?

src/univariate/continuous/noncentralbeta.jl
```diff
@@ -35,11 +35,22 @@ partype(::NoncentralBeta{T}) where {T} = T
 
 ### Evaluation & Sampling
 
-# TODO: add mean and var
+function mean(d::NoncentralBeta)
+    α, β, λ = params(d)
+    return α / (α + β) * exp(-λ / 2) * pFq([1 + α, α + β], [α, α + β + 1], λ / 2)
```
devmotion, line 40: I wonder if this can be simplified further, given e.g. the formula in https://en.wikipedia.org/wiki/Generalized_hypergeometric_function#Generalization_of_Kummer's_transformations_and_identities_for_2F2. It seems, however, that the first identity leads to incorrect results here since one would end up with just `alpha / (alpha  + beta)` - I guess there is some assumption on a, b, c, and d that is not satisfied here.

(Not our fault but also a bit annoying that one has to call `pFq` with vectors instead of tuples)

ArunSanganal: [Wolfram](https://reference.wolfram.com/language/ref/NoncentralBetaDistribution.html) is the reference I used. Can you think of any edge case where the exception might be triggered?

test/univariate/continuous/noncentralbeta.jl
```diff
@@ -0,0 +1,15 @@
+using Distributions
+using Test
+
+
+@testset "NoncentralBeta" begin 
+    α = 11.0 ; β = 8.0; λ = 5.0
```
devmotion, line 6: Could we use non-integers?

PR #1703, opened 2023-03-30
Adding stable distributions

jaksle:
Please be patient, I have zero experience with open source projects.

As for the code:
- Type-1 parametrization is used, the same as explained in Wikipedia and used in SciPy, however Matlab uses type-0 parametrization.  Type-1 one is usually the most convenient one for human use.
- Pdf and cdf are calculated using integral representations. The integrated functions converge to 0 like exp(-inf) and inf*exp(-inf) which returns NaNs, they are truncated by hand in line 83 and 127. Seems to be good enough.
- The code uses Interpolations.jl for 2D linear interpolation. If adding such dependence is a problem I may replace it.
- Fitting is two step. First quantile method is used which returns initial estimate used in the better ecdf method. According to John P. Nolan MLE is best, but ecdf method is only slightly worse and is much, much quicker. This method fits ecdf to the theoretical one, but internally in type-0 parametrization. This is because type-1 is discontinuous at alpha = 1 which causes numerical instability, essentially location parameter becomes meaningless. Only at the end the result it is converted to type-1. 
- If I understand correctly `@quantile_newton Stable` should provide default implementation of cdf inversion, but it does not seem to work when I'm checking on my local version...


jaksle:
Bump. Is there an interest in this? Stable distributions seem an important class for me (depends on the background I guess) and they are provided e.g., in SciPy and Matlab.

simsurace:
That's nice! I'd definitely find this useful.
I think it would need some unit tests though.

jaksle:
@simsurace I've added tests. I hope there will be someone with will/time to supervise merging it. At least, the "competition" in other languages have this feature.

simsurace:
To me everything looks very clean, thanks a lot for the work. However, I don't have merge rights.

simsurace:
@devmotion are you able to review this, or do you know someone who is able to?

jaksle:
@devmotion Could I ask for an update on this? Currently I don't know even if this commit fits and may potentially be merged. If there'd be no reasonable horizon on making this public I'd probably change it to a separate package. Sub-optimal I my view, but better than nothing.

jaksle:
Update: this functionality is currently available using [StableDistributions.jl](https://github.com/jaksle/StableDistributions.jl) package.

devmotion:
I just came across this PR in my backlog, sorry 🙂 I'm not strictly against adding new dependencies to Distributions but my general feeling is that the size of Distributions slows down development (and generally I tend to prefer smaller packages as a user since it allows to reduce compilation and load times), so I think it's fine to define distributions outside of Distributions.jl. For users, I think it could be helpful to list such more "official" downstream packages in the docs. The advantage of Distributions, of course, is that potentially - but not always as you can see... - more people are able to check your code and to contribute. Possibly this could be improved by moving some of such downstream packages to JuliaStats - but just moving it there without actually increasing the number of collaborators isn't very helpful either.

Issue #1378, opened 2021-08-06
qqbuild automatically throws out min and max values


HiramTheHero:
I originally posted this issue on StatsPlots.jl, and they referred me to post the issue here.

Here is the link explaining the issue: [https://github.com/JuliaPlots/StatsPlots.jl/issues/461#issue-956142608](https://github.com/JuliaPlots/StatsPlots.jl/issues/461#issue-956142608)

Here is the comment that explains why the issue should be posted here: [https://github.com/JuliaPlots/StatsPlots.jl/issues/461#issuecomment-891113224](https://github.com/JuliaPlots/StatsPlots.jl/issues/461#issuecomment-891113224)

JoshuaWhittemore:
Thank you @HiramTheHero for filing this.  I noticed it too.  I think the issue name should be changed to `qqbuild automatically throws out min and max values`, or something like that.  @sethaxen 's nice example shows the issue:
```julia
julia> using Distributions

julia> qqpair = Distributions.qqbuild(Normal(), [1,2,3])
QQPair{Array{Float64,1},Array{Float64,1}}([0.0], [2.0])

julia>
```

scheidan:
The problem is how `qqbuild` selects the grid points:
https://github.com/JuliaStats/Distributions.jl/blob/35125579a77b938f8d4a8bd2be23cbf1f2ddf225/src/qq.jl#L16
which is shorter than `length(x)`.

This solution would lead to the same behavior as R (ensuring that `length(n) == length(qx) == length(qd)`):
```Julia
"""
Generates the sequence of probability points

Reference for the selection of  `a`:
Becker, R. A., Chambers, J. M. and Wilks, A. R. (1988) _The New S
Language_.  Wadsworth & Brooks/Cole. 
"""
function ppoints(x)
    n = length(x)
    a =  n <= 10 ? 3/8 : 0.5
    start = (1-a)/(n + 1 - 2*a)
    stop = (n-a)/(n + 1 - 2*a)
    range(start, stop, length=n)
end
```

```Julia
function qqbuild(x::AbstractVector, d::UnivariateDistribution)
    grid = ppoints(x)
    qx = quantile(x, grid)
    qd = quantile.(Ref(d), grid)
    return QQPair(qx, qd)
end
```

This is a breaking change but I think the current behavior is _extremely dangerous_ for the intended use to build qqplots as it hides potential outliers.

If we agree that this is an improvement, I'd be happy to put a little PR together

PR #1708, opened 2023-04-18
Make sure `qqbuild` does not drop min and max value

scheidan:
This fixes https://github.com/JuliaStats/Distributions.jl/issues/1378

The current behavior is dangerous because when `qqbuild` is used for QQ-plots (as 
`Plots.jl` and `Makie.jl` do) the most extreme outliers are not shown.

### DIFF
--- a/src/qq.jl
+++ b/src/qq.jl
@@ -18,7 +18,7 @@ end
 Generate a sequence of probability points of length `n`:
 
 ```math
-(k − 0.5)/n, \\; k ∈ 1, ..., n
+(k − 0.5)/n, \\qquad k \\in \\{1, \\ldots, n\\}
 ```
 
 ## References
--- a/src/qq.jl
+++ b/src/qq.jl
@@ -11,9 +11,26 @@ function qqbuild(x::AbstractVector, y::AbstractVector)
     return QQPair(qx, qy)
 end
 
+
+"""
+Generates a sequence of probability points of length `n`:
+
+``
+(k − a)/(n + 1 − 2a), k ∈ 1, ..., n
+``
+
+`a` should be ∈ [0,1]. See the references listed here:
+https://en.wikipedia.org/wiki/Q%E2%80%93Q_plot#Heuristics
+"""
+function ppoints(n, a=0.5)
+    start = (1-a)/(n + 1 - 2*a)
+    stop = (n-a)/(n + 1 - 2*a)
+    range(start, stop, length=n)
+end
+
 function qqbuild(x::AbstractVector, d::UnivariateDistribution)
     n = length(x)
-    grid = [(1 / (n - 1)):(1 / (n - 1)):(1.0 - (1 / (n - 1)));]
+    grid = ppoints(n)
     qx = quantile(x, grid)
     qd = quantile.(Ref(d), grid)
     return QQPair(qx, qd)
@@ -21,7 +38,7 @@ end
 
 function qqbuild(d::UnivariateDistribution, x::AbstractVector)
     n = length(x)
-    grid = [(1 / (n - 1)):(1 / (n - 1)):(1.0 - (1 / (n - 1)));]
+    grid = ppoints(x)
     qd = quantile.(Ref(d), grid)
     qx = quantile(x, grid)
     return QQPair(qd, qx)
--- a/src/qq.jl
+++ b/src/qq.jl
@@ -38,7 +38,7 @@ end
 
 function qqbuild(d::UnivariateDistribution, x::AbstractVector)
     n = length(x)
-    grid = ppoints(x)
+    grid = ppoints(n)
     qd = quantile.(Ref(d), grid)
     qx = quantile(x, grid)
     return QQPair(qd, qx)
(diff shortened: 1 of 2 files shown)


scheidan:
The failing test seems unrelated.

src/qq.jl
```diff
@@ -11,17 +11,34 @@ function qqbuild(x::AbstractVector, y::AbstractVector)
...
+
+"""
+Generates a sequence of probability points of length `n`:
+
+``
+(k − a)/(n + 1 − 2a), k ∈ 1, ..., n
+``
+
+`a` should be ∈ [0,1]. See the references listed here:
+https://en.wikipedia.org/wiki/Q%E2%80%93Q_plot#Heuristics
```
devmotion, line 23: Some style changes to make it consistent:
````suggestion
    ppoints(n::Int, a::Real=0.5)

Generate a sequence of probability points of length `n`:

```math
(k − a)/(n + 1 − 2a), k ∈ 1, ..., n
```

`a` should be a number in ``[0, 1]``.

## References

https://en.wikipedia.org/wiki/Q%E2%80%93Q_plot#Heuristics
````

File: src/qq.jl [hunk 2]
```diff
@@ -11,17 +11,34 @@ function qqbuild(x::AbstractVector, y::AbstractVector)
...
+"""
+function ppoints(n, a=0.5)
+    start = (1-a)/(n + 1 - 2*a)
+    stop = (n-a)/(n + 1 - 2*a)
+    range(start, stop, length=n)
+end
```
devmotion, line 29: Maybe just
```suggestion
ppoints(n::Int, a::Real=0.5) = ((1:n) .- a) ./ (n + 1 - 2*a)
```
?

src/qq.jl
```diff
@@ -11,17 +11,34 @@ function qqbuild(x::AbstractVector, y::AbstractVector)
...
+
 function qqbuild(x::AbstractVector, d::UnivariateDistribution)
     n = length(x)
-    grid = [(1 / (n - 1)):(1 / (n - 1)):(1.0 - (1 / (n - 1)));]
+    grid = ppoints(n)
```
devmotion, line 33: Is parameter `a` in `ppoints` really needed if the function is only called with one argument?

scheidan: Yes, I was not sure about this one.
Different software use different values for `a`:

- R: `a = n <= 10 ? 3/8 : 0.5`
- Matlab: `a = 0.5`
- Scipy: `a = 0.3175` (but the first and the last point are computed differently)
- Minitab: `a = 3/8`

Maybe somebody would like to use `ppoints` independently of `qqbuild` with more flexibility?  Or should `a` be even an argument of `qqbuild`?

The value hardly matter practically,  I do not have a preference here.

devmotion: I think I'd prefer to get rid of the function and just hardcode the range here (with a comment and references). We can generalize once it's needed and deemed useful but I think the generality is not needed right now.

devmotion: Ah, I missed that it is used in the function below as well. So I suggest just removing parameter `a` then.

src/qq.jl
```diff
@@ -11,17 +11,35 @@ function qqbuild(x::AbstractVector, y::AbstractVector)
     return QQPair(qx, qy)
 end
 
+
+"""
+    ppoints(n::Int, a::Real=0.5)
```
devmotion, line 16:
```suggestion
    ppoints(n::Int)
```

src/qq.jl
```diff
@@ -11,17 +11,35 @@ function qqbuild(x::AbstractVector, y::AbstractVector)
...
+
+Generate a sequence of probability points of length `n`:
+
+```math
+(k − a)/(n + 1 − 2a), k ∈ 1, ..., n
```
devmotion, line 21:
```suggestion
(2k − 1)/(2n), \\qquad k \\in \\{1, \\ldots, n\\}
```

src/qq.jl
```diff
@@ -11,17 +11,35 @@ function qqbuild(x::AbstractVector, y::AbstractVector)
...
+```
+
+`a` should be a number in ``[0, 1]``.
```
devmotion, line 24:
```suggestion
```

src/qq.jl
```diff
@@ -11,17 +11,35 @@ function qqbuild(x::AbstractVector, y::AbstractVector)
...
+
+## References
+
+https://en.wikipedia.org/wiki/Q%E2%80%93Q_plot#Heuristics
+"""
+ppoints(n::Int, a::Real=0.5) = ((1:n) .- a) ./ (n + 1 - 2*a)
```
devmotion, line 30:
```suggestion
function ppoints(n::Int)
    m = 2 * n
    return (1:2:(m - 1)) ./ m
end
```

src/qq.jl
scheidan: I agree, this leads to the smallest change. Will update it accordingly.

scheidan:
Sorry @devmotion, I didn't see your comment in time and just pushed a version without `ppoints`...

 I will make a new clean PR with:

- keep `ppoints`
- but remove parameter `a`
- add reference to Matlab doc (https://ch.mathworks.com/help/stats/qqplot.html)

Anything I've missed?

devmotion (approved):

devmotion:
Thank you @scheidan!

PR #1713, opened 2023-04-30
Distribution: Generalized Inverse Gaussian

kunzaatko:


msainsburydale:
Hi, just wondering the status of this pull request is? It would very useful to have the GIG distribution implemented in Julia.

kunzaatko:
Hi! Sorry to say that my I haven't worked on this pull for quite a while... Currently I am in quite a time pressure, so I presume that I can complete the work no-earlier than at the beginning of February. If you can wait till February, I can put a reminder in my calendar and complete it then.
If you cannot wait, I can offer you any support that is necessary so that you can complete the pull request yourself if that would be a preferable option for you :)
In summary: 
- implemented is `pdf`, `logpdf`, `mgf`, `cf`, `quantile` evaluations and `mode`, `mean`, `var` statistics.
- necessary to implement is
  - [ ] sampling (2 articles above, both use different methods... It is necessary to evaluate, which one fits better and to implement it)
  - [ ] `cdf` (presumably using the marco for newton method estimate, or the approximation given in the paper above)
  - [ ] MLE (given in the book above as formulas. Need to be converted to the parametrization in use in the implementation)
  - [ ] testing against the `R` library (there is a reference testing for other distributions that can be basically copied)
  - [ ] `entropy` (there is a need for evaluation of derivatives of `besselk` and there is an formula for that in DLMF... link above. They should be a part of SpecialFunctions.jl and used instead of being defined directly. This would mean, that the compat entry for SpecialFunctions.jl would be bumped and it should be discussed with the lead developers of Distributions.jl if this is desirable.)
  - [ ] documentation (again, this could basically be copied and transformed for GIG)
 
With this, all the basic functionality should be available (it would be best to check the implementation documentation of Distributions.jl if I missed something). Nothing should be complicated, but it will require some time... Let me help what you think about this and if you will attempt to continue in the pull yourself :wink:

msainsburydale:
Hi, thanks very much, happy to wait until you have time. My comment was mainly to emphasise that the GIG distribution is indeed important (particularly in financial modelling) so it is worthwhile to implement it in Distributions.jl.

PR #1717, opened 2023-05-06
Improve type stability for truncated normal moments

ararslan:
The internal functions used for computing the mean and variance of the truncated normal distribution were implemented using literals like `√2`, which are always `Float64`. However, some branches of the functions based the return type off of the types of the arguments, which needn't be `Float64`. This led to some instability that can be seen via `@code_warntype`. For example, the inferred result of `_tnmom1` for `Float32` inputs was `Union{Float32,Float64}`.

We can instead use the irrational constants defined for square roots of pi and 2 and ratios thereof in order to avoid promoting to `Float64` unnecessarily. This makes the return type concretely inferrable and provides a small performance improvement, about 12% better (-7ns) for `mean` and 14% better (-20ns) for `var` on my machine as compared to current master.

### DIFF
--- a/src/truncated/normal.jl
+++ b/src/truncated/normal.jl
@@ -27,56 +27,59 @@ modes(d::Truncated{Normal{T},Continuous}) where {T <: Real} = [mode(d)]
 # do not export. Used in mean
 # computes mean of standard normal distribution truncated to [a, b]
 function _tnmom1(a, b)
+    mid = middle(a, b)
     if !(a ≤ b)
-        return oftype(middle(a, b), NaN)
+        return oftype(mid, NaN)
     elseif a == b
-        return middle(a, b)
+        return mid
     elseif abs(a) > abs(b)
         return -_tnmom1(-b, -a)
     elseif isinf(a) && isinf(b)
-        return zero(middle(a, b))
+        return zero(mid)
     end
-    Δ = (b - a) * middle(a, b)
+    Δ = (b - a) * mid
+    a′ = a * invsqrt2
+    b′ = b * invsqrt2
     if a ≤ 0 ≤ b
-        m = √(2/π) * expm1(-Δ) * exp(-a^2 / 2) / (erf(a/√2) - erf(b/√2))
+        m = expm1(-Δ) * exp(-a^2 / 2) / (erf(a′) - erf(b′))
     elseif 0 < a < b
-        z = exp(-Δ) * erfcx(b/√2) - erfcx(a/√2)
-        iszero(z) && return middle(a, b)
-        m = √(2/π) * expm1(-Δ) / z
+        z = exp(-Δ) * erfcx(b′) - erfcx(a′)
+        iszero(z) && return mid
+        m = expm1(-Δ) / z
     end
-    return clamp(m, a, b)
+    return clamp(m / sqrthalfπ, a, b)
 end
 
 # do not export. Used in var
 # computes 2nd moment of standard normal distribution truncated to [a, b]
 function _tnmom2(a::Real, b::Real)
+    mid = middle(a, b)
     if !(a ≤ b)
-        return oftype(middle(a, b), NaN)
+        return oftype(mid, NaN)
     elseif a == b
-        return middle(a, b)^2
+        return mid^2
     elseif abs(a) > abs(b)
         return _tnmom2(-b, -a)
     elseif isinf(a) && isinf(b)
-        return one(middle(a, b))
+        return one(mid)
     elseif isinf(b)
-        return 1 + √(2 / π) * a / erfcx(a / √2)
+        return 1 + a / erfcx(a * invsqrt2) / sqrthalfπ
     end
-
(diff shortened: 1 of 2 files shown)


src/truncated/normal.jl
```diff
@@ -27,56 +27,59 @@ modes(d::Truncated{Normal{T},Continuous}) where {T <: Real} = [mode(d)]
 # do not export. Used in mean
 # computes mean of standard normal distribution truncated to [a, b]
 function _tnmom1(a, b)
+    mid = middle(a, b)
```
devmotion, line 30: I guess we should use
```suggestion
    mid = float(middle(a, b))
```
since `middle` does not necessarily return a floating point number:
```julia
julia> StatsBase.middle(0//1, 1//1)
1//2
```

src/truncated/normal.jl
```diff
@@ -27,56 +27,59 @@ modes(d::Truncated{Normal{T},Continuous}) where {T <: Real} = [mode(d)]
...
-        return zero(middle(a, b))
+        return zero(mid)
     end
-    Δ = (b - a) * middle(a, b)
+    Δ = (b - a) * mid
+    a′ = a * invsqrt2
+    b′ = b * invsqrt2
     if a ≤ 0 ≤ b
-        m = √(2/π) * expm1(-Δ) * exp(-a^2 / 2) / (erf(a/√2) - erf(b/√2))
+        m = expm1(-Δ) * exp(-a^2 / 2) / (erf(a′) - erf(b′))
```
devmotion, line 44:
```suggestion
        m = expm1(-Δ) * exp(-a^2 / 2) / erf(b′, a′)
```

src/truncated/normal.jl
```diff
@@ -27,56 +27,59 @@ modes(d::Truncated{Normal{T},Continuous}) where {T <: Real} = [mode(d)]
...
+        m = expm1(-Δ) / z
     end
-    return clamp(m, a, b)
+    return clamp(m / sqrthalfπ, a, b)
 end
 
 # do not export. Used in var
 # computes 2nd moment of standard normal distribution truncated to [a, b]
 function _tnmom2(a::Real, b::Real)
+    mid = middle(a, b)
```
devmotion, line 56:
```suggestion
    mid = float(middle(a, b))
```

src/truncated/normal.jl
```diff
@@ -27,56 +27,59 @@ modes(d::Truncated{Normal{T},Continuous}) where {T <: Real} = [mode(d)]
...
-
+    a′ = a * invsqrt2
+    b′ = b * invsqrt2
     if a ≤ 0 ≤ b
-        ea = √(π/2) * erf(a / √2)
-        eb = √(π/2) * erf(b / √2)
+        ea = sqrthalfπ * erf(a′)
+        eb = sqrthalfπ * erf(b′)
         fa = ea - a * exp(-a^2 / 2)
         fb = eb - b * exp(-b^2 / 2)
```
devmotion, line 74:
```suggestion
        eb_ea = sqrthalfπ * erf(a′, b′)
        fb_fa = eb_ea + a * exp(-a^2 / 2) - b * exp(-b^2 / 2)
```

src/truncated/normal.jl
```diff
@@ -27,56 +27,59 @@ modes(d::Truncated{Normal{T},Continuous}) where {T <: Real} = [mode(d)]
...
-        m2 = (fb - fa) / (eb - ea)
-        return m2
+        return (fb - fa) / (eb - ea)
```
devmotion, line 75:
```suggestion
        return fb_fa / eb_ea
```

src/truncated/normal.jl
ararslan: Ah, good call, thanks!

src/truncated/normal.jl
```diff
@@ -27,56 +27,57 @@ modes(d::Truncated{Normal{T},Continuous}) where {T <: Real} = [mode(d)]
...
+        return mid
     elseif abs(a) > abs(b)
         return -_tnmom1(-b, -a)
     elseif isinf(a) && isinf(b)
-        return zero(middle(a, b))
+        return zero(mid)
     end
-    Δ = (b - a) * middle(a, b)
+    Δ = (b - a) * mid
+    a′ = a * invsqrt2
```
ararslan, line 41: @devmotion, I'm curious if you have any thoughts on this approach vs. `a / sqrt2`. They seem to be extremely close but not exactly identical (but close enough to not affect the tests). Similar situation for the division of things by `sqrthalfπ` vs. defining a separate irrational for sqrt(2/π) and multiplying.

devmotion: Hm no, to me both approaches seem fine. I guess if an irrational of the inverse exists, multiplication might be faster and therefore preferable.

ararslan: That was my thought as well, just wasn't sure whether there was something I might be missing in terms of potential loss of precision or things like that

PR #1719, opened 2023-05-08
Implement `StatsAPI.pvalue`

devmotion:
Currently `pvalue(::UnivariateDistribution, ...)` is defined in HypothesisTests and owned by HypothesisTests. However, when switching HypothesisTests to StatsAPI 1.6 which defines `StatsAPI.pvalue` such definitions would be type piracy. Hence this PR adds them to Distributions.

Note that StatsAPI is already an indirect dependency of Distributions via StatsBase, so this PR does not add packages to the dependency tree.

TODO:
- [x] Add tests


test/statsapi.jl
```diff
@@ -0,0 +1,26 @@
+using Distributions
+using StatsAPI: pvalue
```
ararslan, line 2: Would it make sense to have Distributions reexport `pvalue`? I'd be generally in favor but don't feel too strongly.

ararslan (approved):

test/statsapi.jl
devmotion: Wouldn't this break HypothesisTests?

ararslan: Not after HypothesisTests extends `pvalue` from StatsAPI.

devmotion: Sure, but all existing releases of HypothesisTests would suddenly be broken, wouldn't they? Certainly, this could be fixed by modifying the registry but I think this makes it less attractive to re-export `pvalue`.

devmotion: I'll merge and release without re-exporting `pvalue` for now but this can be reconsidered and changed in a subsequent release if desired.

PR #1721, opened 2023-05-11
Fix `rand` for truncated normal with 0 variance

### LINKED ISSUE #1712: Infinite loop in truncated normal distribution
Using Distributions v0.25.87, the following results in an infinite loop:

```julia
julia> d = Normal(204.764679086758, 0.0)
Normal{Float64}(μ=204.764679086758, σ=0.0)

julia> rand(d)
204.764679086758

julia> d2 = truncated(d, 204.764679086758, 204.764679086758)
Truncated(Normal{Float64}(μ=204.764679086758, σ=0.0); lower=204.764679086758, upper=204.764679086758)

julia> rand(d2)
# hangs
```

Perhaps setting standard deviation to `0.0` is an odd thing to do, but I was expecting it to just return `204.764679086758` on all calls to `rand`.

ararslan:
Fixes #1712.

### DIFF
--- a/src/truncated/normal.jl
+++ b/src/truncated/normal.jl
@@ -143,9 +143,18 @@ function rand(rng::AbstractRNG, d::Truncated{Normal{T},Continuous}) where T <: R
     μ = mean(d0)
     σ = std(d0)
     if isfinite(μ)
-        a = (d.lower - μ) / σ
-        b = (d.upper - μ) / σ
-        z = randnt(rng, a, b, d.tp)
+        a, b = extrema(d)
+        if iszero(σ)
+            if a <= μ <= b
+                z = 0.0
+            else
+                return NaN
+            end
+        else
+            a′ = (a - μ) / σ
+            b′ = (b - μ) / σ
+            z = randnt(rng, a′, b′, d.tp)
+        end
         return μ + σ * z
     else
         return clamp(μ, d.lower, d.upper)
--- a/test/truncated/normal.jl
+++ b/test/truncated/normal.jl
@@ -45,7 +45,7 @@ end
         [(r = rand, r! = rand!),
          (r = ((d, n) -> rand(rng, d, n)), r! = ((d, X) -> rand!(rng, d, X)))]
         repeats = 1000000
-        
+
         @test abs(mean(func.r(trunc, repeats))) < 0.01
         @test abs(median(func.r(trunc, repeats))) < 0.01
         @test abs(var(func.r(trunc, repeats)) - var(trunc)) < 0.01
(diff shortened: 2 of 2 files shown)


src/truncated/normal.jl
```diff
@@ -143,9 +143,18 @@ function rand(rng::AbstractRNG, d::Truncated{Normal{T},Continuous}) where T <: R
...
     if isfinite(μ)
-        a = (d.lower - μ) / σ
-        b = (d.upper - μ) / σ
-        z = randnt(rng, a, b, d.tp)
+        a, b = extrema(d)
+        if iszero(σ)
+            if a <= μ <= b
+                z = 0.0
+            else
+                return NaN
```
devmotion, line 151: I think we would want to catch this case earlier in `truncated`: A distribution with zero mass is by definition not a distribution. See https://github.com/JuliaStats/Distributions.jl/issues/843.

ararslan:
Seems I was beaten to the punch by about 2 years: #1273 🙂

src/truncated/normal.jl
devmotion: Maybe let's throw an error here? I don't think returning `NaN` is particularly useful.

ararslan: I've opened https://github.com/JuliaStats/Distributions.jl/pull/1723 as a means of catching this case earlier, but I'd also be fine throwing an error here.

HakanBrian:
Fixes #1867

HakanBrian:
This should also resolve #1867

src/truncated/normal.jl
```diff
@@ -128,14 +128,22 @@ function rand(rng::AbstractRNG, d::Truncated{<:Normal{<:Real},Continuous})
...
     σ = std(d0)
+    a, b = extrema(d)
     if isfinite(μ)
-        lower, upper = extrema(d)
-        a = (lower - μ) / σ
-        b = (upper - μ) / σ
-        z = randnt(rng, a, b, d.tp)
+        if iszero(σ)
+            if a <= μ <= b
+                z = 0.0
```
devmotion, line 135: I assume this will cause type instabilities, e.g., if parameters of `Normal` and `Truncated` are `Float32`?

ararslan: `z` is always `Float64`, both here and the branch that calls `randnt` (that function requires all inputs be `Float64` and it returns a `Float64`). But if `µ` and the truncation points are wider, say `BigFloat`, then you'll get a `MethodError` from `randnt` when `isfinite(µ) && σ > 0` and a `BigFloat` result otherwise. `Float32` works because `extrema` promotes the truncation points to `Float64`, so every branch will return `Float64`.

PR #1723, opened 2023-05-18
Throw an error when truncating outside of support

### LINKED ISSUE #843: Truncated should catch truncations outside the support

ararslan:
Fixes #843 using option 2 as described [here](https://github.com/JuliaStats/Distributions.jl/issues/843#issuecomment-778815561). The implementation checks for 0 probability within the truncated region but continues to allow coincident truncation bounds if they're within the support. That avoids breaking cases like that shown in #1712.


src/truncate.jl
```diff
@@ -95,6 +95,9 @@ struct Truncated{D<:UnivariateDistribution, S<:ValueSupport, T<: Real, TL<:Union
     logtp::T      # log(tp), i.e. log(ucdf - lcdf)
 
     function Truncated(d::UnivariateDistribution, l::TL, u::TU, loglcdf::T, lcdf::T, ucdf::T, tp::T, logtp::T) where {T <: Real, TL <: Union{T,Nothing}, TU <: Union{T,Nothing}}
+        if iszero(tp) && !(l == u && insupport(d, l))
```
devmotion, line 98: Can we check `logtp` instead? Seems safer if the untruncated has tiny non-zero mass on the truncated interval. Also, I think the second condition is not needed since `tp = P(lower <= X <= upper)` with supposedly correct handling of discrete and degenerate distributions (but maybe it's not computed correctly in some cases currently?).
```suggestion
        if logtp == -Inf
```

devmotion (commented):
I think for AD backends such as Zygote we should allow to disable (and skip automatically) the check by uaing the `@check_args` macro - in other distributions standard checks (without `@check_args`) impacted performance severely.

ararslan:
Looks like this approach doesn't quite do it, as evidenced by the large number of test failures 🙃

devmotion:
Some of them at least are due to tests that have to be fixed/removed: https://github.com/JuliaStats/Distributions.jl/actions/runs/5017513840/jobs/8995725428?pr=1723#step:6:67

Some others seem to indicate that `logtp = -Inf` currently for normal distributions truncated far away from the mean. Even though some of their traits (e.g., mean and variance) are tested they are basically useless:
```julia
julia> d = truncated(Normal(0, 1), 100, 115)
Truncated(Normal{Float64}(μ=0.0, σ=1.0); lower=100.0, upper=115.0)

julia> mean(d)
100.00999800099926

julia> var(d)
9.994005086296622e-5

julia> pdf(d, 101)
NaN

julia> logpdf(d, 101)
Inf

julia> d.logtp
-Inf

julia> d.tp
0.0
```
I think we might be able to fix these existing numerical issues for normal distributions (at least) by improving the numerical accuracy in the calculation of `logtp`. IMO this should be possible since
```julia
julia> SpecialFunctions.logerf(100/sqrt(2), 115/sqrt(2)) - log(2)
-5005.5242086942035
```
It seems one part could consist of improving `logdiffcdf` for normal distributions as currently
```julia
julia> logdiffcdf(Normal(0, 1), 115, 100)
-Inf
```
And then maybe `logdiffcdf` could be used in `truncated`?

devmotion:
I opened #1728 to address the `logdiffcdf` issue for normal distributions.

src/truncate.jl
```diff
@@ -72,7 +72,7 @@ function truncated(d::UnivariateDistribution, l::T, u::T) where {T <: Real}
     ucdf = exp(logucdf)
 
     # (log)tp = (log) P(l ≤ X ≤ u) where X ∼ d
-    logtp = logsubexp(loglcdf, logucdf)
+    logtp = logdiffcdf(d, u, l)
```
devmotion, line 75: This is only correct if the probability mass of `d` at `l` is zero. Hence, while for normal distributions it is correct, for discrete distributions with `pdf(d, l) > 0` it is wrong and should instead be `logaddexp(logdiffcdf(d, u, l), logpdf(d, l))`. Currently, this is handled by excluding these probability masses in `loglcdf` (via `_logcdf_noninclusive`: https://github.com/JuliaStats/Distributions.jl/blob/fd58ce19ba18e15cd0c766893c2928e91ad61ad6/src/censored.jl#L439-L442). A similar fix should be applied here as well:
```suggestion
    logtp = if variate_form(d) === Discrete
        logaddexp(logdiffcdf(d, u, l), logpdf(d, l))
    else
        logdiffcdf(d, u, l)
    end    
```

ararslan:
You're right, the `truncated(Normal(0, 1), 0, 0)` case is unintentionally broken by the change in https://github.com/JuliaStats/Distributions.jl/pull/1723/commits/94d28172699dd62ab4d160ce54ba741905e60617.

Issue #1715, opened 2023-05-04
Adding the Wrapped Normal distribution


sethaxen:
The Wrapped Normal distribution on the circle was mentioned in #224. There's a Distributions-compatible implementation that takes care to avoid excess work at https://gist.github.com/sethaxen/2ce75fa84755e7e0cfd54cc9555d120b.

Based on the discussion in the linked issue though, I'm not certain whether such distributions should be contributed here directly. Limiting the support of `VonMises` to an interval of 2pi centered around the mean seems controversial, and `VonMises` is last I checked the only circular distribution here.

If there's interest in including this here from the maintainers, I'll open a PR. To my knowledge, there are no specialized methods for fitting the parameters.

sethaxen:
Other common wrapped distributions include:
- Wrapped Poisson
- Wrapped Cauchy
- k-wrapped distributions (i.e. wrapping a circular distribution onto itself to represent k-fold symmetry).
One could potentially support these with a "wrapper" distribution `Wrapped`, that could take a distribution with support on the real line and an interval around which to wrap it. e.g.

```julia
using Distributions

struct Wrapped{D<:UnivariateDistribution, S<:ValueSupport, T <: Real} <: UnivariateDistribution{S}
    unwrapped::D
    lower::T      # lower bound
    upper::T      # upper bound
    k::Int
    function Wrapped(d::UnivariateDistribution, l::T, u::T, k::Int=1) where {T <: Real}
        new{typeof(d), Distributions.value_support(typeof(d)), T}(d, l, u, k)
    end
end

wrapped(d::UnivariateDistribution, l::T, u::T, k::Int=1) where {T<:Real} = Wrapped(d, l, u, k)
wrapped(d::UnivariateDistribution, l::Real, u::Real, k::Int=1) = Wrapped(d, Base.promote(l, u)..., k)

Base.minimum(d::Wrapped) = d.lower
Base.maximum(d::Wrapped) = d.upper

function Distributions.pdf(d::Wrapped, x::Real; tol=sqrt(eps(float(typeof(x)))), maxiter=1_000)
    u = d.upper
    l = d.lower
    period = u - l
    z = mod(x, period)
    p = pdf(d.unwrapped, z)
    for i in 1:maxiter
        r = i // d.k
        zl = muladd(-r, period, z)
        zu = muladd(r, period, z)
        pl = insupport(d.unwrapped, zl) ? pdf(d.unwrapped, zl) : zero(p)
[...]

PR #1724, opened 2023-05-18
Add Wrapped distribution wrapper

### LINKED ISSUE #1716: Generically supporting wrapped distributions
#1715 proposes adding the Wrapped Normal distribution. Other common named wrapped distributions include:
- Wrapped Poisson
- Wrapped Cauchy
- k-wrapped distributions (i.e. wrapping a circular distribution onto itself to represent k-fold symmetry).

One could support these generically, as well as arbitrary wrappings, with a "wrapper" distribution `Wrapped`, that could take a distribution with support on the real line and an interval around which to wrap it. e.g.

```julia
using Distributions

struct Wrapped{D<:UnivariateDistribution, S<:ValueSupport, T <: Real} <: UnivariateDistribution{S}
    unwrapped::D
    lower::T      # lower bound
    upper::T      # upper bound
    k::Int
    function Wrapped(d::UnivariateDistribution, l::T, u::T, k::Int=1) where {T <: Real}
        new{typeof(d), Distributions.value_support(typeof(d)), T}(d, l, u, k)
    end
end

wrapped(d::UnivariateDistribution, l::T, u::T, k::Int=1) where {T<:Real} = Wrapped(d, l, u, k)
wrapped(d::UnivariateDistribution, l::Real, u::Real, k::Int=1) = Wrapped(d, Base.promote(l, u)..., k)

Base.minimum(d::Wrapped) = d.lower
Base.maximum(d::Wrapped) = d.upper

function Distributions.pdf(d::Wrapped, x::Real; tol=sqrt(eps(float(typeof(x)))), maxiter=1_000)
    u = d.upper
    l = d.lower
    period = u - l
    z = mod(x, period)
    p = pdf(d.unwrapped, z)
    for i in 1:maxiter
        r = i // d.k
        zl = muladd(-r, period, z)
        zu = muladd(r, period, z)
        pl = insupport(d.unwrapped, zl) ? pdf(d.unwrapped, zl) : zero(p)
        pu = insupport(d.unwrapped, zu) ? pdf(d.unwrapped, zu) : zero(p)
        Δp = pl + pu
        p += Δp
        abs(Δp) < p * tol && break
    end
    return p
end
```

Advantages
- resembling what we currently do with `truncated` and `censored`
- supporting wrapping to arbitrary periods, including integer periods
- still supporting dispatch for cases like `WrappedNormal` where we have more efficient ways to evaluate the series

Disadvantages:
- In some cases, like `WrappedNormal` in #1715, depending on the parameters of `unwrapped`, we might at construction time be able to compute some heuristic to decide which method to use e.g. for density evaluation, which the above approach would require us do during every density evaluation. One solution would be to add some field of arbitrary type to `Wrapped` that could in an overloaded constructor be filled with any relevant precomputations that would be used in `pdf`, etc.  

An example usage
```julia
using StatsPlots

fig = plot(
    plot(wrapped(Normal(π/3, 1), -π, π); title="Wrapped Normal", proj=:polar),
    plot(wrapped(Cauchy(π/3, 1), 0, 2π); title="Wrapped Cauchy", proj=:polar),
    plot(wrapped(Uniform(-1, 1), -π, π, 8); title="8-times Wrapped Uniform", proj=:polar),
    plot(wrapped(Poisson(10), 0, 15); title="Wrapped Poisson");
    label="", dpi=180, titlefontsize=10
)
```
tmp

_adapted from original post by @sethaxen in https://github.com/JuliaStats/Distributions.jl/issues/1715#issuecomment-1536129396_

sethaxen:
Fixes #1716 and #1715


sethaxen:
This version of `Wrapped` allows for any range and periodicity to be specified as fields, but it comes with the trade-off that for distributions like Wrapped Cauchy, where we have algorithms to fit it well, we're not able to because the usual `fit(::Type{<:Distribution}, x)` interface doesn't allow specifying the upper and lower bounds or `k`. It would be nice if there was an alternate interface for wrapper distributions like these.

sethaxen:
Also #1665 adds a `WrappedCauchy` distribution. This implementation is a little more general because it also allows for `k`-wrapping and allows other intervals than those of length `2pi` to be used.

sethaxen:
Before I spend more time polishing this and adding a test suite, a few questions:
- Is the generalization from intervals of length 2π to arbitrary intervals welcome?
- Is the generalization to build distributions with `k-fold` symmetry welcome?
- For the Wrapped Normal, Wrapped Cauchy, and Wrapped Exponential, is this approach of overloading the functions for `Wrapped` preferred, or is it preferred to defined special named distributions as in #1665?
- How to define `fit` methods? (see https://github.com/JuliaStats/Distributions.jl/pull/1724#issuecomment-1553668551)

@devmotion @ararslan

ararslan:
I should preface this by saying that I know almost nothing about wrapped distributions, so any opinions expressed here are not strongly held and may be uninformed.

> Is the generalization from intervals of length 2π to arbitrary intervals welcome?

I don't see why not. ¯\\\_(ツ)\_/¯ I assume that's something that's not uncommon in practice and/or would be difficult to achieve with a simple data transformation or similar?

> Is the generalization to build distributions with `k-fold` symmetry welcome?

I don't see why not. ¯\\\_(ツ)\_/¯ Same general question though.

> For the Wrapped Normal, Wrapped Cauchy, and Wrapped Exponential, is this approach of overloading the functions for `Wrapped` preferred, or is it preferred to defined special named distributions as in #1665?

Defining methods for the `Wrapped` wrapper has prior art with e.g. `Truncated`. In fact, there used to be a `TruncatedNormal` that was deprecated in favor of `Truncated{Normal}`. So I think your approach here is more consistent and extensible.

> How to define `fit` methods?

Not all distributions have `fit` (or rather, `fit_mle`) methods, notably including `Truncated`-wrapped distributions, so I personally think it would be okay to punt on a decision for now and add it at a later time after some more extended design discussion.

> This version of `Wrapped` allows for any range and periodicity to be specified as fields, but it comes with the trade-off that for distributions like Wrapped Cauchy, where we have algorithms to fit it well, we're not able to because the usual `fit(::Type{<:Distribution}, x)` interface doesn't allow specifying the upper and lower bounds or `k`.

Could `k` be made a type parameter, similar to the dimensionality parameter for `Array`? Then the `fit` method could be something like `fit(Wrapped{Cauchy,1}, x)`. That doesn't solve the bounds though.

src/wrapped.jl
```diff
@@ -0,0 +1,317 @@
...
+    lower::T              # lower bound
+    upper::T              # upper bound
+    characteristic::Bool  # whether to use characteristic function
+    k::Int                # k-fold symmetry
+    function Wrapped(
+        d::UnivariateDistribution,
+        l::T,
+        u::T;
+        tol::Real=sqrt(eps(float(T))),
+        characteristic::Bool=_wrapped_use_characteristic(d, l, u, tol),
```
ararslan, line 57: This seems like an odd thing to store in a field and expose to the user. Naively, it seems like an implementation detail that an average user would not need to necessarily care about or modify. I would expect that whether the characteristic function is used would be up to methods defined for particular `D`s in `Wrapped{D}`.

src/wrapped.jl
```diff
@@ -0,0 +1,317 @@
...
+        return new{typeof(d),value_support(typeof(d)),T}(
+            d, l, u, characteristic, k
+        )
+    end
+end
+
+_wrapped_use_characteristic(d::UnivariateDistribution, l, u, tol) = false
+
+minimum(d::Wrapped) = d.lower
+maximum(d::Wrapped) = d.upper - eps(float(d.upper))
```
ararslan, line 69: For `Wrapped{D,S,Int}`, this will produce different types for `minimum` and `maximum`, which seems weird to me. In other cases we use the limit, e.g. when it's unbounded, which I think (again naively) would be acceptable here.

sethaxen: For a wrapped distribution, `d.lower` and `d.upper` are identified as the same point, so for a discrete distribution, including them both will always give the wrong answer if `d.upper` is in the support of the wrapped distribution. Ideally `maximum(d)` would return the largest point strictly less than `d.upper`, but AFAICT there is no function for this in the Distributions API.

See for example the wrapped Poisson example in https://github.com/JuliaStats/Distributions.jl/issues/1716#issue-1697550141. Because 0 and 15 are the same point, the PMF sums to greater than 1 because `maximum(d)` returned `d.upper`.

A compromise could be returning `d.upper` for continuous distributions and this result for discrete ones.

ararslan: > Ideally `maximum(d)` would return the largest point strictly less than `d.upper`

Distributions defines the support of a `DiscreteUnivariateDistribution` to be `round(Int, minimum(d)):round(Int, maximum(d))`, which means that it doesn't support supports (🙃) with non-integer steps. That implies that the largest point that's strictly less than `d.upper` is `ceil(Int, d.upper) - 1`. Would that work here?

src/wrapped.jl
sethaxen: I'm not sure I understand the question. There is no one number of terms. For a given wrapped distribution `dist` with values of `params`, one or both of the series will have fewer terms. But if you change any of the `params` values or change the `dist`, then the other series might now have fewer terms. The difference can be quite extreme. For wrapped normal, for example, if `σ=1e+3`, the series using `cf` needs only 1 term, while the series using pdf needs  >2,000 terms. But for `σ=1e-3`, the series using `cf` needs >6,000 terms, while the one using `pdf` needs 1 term. So we need to choose whether to use the `pdf` series or `cf` series. Options:
- always use one or the other. This is bad for the reasons noted above
- store a field like we do here. The user can select which strategy they want to use, or a developer can overloaded `wrapped(::SomeDistribution, ...)` if they have a heuristic for choosing the strategy
- Instead of storing a field, the strategy coul be picked at every evaluation of `pdf`, `cdf`, etc. Likely wasteful
- Instead of choosing, every function evaluates  both series in parallel until one of them converges. Potentially runs into issues when `cf` is not implemented, and does twice as much works as might be necessary.

src/wrapped.jl
sethaxen: > Distributions defines the support of a `DiscreteUnivariateDistribution` to be `round(Int, minimum(d)):round(Int, maximum(d))`, which means that it doesn't support supports (upside_down_face) with non-unit steps.

Oh really? But this doesn't seem to be consistently enforced, e.g.

```julia
julia> d = Binomial(10, 0.5) * 1//2;

julia> support(d)
0//1:1//2:5//1

julia> d = DiscreteNonParametric(randn(5), normalize(rand(5), 1));

julia> support(d)
5-element Vector{Float64}:
 -0.9108476704164078
 -0.5194957770100691
 -0.024410217492521903
  0.15370458561285605
  1.338024968131234
```

ararslan: Hah, amazing. 🤦 I was going based on the generic definition for `DiscreteUnivariateDistribution`, I had forgotten about `LocationScale` and I don't think I even knew `DiscreteNonParametric` existed. To be fair, I did say I was uninformed!

src/wrapped.jl
ararslan: > I'm not sure I understand the question. There is no one number of terms.

Sorry, I was referring to what you had said in the preceding message:

> this requires someone to sit down and work out bounds for the number of terms needed for each series representation.

sethaxen:
> > Is the generalization from intervals of length 2π to arbitrary intervals welcome?
> 
> I don't see why not. ¯\_(ツ)_/¯ I assume that's something that's not uncommon in practice and/or would be difficult to achieve with a simple data transformation or similar?

It's certainly useful. e.g. one might want to use `[0, 2π)` or `[-π, π)`, or one might want to use degrees or days of the year. To support discrete distributions like wrapped Poisson it becomes necessary.

> > Is the generalization to build distributions with `k-fold` symmetry welcome?
> 
> I don't see why not. ¯\_(ツ)_/¯ Same general question though.

I did some searching, and the only mention I can find of k-times wrapping is in Directional Statistics by Mardia and Jupp, where they give no references for papers that use it. I suspect the most likely version to use besides `k=1` is `k=2`, which turns any circular distribution into an axial one.

> Could `k` be made a type parameter, similar to the dimensionality parameter for `Array`? Then the `fit` method could be something like `fit(Wrapped{Cauchy,1}, x)`. That doesn't solve the bounds though.

Yes I think this is a reasonable solution. For bounds, I propose the fallback `wrapped(d::ContinuousDistribution) = wrapped(d, -π, π)`  (taking care to not promote types unnecessarily; too bad there's no `-π` irrational). Then if one has data with a different period than 2π, they can scale it before fitting. Not perfect, but works.

If we want `Wrapped{<:Cauchy}` and `Wrapped{<:Normal}` to be like `VonMises`, where the support is `[μ-π, μ+π)`, we can change the default e.g. `wrapped(d::Normal) = wrapped(d, d.μ-π, d.μ+π)`

PR #1725, opened 2023-05-21
Fix wrong parameter on chisq.jl documentation

ynhame:
there were a "k" left from the old documentation of the chisq.

this commit changes the "k" to a \\nu, making it consistent with the rest of the distribution.


devmotion (approved):
Thank you!

Issue #1726, opened 2023-05-23
Wrong quantiles calculated for some Truncated


waldie11:
Conform/Expected behavior:
```
julia > using Distributions
julia > quantile( truncated(Normal(),0,1), 1)
1.0 # == d.upper
```

Faulty behavior though there is better information available within `Truncated.d.upper` 
```
julia > quantile( truncated(Normal(0.1,0.01),0,1), 1)
Inf # == quantile(Normal(), 1)
```
Another apparent rounding error example
```
julia > quantile(truncated(Normal(-0.1,2),0,1),1)
1.0000000000000002 # still bigger than expected 1.0
```

Researching the line in `src/truncate.jl`
```
quantile(d::Truncated, p::Real) = quantile(d.untruncated, d.lcdf + p * d.tp)
```
offers more insight: it suffers from non-neglictible rounding errors in the sum of  `d.lcdf+d.tp`$\rightarrow 1$ as $p=1$

Iam suggesting a fix as in #1727 

Affected versions are probably all since `@truncate` macro was [introduced](https://github.com/JuliaStats/Distributions.jl/commit/06a268405f49b648665027c85ad75d26100fad47) . It can be reproduced at least on current master branch, 0.25.64 and 0.25.94

waldie11:
fixed by #1727

waldie11:
fixed by #1727

PR #1727, opened 2023-05-23
Fix Wrong quantiles calculated for some Truncated #1726

waldie11:
As described in
[#1726](https://github.com/JuliaStats/Distributions.jl/issues/1726), there occur small numbers-big number problems while evaluating quantiles on `Truncated`.

This Fix uses the knowledge on the upper bound of `Truncated` to properly limit the output.

### DIFF
--- a/src/truncate.jl
+++ b/src/truncate.jl
@@ -146,7 +146,7 @@ end
 
 ### evaluation
 
-quantile(d::Truncated, p::Real) = quantile(d.untruncated, d.lcdf + p * d.tp)
+quantile(d::Truncated, p::T) where {T<:Real} = min(T(d.upper), quantile(d.untruncated, d.lcdf + p * d.tp))
 
 function pdf(d::Truncated, x::Real)
     result = pdf(d.untruncated, x) / d.tp
--- a/src/truncate.jl
+++ b/src/truncate.jl
@@ -146,7 +146,11 @@ end
 
 ### evaluation
 
-quantile(d::Truncated, p::T) where {T<:Real} = min(T(d.upper), quantile(d.untruncated, d.lcdf + p * d.tp))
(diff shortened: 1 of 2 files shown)


waldie11:
Closes #1726

src/truncate.jl
```diff
@@ -146,7 +146,7 @@ end
 
 ### evaluation
 
-quantile(d::Truncated, p::Real) = quantile(d.untruncated, d.lcdf + p * d.tp)
+quantile(d::Truncated, p::T) where {T<:Real} = min(T(d.upper), quantile(d.untruncated, d.lcdf + p * d.tp))
```
devmotion, line 149: There are a few problems with the suggested code:
- `d.upper` might be `nothing`
- it might not be possible to convert `d.upper` to `T` (e.g., if `T = Int` and `d.upper = 1.2`)
- it does not cover `p = 0`.

devmotion: I think we might want to use something like
```suggestion
function quantile(d::Truncated, p::Real)
    x = quantile(d.untruncated, d.lcdf + p * d.tp)
    min_x, max_x = extrema(d)
    return clamp(x, oftype(x, min_x), oftype(x, max_x))
end
```

devmotion (commented):
Thank you for the PR! I made a suggestion for how to improve it a bit :slightly_smiling_face: 

Can you also add tests of the fixed issue?

waldie11:
@devmotion you are welcome :wink:

I think the examples resemble the discussed issues covered by this patch.

waldie11:
@devmotion bump

devmotion (approved):
Sorry, I was busy the last weeks with other things. The PR looks good to me, thanks!

PR #1728, opened 2023-05-25
Improve accuracy of `logdiffcdf(::Normal, x, y)`

devmotion:
Currently evaluating `logdiffcdf(::Normal, x, y)` for `x`, `y` in the tails of the distribution returns `-Inf`. We can do better by using `SpecialFunctions.logerf`.

This issue came up in #1723 and will be helpful for that PR I assume.


ararslan (approved):
Nice!

PR #1730, opened 2023-05-28
Handle truncation endpoints equal to support endpoints

ararslan:
Consider `truncated(Normal(), -Inf, Inf)`. Currently, this returns a `Truncated`-wrapped `Normal`. However, truncating a normal distribution to ±∞ is... not actually truncating it at all. We handle this case when the bounds are specified as `nothing` and return the untruncated distribution, but not when one or both bounds are equal to or beyond the respective bounds of the support. If we recode `lower <= minimum(d)` and/or `upper >= maximum(d)` to `nothing`, we're able to avoid a `Truncated` wrapper in more cases. And when we do make a `Truncated`, equivalent truncations can now compare equal and hit the same code paths for dispatch on the truncation using `nothing` as appropriate.

Two things to note with this implementation:
- `truncated(d::D, a, b)` is no longer type stable since it can now return either a `D` or a `Truncated{D}` whereas before it only returned a `D` if `a === b === nothing`.
- Nonsensical truncation, e.g. `truncated(Poisson(), -10, -1)`, still works and is still nonsensical.

### DIFF
--- a/src/truncate.jl
+++ b/src/truncate.jl
@@ -2,13 +2,19 @@
     truncated(d0::UnivariateDistribution; [lower::Real], [upper::Real])
     truncated(d0::UnivariateDistribution, lower::Real, upper::Real)
 
-A _truncated distribution_ `d` of a distribution `d0` to the interval
+A _truncated distribution_ `d` of a distribution `d0` to the closed interval
 ``[l, u]=```[lower, upper]` has the probability density (mass) function:
 
 ```math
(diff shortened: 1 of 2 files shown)


devmotion (commented):
I can see why this behaviour might be convenient (and this approach actually came up in some discussion in a private downstream repo not too long ago). But currently I think that

> truncated(d::D, a, b) is no longer typestable

is a major drawback, and that the increased convenience does not justify making `truncated` type unstable. In particular, since users can easily avoid these problems by specifying correct bounds or `nothing` (or, IMO more conveniently, the keyword argument syntax).

PR #1734, opened 2023-06-05
Update docstring of NegativeBinomial

heliosdrm:
Remove note about different notation in [Wikipedia](https://en.wikipedia.org/wiki/Negative_binomial_distribution) (no longer true).


simsurace (approved):
LGTM

PR #1738, opened 2023-06-18
UPDATED:  Multinomial distribution rand does not support Float32 probability vectors

### LINKED ISSUE #1032: Multinomial distribution rand does not support Float32 probability vectors

### LINKED ISSUE #1733: Multinomial sampling not working with Float32 vector
MWE:
```
probabilities = Vector{Float32}([0.3, 0.3, 0.2, 0.15, 0.05])
mnd = Multinomial(1, probabilities)
for i in 1:20
    println(rand(mnd))
end
```

nomadbl:
See issue #1032
This is virtually a copy of  #1033
I forked the code from @darsnack 's repositoty and rebased to master, and ran the tests.

Fixes #1032. Fixes #1733. Closes #1033.

### DIFF
--- a/src/samplers/multinomial.jl
+++ b/src/samplers/multinomial.jl
@@ -1,5 +1,5 @@
-function multinom_rand!(rng::AbstractRNG, n::Int, p::AbstractVector{Float64},
-                         x::AbstractVector{<:Real})
+function multinom_rand!(rng::AbstractRNG, n::Int, p::AbstractVector{PT},
+                         x::AbstractVector{T}) where {T<:Real, PT<:Real}
     k = length(p)
     length(x) == k || throw(DimensionMismatch("Invalid argument dimension."))
 
--- a/src/samplers/multinomial.jl
+++ b/src/samplers/multinomial.jl
@@ -1,9 +1,10 @@
-function multinom_rand!(rng::AbstractRNG, n::Int, p::AbstractVector{Tp},
-                         x::AbstractVector{Tx}) where {Tx <: Real, Tp <: Real}
+function multinom_rand!(rng::AbstractRNG, n::Int, p::AbstractVector{<:Real},
+                         x::AbstractVector{<:Real})
     k = length(p)
     length(x) == k || throw(DimensionMismatch("Invalid argument dimension."))
 
-    rp = Tp(1.0)  # remaining total probability
+    z = zero(eltype(p))
+    rp = oftype(z + z, 1) # remaining total probability (widens type if needed)
     i = 0
     km1 = k - 1
 
@@ -13,21 +14,21 @@ function multinom_rand!(rng::AbstractRNG, n::Int, p::AbstractVector{Tp},
         if pi < rp
             xi = Tx(rand(rng, Binomial(n, Float64(pi / rp))))
             @inbounds x[i] = xi
-            n -= Int64(xi)
+            n -= xi
             rp -= pi
         else
             # In this case, we don't even have to sample
             # from Binomial. Just assign remaining counts
             # to xi.
 
-            @inbounds x[i] = Tx(n)
+            @inbounds x[i] = n
             n = 0
             # rp = 0.0 (no need for this, as rp is no longer needed)
         end
     end
 
     if i == km1
-        @inbounds x[k] = Tx(n)
+        @inbounds x[k] = n
     else  # n must have been zero
         z = zero(eltype(x))
         for j = i+1 : k
--- a/src/samplers/multinomial.jl
+++ b/src/samplers/multinomial.jl
@@ -1,5 +1,5 @@
-function multinom_rand!(rng::AbstractRNG, n::Int, p::AbstractVector{PT},
-                         x::AbstractVector{T}) where {T<:Real, PT<:Real}
+function multinom_rand!(rng::AbstractRNG, n::Int, p::AbstractVector{<:Real},
+                         x::AbstractVector{<:Real})
     k = length(p)
     length(x) == k || throw(DimensionMismatch("Invalid argument dimension."))
 
--- a/src/samplers/multinomial.jl
+++ b/src/samplers/multinomial.jl
@@ -1,9 +1,9 @@
-function multinom_rand!(rng::AbstractRNG, n::Int, p::AbstractVector{<:Real},
-                         x::AbstractVector{<:Real})
+function multinom_rand!(rng::AbstractRNG, n::Int, p::AbstractVector{Tp},
+                         x::AbstractVector{Tx}) where {Tx <: Real, Tp <: Real}
     k = length(p)
     length(x) == k || throw(DimensionMismatch("Invalid argument dimension."))
 
-    rp = 1.0  # remaining total probability
+    rp = Tp(1.0)  # remaining total probability
     i = 0
     km1 = k - 1
 
@@ -11,23 +11,23 @@ function multinom_rand!(rng::AbstractRNG, n::Int, p::AbstractVector{<:Real},
         i += 1
         @inbounds pi = p[i]
         if pi < rp
-            xi = rand(rng, Binomial(n, pi / rp))
+            xi = Tx(rand(rng, Binomial(n, Float64(pi / rp))))
             @inbounds x[i] = xi
-            n -= xi
+            n -= Int64(xi)
             rp -= pi
         else
             # In this case, we don't even have to sample
             # from Binomial. Just assign remaining counts
             # to xi.
 
-            @inbounds x[i] = n
+            @inbounds x[i] = Tx(n)
             n = 0
             # rp = 0.0 (no need for this, as rp is no longer needed)
         end
     end
 
     if i == km1
-        @inbounds x[k] = n
+        @inbounds x[k] = Tx(n)
     else  # n must have been zero
         z = zero(eltype(x))
         for j = i+1 : k
--- a/src/samplers/multinomial.jl
+++ b/src/samplers/multinomial.jl
@@ -12,7 +12,7 @@ function multinom_rand!(rng::AbstractRNG, n::Int, p::AbstractVector{<:Real},
         i += 1
         @inbounds pi = p[i]
         if pi < rp
-            xi = Tx(rand(rng, Binomial(n, Float64(pi / rp))))
+            xi = rand(rng, Binomial(n, Float64(pi / rp)))
             @inbounds x[i] = xi
             n -= xi
             rp -= pi
--- a/test/multivariate/multinomial.jl
+++ b/test/multivariate/multinomial.jl
@@ -2,7 +2,6 @@
 
 using Distributions, Random, StaticArrays
 using Test
-import Distributions: multinom_rand!
 
 
 p = [0.2, 0.5, 0.3]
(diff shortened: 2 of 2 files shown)


src/samplers/multinomial.jl
```diff
@@ -1,4 +1,4 @@
-function multinom_rand!(rng::AbstractRNG, n::Int, p::AbstractVector{Float64},
+function multinom_rand!(rng::AbstractRNG, n::Int, p::AbstractVector{<:Real},
```
devmotion, line 1: AFAICT one has to update the body of the function as well to avoid type stability issues.

devmotion: More concretely, it must not contain any Float64 literals.

test/multivariate/multinomial.jl
```diff
@@ -1,6 +1,6 @@
 # Tests for Multinomial
 
-using Distributions, Random, StaticArrays
+using Distributions, Random, StaticArrays, ForwardDiff
```
devmotion, line 3: Not needed for the changes below it seems:
```suggestion
using Distributions, Random, StaticArrays
```

test/multivariate/multinomial.jl
```diff
@@ -210,3 +210,8 @@ p_v = [0.1, 0.4, 0.3, 0.8]
 @test_throws DomainError Multinomial(10, p_v)
 @test_throws DomainError Multinomial(10, p_v; check_args=true)
 Multinomial(10, p_v; check_args=false) # should not warn
+
+# check different prob vector types
+p = [0.2, 0.4, 0.3, 0.1]
+@test (rand(Multinomial(10, p)); true)
+@test (rand(Multinomial(10, convert.(Float32, p))); true)
```
devmotion, line 217: Can we use better tests that test correctness instead of only "not erroring"? For instance by looping over `Float64` and `Float32` probabilities in the tests above?

devmotion (commented):
Thank you for the PR! I think it looks quite good but could be improved a bit. I made some comments below :slightly_smiling_face:

test/multivariate/multinomial.jl
nomadbl: I'll remove it happily, as I don't personally think it really belongs here.
But at least for completeness sake, here is the source of this addition:
https://github.com/JuliaStats/Distributions.jl/pull/1033#issuecomment-570786766

test/multivariate/multinomial.jl
nomadbl: I'm not sure I follow.
What kind of correctness do you mean?
As for type stability, I'm looking at adding some `@inferred` tests here for that.

src/samplers/multinomial.jl
nomadbl: Isn't is true that the output type of the function can only be `x::AbstractVector{eltype(x)}` ?
I'm adding the type stability tests and it looks ok but I wanted to make sure since I'm pretty new to Julia.

test/multivariate/multinomial.jl
devmotion: That the samples follow the desired distribution - basically, the same tests as for Float64.

src/samplers/multinomial.jl
devmotion: There are internal type instabilities which can cause slow downs.

test/multivariate/multinomial.jl
devmotion: I'd prefer running the existing tests with Float32 as well over a few additional tests.

nomadbl: took care of this :heavy_check_mark:

src/samplers/multinomial.jl
nomadbl: I took care of this aspect.
While doing so I noticed that Binomial is restricted to using Float64, so it may be slower/more accurate than necessary.
This also would affect Multinomial.
This is not needed for me, but it can be a good addition in future.

test/multivariate/multinomial.jl
devmotion: It seems it was added because someone wanted dual numbers to be tested as well - but later it was figured out that for other reasons even with the changes in the PR they don't work and the tests were removed again.

So it seems fine to remove this import.

src/samplers/multinomial.jl
```diff
@@ -1,33 +1,33 @@
-function multinom_rand!(rng::AbstractRNG, n::Int, p::AbstractVector{Float64},
-                         x::AbstractVector{<:Real})
+function multinom_rand!(rng::AbstractRNG, n::Int, p::AbstractVector{Tp},
+                         x::AbstractVector{Tx}) where {Tx <: Real, Tp <: Real}
     k = length(p)
     length(x) == k || throw(DimensionMismatch("Invalid argument dimension."))
 
-    rp = 1.0  # remaining total probability
+    rp = Tp(1.0)  # remaining total probability
```
devmotion, line 6: I'd suggest something like
```suggestion
    z = zero(eltype(p))
    rp = oftype(z + z, 1) # remaining total probability (widens type if needed)
```
This will also work better with `p::Vector{Bool}`.

src/samplers/multinomial.jl
```diff
@@ -1,33 +1,33 @@
...
     i = 0
     km1 = k - 1
 
     while i < km1 && n > 0
         i += 1
         @inbounds pi = p[i]
         if pi < rp
-            xi = rand(rng, Binomial(n, pi / rp))
+            xi = Tx(rand(rng, Binomial(n, Float64(pi / rp))))
```
devmotion, line 14: This is not needed, updating `x[i]` in the line below will convert to the correct type anyway; additionally, the constructor of `Binomial` (should) take care of promotions/conversions (and we just continue with using the type of `p`/`rp`):
```suggestion
            xi = rand(rng, Binomial(n, pi / rp))
```

src/samplers/multinomial.jl
```diff
@@ -1,33 +1,33 @@
...
             @inbounds x[i] = xi
-            n -= xi
+            n -= Int64(xi)
```
devmotion, line 16: Not needed, as `rand` above returns an `Int` (generally, also one should not use `Int64` explicitly if there is not a very good reason - on 32bit machines you end up with a (very slow) mix of `Int32` and `Int64` otherwise):
```suggestion
            n -= xi
```

src/samplers/multinomial.jl
```diff
@@ -1,33 +1,33 @@
...
             rp -= pi
         else
             # In this case, we don't even have to sample
             # from Binomial. Just assign remaining counts
             # to xi.
 
-            @inbounds x[i] = n
+            @inbounds x[i] = Tx(n)
```
devmotion, line 23:
```suggestion
            @inbounds x[i] = n
```

src/samplers/multinomial.jl
```diff
@@ -1,33 +1,33 @@
...
             n = 0
             # rp = 0.0 (no need for this, as rp is no longer needed)
         end
     end
 
     if i == km1
-        @inbounds x[k] = n
+        @inbounds x[k] = Tx(n)
```
devmotion, line 30:
```suggestion
        @inbounds x[k] = n
```

nomadbl:
@devmotion 
Thanks for the input.
Applying your suggestions and running the tests yields failures due to this:
```
Type stability: Error During Test at /home/lior/Distributions.jl/test/multivariate/multinomial.jl:216
  MethodError: no method matching Distributions.BinomialGeomSampler(::Int64, ::Float32)
  
  Closest candidates are:
    Distributions.BinomialGeomSampler(::Any, ::Any, ::Any)
     @ Distributions ~/Distributions.jl/src/samplers/binomial.jl:29
    Distributions.BinomialGeomSampler(::Int64, ::Float64)
     @ Distributions ~/Distributions.jl/src/samplers/binomial.jl:36
```
Or the equivalent message using `Float16`.

That was the reason I made the type conversion in `Binomial(n, Float64(pi / rp))`.

As to preferring the implicit over explicit type conversions, I was thinking it would improve performance (saving the type check at run time) but I can also see the point of not using hard coded types when possible. The only objection I guess would be that in 32bit systems the user would not use Float64 to begin with, so that's not a real issue right?

nomadbl:
@devmotion 
Putting out a reminder here. Everything should be good to go as far as I can tell.

devmotion:
I'll have another look at the PR tomorrow :slightly_smiling_face:

devmotion:
I tried to simplify the tests a bit (e.g., replaced hard-coded `Float64` and `Float32` with loops over types and included `@inferred` in the existing tests instead of adding new tests). If tests pass (apart from Tracker+ReverseDiff), the PR looks good to me :slightly_smiling_face:  Thanks a lot!

devmotion (approved):
Looks good to me, thank you!

mrazomej:
I don't know if I'm wrong about this, but shouldn't the `multinom_rand!` return a `Vector{Int32}` if the probability vector is `Vector{Float32}`?

For example, if I do

```
using Distributions

rand(Normal(0f0, 1f0))
```
I get a `Float32`. But if I do
```
rand(Multinomial(1, Float32[0.3, 0.3, 0.2, 0.15, 0.05]))
```
I get a `Vector{Int64}`. And if I try
```
rand(Multinomial(Int32(1), Float32[0.3, 0.3, 0.2, 0.15, 0.05]))
```
I get an error
```
ERROR: MethodError: no method matching Multinomial{Float32, Vector{Float32}}(::Int32, ::Vector{Float32})

Closest candidates are:
  Multinomial{T, TV}(::Int64, ::TV) where {T<:Real, TV<:AbstractVector{T}}
   @ Distributions ~/.julia/packages/Distributions/GrN7f/src/multivariate/multinomial.jl:25

Stacktrace:
 [1] Multinomial(n::Int32, p::Vector{Float32}; check_args::Bool)
   @ Distributions ~/.julia/packages/Distributions/GrN7f/src/multivariate/multinomial.jl:34
 [2] Multinomial(n::Int32, p::Vector{Float32})
   @ Distributions ~/.julia/packages/Distributions/GrN7f/src/multivariate/multinomial.jl:28
 [3] top-level scope
   @ REPL[49]:1
```

devmotion:
No, that would seem very surprising and unintended to me. Typically, `Int32` mainly shows up on 32bit architectures where `Int = Int32` but it does not correspond to and is not related to `Float32` (eg, also on 32bit machines the most common floating point number type is `Float64`).

This also explains the error you see: The constructor is only defined for `Int`, so for `Int64` on 64bit architectures and `Int32` on 32bit.

mrazomej:
I see; thank you very much for the clarification. I need to understand the differences better. Naively, I thought that working with 32-bit numbers would speed up my computations when doing something such as with `Turing.jl` or `Flux.jl`.

devmotion:
Yes, probably you'll get a speedup from using `Float32` (as long as there are no hardcoded `Float64` literals in the called functions or other problems that cause automatic promotion to `Float64`) and you also want (have?) to use it on GPUs. But usually you don't want to use `Int32` on 64bit architectures (if it is supported at all since many functions are defined for `Int` only which is `Int64` on 64bit).

PR #1744, opened 2023-06-30
Fix MultinomialSampler overflow

0:
`n^2` doesn't fit into `Int64` for `n` larger than `3_037_000_499`.

### DIFF
--- a/src/samplers/multinomial.jl
+++ b/src/samplers/multinomial.jl
@@ -52,7 +52,7 @@ function _rand!(rng::AbstractRNG, s::MultinomialSampler,
                 x::AbstractVector{<:Real})
     n = s.n
     k = length(s)
-    if n^2 > k
+    if n > isqrt(k)
         multinom_rand!(rng, n, s.prob, x)
     else
         # Use an alias table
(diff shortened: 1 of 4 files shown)


src/samplers/multinomial.jl
```diff
@@ -52,7 +52,7 @@ function _rand!(rng::AbstractRNG, s::MultinomialSampler,
                 x::AbstractVector{<:Real})
     n = s.n
     k = length(s)
-    if n^2 > k
+    if n > sqrt(k)
```
devmotion, line 55: It's just a heuristic but to ensure that we get the same behaviour as before I think we need
```suggestion
    if n > isqrt(k)
```
here (see, e.g., the discussion here: https://github.com/JuliaLang/julia/issues/4884).

0:
Changed `sqrt` to `isqrt`.

devmotion:
OK :slightly_smiling_face: Now the main remaining question is: is it possible to add a test for the fix? I assume it could be a bit challenging since the change only affects whether an alias table is used or not, but should not affect correctness (I assume).

0:
Right, this change is about performance and not correctness, since the overflow only affects the heuristic. I initially wrote a test along the lines of
```julia
for n in [3_037_000_499, 3_037_000_500]
    @test all(n .== rand(Multinomial(n, 1), 2))
end
```
but didn't end up including it, because it should pass with both versions of the code anyway. It takes significantly longer using the alias table, but that seems like the wrong thing to check in a unit test.

Issue #1701, opened 2023-03-22
Documentation of the fit method


jaksle:
Only Beta distribution has a documentation of the fit method. This causes `?fit` to show this documentation whereas the intention of the user would be rather to learn how to fit a general one. A possible solution is to remove it and write a general one for `fit(::Type{<:Distribution}, x::AbstractArray{T})`.

PR #1746, opened 2023-07-07
Fix `fit` docs

ParadaCarleton:
`fit` doesn't need a specific docstring just for `Beta`.


ParadaCarleton:
Any objections to merging this?

jaksle:
I mentioned this in Issue #1701. For sure docstring for Beta should be removed, but currently fit has no general docstring for a distribution fit. The question is should it be added.

i9e1:
maybe related #1748

I assume now that `Poisson` is not supported as stated in doc of `fit_mle`.

src/genericfit.jl
```diff
@@ -30,5 +30,15 @@ fit_mle(dt::Type{D}, x::AbstractArray, w::AbstractArray) where {D<:UnivariateDis
 fit_mle(dt::Type{D}, x::AbstractMatrix) where {D<:MultivariateDistribution} = fit_mle(D, suffstats(D, x))
 fit_mle(dt::Type{D}, x::AbstractMatrix, w::AbstractArray) where {D<:MultivariateDistribution} = fit_mle(D, suffstats(D, x, w))
 
+"""
+    fit(D, x)
```
devmotion, line 34: Apparently also `fit(D, args...)` is defined below...

src/genericfit.jl
```diff
@@ -30,5 +30,15 @@ fit_mle(dt::Type{D}, x::AbstractArray, w::AbstractArray) where {D<:UnivariateDis
...
+
+Fit a distribution of type `D` to a dataset `x` using an arbitrary algorithm.
```
devmotion, line 36: I guess
```suggestion
Fit a distribution of type `D` to a dataset `x`.
```
would be sufficient?

Additionally, I think the term `dataset` is very unclear - sounds a bit like it should be a data frame or table even though usually just vectors of samples are suppported.

src/genericfit.jl
```diff
@@ -30,5 +30,15 @@ fit_mle(dt::Type{D}, x::AbstractArray, w::AbstractArray) where {D<:UnivariateDis
...
+
+The fit function will choose a reasonable way to fit the distribution, which,
+in most cases, is maximum likelihood estimation. Note that this algorithm may
```
devmotion, line 39: Maybe this can be formulated in a stronger way - the default fallback is `fit_mle`, so `fit` is always the MLE fit if not explicitly defined otherwise.

ParadaCarleton:
Anything to do before merging?

src/genericfit.jl
```diff
@@ -30,5 +30,18 @@ fit_mle(dt::Type{D}, x::AbstractArray, w::AbstractArray) where {D<:UnivariateDis
 fit_mle(dt::Type{D}, x::AbstractMatrix) where {D<:MultivariateDistribution} = fit_mle(D, suffstats(D, x))
 fit_mle(dt::Type{D}, x::AbstractMatrix, w::AbstractArray) where {D<:MultivariateDistribution} = fit_mle(D, suffstats(D, x, w))
 
+"""
+    fit(D, args...)
+
+Fit a distribution of type `D` to a dataset `x`.
```
devmotion, line 36: What's `x`? It does not appear in the function signature two lines above.

ParadaCarleton: Whoops, thanks for pointing that out.

ParadaCarleton:
Anything else or can I merge?

ParadaCarleton:
Anything else?

ParadaCarleton:
@devmotion should I merge?

ViralBShah:
Merge?

devmotion:
Why was this merged? Docs seem to be broken (missing docstrings): https://juliastats.org/Distributions.jl/dev/fit/

PR #1758, opened 2023-08-07
Import Base.in

### LINKED ISSUE #1756: Why not `Base.in` for the `support` (e.g. `RealInterval`)
The following fails, but it should not really

```julia
julia> using Distributions

julia> support(Gamma)
RealInterval{Float64}(0.0, Inf)

julia> 1.0 ∈ support(Gamma)
ERROR: MethodError: no method matching iterate(::RealInterval{Float64})

Closest candidates are:
  iterate(::Union{LinRange, StepRangeLen})
   @ Base range.jl:880
  iterate(::Union{LinRange, StepRangeLen}, ::Integer)
   @ Base range.jl:880
  iterate(::T) where T<:Union{Base.KeySet{<:Any, <:Dict}, Base.ValueIterator{<:Dict}}
   @ Base dict.jl:698
  ...

Stacktrace:
 [1] in(x::Float64, itr::RealInterval{Float64})
   @ Base ./operators.jl:1282
 [2] top-level scope
   @ REPL[4]:1
```

The problem is that 
```
julia> Distributions.in === Base.in
false
```

It can be easily fixed if added in the `Distributions`
```
import Base: in 
```

bvdmitri:
Fixes #1756

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -4,7 +4,7 @@ using StatsBase, PDMats, StatsFuns, Statistics
 using StatsFuns: logtwo, invsqrt2, invsqrt2π
 
 import QuadGK: quadgk
-import Base: size, length, convert, show, getindex, rand, vec, inv
+import Base: in, size, length, convert, show, getindex, rand, vec, inv
 import Base: sum, maximum, minimum, extrema, +, -, *, ==
 import Base.Math: @horner
 
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -4,7 +4,7 @@ using StatsBase, PDMats, StatsFuns, Statistics
 using StatsFuns: logtwo, invsqrt2, invsqrt2π
 
 import QuadGK: quadgk
-import Base: in, size, length, convert, show, getindex, rand, vec, inv
+import Base: size, length, convert, show, getindex, rand, vec, inv
 import Base: sum, maximum, minimum, extrema, +, -, *, ==
 import Base.Math: @horner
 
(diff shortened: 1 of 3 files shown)


src/Distributions.jl
```diff
@@ -4,7 +4,7 @@ using StatsBase, PDMats, StatsFuns, Statistics
 using StatsFuns: logtwo, invsqrt2, invsqrt2π
 
 import QuadGK: quadgk
-import Base: size, length, convert, show, getindex, rand, vec, inv
+import Base: in, size, length, convert, show, getindex, rand, vec, inv
```
devmotion, line 7: Can you instead qualify `in` in the definition of `in(::Real, ::RealInterval)`? IMO we should try to avoid generic `import`s to make it clearer when and which functions are overloaded.

bvdmitri: Should be better now

test/testutils.jl
```diff
@@ -341,13 +341,29 @@ function test_support(d::UnivariateDistribution, vs::AbstractVector)
...
         end
+        if isupperbounded(d)
+            @test maximum(d) ∈ s
+            @test (maximum(d) + 1) ∉ s
+        end
+    end
+
+    if isbounded(d) && isa(d, DiscreteUnivariateDistribution)
+        s = support(d)
+        @test isa(s, AbstractUnitRange)
```
devmotion, line 364: That's not true in general. E.g., for `Dirac` the support is a one-element tuple.

bvdmitri: The change just shows that I removed two `if` statements into a single `if` statement

devmotion (approved):

Issue #1765, opened 2023-09-05
AffineTransformation does not preserve eltype Float32


bgctw:
For some number-crunching project we work with Float32 instead of Float64.
Most distributions in Distributions.jl allow to construct a Float32-based version by providing a Type parameter, nice.

However, when I transform a Float32 based LogitNormal distribution (e.g. shift mean by +) the mean applied to the distribution gives me a Float64 instead of a Float32.

The reason is
- AffineTransformation uses `eltype(basedist)` to determine its number and parameter type
- default eltype of an uniform continuous distribution is Float64

The normal distribution already defines
`Base.eltype(::Type{Normal{T}}) where {T} = T` 

Should we add such a method to all the other distributions that are based on a parametric type?

Here is a minimal (not) working example

```
    dln = LogitNormal{Float32}(0f0, sqrt(2f0))
    dshifted = dln * 2f0 + 1f0
    @test params(dshifted)[1] isa Float32
    @test mean(dshifted) isa Float32

```

bgctw:
@devmotion commented on a commit: "The parameter type of a distribution can be retrieved by partype, eltype is supposed to be the (element) type of rand"

Computing the mean of distribution, to my understanding, should return the same type as a random number drawn. Moreover, all the uniform continuous distributions which I looked at, enforce the same parametric type as the type of the parameters.

This also leads to some strange behaviour such as:
`d = LogUniform(1,10)`
leading to parametric type Int64
but the Tests still check
`@test eltype(d) === Float64`

Are the terms, concepts, and types involved described somewhere? My expectation would be the [Type Hierarchy](https://juliastats.org/Distributions.jl/latest/types/) section of the documentation. However, the parametric types of the distributions and the types of the parameters are not really clarified there.
Could you, please, point me to where I can read and understand these concepts?

devmotion:
Presumably there's no documentation because people strongly disagree about the design and its implications, as shown in #1071 and its many duplicates. I had a go at it (and some other issues) in https://github.com/JuliaStats/Distributions.jl/pull/1433 but the PR got stuck since I had to work on other things.

ParadaCarleton:
```julia
dln = LogitNormal{Float32}(0f0, sqrt(2f0))
dshifted = dln * 2f0 + 1f0
julia> typeof(rand(dln))
Float64
```
So it seems like the bug is in `LogitNormal`, not in `AffineDistribution`.

bgctw:
[...]

Issue #1402, opened 2021-10-06
Strange `eltype` bahavior


cscherrer:
This seems problematic:
```julia
julia> using Distributions

julia> eltype(Normal())
Float64

julia> eltype(MvNormal(ones(3)))
Float64
```

I found this very surprising, as I'd guess many others have/will. In general, it seems reasonable to expect a law that for any `d::Distribution`,
```julia
typeof(rand(d)) <: eltype(d)
```

My biggest concern here is that I'd like to implement `eltype` in MeasureTheory.jl, and I don't know if people might expect it to behave as it does in Distributions. 

Could there be a path to changing this?

cscherrer:
Thanks for the link! Looks like this has been quite a saga. This design decision for `eltype` to refer to the parameters is very confusing, and seems inconsistent with the rest of Julia. I think it may be best for MeasureTheory to ignore this quirk and just focus of having its `eltype` consistent with Base.

devmotion:
> which should then match the type that comes out of rand.

The main issue here is that it is not always possible to compute the type of `rand(d)` in a nice way, e.g., if sampling is based on a complicated push forward operation or you want to use different "elementary" random samples, e.g. `rand()`, `rand(Float32)`, `rand(Float16)`, or `rand(BigFloat)` for different precision.

mschauer:
Yeah, you know there is a parallel infrastructure where the type of `rand(d)` is derived from `ValueSupport` and `VariateForm` *in most cases* https://github.com/JuliaStats/Distributions.jl/blob/625b72237a342c8d3bf60ec05541f8cb4a78faff/src/common.jl

devmotion:
Sure but unfortunately it is only a heuristic and requires e.g. `float` in https://github.com/JuliaStats/Distributions.jl/blob/27fb31ce6098149fa051613483edfafd616de310/src/univariates.jl#L140.

devmotion:
> It wouldn't return an element type, but the full type of a single variate, so Float64, or Int, or Vector{Float32}, etc.

I think the only reliable way that always works is to call `typeof(rand(d))` :shrug: I guess for many simple cases a heuristic such as `float(eltype(d))` for continuous univariate distributions will work but it's not guaranteed to be correct.

oschulz:
> I think the only reliable way that always works ... I guess for many simple cases a heuristic

[...]

PR #1766, opened 2023-09-05
Base.eltype for uniform continuous distributions

bgctw:
deals with #1765


devmotion:
The parameter type of a distribution can be retrieved by `partype`, `eltype` is supposed to be the (element) type of `rand`.

devmotion:
I'll close this PR since it does not seem to be what we want to do right now. Maybe we even want to remove `eltype` completely (possibly supporting `gentype` instead? https://github.com/JuliaStats/Distributions.jl/issues/1402#issuecomment-975168414). In any case, this needs some design considerations and decisions first.

PR #1769, opened 2023-09-07
allow specify parameter type in AffineDistribution

bgctw:
second suggestion to workaround #1765

### DIFF
--- a/src/univariate/locationscale.jl
+++ b/src/univariate/locationscale.jl
@@ -49,10 +49,17 @@ struct AffineDistribution{T<:Real, S<:ValueSupport, D<:UnivariateDistribution{S}
     end
 end
 
-function AffineDistribution(μ::T, σ::T, ρ::UnivariateDistribution; check_args::Bool=true) where {T<:Real}
+function AffineDistribution(μ::T, σ::T, ρ::UnivariateDistribution, _T::Type; check_args::Bool=true) where {T<:Real}
+    #workaround ##1765 to allow specifying Float32 as parametric type
     @check_args AffineDistribution (σ, !iszero(σ))
-    _T = promote_type(eltype(ρ), T)
     return AffineDistribution{_T}(_T(μ), _T(σ), ρ)
+end
+
+function AffineDistribution(μ::T, σ::T, ρ::UnivariateDistribution; check_args::Bool=true) where {T<:Real}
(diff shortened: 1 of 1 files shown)


src/univariate/continuous/logitnormal.jl
```diff
@@ -72,6 +72,9 @@ LogitNormal(μ::Real=0.0) = LogitNormal(μ, one(μ); check_args=false)
 @distr_support LogitNormal 0.0 1.0
 #insupport(d::Union{D,Type{D}},x::Real) where {D<:LogitNormal} = 0.0 < x < 1.0
 
+# # TODO change @distr_support to generate functions that respect the type parameter
+# minimum(::LogitNormal{T}) where T = T(0)
+# maximum(::LogitNormal{T}) where T = T(1)
```
devmotion, line 77:
```suggestion
```

src/univariate/locationscale.jl
```diff
@@ -49,20 +49,29 @@ struct AffineDistribution{T<:Real, S<:ValueSupport, D<:UnivariateDistribution{S}
     end
 end
 
-function AffineDistribution(μ::T, σ::T, ρ::UnivariateDistribution; check_args::Bool=true) where {T<:Real}
+function AffineDistribution(μ::T, σ::T, ρ::UnivariateDistribution, _T::Type; check_args::Bool=true) where {T<:Real}
```
devmotion, line 52: I'm not a fan of this approach - if users want to set the types specifically, they can just call one of the internal constructors.

Can't we just replace line 54 with
```julia
_T = promote_type(partype(ρ), T)
```

BTW note that `Type` is one of the special cases where the Julia compiler does not specialize on the argument. If you want to specialize on the type, then you have to use `::Type{_T}`.

bgctw: Thanks @devmotion for this aid.

I already tried making AffineTransformation using partype instead of eltype and it caused problems/failing tests with the discretenonparametric distribution. 

I can prepare a pull request with the change and we can look, whether the failing discretenonparametric tests are reasonable (expecting rationals that are promoted to float with the change)

The deeper problem to me is that AffineTransformation transforms both, parameter-based quantities and samples. I think we should keep these two things separate (#1433)

Issue #1771, opened 2023-09-07
Adding MvLogitNormal


sethaxen:
This would be the multivariate generalization of `LogitNormal` described in https://en.m.wikipedia.org/wiki/Logit-normal_distribution, better known as logistic normal. It's an alternative to Dirichlet for distributions of probability vectors.

The implementation approach should probably be similar to `MvLogNormal`.

sethaxen:
Actually, it seems a better approach than `MvLogNormal`'s would be to simply wrap an `AbstractMvNormal` and for its constructor to forward arguments to `MvNormal`. Then its constructor supports all the variants of `AbstractMvNormal`, including `MvNormalCanon`, not just `MvNormal`.

sethaxen:
Implemented in #1774

PR #1774, opened 2023-09-08
Add MvLogitNormal

sethaxen:
Implements #1771

### DIFF
--- a/src/multivariate/mvlogitnormal.jl
+++ b/src/multivariate/mvlogitnormal.jl
@@ -1,5 +1,5 @@
 """
-    MvLogitNormal
+    MvLogitNormal{<:AbstractMvNormal}
 
 The [multivariate logit-normal distribution](https://en.wikipedia.org/wiki/Logit-normal_distribution#Multivariate_generalization)
 is the multivariate generalization of [`LogitNormal`](@ref).
@@ -20,6 +20,10 @@ MvLogitNormal{MvNormal}(μ, Σ)       # same as above
 MvLogitNormal(MvNormalCanon(μ, J))  # MvLogitNormal with y ~ MvNormalCanon(μ, J)
 MvLogitNormal{MvNormalCanon}(μ, J)  # same as above
 ```
+
+# Fields
+
+- `normal::AbstractMvNormal`: contains the ``d-1``-dimensional distribution of ``y``
 """
 struct MvLogitNormal{D<:AbstractMvNormal} <: ContinuousMultivariateDistribution
     normal::D
--- a/src/multivariate/mvlogitnormal.jl
+++ b/src/multivariate/mvlogitnormal.jl
@@ -2,9 +2,8 @@
     MvLogitNormal{<:AbstractMvNormal}
 
 The [multivariate logit-normal distribution](https://en.wikipedia.org/wiki/Logit-normal_distribution#Multivariate_generalization)
-is the multivariate generalization of [`LogitNormal`](@ref).
-
-It is a more general alternative to [`Dirichlet`](@ref) for probability vectors.
+is a multivariate generalization of [`LogitNormal`](@ref) capable of handling correlations
+between variables.
 
 If ``\\mathbf{y} \\sim \\mathrm{MvNormal}(\\boldsymbol{\\mu}, \\boldsymbol{\\Sigma})`` is a
 length ``d-1`` vector, then
--- a/src/multivariate/mvlogitnormal.jl
+++ b/src/multivariate/mvlogitnormal.jl
@@ -32,11 +32,11 @@ end
 # Conversions
 
 function convert(::Type{MvLogitNormal{D}}, d::MvLogitNormal) where {D}
-    return MvLogNormal(convert(D, d.normal))
+    return MvLogitNormal(convert(D, d.normal))
 end
 Base.convert(::Type{MvLogitNormal{D}}, d::MvLogitNormal{D}) where {D} = d
 function convert(::Type{MvLogitNormal{D}}, pars...) where {D}
-    return MvLogNormal(convert(D, MvNormal(pars...)))
+    return MvLogitNormal(convert(D, pars...))
 end
 
 function Base.show(io::IO, d::MvLogitNormal)
--- a/src/multivariate/mvlogitnormal.jl
+++ b/src/multivariate/mvlogitnormal.jl
@@ -35,7 +35,7 @@ MvLogitNormal(args...) = MvLogitNormal(MvNormal(args...))
 function Base.show(io::IO, d::MvLogitNormal)
     # extract params lines from multiline show of d.normal
     norm_str = sprint(show, d.normal; context=IOContext(io))
-    params_str = strip(replace(norm_str, r"^.*?\(" => "", r"\).*?$" => ""))
+    params_str = strip(replace(replace(norm_str, r"^.*?\(" => ""), r"\).*?$" => ""))
     # show with this distribution name
     print(io, distrname(d))
     println(io, "(")
--- a/src/multivariate/mvlogitnormal.jl
+++ b/src/multivariate/mvlogitnormal.jl
@@ -16,12 +16,20 @@ is a length ``d`` probability vector.
 ```julia
 MvLogitNormal(μ, Σ)                 # MvLogitNormal with y ~ MvNormal(μ, Σ)
 MvLogitNormal(MvNormal(μ, Σ))       # same as above
+MvLogitNormal{MvNormal}(μ, Σ)       # same as above
 MvLogitNormal(MvNormalCanon(μ, J))  # MvLogitNormal with y ~ MvNormalCanon(μ, J)
+MvLogitNormal{MvNormalCanon}(μ, J)  # same as above
 ```
 """
 struct MvLogitNormal{D<:AbstractMvNormal} <: ContinuousMultivariateDistribution
     normal::D
+    MvLogitNormal{D}(normal::D) where {D<:AbstractMvNormal} = new{D}(normal)
 end
+function MvLogitNormal{D}(args...) where {D<:AbstractMvNormal}
+    normal = D(args...)
+    return MvLogitNormal{typeof(normal)}(normal)
+end
+MvLogitNormal(d::AbstractMvNormal) = MvLogitNormal{typeof(d)}(d)
 MvLogitNormal(args...) = MvLogitNormal(MvNormal(args...))
 
 function Base.show(io::IO, d::MvLogitNormal)
--- a/src/multivariate/mvlogitnormal.jl
+++ b/src/multivariate/mvlogitnormal.jl
@@ -15,19 +15,19 @@ is a length ``d`` probability vector.
 
 ```julia
 MvLogitNormal(μ, Σ)                 # MvLogitNormal with y ~ MvNormal(μ, Σ)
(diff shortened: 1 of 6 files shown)


sethaxen:
Currently it takes a different approach than `MvLogNormal`. `MvLogNormal` wraps an `MvNormal`, so it cannot be parameterized like an `MvNormalCanon` or other `AbstractMvNormal` types. `MvLogitNormal` instead wraps an `AbstractMvNormal`, which allows it to also wrap `MvNormalCanon` or any user-defined `AbstractMvNormal` type.

src/multivariate/mvlogitnormal.jl
```diff
@@ -0,0 +1,146 @@
+"""
+    MvLogitNormal
+
+The [multivariate logit-normal distribution](https://en.wikipedia.org/wiki/Logit-normal_distribution#Multivariate_generalization)
+is the multivariate generalization of [`LogitNormal`](@ref).
+
+It is a more general alternative to [`Dirichlet`](@ref) for probability vectors.
```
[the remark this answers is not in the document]
sethaxen: Why this removal? This is primarily explaining its significance and utility. The compositional data analysis literature describes 3 different notions of independence that exist in distributions on the simplex. Dirichlet exhibits the most extreme versions of all 3 and is provably the only distribution that does so. In that sense, Dirichlet is very weird. It can't even have non-convex equi-probability contours.

While MvLogitNormal does not contain Dirichlet as a special case, it comes close but also for a dense covariance matrix exhibits none of the notions of dependence, and it can have concave contours.

src/multivariate/mvlogitnormal.jl
```diff
@@ -0,0 +1,146 @@
...
+    MvLogitNormal{D}(normal::D) where {D<:AbstractMvNormal} = new{D}(normal)
+end
+function MvLogitNormal{D}(args...) where {D<:AbstractMvNormal}
+    normal = D(args...)
+    return MvLogitNormal{typeof(normal)}(normal)
+end
+MvLogitNormal(d::AbstractMvNormal) = MvLogitNormal{typeof(d)}(d)
+MvLogitNormal(args...) = MvLogitNormal(MvNormal(args...))
+
+function Base.show(io::IO, d::MvLogitNormal)
```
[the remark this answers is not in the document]
sethaxen: The generic methods treat the field names as parameters, which here looks weird because the field contains another distribution:
```julia
julia> using Distributions, LinearAlgebra

julia> d = MvLogitNormal(randn(3), exp(Symmetric(randn(3, 3))))
MvLogitNormal{FullNormal}(
dim: 3
μ: [-0.6695883087710813, -0.4262204488438936, 1.5508073295811842]
Σ: [1.8911987804702053 0.038814771826912765 1.2907448948413436; 0.038814771826912765 0.7054789545738211 0.06748072828787553; 1.2907448948413436 0.06748072828787553 0.9560938804684098]
)


julia> invoke(Base.show, Tuple{IO,Distribution}, stdout, d)
MvLogitNormal{FullNormal}(
normal: FullNormal(
dim: 3
μ: [-0.6695883087710813, -0.4262204488438936, 1.5508073295811842]
Σ: [1.8911987804702053 0.038814771826912765 1.2907448948413436; 0.038814771826912765 0.7054789545738211 0.06748072828787553; 1.2907448948413436 0.06748072828787553 0.9560938804684098]
)

)

julia> d = MvLogitNormal{MvNormalCanon}(randn(3), exp(Symmetric(randn(3, 3))))
MvLogitNormal{FullNormalCanon}(
μ: [-12.008515816447346, 9.221467028160104, 0.9303379007669893]
h: [-1.3761114231540177, 2.320365757512754, -0.05647292940406612]
J: [0.7714588053672484 0.9492070250024665 -0.929895944697952; 0.9492070250024665 1.6622992471750913 -1.7304512340037166; -0.929895944697952 -1.7304512340037166 5.088641347774431]
)


julia> invoke(Base.show, Tuple{IO,Distribution}, stdout, d)
MvLogitNormal{FullNormalCanon}(
normal: FullNormalCanon(
μ: [-12.008515816447346, 9.221467028160104, 0.9303379007669893]
h: [-1.3761114231540177, 2.320365757512754, -0.05647292940406612]
J: [0.7714588053672484 0.9492070250024665 -0.929895944697952; 0.9492070250024665 1.6622992471750913 -1.7304512340037166; -0.929895944697952 -1.7304512340037166 5.088641347774431]
)

)

```

We could use `show_multiline`, but this requires knowledge of the names of the parameters of the `AbstractMvNormal`, which we don't know, since Distributions doesn't have an API function for parameter names, just values.

src/multivariate/mvlogitnormal.jl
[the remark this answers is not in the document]
ParadaCarleton: It's mostly because I think the other change I recommended conveys the same information--"is a multivariate generalization of [`LogitNormal`](@ref) capable of handling correlations between variables"--but without the possibility of confusing users into thinking this is related to the [generalized Dirichlet distribution](https://www.sciencedirect.com/science/article/pii/S0096300397101400), or into thinking this shares some of the key properties of the Dirichlet distribution (like neutrality).

src/multivariate/mvlogitnormal.jl
```diff
@@ -0,0 +1,150 @@
...
+- `normal::AbstractMvNormal`: contains the ``d-1``-dimensional distribution of ``y``
+"""
+struct MvLogitNormal{D<:AbstractMvNormal} <: ContinuousMultivariateDistribution
+    normal::D
+    MvLogitNormal{D}(normal::D) where {D<:AbstractMvNormal} = new{D}(normal)
+end
+function MvLogitNormal{D}(args...) where {D<:AbstractMvNormal}
+    normal = D(args...)
+    return MvLogitNormal{typeof(normal)}(normal)
+end
```
devmotion, line 35: I'm not sure if this method is a good idea. Maybe at least initially it could be omitted?
```suggestion
```
My main concern is that it allows users to input all kinds of subtypes of `AbstractMvNormal`, possibly also with type parameters, and it is not clear in which cases these will work and when the desired type is actually respected (similar to the confusions/inconsistencies of the type argument in the `fit` methods). To me the alternative `MvLogitNormal(::AbstractMvNormal)` seems to be not susceptible to such issues, and the type version is not even shorter.

src/multivariate/mvlogitnormal.jl
```diff
@@ -0,0 +1,142 @@
...
+    LinearAlgebra.normalize!(x, 1)
+    return x
+end
+function _softmax1!(x::AbstractMatrix, y::AbstractMatrix)
+    map(_softmax1!, eachcol(x), eachcol(y))
+    return x
+end
+
+_drop1(x::AbstractVector) = @views x[firstindex(x, 1):(end - 1)]
+_drop1(x::AbstractMatrix) = @views x[firstindex(x, 1):(end - 1), :]
```
devmotion, line 127: I assume you did not use `begin` because it's not available in the oldest supported Julia version?

test/multivariate/mvlogitnormal.jl
```diff
@@ -0,0 +1,163 @@
...
+
+function test_mvlogitnormal(d::MvLogitNormal; nsamples::Int=10^6)
+    @test d.normal isa AbstractMvNormal
+    dnorm = d.normal
+
+    @testset "properties" begin
+        @test length(d) == length(dnorm) + 1
+        @test params(d) == params(dnorm)
+        @test partype(d) == partype(dnorm)
+        @test eltype(d) == eltype(dnorm)
```
devmotion, line 18: With the suggestion above, we should test both the type and instance based definition:
```suggestion
        @test eltype(d) == eltype(dnorm)
        @test eltype(typeof(d)) == eltype(typeof(dnorm))
```

test/multivariate/mvlogitnormal.jl
```diff
@@ -0,0 +1,163 @@
...
+    @testset "conversions" begin
+        @test convert(typeof(d), d) === d
+        T = partype(d) <: Float64 ? Float32 : Float64
+        if dnorm isa MvNormal
+            @test convert(MvLogitNormal{MvNormal{T}}, d).normal ==
+                convert(MvNormal{T}, dnorm)
+            @test partype(convert(MvLogitNormal{MvNormal{T}}, d)) <: T
+            @test convert(MvLogitNormal{MvNormal{T}}, params(d)...).normal ==
+                convert(MvNormal{T}, dnorm)
+            @test partype(convert(MvLogitNormal{MvNormal{T}}, params(d)...)) <: T
```
devmotion, line 39:
```suggestion
```

test/multivariate/mvlogitnormal.jl
```diff
@@ -0,0 +1,163 @@
...
+            @test canonform(d) isa MvLogitNormal{<:MvNormalCanon}
+            @test canonform(d).normal == canonform(dnorm)
+        elseif dnorm isa MvNormalCanon
+            @test convert(MvLogitNormal{MvNormalCanon{T}}, d).normal ==
+                convert(MvNormalCanon{T}, dnorm)
+            @test partype(convert(MvLogitNormal{MvNormalCanon{T}}, d)) <: T
+            @test convert(MvLogitNormal{MvNormalCanon{T}}, params(d)...).normal ==
+                convert(MvNormalCanon{T}, dnorm)
+            @test partype(convert(MvLogitNormal{MvNormalCanon{T}}, params(d)...)) <: T
```
devmotion, line 48:
```suggestion
```

test/multivariate/mvlogitnormal.jl
```diff
@@ -0,0 +1,163 @@
...
+        end
+    end
+
+    @testset "sampling" begin
+        X = rand(d, nsamples)
+        Y = @views log.(X[1:(end - 1), :]) .- log.(X[end, :]')
+        Ymean = vec(mean(Y; dims=2))
+        Ycov = cov(Y; dims=2)
+        for i in 1:(length(d) - 1)
+            @test isapprox(
```
devmotion, line 60: I'm not completely sure anymore but I assume we don't have any statistical test utilities for e.g. hypothesis tests of samples? IIRC KS tests are used for the univariate distributions but I assume we don't have anything for multi- or matrixvariate distributions.

devmotion (commented):
Looks very good, thank you! I only had a few minor comments.

test/multivariate/mvlogitnormal.jl
sethaxen: This is the closest I can find, which just checks that a 1x1 matrix distribution is consistent with the univariate distribution it reduces to:
https://github.com/JuliaStats/Distributions.jl/blob/b21e515dd4684d2d0f2c706c573ed6eb98ce537f/test/matrixvariates.jl#L193-L199

src/multivariate/mvlogitnormal.jl
sethaxen: Yes that's right. This isn't available in v1.3

test/multivariate/mvlogitnormal.jl
devmotion: Thank you for checking! Would be useful to have something similar for multivariate distributions (generally, IMO we should provide test utilities such that you can check more easily that you implement the interface correctly) but better to do this in a separate PR.

devmotion (approved):

PR #1780, opened 2023-09-19
add loglogistic distribution

yunzli:
This PR adds [log-logistic](https://en.wikipedia.org/wiki/Log-logistic_distribution) distribution to the package. 

The log-logistic distribution has two parameters: scale and shape. It has been widely used in survival analysis.

The model implementation is in `src/univariate/continuous/loglogistic.jl`
The test is in `test/univariate/continuous/loglogistic.jl`


ParadaCarleton:
This looks like a pretty solid implementation, but I do wonder if we should be adding Log* distributions piecemeal like this, rather than providing a more general interface for transformed variables.

ParadaCarleton (changes requested):

src/univariate/continuous/loglogistic.jl
```diff
@@ -0,0 +1,142 @@
...
+        z = log(zero(x))
+    else
+        z = logpdf(Logistic(log(d.θ), 1/d.ϕ), log(x)) + log(x)
+    end
+    return z 
+end
+
+function cdf(d::LogLogistic, x::Real)
+    if x <= 0
+        return 0.0
```
devmotion, line 106: Same here (as a general comment, usually you should avoid Float64 literals).

src/univariate/continuous/loglogistic.jl
```diff
@@ -0,0 +1,142 @@
...
+function mode(d::LogLogistic)
+	if d.ϕ ≤ 1
+		error("mode is defined only when ϕ > 1")
+	end 
+	return d.θ*((d.ϕ-1)/(d.ϕ+1))^(1/d.ϕ)
+end
+
+function var(d::LogLogistic)
+	if d.ϕ ≤ 2
+		erros("var is defined only when ϕ > 2")
```
devmotion, line 74: Same here.

src/univariate/continuous/loglogistic.jl
```diff
@@ -0,0 +1,142 @@
...
+    if x <= 0
+        return 0.0
+    end 
+    z = cdf(Logistic(log(d.θ), 1/d.ϕ), log(x)) 
+    return z
+end
+
+function logcdf(d::LogLogistic, x::Real)
+    if x <= 0
+        -Inf
```
devmotion, line 114: I guess it was supposed to be
```julia
        return -Inf
```
But that would again create a type instability.

src/univariate/continuous/loglogistic.jl
```diff
@@ -0,0 +1,142 @@
...
+    end
+    z = logcdf(Logistic(log(d.θ), 1/d.ϕ), log(x)) 
+	return z
+end
+
+function ccdf(d::LogLogistic, x::Real)
+    if x <= 0
+        return 1
```
devmotion, line 122: Same here.

src/univariate/continuous/loglogistic.jl
```diff
@@ -0,0 +1,142 @@
...
+    end 
+    z = ccdf(Logistic(log(d.θ), 1/d.ϕ), log(x)) 
+	return z 
+end
+
+function logccdf(d::LogLogistic, x::Real)
+    if x <= 0
+        return 0.0
```
devmotion, line 130: Same here.

devmotion (commented):
Thank you for the PR!

I haven't checked the math yet but commented on a few Julia issues that we should fix before merging the PR.

It's good that you added tests! We already have quite extensive automatic tests for distributions but that requires that you add reference values using R to test/refs. Can you fix that?

devmotion:
Fixed by #2008

PR #1786, opened 2023-10-03
Create a compact `show` for `MvNormal`.

Vilin97:
```
julia> using LinearAlgebra, Distributions

julia> dist = MvNormal(I(2)) # plaintext unchanged
MvNormal{Bool, PDMats.PDiagMat{Bool, Vector{Bool}}, FillArrays.Falses{1, Tuple{Base.OneTo{Int64}}}}(
dim: 2
μ: Zeros{Bool}(2)
Σ: Bool[1 0; 0 1]
)

julia> "$dist" # added compact representation
"MvNormal(Zeros{Bool}(2), Bool[1 0; 0 1])"
```

### DIFF
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -245,8 +245,10 @@ distrname(d::ZeroMeanIsoNormal) = "ZeroMeanIsoNormal"
 distrname(d::ZeroMeanDiagNormal) = "ZeroMeanDiagNormal"
 distrname(d::ZeroMeanFullNormal) = "ZeroMeanFullNormal"
 
-Base.show(io::IO, d::MvNormal) =
+Base.show(io::IO, ::MIME"text/plain", d::MvNormal) =
     show_multline(io, d, [(:dim, length(d)), (:μ, mean(d)), (:Σ, cov(d))])
+Base.show(io::IO, d::MvNormal) =
+    print(io, "MvNormal($(d.μ), $(d.Σ))")
 
 ### Basic statistics
 


Vilin97:
@devmotion , I'm not sure who is the right person to ping, sorry. 
Let me know if this is a kind of change that you would be open to. This closes my issue https://github.com/JuliaStats/Distributions.jl/issues/1785.

src/multivariate/mvnormal.jl
```diff
@@ -245,8 +245,10 @@ distrname(d::ZeroMeanIsoNormal) = "ZeroMeanIsoNormal"
 distrname(d::ZeroMeanDiagNormal) = "ZeroMeanDiagNormal"
 distrname(d::ZeroMeanFullNormal) = "ZeroMeanFullNormal"
 
-Base.show(io::IO, d::MvNormal) =
+Base.show(io::IO, ::MIME"text/plain", d::MvNormal) =
     show_multline(io, d, [(:dim, length(d)), (:μ, mean(d)), (:Σ, cov(d))])
+Base.show(io::IO, d::MvNormal) =
+    print(io, "MvNormal($(d.μ), $(d.Σ))")
```
devmotion, line 251: There's already a `show_oneline` implementation in Distributions for compact representation. IMO it should actually be the default for `show(io, dist)`, and only `show(io, ::MIME"text/plain", dist)` should use the current path which returns a one- or multiline representation depending on the number of parameters.

So I wonder what the implications of changing https://github.com/JuliaStats/Distributions.jl/blob/e407fa5fd098e50df51801c6d062946eac7a7d0f/src/show.jl#L12 to
```julia
show(io::IO, d::Distribution) = show_oneline(io, d, fieldnames(typeof(d)))
show(io::IO, ::MIME"text/plain", d::Distribution) = show(io, d, fieldnames(typeof(d)))
```
would be.

Vilin97: `show_oneline` looks messy. Compare the two:
```
julia> "$dist" # show_oneline
"MvNormal{Bool, PDMats.PDiagMat{Bool, Vector{Bool}}, FillArrays.Falses{1, Tuple{Base.OneTo{Int64}}}}(dim=2, μ=Zeros{Bool}(2), Σ=Bool[1 0; 0 1])"

julia> "$dist" # the new show method
"MvNormal(Zeros{Bool}(2), Bool[1 0; 0 1])"
```

devmotion: Yeah, maybe the type parameters should be omitted in `show_oneline`. Additionally, I don't think you'd want to add the `dim` output to `MvNormal`, the vanilla `show_oneline` with the fields should be sufficient.

PR #1791, opened 2023-10-20
Move test utilities to an extension

devmotion:
This PR moves the test utilities to an extension to make Test an optional dependency on Julia >= 1.9, in the same way as done e.g. for AbstractFFTs.


devmotion:
The test failures on Julia nightly are the same as on the master branch.

Issue #1783, opened 2023-09-27
Inconsistent Type Behavior for Uniform Distribution


arthur-bizzi:
Hey all.
Not only do Distributions still behave differently with regards to types when generating one or multiple samples ( #1252 ), but the behavior is also inconsistent for different types:

```
using Random, Distributions 

UBig = Uniform(BigFloat(0),BigFloat(1)) #Uniform{BigFloat}(a=0.0, b=1.0)
rand(UBig) #BigFloat
rand(UBig,2) #Vector{Float64}

U32 = Uniform(Float32(0),Float32(1)) #Uniform{Float32}(a=0.0f0, b=1.0f0)
rand(U32) #Float64
rand(U32,2) #Vector{Float64}
```
This follows from type promotion rules. An attempted solution would be something along the lines of:
```
import Base.rand
rand(rng::AbstractRNG, d::Uniform{T}) where T<:Real = d.a + (d.b - d.a) * rand(rng,T)
```
But the type of single numbers would still differ from that of arrays:
```
rand(U32) #Float32
rand(U32,2) #Vector{Float64}
```
On the other hand, `rand!` seems to ignore the type parameter completely:
```
v = rand(Float32,4)
rand!(UBig,v) #Float32
```
Is there a mental model for this behavior? If so, wouldn't it be a good idea to include it in the docs? They currently contain no references whatsoever to type precision.

devmotion:
The main reason for the current behaviour is rooted in the historical development process. Initially, only Float64 was supported and was hardcoded in many places. Over time these restrictions were lifted in some places but thereby consistency was lost. There is no clear mental model, unfortunately, but I think currently you should always expect samples of contonuous distributions to be of type `Float64` - when promotion happens, it's an unintended consequence of generalizing types.

There is a strong incentive though to fix these issues soonish. Therefore I think it's not worth documenting the current buggy behaviour.

arthur-bizzi:
I see, thank you very much. 
Can you think of any relatively quick fixes I could implement on my end until these changes are merged? 
More specifically: could I get the fix I mentioned above working for vectors? At least for basic univariate distributions.

jondeuce:
[...]

PR #1796, opened 2023-11-06
Computed field len and rand now returns type T in Uniform{T}

f-ij:
Added computed field len which is the length of the interval to be sampled from. Also replaced all instances of (d.b-d.a) by the new field d.len.

Fixed an issue where rand(::Uniform{Float32}) (and other, non Float64 types) did not in fact return a Float32, but a Float64, which caused some performance issues. This was done by modifiying the method rand(rng::AbstractRNG, d::Uniform) to include the subtype of Uniform and pass it to rand.

This was discussed on [discourse](https://discourse.julialang.org/t/performance-of-generating-uniform-float32-numbers/105863)

### DIFF
--- a/src/univariate/continuous/uniform.jl
+++ b/src/univariate/continuous/uniform.jl
@@ -26,7 +26,8 @@ External links
 struct Uniform{T<:Real} <: ContinuousUnivariateDistribution
     a::T
     b::T
-    Uniform{T}(a::Real, b::Real) where {T <: Real} = new{T}(a, b)
+    len::T
+    Uniform{T}(a::Real, b::Real) where {T <: Real} = new{T}(a, b, b-a)
 end
 
 function Uniform(a::T, b::T; check_args::Bool=true) where {T <: Real}
@@ -47,11 +48,11 @@ Base.convert(::Type{Uniform{T}}, d::Uniform{T}) where {T<:Real} = d
 
 #### Parameters
 
-params(d::Uniform) = (d.a, d.b)
+params(d::Uniform) = (d.a, d.b, d.len)
 partype(::Uniform{T}) where {T<:Real} = T
 
 location(d::Uniform) = d.a
-scale(d::Uniform) = d.b - d.a
+scale(d::Uniform) = d.len
 
 
 #### Statistics
@@ -61,12 +62,12 @@ median(d::Uniform) = mean(d)
 mode(d::Uniform) = mean(d)
 modes(d::Uniform) = Float64[]
 
-var(d::Uniform) = (w = d.b - d.a; w^2 / 12)
+var(d::Uniform) = (w = d.len; w^2 / 12)
 
 skewness(d::Uniform{T}) where {T<:Real} = zero(T)
 kurtosis(d::Uniform{T}) where {T<:Real} = -6/5*one(T)
 
-entropy(d::Uniform) = log(d.b - d.a)
+entropy(d::Uniform) = log(d.len)
 
 
 #### Evaluation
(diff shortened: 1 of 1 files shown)


devmotion:
Regarding `len`: IMO we should not add such a field since it is not an actual parameter, could easily cause inconsistencies and I think your example should be fixed differently. As in Random, you should construct a sampler when you perform repeated sampling (unfortunately, the syntax is a bit different in Distributions, possibly because it predates (?) the interface in Random). `len` should only be computed when constructing such a sampler.

src/univariate/continuous/uniform.jl
```diff
@@ -151,7 +152,7 @@ Base.:*(c::Real, d::Uniform) = Uniform(minmax(c * d.a, c * d.b)...)
 
 #### Sampling
 
-rand(rng::AbstractRNG, d::Uniform) = d.a + (d.b - d.a) * rand(rng)
+rand(rng::AbstractRNG, d::Uniform{T}) where T = d.a + d.len * rand(T, rng)
```
stevengj, line 155:
```suggestion
rand(rng::AbstractRNG, d::Uniform{T}) where T = d.a + d.len * rand(float(T), rng)
```
You need to promote `T` to a floating-point type.

stevengj:
In particular, this corresponds to issue #1783.

devmotion:
With the current design, the "correct" fix would be to always return Float64 as the variates are supposed to be of type Float64. But as said, I'd rather fix this design. I'll try to write up our suggestions and draft a PR within the next few days.

f-ij:
> The `rand` method follows the current design, currently almost all distributions return samples of type `Float64` and `Int` and intentionally the random variate type is independent of the parameter types (there are many issues and discussions in the repo). I would like to improve this design, and I've discussed a better design with some other contributors, but this requires more and different changes. We should not introduce an additional inconsistency by making `rand` of `Uniform` special.

I haven't used Distributions extensively, but currently calling rand on Uniform{Float32} does indeed return a Float32. Regarding your comment on the len field, I don't see why this would be a problem. For most design patterns in Julia we may assume a user won't directly change any struct fields, right? If so then inconsistencies shouldn't be a problem.

src/univariate/continuous/uniform.jl
f-ij: Thanks, I missed that one.

stevengj:
> currently calling rand on Uniform{Float32} does also return a Float32

@f-ij, no, it returns a `Float64` currently:
```jl
julia> u = Uniform(1.0f0, 2.0f0)
Uniform{Float32}(a=1.0f0, b=2.0f0)

julia> rand(u)
1.8944514300368434

julia> typeof(ans)
Float64
```

f-ij:
> > currently calling rand on Uniform{Float32} does also return a Float32
> 
> @f-ij, was this a typo? It returns a `Float64` currently:
> 
> ```julia
> julia> u = Uniform(1.0f0, 2.0f0)
> Uniform{Float32}(a=1.0f0, b=2.0f0)
> 
> julia> rand(u)
> 1.8944514300368434
> 
> julia> typeof(ans)
> Float64
> ```

Sorry that was indeed a typo! I meant Normal{Float32}..

devmotion:
`Normal` is the inconsistent one that does not follow the design. `eltype` is supposed to specify the element type of the random variates, `partype` is the type of the parameters. By design, they are intended to be distinct and `eltype` is `Float64` for continuous distributions and `Int` for discrete distributions by default.

Changing `rand` for `Uniform` as suggested in this PR is not a fix since it just introduces another inconsistent distribution.

There are clear ideas for a better design and a push for making these changes.

simsurace:
What is the issue/PR to track/contribute to? Sorry @f-ij for sending you in the wrong direction. I was under the false impression that this would be uncontroversial.

devmotion:
There are two different aspects: The `len` precomputation for repeated calls of `rand` should be addressed separately from the general `rand` interface design question.

IMO the most natural way to address the first one is to create a `Distributions.Sampler` type for `Uniform` that can be used for repeated sampling, since the same approach is also used in Random: If you want to perform repeated sampling with the same object, you use
```julia
spl = sampler(dist)
x = rand(spl)
for _ in 1:1_000_000
    x += rand(spl)
end
```
instead of
```julia
x = rand(dist)
for _ in 1:1_000_000
    x += rand(dist)
end
```
(FYI these samplers are already used internally when you create an array of samples; if there exists no sampler for a distribution, `sampler(dist)` just falls back to `dist`, ie. no special pre-computations are performed.) AFAIK there exists no issue or other PR for this possible performance improvement yet.

The design question is discussed in multiple issues and even PRs (#1433), but it's a bit scattered since multiple duplicate issues have been opened and discussions started from scratch or recircled to already existing explanations.

simsurace:
I did not know about `sampler` yet, but that's a very good point.

f-ij:
> What is the issue/PR to track/contribute to? Sorry @f-ij for sending you in the wrong direction. I was under the false impression that this would be uncontroversial.

Not a problem at all, it was not much work and I gained me some insight out of it.

PR #1797, opened 2023-11-06
Fix type unstable `Uniform`

prbzrg:
this fixes:
```julia
julia> dist = Uniform{Float32}(-0.1f0, 0.1f0)
Uniform{Float32}(a=-0.1f0, b=0.1f0)

julia> typeof(rand(dist))
Float64
```


devmotion:
See https://github.com/JuliaStats/Distributions.jl/pull/1796 and other discussions in the repo for why this is inconsistent with the current design of Distributions.

Issue #1762, opened 2023-09-02
Adding a NamedTupleVariate


sethaxen:
It could be useful to add a `NamedTupleVariate` with necessary defaults to this package. A concrete use case is that one may want a product distribution with easy access to the individual components.

Here's a barebones implementation:

```julia
using Distributions, Random

abstract type NamedTupleVariate <: VariateForm end

struct NamedTupleProductDistribution{Tnames,Tdists,eltypes,S<:ValueSupport} <:
       Distribution{NamedTupleVariate,S}
    dists::NamedTuple{Tnames,Tdists}
end
function NamedTupleProductDistribution(
    dists::NamedTuple{K,V}
) where {K,V<:Tuple{Vararg{Distribution}}}
    eltypes = Tuple{map(eltype, values(dists))...}
    # would be better to allow mixed ValueSupports here
    vs = Distributions._product_valuesupport(dists)
    return NamedTupleProductDistribution{K,V,eltypes,vs}(dists)
end

function Distributions.product_distribution(
    dists::NamedTuple{K,V}
) where {K,V<:Tuple{Vararg{Distribution}}}
    return NamedTupleProductDistribution(dists)
end

function Distributions.eltype(::NamedTupleProductDistribution{K,<:Any,V}) where {K,V}
    return NamedTuple{K,V}
end

function Distributions.insupport(
    dist::NamedTupleProductDistribution{K}, x::NamedTuple{K}
) where {K}
    return all(Base.splat(insupport), zip(dist.dists, x))
end

function Distributions.pdf(
    dist::NamedTupleProductDistribution{K}, x::NamedTuple{K}
) where {K}
    return exp(logpdf(dist, x))
end
function Distributions.logpdf(
    dist::NamedTupleProductDistribution{K}, x::NamedTuple{K}
) where {K}
    return mapreduce(logpdf, +, dist.dists, x)
end

function Distributions.rand(
    rng::AbstractRNG, dist::NamedTupleProductDistribution{K}
) where {K}
    return NamedTuple{K}(map(Base.Fix1(rand, rng), dist.dists))
end
```

and example usage:

```julia
julia> dist = product_distribution((; x=Normal(), y=Dirichlet(3, 1)));

julia> x = rand(dist)
(x = 0.35877545055901744, y = [0.5772411031132897, 0.2856859832919011, 0.13707291359480941])

julia> insupport(dist, x)
true

julia> pdf(dist, x)
0.7481503903444227
```

[...]

PR #1803, opened 2023-11-23
Add NamedTupleVariate and ProductNamedTupleDistribution

sethaxen:
Implements #1762


sethaxen:
One challenge is that `eltype` for a distribution `d` doesn't tell us what type `rand(d)` will return. The worst case here is `LKJCholesky`, for which the eltype is `Float64`, but `rand` returns a `Cholesky{Float64}`. So without calling `rand` recursively on all distributions once, we cannot know the exact types a call to `rand(::ProductNamedTupleDistribution)` should return.

devmotion:
> One challenge is that eltype for a distribution d doesn't tell us what type rand(d) will return. The worst case here is LKJCholesky, for which the eltype is Float64, but rand returns a Cholesky{Float64}. So without calling rand recursively on all distributions once, we cannot know the exact types a call to rand(::ProductNamedTupleDistribution) should return.

IMO `eltype` is a design flaw and should be removed completely. So I don't think it should block the PR.

sethaxen:
> IMO `eltype` is a design flaw and should be removed completely. So I don't think it should block the PR.

Without `eltype`, how would you recommend container eltype be determined for `rand` calls such as `rand(d, 2, 3)`? So far I've assumed this is what `eltype` should return, but then of course there's `LKJCholesky`, which returns `Float64` when the container eltype is `Cholesky{Float64}`.

devmotion:
No, that's not what `eltype` is supposed to give you currently. It is intended to return the element type of a single variate (e.g., `Float64` for a `MvNormal` whose variates are `Vector{Float64}`). But IMO there are at least three major problems: 1) It's implemented inconsistently and everyone expects something different; 2) It's a misnomer (IMO it was a suboptimal decision in retrospect to re-use `eltype` in Random; the less known `randtype` seems much clearer); 3) The concept breaks down if you try to allow variates of dynamic types (depending on the return type of the `rand(rng)` calls and/or the type of the parameters). In my experience any heuristic for addressing 3) is doomed to fail in some cases, so my suggestion would be to call `rand` if you want to know the type of its return value (and if you're adventurous you could try to infer it with `return_type` but IMO that's too brittle and internal for a default definition in Distributions).

sethaxen:
> the less known `randtype` seems much clearer

In which package is `randtype` defined? I have not heard of it.

> It is intended to return the element type of a single variate (e.g., `Float64` for a `MvNormal` whose variates are `Vector{Float64}`).

Ah, okay, then this PR currently implements it incorrectly. Will fix.

> my suggestion would be to call `rand` if you want to know the type of its return value

Cool, this is the approach I took, so will leave as is.

devmotion:
> In which package is randtype defined? I have not heard of it.

Sorry, I misremembered, it's called `gentype`. It's defined in Random, it also appeared in some discussion in some issue in Distributions, but IIRC there were PRs that tried to remove it, so not everyone shares my opinion 😄.

sethaxen:
I added a new API function `marginal`. While we have `component` for mixture distributions, we didn't have any API function for accessing marginals of a product distribution. And there are other cases where we know the marginals and users might want them (e.g. slices of `MvNormal`). It's an optional addition, and I'm fine with removing if maintainers think this should be in its own PR.

RE docs, currently `product_distribution` is documented in the multivariate distributions page, but with this PR, sometimes `product_distribution` will return a different variate form, so I think it makes sense to move all docs related to product distributions to their own page. `marginal` (if kept) could be documented here as well, though it's more general than that.

sethaxen:
@devmotion this is ready for a review

Red-Portal:
Hi all, would it be possible to get this moving? I think it would be really great to have this feature!

sethaxen:
This is really just waiting on a review. @devmotion?

Red-Portal:
Hi @devmotion, could we make this happen?

src/common.jl
```diff
@@ -16,6 +16,12 @@ const Univariate    = ArrayLikeVariate{0}
 const Multivariate  = ArrayLikeVariate{1}
 const Matrixvariate = ArrayLikeVariate{2}
 
+"""
+`F <: NamedTupleVariate{K}` specifies that the variate or a sample is of type
+`NamedTuple{K}`.
+"""
+abstract type NamedTupleVariate{K} <: VariateForm end
```
devmotion, line 23: Couldn't/Shouldn't this be a concrete subtype?
```suggestion
struct NamedTupleVariate{K} <: VariateForm end
```

src/common.jl
```diff
@@ -464,6 +470,16 @@ Base.@propagate_inbounds function loglikelihood(
...
+"""
+    marginal(d::Distribution, k...) -> Distribution
+
+Return the marginal distribution of `d` at the indices `k...`.
+
+The result is the distribution of the variate `rand(d)[k...]` that one would obtain by
+integrating over all other indices.
+"""
+marginal(d::Distribution, k...)
+
```
devmotion, line 482: Maybe this should go into a separate PR? It seems this is a bit orthogonal to NamedTupleVariate.

src/namedtuple/productnamedtuple.jl
```diff
@@ -0,0 +1,131 @@
...
+    independent distributions instead of constructing a `ProductNamedTupleDistribution`
+    directly.
+"""
+struct ProductNamedTupleDistribution{Tnames,Tdists,S<:ValueSupport,eltypes} <:
+       Distribution{NamedTupleVariate{Tnames},S}
+    dists::NamedTuple{Tnames,Tdists}
+end
+function ProductNamedTupleDistribution(
+    dists::NamedTuple{K,V}
+) where {K,V<:Tuple{Vararg{Distribution}}}
```
devmotion, line 18:
```suggestion
) where {K,V<:Tuple{Distribution,Vararg{Distribution}}}
```
to ensure that there is at least one distribution

src/namedtuple/productnamedtuple.jl
```diff
@@ -0,0 +1,131 @@
...
+function Base.rand(rng::AbstractRNG, d::ProductNamedTupleDistribution, dims::Dims)
+    xs = return map(CartesianIndices(dims)) do _
+        return rand(rng, d)
+    end
+    return xs
+end
+
+function _rand!(rng::AbstractRNG, d::ProductNamedTupleDistribution, xs::AbstractArray)
+    for i in eachindex(xs)
+        xs[i] = Random.rand(rng, d)
```
devmotion, line 128:
```suggestion
        xs[i] = rand(rng, d)
```

src/namedtuple/productnamedtuple.jl
```diff
@@ -0,0 +1,131 @@
...
+
+# Sampling
+
+function Base.rand(rng::AbstractRNG, d::ProductNamedTupleDistribution{K}) where {K}
+    return NamedTuple{K}(map(Base.Fix1(rand, rng), d.dists))
+end
+function Base.rand(rng::AbstractRNG, d::ProductNamedTupleDistribution, dims::Dims)
+    xs = return map(CartesianIndices(dims)) do _
+        return rand(rng, d)
+    end
```
devmotion, line 122: I think for efficiency reasons you might want to use something like
```suggestion
sampler(d::ProductNamedTupleDistribution{K}) where K = ProducNamedTupleSampler{K}(map(sampler, d.dists))
function Base.rand(rng::AbstractRNG, d::ProductNamedTupleDistribution, dims::Dims)
    spl = sampler(d)
    xs = return map(CartesianIndices(dims)) do _
        return rand(rng, spl)
    end
```
Since `sampler` can be type unstable, probably it would be good to introduce a function barrier though.

src/namedtuple/productnamedtuple.jl
```diff
@@ -0,0 +1,131 @@
...
+    return xs
+end
+
+function _rand!(rng::AbstractRNG, d::ProductNamedTupleDistribution, xs::AbstractArray)
```
devmotion, line 126: Also here, probably you should use `sampler` + a function barrier.

src/common.jl
sethaxen: Perhaps, but this would be inconsistent with other `VariateForm` subtypes. Both `ArrayLikeVariate` and `CholeskyVariate` could be concrete subtypes but are currently defined as abstract subtypes.

sethaxen:
Bump @devmotion

devmotion:
> what do you think about this proposed docs restructure?

I think it's reasonable to move it.

src/common.jl
devmotion: I see. I'm not sure there's a good reason for the current implementation but maybe let's be consistent for now then.

sethaxen:
@devmotion I restructured the docs. I also noticed that, while `Product` is deprecated, its docstring was missing a deprecation warning, so I added it. I also figured the behavior of `mean(::ProductNamedTuple)`, `mode`, `var`, etc may not be obvious, so I added a doctested example.

sethaxen:
Bump @devmotion

src/common.jl
devmotion: I thought a bit more about this and I think the existing definitions might just be a bug - I've never encountered a subtype of these variate forms, and I wouldn't expect anyone to define a subtype for NamedTupleVariate either.

I also wonder if abstract type parameters might actually be worse for the compiler.

So I would suggest making it a concrete type.

src/namedtuple/productnamedtuple.jl
```diff
@@ -0,0 +1,152 @@
...
+var(d::ProductNamedTupleDistribution) = map(var, d.dists)
+
+std(d::ProductNamedTupleDistribution) = map(std, d.dists)
+
+entropy(d::ProductNamedTupleDistribution) = sum(entropy, values(d.dists))
+
+function kldivergence(
+    d1::ProductNamedTupleDistribution{K}, d2::ProductNamedTupleDistribution{K}
+) where {K}
+    return mapreduce(kldivergence, +, d1.dists, d2.dists)
```
devmotion, line 130: Maybe be consistent with the definitions above and use

```suggestion
    return sum(map(kldivergence, d1.dists, d2.dists))
```
?

src/namedtuple/productnamedtuple.jl
```diff
@@ -0,0 +1,152 @@
...
+function _gentype(d::Distribution{CholeskyVariate})
+    T = eltype(d)
+    return LinearAlgebra.Cholesky{T,Matrix{T}}
+end
+_gentype(::Distribution) = Any
+
+_product_namedtuple_eltype(dists) = typejoin(map(_gentype, dists)...)
+
+function Base.show(io::IO, d::ProductNamedTupleDistribution)
+    return show_multline(io, d, collect(pairs(d.dists)))
```
devmotion, line 61: Is this the standard definition in Distributions? I though `show` with two arguments should print a short one-line representation and only the three argument version should be more verbose.

src/namedtuple/productnamedtuple.jl
```diff
@@ -0,0 +1,152 @@
...
+
+_gentype(d::UnivariateDistribution) = eltype(d)
+_gentype(d::Distribution{<:ArrayLikeVariate{S}}) where {S} = Array{eltype(d),S}
+function _gentype(d::Distribution{CholeskyVariate})
+    T = eltype(d)
+    return LinearAlgebra.Cholesky{T,Matrix{T}}
+end
+_gentype(::Distribution) = Any
+
+_product_namedtuple_eltype(dists) = typejoin(map(_gentype, dists)...)
```
devmotion, line 58: This is a bit inconsistent with `ProductDistribution` which uses `promote_type`, not `typejoin`. Maybe similarly one could use something like

```suggestion
_product_namedtuple_eltype(dists::NamedTuple{K,V}) where {K,V} = __product_promote_type(eltype, V)
```
?

sethaxen: I think we want this inconsistency right?, since the eltype of the concatenation of 2 vectors uses promotion, while the eltype of the merging of 2 (named-)tuples uses a typejoin (ultimately calling https://github.com/JuliaLang/julia/blob/v1.10.5/base/tuple.jl#L228-L237). Shouldn't `eltype(::Distribution)` in general agree with `eltype(rand(::Distribution))`?

src/namedtuple/productnamedtuple.jl
sethaxen: There's only one 3-arg definition in Distributions (for `Censored`). Otherwise it's always 2-arg. And a bunch of distributions override the default `show` method to explicitly call `show_multline`: `Truncated`, `MvNormal`, `MvLogNormal`, `Censored`, and all matrix distributions.

Without this override, `show_multline` gets used anyways but with the only parameter being the field `dists`.

devmotion: I see, we don't want to touch this issue in this PR here 🙂

src/namedtuple/productnamedtuple.jl
devmotion: > Shouldn't eltype(::Distribution) in general agree with eltype(rand(::Distribution))?

I guess that's what distributions should satisfy currently (even though it's quite inconsistent). So I think you're right, we should use the same approach that is used for computing `eltype(::NamedTuple)`, ie. we should use `typejoin`.

sethaxen:
@devmotion alright, I've implemented the requested changes.

test/namedtuple/productnamedtuple.jl
```diff
@@ -0,0 +1,195 @@
...
+            z=DiscreteUniform(1, 10),
+            w=LKJCholesky(3, 2.0),
+        )
+        d = @inferred product_distribution(nt)
+        @test d === ProductNamedTupleDistribution(nt)
+    end
+
+    @testset "show" begin
+        d = ProductNamedTupleDistribution((x=Gamma(1.0, 2.0), y=Normal()))
+        @test sprint(show, d) == """
```
devmotion, line 46: FYI an alternative would be
```suggestion
        @test repr(d) == """
```

devmotion (approved):
Looks good to me, thank you 🙂

sethaxen:
@devmotion can you merge this and make a release?

src/namedtuple/productnamedtuple.jl
```diff
@@ -0,0 +1,155 @@
...
+ProductNamedTupleDistribution{(:x, :y)}(
+x: Normal{Float64}(μ=0.0, σ=1.0)
+y: Dirichlet{Int64, Vector{Int64}, Float64}(alpha=[2, 4])
+)
+
+
+julia> nt = rand(d)
+(x = 1.5155385995160346, y = [0.533531876438439, 0.466468123561561])
+
+julia> pdf(d, nt)
```
devmotion, line 25: Just today I was dealing with ordering of `NamedTuple` in TransformVariables transforms: https://github.com/tpapp/TransformVariables.jl/pull/120

So I wonder: Should we have
```julia
julia> pdf(d, (; x = 1.5155385995160346, y = [0.533531876438439, 0.466468123561561])) == pdf(d, (; y = [0.533531876438439, 0.466468123561561], x = 1.5155385995160346))
true
```
?

sethaxen: Good point, I modified the code to support permutations of the names.

devmotion (approved):

PR #1805, opened 2023-11-23
Fix inconsistent return types of `quantile` etc.

### LINKED ISSUE #1798: Inconsistent behaviour of `truncate` for discrete distributions
When truncating a `Geometric` distributions, `rand` will return different types depending on the bounds chosen. For example, values sampled from `truncated(Geometric(0.6); lower = 2)` are floats, while values sampled from `truncated(Geometric(0.6); upper = 4)` are integers.
I guess this is due to https://github.com/JuliaStats/Distributions.jl/blob/7e232ca86c56c4b2e1e7a44b16aa23c26668af3d/src/truncate.jl#L211-L229 using the float-returning `quantile` / `invlogcdf` in special cases.

devmotion:
Currently, the return type of `quantile` etc. is inconsistent, in particular for discrete distributions. For many distributions (and in particular those for which we fall back to StatsFuns/Rmath) the return type is explicitly to converted to `Int` to match the type of the variates. However, the behaviour of some distributions such as `Geometric` or `Bernoulli` is inconsistent: The return type of `quantile` does not match the type of the variates.

This PR fixes the inconsistency. Moreover, I added tests to avoid regressions and made an existing test a bit stricter. These tests revealed a few additional problems (e.g. caused by the Rmath behaviour that `StatsFuns.RFunctions.binominvcdf(10, 1.0, 0.0) = 0` and `StatsFuns.RFunctions.hyperinvlogccdf(2, 2, 2, -Inf) = NaN` which so far we have hidden by skipping them in the tests).

Fixes #1798.

### DIFF
--- a/src/univariate/discrete/negativebinomial.jl
+++ b/src/univariate/discrete/negativebinomial.jl
@@ -112,10 +112,30 @@ cdf(d::NegativeBinomial, x::Real) = nbinomcdf(d.r, d.p, x)
 ccdf(d::NegativeBinomial, x::Real) = nbinomccdf(d.r, d.p, x)
 logcdf(d::NegativeBinomial, x::Real) = nbinomlogcdf(d.r, d.p, x)
 logccdf(d::NegativeBinomial, x::Real) = nbinomlogccdf(d.r, d.p, x)
-quantile(d::NegativeBinomial, q::Real) = convert(Int, nbinominvcdf(d.r, d.p, q))
-cquantile(d::NegativeBinomial, q::Real) = convert(Int, nbinominvccdf(d.r, d.p, q))
-invlogcdf(d::NegativeBinomial, lq::Real) = convert(Int, nbinominvlogcdf(d.r, d.p, lq))
-invlogccdf(d::NegativeBinomial, lq::Real) = convert(Int, nbinominvlogccdf(d.r, d.p, lq))
+function quantile(d::NegativeBinomial, q::Real)::Int
+    _check_quantile_arg(q)
+    return nbinominvcdf(d.r, d.p, q)
+end
+function cquantile(d::NegativeBinomial, q::Real)::Int
+    _check_cquantile_arg(q)
+    return nbinominvccdf(d.r, d.p, q)
+end
+function invlogcdf(d::NegativeBinomial, lq::Real)::Int
+    _check_invlogcdf_arg(lq)
+    if isinf(lq)
+        return nbinominvcdf(d.r, d.p, zero(lq))
+    else
+        return nbinominvlogcdf(d.r, d.p, lq)
+    end
+end
+function invlogccdf(d::NegativeBinomial, lq::Real)::Int
+    _check_invlogccdf_arg(lq)
+    if isinf(lq)
+        return nbinominvccdf(d.r, d.p, zero(lq))
+    else
+        return nbinominvlogccdf(d.r, d.p, lq)
+    end
+end
 
 ## sampling
 function rand(rng::AbstractRNG, d::NegativeBinomial)
--- a/src/univariate/discrete/noncentralhypergeometric.jl
+++ b/src/univariate/discrete/noncentralhypergeometric.jl
@@ -9,23 +9,23 @@ abstract type NoncentralHypergeometric{T<:Real} <: DiscreteUnivariateDistributio
 
 # Functions
 
-function quantile(d::NoncentralHypergeometric{T}, q::Real) where T<:Real
-    if !(zero(q) <= q <= one(q))
-        T(NaN)
-    else
-        range = support(d)
-        if q > 1/2
-            q = 1 - q
-            range = reverse(range)
-        end
+function quantile(d::NoncentralHypergeometric, q::Real)
+    _check_quantile_arg(q)
 
-        qsum, i = zero(T), 0
-        while qsum < q
-            i += 1
-            qsum += pdf(d, range[i])
-        end
-        range[i]
+    range = support(d)
+    if q > 1/2
+        q = 1 - q
+        range = reverse(range)
     end
+
+    # Support must not be empty
+    r, state = iterate(range)
+    qsum = pdf(d, r)
+    while qsum < q && (r_state = iterate(range, state)) !== nothing
+        r, state = r_state
+        qsum += pdf(d, r)
+    end
+    return r
 end
 
 params(d::NoncentralHypergeometric) = (d.ns, d.nf, d.n, d.ω)
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -392,21 +392,30 @@ quantile(d::UnivariateDistribution, p::Real)
 
 The complementary quantile value, i.e. `quantile(d, 1-q)`.
 """
-cquantile(d::UnivariateDistribution, p::Real) = quantile(d, 1.0 - p)
+function cquantile(d::UnivariateDistribution, p::Real)
+    _check_cquantile_arg(p)
+    quantile(d, 1 - p)
+end
 
 """
     invlogcdf(d::UnivariateDistribution, lp::Real)
 
 The inverse function of logcdf.
 """
-invlogcdf(d::UnivariateDistribution, lp::Real) = quantile(d, exp(lp))
+function invlogcdf(d::UnivariateDistribution, lp::Real)
+    _check_invlogcdf_arg(lp)
+    quantile(d, exp(lp))
+end
 
 """
     invlogccdf(d::UnivariateDistribution, lp::Real)
 
 The inverse function of logccdf.
 """
-invlogccdf(d::UnivariateDistribution, lp::Real) = quantile(d, -expm1(lp))
+function invlogccdf(d::UnivariateDistribution, lp::Real)
+    _check_invlogccdf_arg(lp)
+    quantile(d, -expm1(lp))
+end
 
 # gradlogpdf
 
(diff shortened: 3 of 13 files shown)


src/univariates.jl
```diff
@@ -619,6 +628,38 @@ function integerunitrange_logccdf(d::DiscreteUnivariateDistribution, x::Integer)
...
 
+### Error messages
+
+const QUANTILE_ARG_ERROR = "`Distributions.quantile(d, p)` only accepts arguments `p` in `[0, 1]`." 
+const CQUANTILE_ARG_ERROR = "`Distributions.cquantile(d, p)` only accepts arguments `p` in `[0, 1]`." 
+const INVLOGCDF_ARG_ERROR = "`Distributions.invlogcdf(d, lp)` only accepts non-negative arguments `lp`." 
+const INVLOGCCDF_ARG_ERROR = "`Distributions.invlogccdf(d, lp)` only accepts non-negative arguments `lp`." 
+
+function _check_quantile_arg(p::Real)
+    if !(zero(p) <= p <= oneunit(p))
```
nalimilan, line 639: Does `oneunit` really make sense here? For `Real`s AFAIK it's the same as `one`. (I also suspect that for the compiler it's equivalent to `0 <= p <= 1` but I guess it's safer to use `zero` and `one`.)

src/univariates.jl
```diff
@@ -619,6 +628,38 @@ function integerunitrange_logccdf(d::DiscreteUnivariateDistribution, x::Integer)
     return result
 end
 
+### Error messages
+
+const QUANTILE_ARG_ERROR = "`Distributions.quantile(d, p)` only accepts arguments `p` in `[0, 1]`."
```
nalimilan, line 633: While you're at it, how about adding `(got $p)`? That's generally very useful for users.

Also, I just noticed that `?quantile` calls this argument `q` so it would be good to use the same name (maybe `p` for consistency with Statistics?).

src/univariates.jl
```diff
@@ -619,6 +628,38 @@ function integerunitrange_logccdf(d::DiscreteUnivariateDistribution, x::Integer)
     return result
 end
 
+### Error messages
+
+const QUANTILE_ARG_ERROR = "`Distributions.quantile(d, p)` only accepts arguments `p` in `[0, 1]`." 
+const CQUANTILE_ARG_ERROR = "`Distributions.cquantile(d, p)` only accepts arguments `p` in `[0, 1]`." 
+const INVLOGCDF_ARG_ERROR = "`Distributions.invlogcdf(d, lp)` only accepts non-negative arguments `lp`." 
+const INVLOGCCDF_ARG_ERROR = "`Distributions.invlogccdf(d, lp)` only accepts non-negative arguments `lp`."
```
nalimilan, line 636: "Non-negative" doesn't seem to reflect the check `!(lp <= zero(lp))`. Isn't that "strictly positive" instead?

BTW, for some distributions (e.g. `Geometric`) it was allowed to pass zero and now it throws an error. Is that intended?

src/univariate/discrete/negativebinomial.jl
```diff
@@ -112,10 +112,30 @@ cdf(d::NegativeBinomial, x::Real) = nbinomcdf(d.r, d.p, x)
...
+    _check_quantile_arg(q)
+    return nbinominvcdf(d.r, d.p, q)
+end
+function cquantile(d::NegativeBinomial, q::Real)::Int
+    _check_cquantile_arg(q)
+    return nbinominvccdf(d.r, d.p, q)
+end
+function invlogcdf(d::NegativeBinomial, lq::Real)::Int
+    _check_invlogcdf_arg(lq)
+    if isinf(lq)
```
nalimilan, line 125:
```suggestion
    if isinf(lq) # -Inf
```
(I didn't get it at first.)

src/univariate/discrete/negativebinomial.jl
```diff
@@ -112,10 +112,30 @@ cdf(d::NegativeBinomial, x::Real) = nbinomcdf(d.r, d.p, x)
...
+        return nbinominvcdf(d.r, d.p, zero(lq))
+    else
+        return nbinominvlogcdf(d.r, d.p, lq)
+    end
+end
+function invlogccdf(d::NegativeBinomial, lq::Real)::Int
+    _check_invlogccdf_arg(lq)
+    if isinf(lq)
```
nalimilan, line 133:
```suggestion
    if isinf(lq) # -Inf
```

src/univariate/discrete/noncentralhypergeometric.jl
```diff
@@ -9,23 +9,23 @@ abstract type NoncentralHypergeometric{T<:Real} <: DiscreteUnivariateDistributio
...
+        range = reverse(range)
+    end
+
+    # Support must not be empty
+    r, state = iterate(range)
+    qsum = pdf(d, r)
+    while qsum < q && (r_state = iterate(range, state)) !== nothing
+        r, state = r_state
+        qsum += pdf(d, r)
     end
```
nalimilan, line 27: Wouldn't it be simpler to keep the current approach and just check that `!isempty(range)`? I don't find the `iterate` approach super readable.

src/univariates.jl
```diff
@@ -396,7 +396,10 @@ quantile(d::UnivariateDistribution, p::Real)
 
 The complementary quantile value, i.e. `quantile(d, 1-q)`.
 """
-cquantile(d::UnivariateDistribution, p::Real) = quantile(d, 1.0 - p)
+function cquantile(d::UnivariateDistribution, p::Real)
+    _check_cquantile_arg(p)
+    quantile(d, 1 - p)
```
nalimilan, line 401: Or even:
```suggestion
    quantile(d, one(p) - p)
```

src/univariates.jl
```diff
@@ -649,10 +690,34 @@ macro _delegate_statsfuns(D, fpre, psyms...)
...
+            _check_cquantile_arg(q)
+            return $(finvccdf)($(pargs...), q)
+        end
+
+        # There is a bug in some functions in StatsFuns.RFunctions/Rmath:
+        # For `lq = -Inf` they return `NaN`
+        # We work around this issue by falling back to quantile/cquantile
+        function $Distributions.invlogcdf(d::$D, lq::Real)::$T
+            _check_invlogcdf_arg(lq)
+            if isinf(lq)
```
nalimilan, line 707:
```suggestion
            if isinf(lq) # -Inf
```

src/univariates.jl
```diff
@@ -649,10 +690,34 @@ macro _delegate_statsfuns(D, fpre, psyms...)
...
+                return $(finvcdf)($(pargs...), zero(lq))
+            else
+                return $(finvlogcdf)($(pargs...), lq)
+            end
+        end
+        function $Distributions.invlogccdf(d::$D, lq::Real)::$T
+            _check_invlogccdf_arg(lq)
+            if isinf(lq)
```
nalimilan, line 715:
```suggestion
            if isinf(lq) # -Inf
```

src/univariate/discrete/negativebinomial.jl
```diff
@@ -112,10 +112,30 @@ cdf(d::NegativeBinomial, x::Real) = nbinomcdf(d.r, d.p, x)
...
-invlogccdf(d::NegativeBinomial, lq::Real) = convert(Int, nbinominvlogccdf(d.r, d.p, lq))
+function quantile(d::NegativeBinomial, q::Real)::Int
+    _check_quantile_arg(q)
+    return nbinominvcdf(d.r, d.p, q)
+end
+function cquantile(d::NegativeBinomial, q::Real)::Int
+    _check_cquantile_arg(q)
+    return nbinominvccdf(d.r, d.p, q)
+end
+function invlogcdf(d::NegativeBinomial, lq::Real)::Int
```
nalimilan, line 123: Aren't this method and the next one redundant with the new general definition from `_delegate_statsfuns`?

test/testutils.jl
```diff
@@ -447,6 +447,7 @@ end
 
 
 function test_evaluation(d::DiscreteUnivariateDistribution, vs::AbstractVector, testquan::Bool=true)
```
nalimilan, line 449: Are there systematic tests somewhere that for all distributions, `p` and `lp` throw when they are not in the expected range? That seems easy to do in a loop for all corner cases (0.0, -0.0, 1, Inf, -Inf, NaN...).

nalimilan (commented):
Since it appears you need a review, here's one, though I'm not familiar with the package. ;-)

Just to make sure everybody is on the same page: currently many common distributions (`Normal`, `Geometric`, `Exponential`, `Cauchy`, `Bernoulli`...) support `quantile(d, NaN) === NaN` and similar, while others throw an error. With this PR it's always an error. That's cleaner, but it's breaking so that requires a new major release (time to tag 1.0?).

src/univariates.jl
dkarrasch: For unitful numbers, `one` gives the unitless 1 so that `1*p = p` and `one(p)` and `p` may not even be comparable. Instead, `oneunit` returns a number of the same unit as `p` and should be comparable.

dkarrasch: I forgot if unitful numbers subtype `Real` or just `Number`, so my comment may be irrelevant...

simsurace: `Number` AFAIK, not `Real`

devmotion: That's not enforced anywhere, so AFAICT in principle also unitful numbers could subtype `Real`. Unitfuls.jl, however, does only subtype `Number`, not `Real`.

dkarrasch: I still think that - regardless whether it makes a computational difference - `oneunit` is conceptually the right thing to use here: `p` and `zero(p)` are of same "unit"/type, `one(p)` may not.

nalimilan: Not a big deal anyway so it shouldn't hold this PR back.

Issue #1443, opened 2021-11-29
Error thrown when 


ParadaCarleton:
```
julia> kldivergence(TDist(1), Normal(0, 1))
    ERROR: DomainError with -0.9999999999999964:
    integrand produced NaN in the interval (-1.0, -0.9999999999999929)
```
It makes sense why this happens (blows up to infinity), but I think it would be more sensible to find some way to deal with cases like this and return `Inf`.

devmotion:
It seems the KL divergence is finite and there's even a simpler form, also for the multivariate generalization: https://rpubs.com/FJRubio/DKLtn

The problem is probably the same as in https://github.com/JuliaMath/QuadGK.jl/issues/38. Maybe applying the suggestion and adjusting `rtol` fixes the issue?

In general, keep in mind that QuadGK is just used as a fallback in the univariate case if no closed form expressions are implemented. I'm actually happy that it does not just return an incorrect result here - even though the error is a bit confusing (confused me at least when I encountered it in some other setting). Maybe QuadGK could throw a more descriptive error in this case (and the one described in the linked issue)?

ParadaCarleton:
> It seems the KL divergence is finite and there's even a simpler form, also for the multivariate generalization: https://rpubs.com/FJRubio/DKLtn
> 
> The problem is probably the same as in [JuliaMath/QuadGK.jl#38](https://github.com/JuliaMath/QuadGK.jl/issues/38). Maybe applying the suggestion and adjusting `rtol` fixes the issue?
> 
> In general, keep in mind that QuadGK is just used as a fallback in the univariate case if no closed form expressions are implemented. I'm actually happy that it does not just return an incorrect result here - even though the error is a bit confusing (confused me at least when I encountered it in some other setting). Maybe QuadGK could throw a more descriptive error in this case (and the one described in the linked issue)?

The divergence is finite in one direction -- `kldivergence(Normal(0, 1), TDist(1))` accurately returns `0.25924453248886237`. It's the other direction that's infinite and throws an error (a normal will always be an infinitely bad approximation of the Cauchy distribution, but a Cauchy distribution can be used as an OK approximation of the normal).

PR #1806, opened 2023-11-27
Add Generalized chi-squared distribution

heliosdrm:
This extends the collection of univariate continuous distributions with the Generalized chi-squared. All required methods are added, and some of the recommended ones, as indicated in https://juliastats.org/Distributions.jl/stable/extends/#Univariate-Distribution

**Some notes:**

1. I started considering [Abhranil Das' code for Matlab](https://github.com/abhranildas/gx2) (MIT-licensed) as reference, although progressively departed from it, so that eventually there is little in the Julia code that can be attributed to him. I have kept a comment where I mention the inspiration to define the `cdf` function, using Davies' algorithm to integrate the characteristic function. But I don't know if a more explicit attribution in the code or in the license should be made.
2. There are several auxiliary functions that should not be part of the API, and I have encapsulated all of them in the module `GChisqComputations`, after the example of the Chernoff distribution.
3. That module includes two functions to calculate the characteristic function: `cf_explicit` that implements the explicit formula, and `cf_inherit` that composes the results of `cf(::Normal)` and `cf(::NoncentralChisq)`. Both are equivalent, and the code of the second is clearer, but `cf(::GeneralizedChisq)` calls the first one, because in a few tests that I have done, it is slightly (but not much) faster.
4. `cdf` and `pdf` are calculated by integration, using the Gil-Pelaez theorem (the algorithm for the CDF is given in Davies' paper, and the algorithm for the PDF has been indirectly derived from it). The integration is done using QuadGK.jl, which was already in the dependencies of Distributions.jl. The default tolerances of `QuadGK.quadgk` made the calculation of `cdf` at the median very slow, because there the integral is zero. Since in the ends of the distribution the integral is `±π/2`, which is in the order of magnitude of the unit, I have used a fixed absolute tolerance of `eps(one(T))` for the integration in the calculation of the CDF. I have not, however, changed the defaults for the calculation of the PDF.
5. `quantile` is calculated using `Distributions.quantile_newton`. However, there is no analytic solution to find the mode, which is the recommended starting point, so this is searched using the following strategy: (1) start at the mean of the distribution, and use it if it meets the convergence criteria; (2) otherwise define a bracket between that point and another at one standard deviation from the mean, in the direction towards the target probability (and if the bracket does not contain the target probability, extend it until it does); (3) bisect the bracket until one of the ends meets the convergence criteria, and then use that as the starting point. This seems to work fine, but it's a self-made algorithm, and perhaps there is a more efficient one that might be used instead.

### DIFF
--- a/docs/src/univariate.md
+++ b/docs/src/univariate.md
@@ -238,6 +238,13 @@ plotdensity((0, 18), Gamma, (7.5, 1)) # hide
 ```
 
 ```@docs
+GeneralizedChisq
+```
+```@example plotdensity
+plotdensity((-20, 20), GeneralizedChisq, ([1,-2], [1,2], [0,2], 10, 1)) # hide
+```
+
+```@docs
 GeneralizedExtremeValue
 ```
 ```@example plotdensity
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -97,6 +97,7 @@ export
     FullNormalCanon,
     Gamma,
     DiscreteNonParametric,
+    GeneralizedChisq,
     GeneralizedPareto,
     GeneralizedExtremeValue,
     Geometric,
--- a/src/univariate/continuous/generalizedchisq.jl
+++ b/src/univariate/continuous/generalizedchisq.jl
@@ -126,20 +126,6 @@ function quantile(d::GeneralizedChisq, p::Real)
     quantile_newton(d, p, x0)
 end
 
-function quantile_b(d::GeneralizedChisq, p::Real)
-    # search starting point meeting convergence criterion
-    x0 = mean(d)
-    error0, curv0, converges = GChisqComputations.newtonconvergence(d, p, x0)
-    if converges
-        return quantile_newton(d, p, x0)
-    else
-        (x1, x2), _, _ = GChisqComputations.definebracket(d, p, x0, error0, curv0, sqrt(var(d)))
-        xl, xr = (x1 < x2) ? (x1, x2) : (x2, x1)
-        return quantile_bisect(d, p, xl, xr)
-    end
-end
-
-
 function minimum(d::GeneralizedChisq{T}) where T
     d.σ > zero(T) || any(<(zero(T)), d.w) ? typemin(T) : d.μ
 end
@@ -200,7 +186,7 @@ module GChisqComputations
     where `θ` and `ρ` are the outputs of this function.
 
     Those terms are related to the characteristic function of the distribution as:
-        exp(θ*im)/ρ = exp(-ux*im)*cf(u) 
+        exp(θ*im)/ρ = exp(-u*x*im)*cf(u) 
 
     They can be also used to calculate the pdf as:
         f(x) = 1/π *∫cos(θ)/ρ du
@@ -244,13 +230,6 @@ module GChisqComputations
         end
         return integral/π
     end
-
-    # function dpdf(d::GeneralizedChisq{T}, x::Real) where T
-    #     # special case
-    #     iszero(d.σ) && !insupport(d, x) && return zero(T)
-    #     # general case
-    #     GChisqComputations.daviesdpdf(x, d.w, d.ν, d.λ, d.μ, d.σ)
-    # end
 
     # functions to look for starting point where
     # the Newton method for quantiles converges:
--- a/src/univariate/continuous/generalizedchisq.jl
+++ b/src/univariate/continuous/generalizedchisq.jl
@@ -1,0 +1,313 @@
+using Distributions, Random # remove
+import Distributions: sampler, cf, cdf, pdf, quadgk, quantile_newton, quantile_bisect # remove
+
+"""
+    GeneralizedChisq(w, ν, λ, μ, σ)
+
+The *Generalized chi-squared distribution* is the distribution of a sum of independent noncentral chi-squared variables and a normal variable:
+
+```math
+\\xi =\\sum_{i}w_{i}y_{i}+x,\\quad y_{i}\\sim \\chi '^{2}(\\nu_{i},\\lambda _{i}),\\quad x\\sim N(\\mu,\\sigma^{2}).
+```
+
+```julia
+GeneralizedChisq(w, ν, λ, μ, σ)
+
+```
+
+External links
+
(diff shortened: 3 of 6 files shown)


mahaa2:
When this is will be added to the main branch ?

mahaa2:
Hi,

I have played around with this extension. 
As a Statistician, I strongly believe it is a **very useful and very important distribution to have in main branch** of the package Distribution.jl. 

In general it seems to work very well and I will use for future research.

However, I notice that while doing my own tests the quadgk.jl package/function used inside the file generalizedchisq.jl
**appears to fail** accusing the error (line 260 of that file)

"
DomainError with 0.9999999999999964:
integrand produced NaN in the interval (0.9999999999999929, 1.0)
" ,

**in the following explained cases.** 

Lets say that X ~ N_(µ, Σ) with particular vector dimension D = 2, 3, 4.
Then we now know that Y = q0 + q1'X + X'QX ~GChisq(w, ν, λ, m, s)
 (the notation in the GChisq package is different, but also irrelevant here).

The parameter values (w, ν, λ, m, s) = T(q0, q1, Q, µ, Σ) are obtained accordingly with the transformation, say T,
following the paper [https://arxiv.org/abs/2012.14331](url) (I present below my own code to replicate the experiments in Julia).

One can just run the below code and check the results. The error also seems to be exactly the same one appearing in here [https://github.com/JuliaMath/QuadGK.jl/issues/38](url).

```julia

mutable struct parGChisq
    w::Vector{<: Real}
    ν::Vector{<: Real}
    λ::Vector{<: Real}
    m::Real
    s::Real
end

function mapGChisqpar(
    q0::Real,
    q1::Vector{<: Real},
    Q ::AbstractMatrix,
    µ ::Vector{<: Real},
    Σ ::AbstractMatrix,
    ch::Bool = false)::parGChisq
    
    """
     if   X ~ N(µ, Σ)
     then Υ = q0 + q1' X + X'QX ~ GChisq(w, ν, λ, m, s)
     
     see https://arxiv.org/pdf/2012.14331.pdf

     the option with ch = true
     does not work. why ? 
    """ 

    # check dimensionality match
    dimQ = size(Q, 1)
    dimq = length(q1)
    dimΣ = size(Σ, 1)
    dimµ = length(µ)

    dimQ !== dimq ? @error("dims mismatch Q, q1") : nothing
    dimΣ !== dimq ? @error("dims mismatch Σ, q1") : nothing
    dimΣ !== dimµ ? @error("dims mismatch Σ, µ") : nothing

    # check matrix conditions
    !issymmetric(Q) ? @error("Q is not symmetric") : nothing
    !isposdef(Σ)    ? @error("Σ is not PD") : nothing

    # take the square root matrix 
    S = !ch ? sqrt(Σ) : cholesky(Σ).L

    # standartize the space
    t_Q² = !ch ? Symmetric(S * Q * S) : Symmetric(S * Q * S')
    t_q1 = S * (2.0 * Q * µ + q1)
    t_q0 = q0 + dot(q1, µ) + dot(µ, Q, µ) 

    # get the generalized Chi-squared parameters
    d, R = eigen(t_Q²)
    d .= trunc.(d, digits = 9)
    # d[abs.(d) .< eps()] .= 0.0

    # compute b terms
    b = R' * t_q1

    # unique nonzeros eigenvalues
    # the unique function returns an ordered set
    iz = d .!= 0.0
    w  = unique(d[iz])
    ew = !isempty(w) 

    # warn if there are only zero eigenvalues
    ew ? nothing : @warn("Y will follow a Gaussian distribution") 

    # total dof of each nonzero eigenvalue
    ν = ew ? [sum(d .== i) for i in w] : Float64[]

    # total non-centrality for each nonzero eigenvalue
    λ = ew ? [sum(b[d .== x].^2.0) for x in w] ./ (4.0.*w.^2.0) : Float64[]
    
    # Gaussian parameters
    m = t_q0 - dot(w, λ)
    s = norm(b[.!iz]) 

    return parGChisq(w, ν, λ, m, s)
end


using Distributions, 
      LinearAlgebra,
      Plots


# tests
# dimension
D = 3 # 2 or 4

# set parameter of the const + linear + quadratic form
q0 = randn()
q1 = rand(D)
Q  = Diagonal(randn(D))
µ  = randn(D)
Σ  = rand(Wishart(D, Matrix(I, D, D)), 1)[];

# set the correspondent parameters of the GChisq
th = mapGChisqpar(q0, q1, Q, µ, Σ)

# set the GChisq distribution
π = GeneralizedChisq(th.w, th.ν, th.λ, th.m, th.s)

# test
@time pdf(π, mean(π))
```

AdaemmerP:
Adding the generalized chi-squared distribution would be highly appreciated!

heliosdrm:
If there is anything blocking the merging of this PR that could be solved, I'd love to help. Regarding the issue reported by @mahaa2, maybe it is related to what I had mentioned:

> The default tolerances of QuadGK.quadgk made the calculation of cdf at the median very slow, because there the integral is zero. Since in the ends of the distribution the integral is ±π/2, which is in the order of magnitude of the unit, I have used a fixed absolute tolerance of eps(one(T)) for the integration in the calculation of the CDF. I have not, however, changed the defaults for the calculation of the PDF.

I see that this problem also happens in other functions of the package, as `kldivergence` (#1443), and in that case @devmotion comments that it's "safer to throw an error than to return a possibly incorrect value". So I'm not sure what should be done in this case: (a) set another tolerance default for the PDF, (b) document the issue around median values, (c) devise a workaround to change the starting point of the integration if this situation is detected, or (d) just leave it as it is.

Let aside that, there were some CI errors for nightly, but those also happened to other unrelated PR when this one was submitted, so I think they are not related to real issues in the code.

heliosdrm:
Hi. Any suggestion about what might be done to improve this PR and get it merged? Thank you!

AdaemmerP:
Any updates on this merge attempt?

sztal:
Wonderful idea! Any decision when this can be released?

AdaemmerP:
Any updates on this PR? It seems everything has passed and should be ready for merging?

heliosdrm:
As a workaround to use this distribution while this PR is open, I have created this package: https://github.com/heliosdrm/GeneralizedChisqDistribution.jl

It is not registered, but [I'm open to discuss it](https://github.com/heliosdrm/GeneralizedChisqDistribution.jl/issues/1).

PR #1808, opened 2023-12-05
Add Asymmetric Exponential Power Distribution

lukketotte:
I have previously added the Skewed Exponential Power distribution (SEPD) to `Distributions.jl`. This pull request implements the Asymmetric Exponential Power Distribution (AEPD) of [Zhu and Zinde-Walsh, 2009](https://www.sciencedirect.com/science/article/pii/S0304407608001668) which is the extension of the SEPD to include different shape parameters on the left and right hand sides of the mode.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -10,7 +10,9 @@ using LinearAlgebra
 import JSON
 import ForwardDiff
 
-const tests = [
+
+
+const temp = [
     "univariate/continuous/loguniform",
     "univariate/continuous/arcsine",
     "univariate/discrete/dirac",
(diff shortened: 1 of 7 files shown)


test/runtests.jl
```diff
@@ -154,7 +157,7 @@ const tests = [
     # "univariate/discrete/geometric",
     # "univariate/discrete/hypergeometric",
     # "univariate/discrete/noncentralhypergeometric",
-    # "univariate/discrete/poisson",
+    # "univariate/discrete/poisson",a
```
lukketotte, line 160: typo

PR #1809, opened 2023-12-09
support intervals

aplavin:
Add extension for IntervalSets (the most popular intervals package) to support its intervals where it makes sense in Distributions.
I defined and used this `Uniform` method a few times myself, others are  included here for completeness.

No tests for now, looking for a general confirmation first.


devmotion:
I wonder if there is general interest for this functionality, the application seems a bit niche to me since most distributions have a fixed support that can't be adjusted based on a given interval. My gut feeling is that it is easy enough to extract the endpoints of an interval and call the standard constructors if one wants to construct e.g. a `Uniform` distribution on a given interval.

chelate:
I just want to point out that intervals are somewhat fundamental objects in probability (the Borel algebra), and that prob(interval,dist) = cdf(dist,interval.upper) - cdf(dist,interval.lower) or similar could be pretty neat. Likewise conditional probabilities, empirical samplers, histograms, all kinds of stuff!

devmotion:
I guess such definitions might be convenient for someone working with IntervalSets, but my feeling is that these should rather be put in an extension of IntervalSets (similar to existing extensions for Random, Statistics, and StatsBase).

devmotion:
My concern is that putting these definitions in an extension in Distributions adds maintenance burden for developers that seemingly so far have never expressed a desire to use IntervalSets for constructing the few univariate distributions whose support is an interval that fully characterizes the parameters of this distribution type. The use case still seems quite limited to me, and hence I think one has not be too worried about new distributions. Since the PR also only uses exported and official APIs, I don't think potential breakages are a problem either.

To me it seems contributors of IntervalSets are much more interested in such a functionality than developers of Distributions, so I guess an extension in IntervalSets would be preferable.

devmotion:
Existing tools in external packages could surely be useful. Note, however, that [my comment](https://github.com/JuliaStats/Distributions.jl/issues/1254#issuecomment-785218433) was about discrete distributions for which AFAICT IntervalSets would be insufficient and we would need DomainSets or something different such as InfiniteArrays.

Regarding the previous changes (constructors with intervals) my opinion remains unchanged (see above).

So if you'd like to replace RealInterval with IntervalSets.Interval I suggest opening a separate PR (or reverting the other changes).

devmotion:
> use intervalsets in some places, but not everywhere where they make sense.

My point is that I think it doesn't make sense to use them to construct distributions but it does make sense to use them as an improvement of `RealInterval`.

aplavin:
Happy 2nd anniversary of this PR :)
Just checking whether the opinion of Distributions.jl maintainers changer wrt integrating with other popular packages in the ecosystem – IntervalSets.jl in this case. Stuff like constructing distributions from intervals (included in this PR), or even adding `uniform(xs)` function (will be happy to add) that returns the proper uniform distribution for anything, from intervals to ranges to arbitrary collections.

PR #1814, opened 2024-01-01
Clarify definitions of inverse CDF functions

### LINKED ISSUE #1813: `inv*` functions are not really inverses
There is an inconsistency between the documentation (and name) of functions like `invlogcdf` and their actual behavior. They are claimed to be inverses (to `logcdf`), so should presumably error when inverting is not possible. However, they still return some result in these cases, which definitely isn't the proper inverse:
```julia
help?> invlogcdf
  invlogcdf(d::UnivariateDistribution, lp::Real)
  The inverse function of logcdf.

julia> d = Uniform(0, 1)

julia> invlogcdf(d, logcdf(d, -5))
0.0
```

sethaxen:
Fixes #1813 by updating docstrings as proposed in https://github.com/JuliaStats/Distributions.jl/issues/1813#issuecomment-1872989317:
- documents in what sense `quantile` and friends invert `cdf` and friends
- documents that inputs to `quantile` and friends are probabilities in `[0, 1]` or log-probabilities in `(-Inf, 0]`. (inputs outside of this range would then lead to undefined behavior)

### DIFF
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -382,7 +382,7 @@ logccdf(d::UnivariateDistribution, x::Real) = log(ccdf(d, x))
 """
     quantile(d::UnivariateDistribution, q::Real)
 
-Evaluate the generalized inverse cumulative distribution function at `q`.
+Evaluate the (generalized) inverse cumulative distribution function at `q`.
 
 For a given `0 ≤ q ≤ 1`, `quantile(d, q)` is the smallest value `x` in the support of `d`
 for which `cdf(d, x) ≥ q`.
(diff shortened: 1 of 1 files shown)


src/univariates.jl
```diff
@@ -381,7 +382,10 @@ logccdf(d::UnivariateDistribution, x::Real) = log(ccdf(d, x))
 """
     quantile(d::UnivariateDistribution, q::Real)
 
-Evaluate the inverse cumulative distribution function at `q`.
+Evaluate the generalized inverse cumulative distribution function at `q`.
+
+For a given `0 ≤ q ≤ 1`, `quantile(d, q)` is the smallest value `x` in the support of `d`
```
andreasnoack, line 387: While I think it good to include the accurate definition, I feel like it would be helpful to state the simplified version before stating the precise but also more complicated definition. For most people, in most situations, the simple but slightly inaccurate definitions is probably more helpful. For the same reason, maybe put "generalized" in parentheses as many people might not know what it means. Same comment applies to the other changes.

sethaxen: Done!

andreasnoack (approved):

devmotion:
Related: https://github.com/JuliaStats/Distributions.jl/pull/1805

Issue #966, opened 2019-09-04
pdf(d) should give a function


dpsanders:
If `d` is a distribution then it would be useful to have `pdf(d)` be the function that maps `x` to `pdf(d, x)`.

dpsanders:
Or maybe it should return a `PDF` type.

andreasnoack:
Could you provide a bit of motivation for this? As you show, it's a very easy to define this on your own.

dpsanders:
If I have a distribution `d`, it's very natural to want to talk about "the pdf of `d`" and expect this to be a function. For example to do the following:

```jl
d = Normal()
r = rand(d, 100)

using Plots
histogram(d)
plot!(pdf(d))
```

It's also not user-friendly for `pdf(d)` to just throw an error:
```jl
julia> d = Normal()
Normal{Float64}(μ=0.0, σ=1.0)

julia> pdf(d)
ERROR: MethodError: no method matching pdf(::Normal{Float64})
Closest candidates are:
  pdf(::Normal, ::Real) at /Users/dpsanders/.julia/packages/Distributions/tfkz4/src/univariate/continuous/normal.jl:112
```
(even though it does hint at what you can do with `pdf`).

What is the argument against defining this in the package?

andreasnoack:
> What is the argument against defining this in the package?

That it's almost a trivial definition, i.e.

```julia
plot!(t -> pdf(d, t))
```

> If I have a distribution d, it's very natural to want to talk about "the pdf of d" ...

Indeed

> and expect this to be a function.

...well, not sure about this part. Generally, I think it would be clearer to use a generic function `pdf` instead of an anonymous function. As you mention, having a `PDF` could be an alternative but there is always implementation overhead when introducing new types so it would be reasonable to come up with more possible method before introducing a new type.

richardreeve:
Maybe a possible advantage of `pdf(d::Dist)` (and `cdf(d)`, `logcdf(d)`, etc.) returning a specified callable type (`DistFun{Dist}`?) would be that plotting it could then take the support correctly and handle it seamlessly, so that types with countable support, bounded continuous types, unbounded continuous types, etc. could all be handled neatly in a consistent way, despite there being several such functions...

PR #1818, opened 2024-01-04
Replace broadcasting over distributions with broadcasting with partially applied functions

### LINKED ISSUE #1815: rand! with Uniform distribution does not work with CUDA arrays
The following code to initialize an existing CUDA matrix results in memory corruption (and crashes):
(Note: run in CUDA's compute sanitizer to catch the bug.)

```julia
using CUDA
CUDA.run_compute_sanitizer()

using CUDA
using Distributions
using Random

m = CUDA.rand(1024, 1024)
s = Uniform(-1.0, 1.0)
rand!(CUDA.default_rng(), s, m)
```

It results in a series of messages that include:

```text
========= Invalid __global__ read of size 1 bytes
=========     at _Z16broadcast_kernel15CuKernelContext13CuDeviceArrayI7Float32Li2ELi1EE11BroadcastedI12CuArrayStyleILi2EE5TupleI5OneToI5Int64ES5_IS6_EE8quantileS4_I12CuRefPointerI7UniformI7Float64EE8ExtrudedIS0_IS1_Li2ELi1EES4_I4BoolS12_ES4_IS6_S6_EEEES6_+0x550 in pointer.jl:119
=========     by thread (99,0,0) in block (0,0,0)
=========     Address 0x7f2a88250976 is out of bounds
=========     and is 139,732,598,458,743 bytes after the nearest allocation at 0x1480000000 of size 4,194,304 bytes
=========     Saved host backtrace up to driver entry point at kernel launch time
```

The problem seems to come from the following:

```text
In materialize!(dest, bc) at broadcast.jl:910
 910  @inline function materialize!(dest, bc::Broadcasted{<:Any})
>911      return materialize!(combine_styles(dest, bc), dest, bc)
 912  end

About to run: <(Base.Broadcast.combine_styles)(Float32[0.64629114 0.78703153 0.52409637 0.6335989 0.8139833 0.772531...>
```
(which is called by `_rand!(rng::AbstractRNG, d::Uniform, A::AbstractArray{<:Real})` in src/univariate/continuous/uniform.jl:156.

The following implementation (which is functionally equivalent to the existing) of `_rand!` seems to work ok:

```julia
function Distributions._rand!(rng::AbstractRNG, d::Uniform, A::AbstractArray{<:Real})
                rand!(rng, A)
                A .*= d.b - d.a
                A .+= d.a
end
```

Any ideas of what may be going wrong?

## Environment

I have tried mostly in Julia 1.10, but have observed the problem in 1.9 (1.9.4 and, I believe, 1.9.3).
I am using Windows and WSL.
I have the following packages:

```text
  [052768ef] CUDA v5.1.1
  [31c24e10] Distributions v0.25.104
```

The CUDA environment is:

```text
CUDA runtime 12.3, artifact installation
CUDA driver 12.2
NVIDIA driver 550.9.0

CUDA libraries:
- CUBLAS: 12.3.4
- CURAND: 10.3.4
- CUFFT: 11.0.12
- CUSOLVER: 11.5.4
- CUSPARSE: 12.2.0
- CUPTI: 21.0.0
- NVML: 12.0.0+550.9

Julia packages:
- CUDA: 5.1.1
- CUDA_Driver_jll: 0.7.0+0
- CUDA_Runtime_jll: 0.10.1+0

Toolchain:
- Julia: 1.10.0
- LLVM: 15.0.7

1 device:
  0: NVIDIA GeForce RTX 2080 Ti (sm_75, 8.745 GiB / 11.000 GiB available)
```

devmotion:
Broadcasting treats `UnivariateDistribution`s as scalars due to https://github.com/JuliaStats/Distributions.jl/blob/9e72f1ff39eb547cc420a278b5d96323a6de5ad5/src/common.jl#L175. However, such `Ref`s are suboptimal since
- they may affect the broadcasting style and/or break broadcasting (see, e.g., #1815 and https://github.com/JuliaGPU/GPUArrays.jl/pull/510#issue-2064061047)
- tuples tend to infer better (see, e.g., https://github.com/JuliaGPU/GPUArrays.jl/pull/510#issuecomment-1875852667 and https://discourse.julialang.org/t/add-scalar-macro-to-base/50255/4)

Unfortunately, replacing https://github.com/JuliaStats/Distributions.jl/blob/9e72f1ff39eb547cc420a278b5d96323a6de5ad5/src/common.jl#L175 with `broadcastable(d::UnivariateDistribution) = (d,)` is not an option since broadcasting with tuples is not identical to broadcasting with `Ref`s (see, e.g., https://discourse.julialang.org/t/add-scalar-macro-to-base/50255/5).

However, broadcasting over distributions is actually not needed as one can just broadcast with a partially applied function that captures the distribution instead. Therefore this PR changes all broadcasts over distributions in Distributions.jl to broadcasts with such partially applied functions (and quite often this implies that broadcasting is not needed anymore and a simple `map` would be sufficient).

Fixes #1815.

---

As a side note, IMO it would be more convenient for users and downstream packages that would like to follow the same pattern if we would define `pdf(d::Distribution) = Base.Fix1(pdf, d)` etc. (similar to `==(x)` etc. in Base) (ref https://github.com/JuliaStats/Distributions.jl/issues/966). Unfortunately, `pdf(d::DiscreteUnivariateDistribution)` is already defined in a different way: https://github.com/JuliaStats/Distributions.jl/blob/9e72f1ff39eb547cc420a278b5d96323a6de5ad5/src/deprecates.jl#L43


theogf:
> Unfortunately, pdf(d::DiscreteUnivariateDistribution) is already defined in a different way

I find your proposal for the use of the single argument `pdf` to be far more useful.
Since it's already deprecated, does that mean we could introduce the new definition in the next breaking release?

PR #1823, opened 2024-01-13
RFC/WIP: coercion "cascades" for MvNormal

### LINKED ISSUE #1544: Fitting does not respect type parameters

timholy:
This redesigns the constructors of `MvNormal` to:
- check argument consistency in the inner constructor
- support "coercion," for example `MvNormal{Float32}(μ, Σ)`
  will enforce that the result has eltype `Float32`
- cascade coercion among the multiple type parameters

The benefits of this design are subtle but significant, as it
allows the user to specify intent, allows the constructors to
accept input arguments that are not (yet) consistent with
the desired output types (e.g., `MvNormal{Float64}(Any[1,2], I)`),
and makes it easy for developers to "propagate" consistent
types across the code base (e.g., fixes #1544).

Among things this fixes, it's currently possible to construct an `MvNormal` with inconsistent dimensions:
```julia
julia> dist = DiagNormal(zeros(3), PDiagMat(zeros(2)));

julia> rand(dist, 5)
ERROR: DimensionMismatch: Inconsistent argument dimensions.
Stacktrace:
```

This is possible because the consistency check is not in the inner constructor.

### DIFF
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -166,13 +166,20 @@ Generally, users don't have to worry about these internal details.
 We provide a common constructor `MvNormal`, which will construct a distribution of
 appropriate type depending on the input arguments.
 """
-struct MvNormal{T<:Real,Cov<:AbstractPDMat,Mean<:AbstractVector} <: AbstractMvNormal
+struct MvNormal{T<:Real,Cov<:AbstractPDMat{T},Mean<:AbstractVector{T}} <: AbstractMvNormal
     μ::Mean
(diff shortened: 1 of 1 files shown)


timholy:
This won't currently pass tests because PDMats needs similar improvements.

PR #1825, opened 2024-01-14
Tweak docs, add tests for extension interface

timholy:
This makes minor documentaion changes that show
users how to extend the interface without generating
ambiguities or StackOverflowErrors.

It turns out to be important to ensure that one's methods "win" specificity, otherwise it's pretty easy to get StackOverflowErrors. In particular, `AbstractArray` often isn't sufficient, it often needs to be `{<:Real}` and/or of specific dimensionality.

Note that a few are `@test_broken`, despite following the documentation. These could be fixed during review or in a follow-up PR.


ViralBShah:
@devmotion @andreasnoack Perhaps this is worth a look to review and merge?

PR #1837, opened 2024-02-19
Remove TruncatedNormal 

andreasnoack:
It has been deprecated for five years at this point.


devmotion (approved):

devmotion:
Seems tests have to be updated.

andreasnoack:
I've updated the tests. Only remaining failed are on nightly and unrelated.

devmotion (commented):
Can https://github.com/JuliaStats/Distributions.jl/blob/3d71a2b0431892d739e3b861d2dd1baf9fdc46d3/test/ref/continuous/truncatednormal.R be removed as well or is it still needed for the tests?

andreasnoack:
It's called by https://github.com/JuliaStats/Distributions.jl/blob/3d71a2b0431892d739e3b861d2dd1baf9fdc46d3/test/ref/gendref.R so as long as we maintain that infrastructure then we should probably keep it. However, I guess those scripts haven't been run in a long time. They rely on exotic R packages of [which some are no longer on CRAN](https://cran.r-project.org/web/packages/LindleyR/index.html). I'll investigate this in a separate PR/issue.

devmotion:
Ah yes, I remember these R scripts now (had to deal with installing dependencies and running them a while ago when I added a few more tests). So then it seems the PR is good to go.

Issue #1845, opened 2024-04-01
Missing docstring for loglikelihood


cgarling:
On the `latest` version of the documentation, the following docstring is missing as there is no method with this signature defined. It looks like this should be `loglikelihood(::UnivariateDistribution, ::Real)` instead.

https://github.com/JuliaStats/Distributions.jl/blob/a9b0e3c99c8dda367f69b2dbbdfa4530c810e3d7/docs/src/univariate.md?plain=1#L76

mschauer:
The vectorised version left us with https://github.com/JuliaStats/Distributions.jl/commit/d6985ad5f08990123b9ba84431626f22f1a408bd#diff-c185a88dade2ce18f68ff0130b5626734ed3cdb341bf5765dbb008fd8a6d91a0

`loglikelihood(::UnivariateDistribution, ::Real)` should have a docstring though, did it get lost there?

cgarling:
I am unfortunately not a git wizard but it looks to me like the relevant method definition is 
https://github.com/JuliaStats/Distributions.jl/blob/f33af97da34a2e772837b63891c0ed52a3a1a989/src/univariates.jl#L329-L330
for which I do not see a docstring.

PR #1847, opened 2024-04-02
Couple documentation fixes

cgarling:
Have some time so fixing some of the documentation issues like #1845. Will try to document things I notice but don't feel qualified to fix.

### DIFF
--- a/docs/src/multivariate.md
+++ b/docs/src/multivariate.md
@@ -31,8 +31,8 @@ entropy(::MultivariateDistribution, ::Real)
 
 ```@docs
 insupport(::MultivariateDistribution, ::AbstractArray)
-pdf(::MultivariateDistribution, ::AbstractArray)
-logpdf(::MultivariateDistribution, ::AbstractArray)
+pdf(::Distribution{ArrayLikeVariate{N}}, x::AbstractArray{<:Real,M}) where {N,M}
+logpdf(::Distribution{ArrayLikeVariate{N}}, x::AbstractArray{<:Real,M}) where {N,M}
 loglikelihood(::MultivariateDistribution, ::AbstractVector{<:Real})
 ```
 **Note:** For multivariate distributions, the pdf value is usually very small or large, and therefore direct evaluation of the pdf may cause numerical problems. It is generally advisable to perform probability computation in log scale.
--- a/docs/src/univariate.md
+++ b/docs/src/univariate.md
@@ -73,7 +73,7 @@ pdfsquaredL2norm
 insupport(::UnivariateDistribution, x::Any)
 pdf(::UnivariateDistribution, ::Real)
 logpdf(::UnivariateDistribution, ::Real)
-loglikelihood(::UnivariateDistribution, ::AbstractArray)
+loglikelihood(::UnivariateDistribution, ::Real)
 cdf(::UnivariateDistribution, ::Real)
 logcdf(::UnivariateDistribution, ::Real)
 logdiffcdf(::UnivariateDistribution, ::Real, ::Real)
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -326,7 +326,13 @@ logpdf(d::UnivariateDistribution, x::Real)
 # extract value from array of zero dimension
 logpdf(d::UnivariateDistribution, x::AbstractArray{<:Real,0}) = logpdf(d, first(x))
 
-# loglikelihood for `Real`
+"""
+    loglikelihood(d::UnivariateDistribution, x::Real)
+
+Evaluate the logarithm of the likelihood at `x`.
+
+See also: [`logpdf`](@ref).
+"""
 Base.@propagate_inbounds loglikelihood(d::UnivariateDistribution, x::Real) = logpdf(d, x)
 
 """


docs/src/multivariate.md
```diff
@@ -31,8 +31,8 @@ entropy(::MultivariateDistribution, ::Real)
 
 ```@docs
 insupport(::MultivariateDistribution, ::AbstractArray)
-pdf(::MultivariateDistribution, ::AbstractArray)
-logpdf(::MultivariateDistribution, ::AbstractArray)
+pdf(::Distribution{ArrayLikeVariate{N}}, x::AbstractArray{<:Real,M}) where {N,M}
+logpdf(::Distribution{ArrayLikeVariate{N}}, x::AbstractArray{<:Real,M}) where {N,M}
```
devmotion, line 35: I'm not sure... This page is about `MultivariateDistribution` but the methods listed here are for generic distributions with array-like variates. I think it would be better to include these generic docstrings in a separate page and only list here `pdf`, `logpdf` etc. as part of the interface for multivariate distributions with a link to the generic docstring.

docs/src/univariate.md
```diff
@@ -73,7 +73,7 @@ pdfsquaredL2norm
 insupport(::UnivariateDistribution, x::Any)
 pdf(::UnivariateDistribution, ::Real)
 logpdf(::UnivariateDistribution, ::Real)
-loglikelihood(::UnivariateDistribution, ::AbstractArray)
+loglikelihood(::UnivariateDistribution, ::Real)
```
devmotion, line 76: You almost never want to call `loglikelihood` with `::Real` but rather with `::AbstractArray`, so documenting the former instead of the latter seems wrong.

src/univariates.jl
```diff
@@ -326,7 +326,13 @@ logpdf(d::UnivariateDistribution, x::Real)
...
 logpdf(d::UnivariateDistribution, x::AbstractArray{<:Real,0}) = logpdf(d, first(x))
 
-# loglikelihood for `Real`
+"""
+    loglikelihood(d::UnivariateDistribution, x::Real)
+
+Evaluate the logarithm of the likelihood at `x`.
+
+See also: [`logpdf`](@ref).
+"""
```
devmotion, line 335: Same here, this is just a fallback for the case of a single variate but not the typical use case of `loglikelihood`.

docs/src/univariate.md
mschauer: Where is that method?

cgarling:
Some other minor docs issues I noted that I could work on with some guidance:
- [ ] Some local links like `[Beta distribution](@ref Beta)` in the docstring for `Kumaraswamy` aren't working for some reason I can't figure out https://github.com/JuliaStats/Distributions.jl/blob/f33af97da34a2e772837b63891c0ed52a3a1a989/src/univariate/continuous/kumaraswamy.jl#L11-L14
- [ ] There are `@docs` entries for `_logpdf` in multivariate.md and matrix.md but no docstrings in the source. There is a commented out method in common.jl, but I'm not sure where a docstring should be added or the references should be removed https://github.com/JuliaStats/Distributions.jl/blob/f33af97da34a2e772837b63891c0ed52a3a1a989/src/common.jl#L276-L279
- [ ] `@docs` entry for `_rand!` but no docstring; not sure what this does so can't write one https://github.com/JuliaStats/Distributions.jl/blob/f33af97da34a2e772837b63891c0ed52a3a1a989/docs/src/matrix.md?plain=1#L28 
- [ ] Warnings for references ```[`Distribution`](@ref)```, ```[`ReshapedDistribution`](@ref)```, ```[`MultivariateDistribution`](@ref)``` in reshape.md. `MultivariateDistribution` appears not to have a docstring. The other two do, but they don't seem to have `@docs` entries and so the linking fails. I feel like `MultivariateDistribution` get a docstring and have an `@docs` entry at the top of multivariate.md and `ReshapedDistribution` should likewise have an `@docs` entry at the top of `reshape.md`. I would guess that `Distribution` should get an `@docs` entry on the "Type Hierarchy" page under the "Distributions" heading.

docs/src/univariate.md
cgarling: Thanks for the review. It is not obvious to me why the `::AbstractArray` signature would be preferred over broadcasting through the `::Real` signature for `UnivariateDistribution`. My naive go-to would be the `::Real` method as it is available for all `UnivariateDistribution` and the `::AbstractArray` signature is not. You could make the docstring more useful by adding a sentence explaining why you may or may not want to use it.

devmotion: The canonical method for single variates is just `logpdf`. The real benefit of `loglikelihood` is that it sums the `logpdf` values for multiple variates, possibly using exploiting constant terms etc.

docs/src/multivariate.md
cgarling: That's a valid concern, I was just following the docstring for `loglikelihood` for array-like variates which is currently listed on the multivariate page, so I thought it would be consistent. https://github.com/JuliaStats/Distributions.jl/blob/f33af97da34a2e772837b63891c0ed52a3a1a989/src/common.jl#L433-L443
If you think these should be taken to a separate page, then `loglikelihood` probably should as well

docs/src/univariate.md
devmotion: `loglikelihood` is defined here: https://github.com/JuliaStats/Distributions.jl/blob/f33af97da34a2e772837b63891c0ed52a3a1a989/src/common.jl#L444-L465

mschauer: Ah, I was confused about "ArrayLikeVariate{N}", does this include univariate distributions?

cgarling: Oh, I see, it calls down to 
https://github.com/JuliaStats/Distributions.jl/blob/f33af97da34a2e772837b63891c0ed52a3a1a989/src/common.jl#L434-L465
because `UnivariateDistribution <: Distribution{ArrayLikeVariate{0}}`
For some reason I thought the univariate case was defined differently...

It seems like the real issue might be how we go about linking these general methods into the separate markdown pages, as the current `@docs` strings no longer work with the new methods. I don't really feel comfortable doing that myself, I don't understand the internals all that well (as shown above...)

PR #1856, opened 2024-05-19
Update CI.yml to include mac aarch64

ViralBShah:


devmotion:
If I understand correctly, this change was unnecessary because recently Github updated `macos-latest` to point to `macos-14` which is arm64 exclusive: https://github.com/actions/runner-images/issues/9255 This can be seen from the CI logs as well, the existing macos job already runs on arm64: https://github.com/JuliaStats/Distributions.jl/actions/runs/9144982709/job/25143434782#step:1:8

ViralBShah:
Oh I didn't realize that they had updated that. I'll revert this.

PR #1857, opened 2024-05-20
Replace deprecated lfact(x)

THargreaves:
`lfact` has been replaced by `logfactorial` since at least version 1.2 of SpecialFunctions.jl, which is the earliest compatible version with Distributions.jl.


devmotion (approved):
Thank you!

PR #1860, opened 2024-05-25
Fix type bug in `Uniform`

Red-Portal:
Fixed the bug where `rand(Uniform{T})` returned a `Float64` instead of `T`.

### DIFF
--- a/src/univariate/continuous/uniform.jl
+++ b/src/univariate/continuous/uniform.jl
@@ -151,7 +151,7 @@ Base.:*(c::Real, d::Uniform) = Uniform(minmax(c * d.a, c * d.b)...)
 
 #### Sampling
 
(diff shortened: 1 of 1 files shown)


andreasnoack (commented):
Please search the issue tracker for discussions about the relationship between the type parameter of the distribution and the type of the random variates. It's a bit complicated. Your version will also fail in several cases. E.g. when the parameters are `Rational{Int}` or `Int` (the latter requiring that you bypass the usual constructor with
```
julia> Uniform{Int}(0, 1)
Uniform{Int64}(a=0, b=1)
```

Red-Portal:
Hmmm... okay, then is the current behavior intended by the current policy?

andreasnoack:
Yes. I'd say so.

Red-Portal:
Okay got it

PR #1861, opened 2024-05-26
Stop testing on nightly

andreasnoack:
There are so many failures that it makes the CI status useless. We should trigger CI upcoming Julia releases in dedicated PRs instead of unconditionally testing on nightly.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -26,7 +26,7 @@ jobs:
         version:
           - '1.3'
           - '1'
-          - 'nightly'
+          # - 'nightly'
         os:
           - ubuntu-latest
           - macos-latest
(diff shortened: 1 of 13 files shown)


devmotion (commented):
I agree, nightly seems too noisy. An alternative we could consider would be to test against the latest tagged version by running one test with
```yaml
version: '1'
include-all-prereleases: true
```
(see https://github.com/julia-actions/setup-julia?tab=readme-ov-file#prereleases).

.github/workflows/CI.yml
```diff
@@ -34,6 +34,7 @@ jobs:
       - uses: julia-actions/setup-julia@v2
         with:
           version: ${{ matrix.version }}
+          include-all-prereleases: true
```
devmotion, line 37: It seems this implies that currently only the prerelease is tested but no stable release anymore...

devmotion:
I think this PR can be closed since the problem was already fixed by #1863 and #1888.

PR #1862, opened 2024-05-27
Fix failing tests on 1.11-beta

andreasnoack:
We need a slightly larger sample size when testing DiscreteNonParametric to ensure that empirical moments are close enough to theoretical moments.

The semicircle tests were also vulnerable to changes to the drawn variates so I've switched from MersenneTwister to StableRNGs.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -2,12 +2,12 @@ name: CI
 
 on:
   pull_request:
-    branches:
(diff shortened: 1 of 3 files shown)


devmotion (approved):

andreasnoack:
This one was against my other PR. I think we should cherry-pick what's here and go with your PR instead.

PR #1863, opened 2024-05-27
Update CI: No nightly tests, improve macOS CI setup + use julia-actions/cache

devmotion:
~~In this draft I see if one can achieve something like #1861 without running tests twice if no prerelease is available. Additionally,~~ This PR removes the nightly tests (see #1861), includes a fix for the mac CI setup (see #1856), and switches to julia-actions/cache for caching.

Edit: The PR also includes https://github.com/JuliaStats/Distributions.jl/pull/1862 which fixes tests on Julia 1.11-beta1 (see https://github.com/JuliaStats/Distributions.jl/pull/1862#issuecomment-2133916510).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,7 +17,7 @@ concurrency:
 
 jobs:
   test:
-    name: Julia ${{ matrix.version }}${{ matrix.prerelease && ' (prerelease)' || '' }} - ${{ matrix.os }} - ${{ github.event_name }}
+    name: "Julia ${{ matrix.version }}${{ matrix.prerelease && ' (prerelease)' || '' }} - ${{ matrix.os }} - ${{ github.event_name }}"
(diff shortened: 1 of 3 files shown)


andreasnoack:
> without running tests twice if no prerelease is available.

I don't think that will happen. There will only be a single run for `julia@1` but that run might be a prerelease, see https://github.com/JuliaStats/Distributions.jl/actions/runs/9250768369/job/25444947763?pr=1862,  and maybe that is not what we want.

devmotion:
> There will only be a single run for julia@1 but that run might be a prerelease, 

Indeed, my comment above was incorrect: I was already assuming that we would want to change that PR such that we continue testing with the latest stable release.

devmotion:
I agree, in particular given that tests seem to fail on the prerelease as well (and the fact that the PR in its current state actually doesn't skip the duplicate runs).

ararslan (approved):

Issue #1317, opened 2021-05-02
What is the breaking change in 0.25?


davidanthoff:
I'm looking at the release notes, and I can't figure out what was breaking in this release?

devmotion:
The types of `MixtureModel` and `LocationScale` were changed.

davidanthoff:
Could there maybe some language be added to the release notes that explains what one has to do to make downstream packages compatible? There are literally probably hundreds of packages that now get a CompatHelper notification that they should mark their package as compatible with v0.25, and at the moment it is really difficult to figure out what I have to do as a downstream package maintainer.

Also, just out of curiosity: is there a chance that we could bunch breaking changes together, and ideally just have one 1.0 release at some point? Distributions.jl is _soo_ deep in the dependency tree and affects so many packages, that the cost in terms of contributor time of downstream packages is enormous when there is a breaking version released here...

devmotion:
I agree, probably it would be good to not only rely on the information about merged PRs and closed issues but add some more detailed release notes. And maybe one should add this information also to a persistent changelog such as a NEWS.md file?

We bundled two completely different breaking PRs in 0.25.0 to avoid having too many breaking releases. I am not opposed to a 1.0 release (there's also an issue that was started almost two years ago: https://github.com/JuliaStats/Distributions.jl/issues/880) but I don't think it will solve the general problem of breaking releases. I anticipate that even after 1.0 (regardless of how many breaking changes one pushes into it), there will be breaking changes quite regularly - Distributions is just too big and there are too many possible improvements. Also IMO the two breaking changes in 0.25.0 are two major improvements (fixes many type inference problems with `MixtureModel` since now all fields are concretely typed and generalizes `LocationScale` to discrete distribution) and it is good to release them timely instead of waiting for other changes that are desired for 1.0 but not worked on yet.

DilumAluthge:
> I anticipate that even after 1.0 (regardless of how many breaking changes one pushes into it), there will be breaking changes quite regularly - Distributions is just too big and there are too many possible improvements.

In that case, perhaps it would be a good idea to break Distributions.jl into multiple smaller packages. Then people could only depend on the individual packages that they need.

@stefankarpinski Any thoughts?

StefanKarpinski:
I dunno, but the breaking changes here are getting quite disruptive so it seems like something has to be done.

davidanthoff:
[...]

PR #1872, opened 2024-06-15
Fix #1870: Rename ValueSupport subtypes for improved clarity

### LINKED ISSUE #1870: Rename `ValueSupport` subtypes to more specific names
Currently the subtypes of `ValueSupport` are `Discrete` and `Continuous` as described here:

https://github.com/JuliaStats/Distributions.jl/blob/b356da03a189d023cdb8467c61806a8a11dcb262/docs/src/types.md?plain=1#L34-L50

These names are too generic to export, and conflict with other mathematical packages.

Would you be ok with renaming these two to `DiscreteSupport` and `ContinuousSupport`?

I can work on a PR that includes a deprecation warning.

juliohm:
This PR fixes #1870.


juliohm:
Tests are failing on Julia nightly for unrelated reasons.

devmotion:
As mentioned in #1870, I'm not sure yet if it's worth to make this change.

juliohm:
The names Continuous and Discrete are too inclusive in meaning. There are
other packages like DataScienceTraits.jl where the word Continuous is used
to refer to Continuous variables in general. In this package the meaning is
that of ContinuousSupport. This PR uses these more specific names.

Alternatively, I can start another PR that doesn't export Continuous nor
Discrete to avoid similar conflicts.

Em sáb., 15 de jun. de 2024, 16:39, David Widmann ***@***.***>
escreveu:

devmotion:
It's clearly a quite general term - but IMO that's not a problem per se. I think the existence of the same (exported) name in some other package is not a sufficient reason for such a major change in Distributions that probably also causes multiple deprecation warnings in downstream packages (in particular since it seems its use in Distributions predates the one in DataScienceTraits).

juliohm:
Is there a problem in not exporting such general names? I don't think users
of Distributions.jl rely on these. And if they do, they can be explicit
about it with an explicit import.

Em sáb., 15 de jun. de 2024, 17:51, David Widmann ***@***.***>
escreveu:

juliohm:
@devmotion are you willing to unexport the `Continuous` and `Discrete` generic names? I can close this PR and open a new one if that is the case.

devmotion:
We can't unexport them in a non-breaking release and there are no plans for a breaking release in the near future.

juliohm:
> there are no plans for a breaking release in the near future.

Because no one is actively leading progress in Distributions.jl or because there is a consensus among the active maintainers that breaking releases should only happen after months?

devmotion:
Because my impression is that downstream developers are annoyed by breaking releases of such a core part of the ecosystem (IIRC there are issues with complaints and last time we did one there were also not too many positive reactions on Slack), hence we try to minimize the number of breaking releases, and there's already a list of important points to change/improve in the next breaking release.

juliohm:
> Because my impression is that downstream developers are annoyed by breaking releases of such a core part of the ecosystem (IIRC there are issues with complaints and last time we did one there were also not too many positive reactions on Slack), hence we try to minimize the number of breaking releases, and there's already a list of important points to change/improve in the next breaking release.

Isn't that always the case? Why shouldn't we fix the issue at hand in the next breaking release? It is breaking, and nothing is required downstream if we provide a deprecation warning, and document the breaking change properly.

Is there a plan somewhere of the next breaking release? Can you please share it?

devmotion:
> Why shouldn't we fix the issue at hand in the next breaking release?

I'm not against changing it in the next breaking release (I still think that it's somewhat questionable to call it a fix) but I'm opposed to making a breaking release where this is the only (?) breaking change and all the breaking changes that have been discussed (more or less publicly) for a possible future breaking release are not addressed.

I also found the issue again that I had in mind: https://github.com/JuliaStats/Distributions.jl/issues/1317

juliohm:
> I'm opposed to making a breaking release where this is the only (?) breaking change

But I did not ask for that. What I want to know is which of the two PRs proposed is preferred for the next breaking release. This one (my preferred choice), or a new one that simply unexports the generic names?

mschauer:
What is the current practice of this dichotomy? I don’t trust the docs very much, I assume it’s a mixture between the eltype of the samples and other notions of „support“ and „discrete“ and „continuous“

juliohm:
@mschauer I understand that it refers to the support of the CDF of the random variable. In any case, the exact meaning here shouldn't affect the issue at hand: the generic `Continuous` and `Discrete` names are exported and conflict with other modeling packages that use the words in the more general sense (e.g., continuous variable, not necessarily random).

devmotion:
> what is the preferred solution to this problem?

My preferred solution at the time being is to not do anything - there doesn't seem to be an obvious bug to me, the names have been exported for a long time without issue, the problem can be avoided easily on the user side by only loading desired functionality (e.g., using `using Distribution: ...` or `import Distributions`), it's unclear how many users are actually affected by inconveniences because other packages started to export the same names, and probably a breaking change wouldn't make it into a new release in the near future anyway.

PR #1874, opened 2024-06-26
Add mean et al. for truncated log normal

### LINKED ISSUE #709: Error calculating mean of Truncated Log Normal
Any idea why I get the following error with this command:

`    mean(Truncated(LogNormal(1.0,5.0),0.0,1.0e5))
`
`MethodError: no method matching start(::Distributions.Truncated{Distributions.LogNormal{Float64},Distributions.Continuous})[0m
Closest candidates are:
  start([91m::SimpleVector[39m) at essentials.jl:258
  start([91m::Base.MethodList[39m) at reflection.jl:560
  start([91m::ExponentialBackOff[39m) at error.jl:107
  ...
mean(::Base.#identity, ::Distributions.Truncated{Distributions.LogNormal{Float64},Distributions.Continuous}) at statistics.jl:19
mean(::Distributions.Truncated{Distributions.LogNormal{Float64},Distributions.Continuous}) at statistics.jl:34
include_string(::String, ::String) at loading.jl:522
include_string(::String, ::String, ::Int64) at eval.jl:30
include_string(::Module, ::String, ::String, ::Int64, ::Vararg{Int64,N} where N) at eval.jl:34
(::Atom.##100#105{String,Int64,String})() at eval.jl:75
withpath(::Atom.##100#105{String,Int64,String}, ::String) at utils.jl:30
withpath(::Function, ::String) at eval.jl:38
didWriteToREPL(::Atom.##99#104{String,Int64,String}) at repl.jl:129
hideprompt(::Atom.##99#104{String,Int64,String}) at repl.jl:65
macro expansion at eval.jl:73 [inlined]
(::Atom.##98#103{Dict{String,Any}})() at task.jl:80
`

ararslan:
This PR adds `mgf` for truncated normal and uses that to implement `mean`, `var`, `skewness`, and `kurtosis` for truncated log normal based on [this observation](https://stats.stackexchange.com/a/511377). `median` is also implemented for truncated log normal.

WIP as I've not yet added tests (nor verified correctness 😅)

Fixes #709


devmotion:
My worry (that was also expressed in issues such as https://github.com/JuliaStats/Distributions.jl/issues/968) is that generally numerical integration is challenging and a fallback might lead to silently incorrect results. It seems such a fallback would be wrong (or at least problematic) e.g. if the moments are not finite (such as e.g. for `Cauchy`).

So my general feeling is that numerical integration should maybe be restricted to a smaller subset of distributions, or maybe even only be available as a separate function. In case we want to use it more broadly, I think it would also be safer to error if the integration error estimate is too large, to reduce the probability of silently incorrect results.

PaulSoderlind:
@devmotion Is there any numerical integration in this code?

ararslan:
@devmotion, perhaps you're thinking of #1875? As @PaulSoderlind noted, there's no integration here.

devmotion:
Oh yes, indeed, it seems I commented on the wrong PR.

ararslan:
CI failures on nightly are unrelated.

src/truncated/lognormal.jl
```diff
@@ -0,0 +1,52 @@
...
+# function of the truncated normal:
+#   Let Y ~ LogNormal(μ, σ) truncated to (a, b). Then log(Y) ~ Normal(μ, σ) truncated
+#   to (log(a), log(b)), and E[Y^n] = E[(e^log(Y))^n] = E[e^(nlog(Y))] = mgf(log(Y), n).
+
+# Given `truncate(LogNormal(μ, σ), a, b)`, return `truncate(Normal(μ, σ), log(a), log(b))`
+function _truncnorm(d::Truncated{<:LogNormal})
+    μ, σ = params(d.untruncated)
+    T = partype(d)
+    a = d.lower === nothing ? nothing : log(T(minimum(d)))
+    b = d.upper === nothing ? nothing : log(T(maximum(d)))
```
devmotion, line 11: Is there a reason to use `minimum`/`maximum` instead of `d.lower`/`d.upper`?

src/truncated/normal.jl
```diff
@@ -118,6 +118,36 @@ function entropy(d::Truncated{<:Normal{<:Real},Continuous})
     0.5 * (log2π + 1.) + log(σ * z) + (aφa - bφb) / (2.0 * z)
 end
 
+function mgf(d::Truncated{<:Normal{<:Real},Continuous}, t::Real)
+    T = promote_type(partype(d), typeof(t))
```
devmotion, line 122: I guess
```suggestion
    T = float(promote_type(partype(d), typeof(t)))
```
would be more correct.

src/truncated/lognormal.jl
```diff
@@ -0,0 +1,52 @@
...
+    tn = _truncnorm(d)
+    m1 = mgf(tn, 1)
+    m2 = mgf(tn, 2)
+    m3 = mgf(tn, 3)
+    m4 = mgf(tn, 4)
+    v = m2 - m1^2
+    return @horner(m1, m4, -4m3, 6m2, 0, -3) / v^2 - 3
+end
+
+median(d::Truncated{<:LogNormal}) = exp(median(_truncnorm(d)))
```
devmotion, line 43: This suggests we should add a similar definition for `quantile` as well?

src/truncated/lognormal.jl
```diff
@@ -0,0 +1,52 @@
...
+    b = d.upper === nothing ? nothing : log(T(maximum(d)))
+    return truncated(Normal(μ, σ), a, b)
+end
+
+mean(d::Truncated{<:LogNormal}) = mgf(_truncnorm(d), 1)
+
+function var(d::Truncated{<:LogNormal})
+    tn = _truncnorm(d)
+    # Ensure the variance doesn't end up negative, which can occur due to numerical issues
+    return max(mgf(tn, 2) - mgf(tn, 1)^2, 0)
```
devmotion, line 20: Repeated evaluation of `mgf` involves repeated calculations of the same (intermediate) quantities. But my fear is that optimizing this further will lead to less readable code...

ararslan: Yeah, I had the same thought and wasn't sure what to do about it so I just... didn't address it, haha

src/truncated/lognormal.jl
ararslan: Oh good call, it turns out that this relation holds in general, not just for the median.

src/truncated/lognormal.jl
ararslan: I feel like maybe I had a reason but who knows what it was

src/truncated/lognormal.jl
ararslan: Actually, defining `median`/`quantile` is unnecessary as there's already a fallback definition for computing quantiles for truncated distributions that would be more efficient than the additional layer of indirection involved with constructing a truncated normal.

src/truncated/lognormal.jl
ararslan: Ah, I remember now: it was to handle truncation points outside of the support, e.g. `truncated(LogNormal(); lower=-1)`. In that case, `d.lower == -1` but `minimum(d) == 0`. I should add a test for that.

ararslan:
Ugh, type inference issue on 1.3 and I can't test on 1.3 locally 😑

src/truncated/lognormal.jl
```diff
@@ -0,0 +1,50 @@
...
+# function of the truncated normal:
+#   Let Y ~ LogNormal(μ, σ) truncated to (a, b). Then log(Y) ~ Normal(μ, σ) truncated
+#   to (log(a), log(b)), and E[Y^n] = E[(e^log(Y))^n] = E[e^(nlog(Y))] = mgf(log(Y), n).
+
+# Given `truncate(LogNormal(μ, σ), a, b)`, return `truncate(Normal(μ, σ), log(a), log(b))`
+function _truncnorm(d::Truncated{<:LogNormal})
+    μ, σ = params(d.untruncated)
+    T = float(partype(d))
+    a = d.lower === nothing || d.lower <= 0 ? nothing : log(T(d.lower))
+    b = d.upper === nothing || isinf(d.upper) ? nothing : log(T(d.upper))
```
devmotion, line 11: Tests on Julia 1.3 pass locally when changing this to
```suggestion
    a = d.lower === nothing ? nothing : log(T(max(d.lower, 0)))
    b = d.upper === nothing ? nothing : log(T(d.upper))
```

ararslan: That isn't functionally equivalent though; IIRC, I was relying on `a` and/or `b` being `nothing` in those cases so that `truncated` would handle it a particular way.

devmotion: But arguably this optimization (`d.lower === nothing` or `d.upper === nothing`) should already have happened, either by a user or internally, when constructing the `d = truncated(LogNormal(...))`? Maybe one shouldn't expect that unoptimized inputs lead to optimized algorithms. AFAICT the optimization of `truncated(Normal(...), ...)` is also only exploited in the case where this returns a `Normal` (`a = b = nothing`); the code for `Truncated{<:Normal}` does not seem to use the fact that a bound might be `nothing`. In the `Normal` case, arguably the `LogNormal` shouldn't be truncated in the first place.

ararslan: I'm currently sick and can only barely comprehend that message but if you'd prefer to go with your suggested change then feel free to apply it, I trust your judgement

ararslan: Did we have this conversation a few months ago? My brain similarly can't comprehend whether this is the same discussion: https://github.com/JuliaStats/Distributions.jl/pull/1874#discussion_r1659372206

I should probably just log off and go sleep

Issue #1687, opened 2023-02-28
Reproducibility with global rng & distributions with Julia 1.8.5


odunbar:
Hi folks - I'm not sure if this is a `Random` or a `Distributions` issue. It could be either/both. At the very least this may be of interest to you.

### Setup
On Julia `1.8.5` , and `Distributions v0.25.85`

### Issue 
Repeated calls to `rand(...,1)`, `N` times, no longer appears to give the same as `rand(...,N)` . 
```julia
using Random, Distributions, LinearAlgebra
seed = 1000
d = MvNormal(zeros(4),I)

Random.seed!(seed)
X = [rand(d) rand(d)]
Random.seed!(seed)
Y = rand(d,2)

X == Y # false
```

### Odd thing 1: The behaviour depends on the distribution
```julia
using Random, Distributions, LinearAlgebra
seed = 1000
d = MvNormal(zeros(3),I)

Random.seed!(seed)
X = [rand(d) rand(d)]
Random.seed!(seed)
Y = rand(d,2)

X == Y # true
```
i.e. changing to a `<4` dimensional `MvNormal` does not display the same behaviour, here we do get consistency between `rand(...,1)` and `rand(...,N)`. (Is this an artifact from a method change when dim `>3`?)

### Odd thing 2: The behaviour depends on the type of rng
```julia 
using Random, Distributions, LinearAlgebra 
seed = 1000
d = MvNormal(zeros(4),I)
rng = Random.MersenneTwister(seed)
Y = rand(rng,d,3) 
rng = Random.MersenneTwister(seed)
X = [rand(rng,d) rand(rng,d) rand(rng,d)]

X == Y # true
```
i.e. using an explicit rng seems to also resolve this in these settings.

Thanks for your time! 
Cheers, 
Ollie

devmotion:
To me it seems this is not a Distributions issue:
```julia
julia> using Random

julia> Random.seed!(1000);

julia> X = [randn() for _ in 1:20];

julia> Random.seed!(1000);

julia> Y = randn(20);

julia> X == Y
false
```

Generally, Julia does not guarantee this kind of consistency. And I assume, due to #1679 and #1680 we observe the same behaviour in Distributions.

IMO it's fine - and not completely surprising - that the results deviate in this example: for performance reasons I would expect (haven't checked if that's actually the case) `randn(n)` and `randn()` to dispatch to different algorithms.

odunbar:
Thanks for clearing this up. 

[...]

PR #1879, opened 2024-07-25
Specialized vector rand! for many distributions

quildtide:
The much-larger sequel to #1269 and #1680

All claims tested on julia 1.11.0-rc1, Windows 10, AMD Ryzen 9 5900X, RTX 3080

I was originally planning on making the normal dispatch path faster for all univariate distributions on `rand!`, but it turns out that the current dispatch path is  probably about as fast as we will ever get with a generic rand -> rand! fallback path. I tested a few alternatives and they were all similar or slower.

The root cause of why this dispatch path is slower than specialized vector methods is because Julia's base `rand`/`randn`/`randexp` functions have `rand!` specializations for both Xoshiro and Mersenne that perform much faster than calling scalar `rand` many times.  [Xoshiro actually has a separate SIMD implementation in Random.jl](https://github.com/JuliaLang/julia/blob/master/stdlib/Random/src/XoshiroSimd.jl) [and both Mersenne and Xoshiro branch based off of vector length to choose optimal algorithms](https://github.com/JuliaLang/julia/blob/master/stdlib/Random/src/normal.jl#L250).

Creating specialized `rand!` methods yields some pretty impressive performance gains as a result. I plan on continuing further with this, but this pull request contains:

## Vectorized `rand!` specializations for:

- Exponential (~67% faster)
- ~~Laplace (\~185% faster)~~
- Pareto (~86% faster)
- ~~InverseGaussian (\~102% faster)~~
- LogitNormal (~46% faster)
- LogNormal (~234% faster)
- NormalCanon (~108% faster)
- ~~PGeneralizedGaussian (\~60% faster)~~

Benchmarking was done in a very ad hoc way; I grabbed 3 values from `@time`, excluding any runs that included compilation or GC time. This is a small sample size, but results were pretty consistent and I don't really feel a need to prove significance when they're all consistently >40% faster.

### Improved GPU compatibility

I don't have any non-CUDA GPUs to test with, so testing has been CUDA-only, but these things are probably true for other GPU types too.

CUDA.jl does not like the for-loop in the generic fallback `rand!` dispatch path in Distributions.jl. However, specialized dispatch paths can help alleviate this issue.

These distributions now have functional out-of-the-box CUDA.jl support for `rand!`:
- LogitNormal
- LogNormal
- NormalCanon

These distributions would probably work with the addition of a `randexp` implementation in CUDA.jl:
- Exponential
- Pareto
- ~~Laplace~~

~~`rand!(::PGeneralizedGaussian)` is dependent on `rand(::Gamma)`. CUDA.jl does not like functions that call `rand(::Gamma)`.~~

~~I'm not sure why `rand(::InverseGaussian)` isn't working. I'll have to revisit that later.~~

### Improved Dagger.jl compatibility

The current default fallback `rand!` methods perform so poorly on Dagger.jl that they cannot even really be benchmarked. This is mostly due to issues on the Dagger.jl side that are currently being addressed. I only did a tiny bit of testing with these new methods, but they do perform better than the generic fallbacks.

### Minor adjustments to scalar `rand` methods:

While I was implementing the specialized `rand!` methods, I noticed opportunities to make small changes to some scalar `rand` methods that would, in theory, yield small performance or accuracy benefits. In practice, none of these changes seem to have significant observable changes in speed:
- LogitNormal uses `muladd` now
- Lognormal uses `muladd` now
- Normal uses `muladd` now
- PGeneralizedGaussian calls `rand(rng, Bool)` instead of `rand(rng) < 0.5` now

## Additional bloating for the test suite

One negative drawback to increased specialization of methods is that there's a larger surface for bugs to emerge. This is doubly so when I'm also making small adjustments to the scalar `rand` methods.

I compensated by adding new functionality to `test_distr` and `test_samples` (enabled by keyword) that allows these functions to call the scalar `rand` functions multiple times. I also added `test_distr` entries for all of the distributions that I've added new methods for.

Unfortunately, I'm pretty sure a lot of the testing I added also duplicates work done by existing tests. :shrug:

I didn't remove any existing tests because I was unsure exactly which parts were redundant. I think code coverage probably went up a bit as a side effect.

I also had to make minor tolerance adjustments to 2 tests.

## Future Plans

A lot of distributions are dependent on `rand(::Gamma)`, which is non-trivial to translate to a vectorized form. Accomplishing this would probably enable out-of-the-box GPU support and large performance gains for all dependent distributions.

Most of the distributions addressed in this pull request were picked because I thought they would be trivial or easy to implement, but there's still a ton of others remaining in this category.


quildtide:
This pull request will probably cause a repeat of #1687 on a larger scale, but I don't think it's really an issue, as this is consistent with Julia behavior. Weird trivia example regarding Random.jl:

`rand(Xoshiro(123), 7) == rand(Xoshiro(123), 7)`

`rand(Xoshiro(123), 7) != rand(Xoshiro(123), 8)[1:7]`

quildtide:
@devmotion Could I get this reviewed if the scope of the pull request is not too wide?

src/univariate/continuous/pgeneralizedgaussian.jl
```diff
@@ -141,9 +141,20 @@ function rand(rng::AbstractRNG, d::PGeneralizedGaussian)
...
     z = d.α * rand(rng, g)^inv_p
-    if rand(rng) < 0.5
+    if rand(rng, Bool)
         return d.μ - z
     else
         return d.μ + z
     end
 end
+
+function rand!(rng::AbstractRNG, d::PGeneralizedGaussian, A::AbstractArray{<:Real})
```
devmotion, line 151: This function suffers from the same problems as mentioned above, can you remove it from the PR?

src/univariate/continuous/pareto.jl
```diff
@@ -112,6 +112,12 @@ quantile(d::Pareto, p::Real) = cquantile(d, 1 - p)
 
 rand(rng::AbstractRNG, d::Pareto) = d.θ * exp(randexp(rng) / d.α)
 
+rand!(rng::AbstractRNG, d::Pareto, A::AbstractArray{<:Real}) =
+    # A .* 1/d.α is faster than A ./ d.α, but less exact
+    # fast-math style optimization; should be fine for PRNG
+    A .=  d.θ .* exp.(randexp!(rng, A) .* (1/d.α))
+
+
```
devmotion, line 120: I think using `randexp!` is an obvious improvement but I'm not sure whether `1/...` and the inconsistency between `rand` and `rand!` is worth it (generally `1/...` can easily lead to undesired promotions but I don't think that's a problem here). I'd use something like

```suggestion
xval(d::Pareto, z::Real) = d.θ * exp(z / d.α)

rand(rng::AbstractRNG, d::Pareto) = xval(d, randexp(rng))
function rand!(rng::AbstractRNG, d::Pareto, A::AbstractArray{<:Real})
    randexp!(rng, A)
    map!(Base.Fix1(xval, d), A, A)
    return A
end

```

src/univariate/continuous/normalcanon.jl
```diff
@@ -89,6 +89,12 @@ invlogccdf(d::NormalCanon, lp::Real) = xval(d, norminvlogccdf(lp))
 
 rand(rng::AbstractRNG, cf::NormalCanon) = cf.μ + randn(rng) / sqrt(cf.λ)
 
+rand!(rng::AbstractRNG, cf::NormalCanon, A::AbstractArray{<:Real}) =
+    # A .* inv(sqrt(cf.λ)) is faster but less accurate than A ./ sqrt(cf.λ)
+    # in floating point. This is a fast-math style optimization.
+    # It shouldn't be a problem for PRNG.
+    A .= muladd.(randn!(rng, A), inv(sqrt(cf.λ)), cf.μ)
```
devmotion, line 96: I think using `randn!` os an obvious improvement but I'm not sure whether `inv` + `muladd` and the inconsistency between `rand` and `rand!` is worth it (generally `inv` can easily lead to undesired promotions but I don't think that's a problem here). I'd use something like


```suggestion
function rand!(rng::AbstractRNG, cf::NormalCanon, A::AbstractArray{<:Real})
    randn!(rng, A)
    A .= cf.μ .+ A ./ sqrt(cf.λ)
    return A
end
```

src/univariate/continuous/exponential.jl
```diff
@@ -107,6 +107,8 @@ cf(d::Exponential, t::Real) = 1/(1 - t * im * scale(d))
 #### Sampling
 rand(rng::AbstractRNG, d::Exponential) = xval(d, randexp(rng))
 
+rand!(rng::AbstractRNG, d::Exponential, A::AbstractArray{<:Real}) =
+    A .= xval.(d, randexp!(rng, A))
```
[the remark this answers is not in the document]
quildtide: I recall testing something similar before with `map!`, and it was slightly slower and caused allocations, but your version has no allocations and is slightly faster.

src/univariate/continuous/inversegaussian.jl
```diff
@@ -169,8 +169,23 @@ function rand(rng::AbstractRNG, d::InverseGaussian)
...
+function rand!(rng::AbstractRNG, d::InverseGaussian, A::AbstractArray{<:Real})
+    # Based off of the non-vectorized code
+    μ, λ = params(d)
+    Z = V = W = X1 = A # prevents extra heap allocs
+    randn!(rng, A)
+    V .*= Z
+    W .*= μ
+    X1 .= μ .+ μ / (2λ) .* (W .- sqrt.(W .* (4λ .+ W)))
+    A .= ifelse.( # TODO Can we avoid heap-allocing a whole float vec?
+        rand(rng, size(A)...) .>= μ ./ (μ .+ X1),
```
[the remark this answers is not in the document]
quildtide: You know, that explains why this wasn't working on CUDA earlier; hadn't thought of that.

quildtide:
@devmotion

test/testutils.jl
```diff
@@ -28,8 +28,9 @@ end
 # testing the implementation of a discrete univariate distribution
 #
 function test_distr(distr::DiscreteUnivariateDistribution, n::Int;
-                    testquan::Bool=true)
-
+                    testquan::Bool=true, rng::AbstractRNG=MersenneTwister(),
+                    test_scalar_rand::Bool=false)
```
devmotion, line 32: I'd like to keep the changes a bit more minimal - I think we can continue working with the `MersenneTwister` and test the the scalar `rand` method by default?
```suggestion
                    testquan::Bool=true)
```

test/testutils.jl
```diff
@@ -28,8 +28,9 @@ end
...
+    println("    testing $(distr)")
```
devmotion, line 33: I think tests are already quite verbose? If needed, they can also be grouped/wrapped in a `@testset`.
```suggestion
```

test/testutils.jl
```diff
@@ -40,7 +41,12 @@ function test_distr(distr::DiscreteUnivariateDistribution, n::Int;
 
     test_stats(distr, vs)
     test_samples(distr, n)
-    test_samples(distr, n, rng=MersenneTwister())
+    test_samples(distr, n, rng=rng)
+    if test_scalar_rand
+        xs = test_samples(distr, n; call_scalar = true)
+        xs = test_samples(distr, n, rng=rng, call_scalar = true)
+    end
+
```
devmotion, line 49: I think we can just revert to
```suggestion
    test_samples(distr, n, rng=MersenneTwister())
```

test/testutils.jl
```diff
@@ -67,7 +73,9 @@ end
 # testing the implementation of a continuous univariate distribution
 #
 function test_distr(distr::ContinuousUnivariateDistribution, n::Int;
-                    testquan::Bool=true, rng::AbstractRNG=MersenneTwister(123))
+                    testquan::Bool=true, rng::AbstractRNG=MersenneTwister(123),
+                    test_scalar_rand::Bool=false)
```
devmotion, line 77:
```suggestion
                    testquan::Bool=true, rng::AbstractRNG=MersenneTwister(123))
```

test/testutils.jl
```diff
@@ -67,7 +73,9 @@ end
...
+    println("    testing $(distr)")
```
devmotion, line 78:
```suggestion
```

test/testutils.jl
```diff
@@ -82,6 +90,13 @@ function test_distr(distr::ContinuousUnivariateDistribution, n::Int;
     allow_test_stats(distr) && test_stats(distr, xs)
     xs = test_samples(distr, n, rng=rng)
     allow_test_stats(distr) && test_stats(distr, xs)
+    if test_scalar_rand
+        xs = test_samples(distr, n; call_scalar = true)
+        allow_test_stats(distr) && test_stats(distr, xs)
+        xs = test_samples(distr, n, rng=rng, call_scalar = true)
+        allow_test_stats(distr) && test_stats(distr, xs)
+    end
+
```
devmotion, line 99:
```suggestion
```

test/testutils.jl
```diff
@@ -101,8 +116,8 @@ function test_samples(s::Sampleable{Univariate, Discrete},      # the sampleable
                       n::Int;                                   # number of samples to generate
                       q::Float64=1.0e-7,                        # confidence interval, 1 - q as confidence
                       verbose::Bool=false,                      # show intermediate info (for debugging)
-                      rng::Union{AbstractRNG, Missing}=missing) # add an rng?
-
+                      rng::Union{AbstractRNG, Missing}=missing,  # add an rng?
+                      call_scalar::Bool=false)                  # directly scall the scalar rand(d) instead
```
devmotion, line 120:
```suggestion
                      rng::Union{AbstractRNG, Missing}=missing) # add an rng?
```

test/testutils.jl
```diff
@@ -146,14 +161,30 @@ function test_samples(s::Sampleable{Univariate, Discrete},      # the sampleable
...
+        end
         Random.seed!(1234)
-        samples = rand(s, n)
-        Random.seed!(1234)
-        samples2 = rand(s, n)
+        samples2 = if !call_scalar
+            rand(s, n)
+        else
+            map((_) -> rand(s), 1:n)
+        end
```
devmotion, line 175: Can we introduce `samples3` and `samples4` unconditionally, that sample with `rand`? And then check their equality (`samples3 == samples4`) and the properties below as well?

test/testutils.jl
```diff
@@ -146,14 +161,30 @@ function test_samples(s::Sampleable{Univariate, Discrete},      # the sampleable
...
+        samples = if !call_scalar
+            rand(rng, s, n)
+        else
+            map((_) -> rand(rng, s), 1:n)
+        end
+        samples2 = if !call_scalar
+            rand(rng2, s, n)
+        else
+            map((_) -> rand(rng2, s), 1:n)
+        end
```
devmotion, line 187: Same here, test these alternatives unconditionally?

test/testutils.jl
```diff
@@ -180,8 +211,8 @@ function test_samples(s::Sampleable{Univariate, Discrete},      # the sampleable
 end
 
 test_samples(distr::DiscreteUnivariateDistribution, n::Int;
-             q::Float64=1.0e-6, verbose::Bool=false, rng=missing) =
-    test_samples(distr, distr, n; q=q, verbose=verbose, rng=rng)
+             q::Float64=1.0e-6, verbose::Bool=false, rng=missing, kwargs...) =
+    test_samples(distr, distr, n; q=q, verbose=verbose, rng=rng, kwargs...)
```
devmotion, line 215:
```suggestion
             q::Float64=1.0e-6, verbose::Bool=false, rng=missing) =
    test_samples(distr, distr, n; q=q, verbose=verbose, rng=rng)
```

test/testutils.jl
```diff
@@ -191,7 +222,8 @@ function test_samples(s::Sampleable{Univariate, Continuous},    # the sampleable
                       nbins::Int=50,                            # divide the main interval into nbins
                       q::Float64=1.0e-6,                        # confidence interval, 1 - q as confidence
                       verbose::Bool=false,                      # show intermediate info (for debugging)
-                      rng::Union{AbstractRNG, Missing}=missing) # add an rng?
+                      rng::Union{AbstractRNG, Missing}=missing, # add an rng?
+                      call_scalar::Bool=false)                  # directly scall the scalar rand(d) instead
```
devmotion, line 226:
```suggestion
                      rng::Union{AbstractRNG, Missing}=missing) # add an rng?
```

test/testutils.jl
```diff
@@ -284,8 +332,8 @@ function test_samples(s::Sampleable{Univariate, Continuous},    # the sampleable
     return samples
 end
 
-test_samples(distr::ContinuousUnivariateDistribution, n::Int; nbins::Int=50, q::Float64=1.0e-6, verbose::Bool=false, rng=missing) =
-    test_samples(distr, distr, n; nbins=nbins, q=q, verbose=verbose, rng=rng)
+test_samples(distr::ContinuousUnivariateDistribution, n::Int; nbins::Int=50, q::Float64=1.0e-6, verbose::Bool=false, rng=missing, kwargs...) =
+    test_samples(distr, distr, n; nbins=nbins, q=q, verbose=verbose, rng=rng, kwargs...)
```
devmotion, line 336:
```suggestion
test_samples(distr::ContinuousUnivariateDistribution, n::Int; nbins::Int=50, q::Float64=1.0e-6, verbose::Bool=false, rng=missing) =
    test_samples(distr, distr, n; nbins=nbins, q=q, verbose=verbose, rng=rng)
```

test/testutils.jl
```diff
@@ -246,14 +278,30 @@ function test_samples(s::Sampleable{Univariate, Continuous},    # the sampleable
     # generate samples using RNG passed or default RNG
     # we also check reproducibility
     if rng === missing
+        Random.seed!(1234)    
+        samples = if !call_scalar
+            rand(s, n)
+        else
+            map((_) -> rand(s), 1:n)
+        end
```
devmotion, line 286: Run these additional branches unconditionally?

test/univariate/continuous/exponential.jl
```diff
@@ -8,3 +7,18 @@
...
+test_cgf(Exponential(10  ), (0.08, -1, -100f0, -1e6))
+
+# Sampling Tests
+@testset "Exponential sampling tests" begin
+    for d in [
+        Exponential(1),
+        Exponential(0.91),
+        Exponential(10)
+    ]
+        test_distr(d, 10^6, test_scalar_rand = true)
```
devmotion, line 22:
```suggestion
        test_distr(d, 10^6)
```

test/univariate/continuous/logitnormal.jl
```diff
@@ -64,3 +64,12 @@ end
     @test convert(LogitNormal{Float32}, d) === d
     @test typeof(convert(LogitNormal{Float64}, d)) == typeof(LogitNormal(2,1))
 end
+
+@testset "Logitnormal Sampling Tests" begin
+    for d in [
+        LogitNormal(-2, 3),
+        LogitNormal(0, 0.2)
+    ]
+        test_distr(d, 10^6, test_scalar_rand = true)
```
devmotion, line 73:
```suggestion
        test_distr(d, 10^6)
```

test/univariate/continuous/lognormal.jl
```diff
@@ -314,3 +314,17 @@ end
...
+    for d in [
+            LogNormal()
+            LogNormal(1.0)
+            LogNormal(0.0, 2.0)
+            LogNormal(1.0, 2.0)
+            LogNormal(3.0, 0.5)
+            LogNormal(3.0, 1.0)
+            LogNormal(3.0, 2.0)
+        ]
+        test_distr(d, 10^6, test_scalar_rand = true)
```
devmotion, line 328:
```suggestion
        test_distr(d, 10^6)
```

test/univariate/continuous/normalcanon.jl
```diff
@@ -0,0 +1,10 @@
+# Sampling Tests
+@testset "NormalCanon sampling tests" begin
+    for d in [
+        NormalCanon()
+        NormalCanon(-1.0, 2.5)
+        NormalCanon(2.0, 0.8)
+    ]
+        test_distr(d, 10^6, test_scalar_rand = true)
```
devmotion, line 8:
```suggestion
        test_distr(d, 10^6)
```

test/univariate/continuous/pareto.jl
```diff
@@ -0,0 +1,10 @@
+@testset "Pareto Sampling Tests" begin
+    for d in [
+            Pareto()
+            Pareto(2.0)
+            Pareto(2.0, 1.5)
+            Pareto(3.0, 2.0)
+        ]
+        test_distr(d, 10^6, test_scalar_rand = true)
```
devmotion, line 8:
```suggestion
        test_distr(d, 10^6)
```

devmotion (commented):
Can you please add newline characters at the end of the newly added (and modified) pages? Thank you!

quildtide:
@devmotion Requested changes have been made.

devmotion (approved):
Thank you, @quildtide!

PR #1882, opened 2024-08-09
Fix InverseGaussian cdf overflows

### LINKED ISSUE #1873: cdf of InverseGaussian
Hi,

I have noticed a problem with the cdf of InverseGaussian distribution which can potentially be equal to Inf. This is a problem since a cdf is probability and consequently must be between 0 and 1. For example,

d = InverseGaussian(1.65,590)
cdf(d, 2.)

returns Inf. In addition, if you plot the empirical histogram, you can see that 2.0 is not so extreme with respect to the simulated values.

histogram(rand(d, 10000),normalize=:pdf)

For comparison, in the same situation R provides the good value of the cdf which is equal to 1 in this case.

Best regards.

Laurent

quildtide:
Closes #1873

Originally I intended to just add `clamp(z, 0, 1)`, but when I found the source of the overflow, I realized there was room for something better.

The original example in the issue, currently:
```julia
julia> d = InverseGaussian(1.65,590)
InverseGaussian{Float64}(μ=1.65, λ=590.0)

julia> cdf(d, 2.)
Inf

julia> exp(logcdf(d, 2.))
0.9998791763271095
```

With this pull request:
```julia
julia> d = InverseGaussian(1.65,590)
InverseGaussian{Float64}(μ=1.65, λ=590.0)

julia> cdf(d, 2.)
0.9998791763271095

julia> exp(logcdf(d, 2.))
0.9998791763271095
```

There is also a backup clamp just in case someone manages to go even further and break the primary fix.

### DIFF
--- a/src/univariate/continuous/inversegaussian.jl
+++ b/src/univariate/continuous/inversegaussian.jl
@@ -102,7 +102,7 @@ function cdf(d::InverseGaussian, x::Real)
     # 2λ/μ and normlogcdf(-u*(v+1)) are similar magnitude, opp. sign
     # truncating to [0, 1] as an additional precaution
     # Ref https://github.com/JuliaStats/Distributions.jl/issues/1873
-    z = clamp(normcdf(u * (v - 1)) + exp(2λ / μ + normlogcdf(-u * (v + 1)), 0, 1))
+    z = clamp(normcdf(u * (v - 1)) + exp(2λ / μ + normlogcdf(-u * (v + 1))), 0, 1)
 
     # otherwise `NaN` is returned for `+Inf`
     return isinf(x) && x > 0 ? one(z) : z
@@ -116,7 +116,7 @@ function ccdf(d::InverseGaussian, x::Real)
     # 2λ/μ and normlogcdf(-u*(v+1)) are similar magnitude, opp. sign
     # truncating to [0, 1] as an additional precaution
     # Ref https://github.com/JuliaStats/Distributions.jl/issues/1873
-    z = clamp(normccdf(u * (v - 1)) - exp(2λ / μ + normlogcdf(-u * (v + 1)), 0, 1))
+    z = clamp(normccdf(u * (v - 1)) - exp(2λ / μ + normlogcdf(-u * (v + 1))), 0, 1)
 
     # otherwise `NaN` is returned for `+Inf`
     return isinf(x) && x > 0 ? zero(z) : z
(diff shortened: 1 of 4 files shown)


src/univariate/continuous/inversegaussian.jl
```diff
@@ -110,7 +112,8 @@ function ccdf(d::InverseGaussian, x::Real)
     y = max(x, 0)
     u = sqrt(λ / y)
     v = y / μ
-    z = normccdf(u * (v - 1)) - exp(2λ / μ) * normcdf(-u * (v + 1))
+    z = normccdf(u * (v - 1)) - exp(2λ / μ + normlogcdf(-u * (v + 1)))
+    z = clamp(z, 0, 1)
```
devmotion, line 116: Same here:
```suggestion
    # 2λ/μ and normlogcdf(-u*(v+1)) are similar magnitude, opp. sign
    # truncating to [0, 1] as an additional precaution
    # Ref https://github.com/JuliaStats/Distributions.jl/issues/1873
    z = clamp(normccdf(u * (v - 1)) - exp(2λ / μ + normlogcdf(-u * (v + 1)), 0, 1)
```

test/univariate/continuous.jl
```diff
@@ -99,6 +99,19 @@ end
...
+        InverseGaussian(0.5, 1000)
+    ]
+        for x in [0.02, 1.0, 20.0, 300.0]
+            p = cdf(d, x)
+            @test 0.0 <= p <= 1.0
+            @test p ≈ exp(logcdf(d, x))
+        end
+    end
+end
+
```
devmotion, line 114: Can you please move the test to test/univariate/continuous/inversegaussian.jl?

quildtide: That was my original plan, but that file doesn't currently exist, and introducing it in this pull request would probably cause a merge conflict with this other recent pull request of mine, #1879 , as that one also introduces that file.

devmotion: I had missed #1879 but to get this PR in I'd suggest creating the test file anyway and then fixing the merge conflicts in #1879 (which should be easy to resolve).

quildtide: Sounds good.

test/univariate/continuous/inversegaussian.jl
```diff
@@ -0,0 +1,12 @@
+@testset "InverseGaussian cdf outside of [0, 1] (#1873)" begin
+    for d in [
+        InverseGaussian(1.65, 590),
+        InverseGaussian(0.5, 1000)
+    ]
+        for x in [0.02, 1.0, 20.0, 300.0]
+            p = cdf(d, x)
+            @test 0.0 <= p <= 1.0
+            @test p ≈ exp(logcdf(d, x))
```
devmotion, line 9: Can you extend the tests to `ccdf`?

devmotion (commented):
Thank you, looks good!

devmotion (approved):

Issue #1884, opened 2024-08-13
Is there a reason `rand(TDist(nu))` does not obey the type of `nu`?


Red-Portal:
Hi, it seems that 
```julia
nu = 3f0
rand(TDist(nu))
```
returns a `Float64` instead of a `Float32`. Is this intended behavior? I find it unusual because the `rand` [implementation](https://github.com/JuliaStats/Distributions.jl/blob/13029c03fa885a2b4512b954e61f9d5a7dfa0612/src/univariate/continuous/tdist.jl#L85) tries to be type-stable half way there by using `one(T)` but ends up calling `randn(rng)` without specifying the type.

devmotion:
Indeed, this seems to be a bug. `randn(rng)` should probably be changed to `randn(rng, typeof(z))`.

Red-Portal:
Okay let me pick this up.

Red-Portal:
Hi @devmotion I opened a PR, could you take a look when you have time?

quildtide:
Has the situation from #1071, #1666, etc. changed?

devmotion:
> I'm not at all against the idea that eltype(d) should equal partype(d)

That's absolutely not the idea here. In the case of `TDist`, we get a sample from ChiSq (of some type, regardless of whether it follows any conventions or not). And IMO regardless of what type it is, it seems wrong to modify/promote its type by a `randn` call (basically a "Float64 literal").

Red-Portal:
Hi @devmotion , unfortunately, it seems that calls to `rand(q, d)` still result in the wrong type. I guess this is because `eltype(q)` returns `Float64`. Should I make a PR that defines `eltype(q) = partype(q)`?

quildtide:
One option within the current interface, if you need more than one random value, is to call the `rand!` method on a vector of the type you want, e.g. `rand!(Exponential(), zeros(Float32, 20))`.

I think there'll still be a double conversion going on (Float32 -> Float64-> Float32) if you call this on the current public version, but I think a side-effect of merging #1879 will cause things to stay `Float32` at all times. Not 100% sure on this though; hadn't given this much thought before.

devmotion:
[...]

PR #1885, opened 2024-08-13
fix type stability of sampling from `Chisq`, `TDist`, `Gamma`

Red-Portal:
This addresses #1884

### DIFF
--- a/src/univariate/continuous/exponential.jl
+++ b/src/univariate/continuous/exponential.jl
@@ -105,7 +105,7 @@ cf(d::Exponential, t::Real) = 1/(1 - t * im * scale(d))
 
 
 #### Sampling
-rand(rng::AbstractRNG, d::Exponential) where {T} = xval(d, randexp(rng, float(T)))
+rand(rng::AbstractRNG, d::Exponential{T}) where {T} = xval(d, randexp(rng, float(T)))
 
 
 #### Fit model
--- a/src/univariate/continuous/exponential.jl
+++ b/src/univariate/continuous/exponential.jl
@@ -105,7 +105,7 @@ cf(d::Exponential, t::Real) = 1/(1 - t * im * scale(d))
 
 
 #### Sampling
-rand(rng::AbstractRNG, d::Exponential{T}) where {T} = xval(d, randexp(rng, T))
+rand(rng::AbstractRNG, d::Exponential) = xval(d, randexp(rng))
 
 
 #### Fit model
--- a/src/univariate/continuous/exponential.jl
+++ b/src/univariate/continuous/exponential.jl
@@ -105,7 +105,7 @@ cf(d::Exponential, t::Real) = 1/(1 - t * im * scale(d))
 
 
 #### Sampling
-rand(rng::AbstractRNG, d::Exponential) = xval(d, randexp(rng))
+rand(rng::AbstractRNG, d::Exponential) = xval(d, randexp(rng, float(T)))
 
 
 #### Fit model
--- a/src/univariate/continuous/exponential.jl
+++ b/src/univariate/continuous/exponential.jl
@@ -105,7 +105,7 @@ cf(d::Exponential, t::Real) = 1/(1 - t * im * scale(d))
 
 
 #### Sampling
-rand(rng::AbstractRNG, d::Exponential) = xval(d, randexp(rng))
+rand(rng::AbstractRNG, d::Exponential{T}) where {T} = xval(d, randexp(rng, T))
 
 
 #### Fit model
--- a/src/univariate/continuous/exponential.jl
+++ b/src/univariate/continuous/exponential.jl
@@ -105,7 +105,7 @@ cf(d::Exponential, t::Real) = 1/(1 - t * im * scale(d))
 
 
 #### Sampling
-rand(rng::AbstractRNG, d::Exponential) = xval(d, randexp(rng, float(T)))
+rand(rng::AbstractRNG, d::Exponential) where {T} = xval(d, randexp(rng, float(T)))
 
 
 #### Fit model
--- a/src/univariate/continuous/tdist.jl
+++ b/src/univariate/continuous/tdist.jl
@@ -82,7 +82,7 @@ end
 function rand(rng::AbstractRNG, d::TDist)
     ν = d.ν
     z = sqrt(rand(rng, Chisq{typeof(ν)}(ν)) / ν)
-    return randn(rng) / (isinf(ν) ? one(z) : z)
+    return randn(rng, typeof(ν)) / (isinf(ν) ? one(z) : z)
 end
 
 function cf(d::TDist{T}, t::Real) where T <: Real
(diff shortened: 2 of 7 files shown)


quildtide:
Should probably add method `eltype(d::Gamma) = partype(d)` and friends.

There are many side effects associated with changing Gamma. The following list is not exhaustive, but I think they will be impacted:

- Beta
- BetaPrime
- InverseGamma
- Chi


Chisq will have some side effects too, e.g.:
- FDist

Red-Portal:
Hi @quildtide ,

Okay so to be precise, I'll do the following:
* Identify all the distributions that are affected by sampling Gammas. If the type of the sampled gamma affects the type of the final sample, add `eltype = partype`
* Add tests for the type stability

Would that be sufficient?

quildtide:
Not a maintainer, so don't take my word for granted. If the route this pull request takes is chosen, then the things you propose doing would indeed be a good idea.

Beta has two sampling paths, IIRC, where it sometimes uses a Gamma sampler, and sometimes uses a different sampler, so you will probably need to alter the other path too, if the Gamma path is returning partype values now (I think it is).

----

I know that some of the maintainers have pushed back to some degree in the past on making `eltype == partype` for Distributions that aren't already doing that (Normal and a few friends), but @devmotion seems to be calling the shots right now, and he seems to back the core idea of this pull req for `TDist`.

src/univariate/continuous/tdist.jl
```diff
@@ -82,7 +82,7 @@ end
 function rand(rng::AbstractRNG, d::TDist)
     ν = d.ν
     z = sqrt(rand(rng, Chisq{typeof(ν)}(ν)) / ν)
-    return randn(rng) / (isinf(ν) ? one(z) : z)
+    return randn(rng, typeof(ν)) / (isinf(ν) ? one(z) : z)
```
devmotion, line 85: IMO it shouldn't depend on the parameter (which could any non-floating point number) but floating point number `z` (whose type we do not want to mess up, similar to also the case of the `Gamma` sampler above):
```suggestion
    return randn(rng, typeof(z)) / (isinf(ν) ? one(z) : z)
```

src/univariate/continuous/exponential.jl
```diff
@@ -105,7 +105,7 @@ cf(d::Exponential, t::Real) = 1/(1 - t * im * scale(d))
 
 
 #### Sampling
-rand(rng::AbstractRNG, d::Exponential) = xval(d, randexp(rng))
+rand(rng::AbstractRNG, d::Exponential{T}) where {T} = xval(d, randexp(rng, T))
```
devmotion, line 108: This is a slightly different case and easily broken (e.g., when `T` is not a floating point number type). In the `TDist` and `Gamma` case we just try to avoid promotions of a sample from another `rand` call, whereas this case goes deeper into the question of how `rand` should behave wrt parameters etc. (see also https://github.com/JuliaStats/Distributions.jl/pull/1433#issuecomment-2293472385).
```suggestion
rand(rng::AbstractRNG, d::Exponential) = xval(d, randexp(rng))
```

test/univariate/continuous/chisq.jl
```diff
@@ -1,2 +1,8 @@
-test_cgf(Chisq(1), (0.49, -1, -100, -1f6))
-test_cgf(Chisq(3), (0.49, -1, -100, -1f6))
+
+@testset "Chisq" begin
+    test_cgf(Chisq(1), (0.49, -1, -100, -1.0f6))
+    test_cgf(Chisq(3), (0.49, -1, -100, -1.0f6))
+
+    @test rand(Chisq(1.0)) isa Float64
+    @test rand(Chisq(1.0f0)) isa Float32
```
devmotion, line 7: This just happens to be the case but IMO it is not really something that we have (or should) generalize in this PR.

Red-Portal: Okay then let me axe this.

src/univariate/continuous/exponential.jl
Red-Portal: If we do this though, the return type of `rand(Gamma(Float32, Float32))` changes depending on the value of the shape parameter because `shape == 1` samples from `Exponential`. Should we let this happen? I imagine some people will be super surprised by such behavior.

Red-Portal: Hi @devmotion could you comment on this?

devmotion: I think that's an argument for using the same approach as for `Normal` here, until we move to a better/different API:
```suggestion
rand(rng::AbstractRNG, d::Exponential{T}) where {T} = xval(d, randexp(rng, float(T)))
```

src/univariate/continuous/exponential.jl
```diff
@@ -105,7 +105,7 @@ cf(d::Exponential, t::Real) = 1/(1 - t * im * scale(d))
 
 
 #### Sampling
-rand(rng::AbstractRNG, d::Exponential) = xval(d, randexp(rng))
+rand(rng::AbstractRNG, d::Exponential) where {T} = xval(d, randexp(rng, float(T)))
```
devmotion, line 108:
```suggestion
rand(rng::AbstractRNG, d::Exponential{T}) where {T} = xval(d, randexp(rng, float(T)))
```

test/univariate/continuous/gamma.jl
```diff
@@ -1,25 +1,35 @@
...
+        @test_throws DimensionMismatch suffstats(Gamma, a, wb)
+    end
+
+    @test rand(Gamma(1.0, 1.0)) isa Float64
+    @test rand(Gamma(0.5, 1.0)) isa Float64
+    @test rand(Gamma(2.0, 1.0)) isa Float64
+
+    @test rand(Gamma(1.0f0, 1.0f0)) isa Float32
+    @test rand(Gamma(0.5f0, 1.0f0)) isa Float32
+    @test rand(Gamma(2.0f0, 1.0f0)) isa Float32
```
devmotion, line 34: Are these inferred? Maybe we could check with `@inferred`?
```suggestion
    for T in (Float32, Float64)
        @test @inferred(rand(Gamma(T(1), T(1)))) isa T
        @test @inferred(rand(Gamma(1/T(2), T(1)))) isa T
        @test @inferred(rand(Gamma(T(2), T(1)))) isa T
    end
```

test/univariate/continuous/tdist.jl
```diff
@@ -3,10 +3,19 @@ using ForwardDiff
...
+        @inferred(rand(TDist(ForwardDiff.Dual(1.0))))
+
     end
-    @inferred(rand(TDist(ForwardDiff.Dual(1.0))))
+
+    @test rand(TDist(1.0)) isa Float64
+    @test rand(TDist(1.0f0)) isa Float32
+
+    @test entropy(TDist(1.0)) isa Float64
+    @test entropy(TDist(1.0f0)) isa Float32
```
devmotion, line 20: Similarly here, maybe also check type inference?
```suggestion
    for T in (Float32, Float64)
        @test @inferred(rand(TDist(T(1)))) isa T
        @test @inferred(entropy(TDist(T(1)))) isa T
    end
```

devmotion (commented):
Can you also add a test for `Exponential`?

Red-Portal:
I also added back the type stability tests for `Chisq` since we now know that the gamma samplers respect the parameter type.

devmotion (approved):

Issue #1810, opened 2023-12-11
random log-gamma for taming underflow issues


chelate:
Hi all sorry about this. I just have this function from an old project, it works really well for log-gamma sampling for small shape parameter

```julia

function randlogGamma(a;scale = 1)
    if a > .2
        return log.(rand(Gamma(a))) #fall back is fine, just not perfect Float-dense
    else
        L = 1 / a - 1
        w = a / MathConstants.e / (1 - a)
        ww = 1 / (1 + w)
        η = z -> ( z >= 0 ? exp(-z) : w * L * exp(L * z))
        h = z -> exp(-z - exp(-z / a))
        function rh(a)
            while true
                U = rand()
                z = (U <= ww) ? -log(U / ww) : log(rand()) / L
                h(z) / η(z) > rand() && return(z)
            end
        end
        return log(scale) - rh(a) / a
    end
end
```

It is especailly useful for stable beta sampling when α/β are super small. I don't know who wrote the code anymore, but it's very useful, I've had to copy it to a few projects now. 

The problem is, without this you don't get the right power laws because of underflow for small a. Ideally, for best possible numerical stability, beta and Dirichlet would always be defined through `logsumexp` acting on vectors of these guys to enforce normalization and keep log(0) at bay.

I know just posting code instead of a pr is terrible manners and I will see if I can figure out how to do that. But we Really need a log-gamma sampler to do gamma and beta properly!

chelate:
see #1003 #1024

Issue #1003, opened 2019-10-29
Issue with support of Beta distribution when alpha < 1 or beta < 1


MFairley:
I am having a numerical issue with sampling from a Beta distribution when alpha or beta is less than 1. E.g. sometimes a sample will come back as exactly 1.0 but the support of the Beta distribution does not include 1 when beta is less than 1 and the pdf evaluates to Inf due to division by 0.

devmotion:
I just ran into the same issue:
```julia
using Distributions
using Random
using Test

Random.seed!(1234)
v = rand(Beta(0.01, 1.01), 10_000)
@test_broken !any(iszero, v)
```
The standard implementation in R does not suffer from the same problem:
```R
set.seed(1234)
v <- rbeta(n=10000, shape1=0.01, shape2=1.01)
0 %in% v && stop("incorrect samples")
```
It seems the [standard implementation in R](https://github.com/wch/r-source/blob/trunk/src/nmath/rbeta.c) uses the algorithms BB and BC from
[R. C. H. Cheng (1978). Generating beta variates with nonintegral shape parameters. Communications of the ACM, 21, 317–322](https://dl.acm.org/citation.cfm?id=359482). In general, it might be interesting to have a look at [Ying-Chao Hung, Narayanaswamy Balakrishnan & Yi-Te Lin (2009). Evaluation of Beta Generation Algorithms. Communications in Statistics - Simulation and Computation, 38:4, 750-770](https://www.tandfonline.com/doi/full/10.1080/03610910802645347) for a comparison of different algorithms for sampling from the Beta distribution.

devmotion:
I think the main issue with the current implementation is that the Gamma sampler returns incorrect values:
```julia
using Distributions
using Random
using Test

Random.seed!(1234)
v = rand(Gamma(0.01, 1.0), 10_000)
@test_broken !any(iszero, v)
```

andreasnoack:
> I think the main issue with the current implementation is that the Gamma sampler returns incorrect values:

Why do you conclude that the values are incorrect? The examples seems similar to
```julia
julia> any(iszero, rand(10000))
false
```

There is zero probability at zero. The probability in a small neighborhood looks okay
```julia
julia> x = rand(Gamma(0.01, 1.0), 10000);

julia> mean(t -> t < eps(), x)
0.7067

julia> cdf(Gamma(0.01, 1.0), eps())
0.7013514054165827
```

devmotion:
[...]

PR #1886, opened 2024-08-16
Fix Dirichlet rand overflows #1702

### LINKED ISSUE #1702: sampling from Dirichlet produces NaN and Inf at extreme alpha
In my code a random process generates alpha vectors for Dirichlet distributions and on rare occasions these can have extreme values. When sampling with an alpha where all values are extremely low the rand() function returns an array of NaN, sometimes with an Inf in there. Here is an example:
```julia
rand(Dirichlet([8e-5, 1e-5, 2e-5]))
```
At slightly less extreme values it produce NaN arrays less frequently, but still sometimes:
```julia
rand(Dirichlet([8e-4, 1e-4, 2e-4]))
```
It seem to only happen when __all__ values are low.

I don't know if this is really a bug. There is no practical difference between these examples and using less extreme values that works (they are so extreme they will always produce certainty in one category). So it is easy to make a work around to makes my code run. However, I thought report it anyway as it might highlight a deeper, less noticeable problem?

I have also experienced it producing a vector where the values doesn't sum to 1. So much that the Categorical distribution throws and error saying that it is not a probability vector. But I failed to make a reproducible example (might be a bug of my own then?). Is this something there could be an explanation for?

quildtide:
Closes #1702 

## Core Issues

The `rand(d::Dirichlet)` calls `Gamma(d.α[i])` i times and writes to `x`.

It then rescales this result by `inv(sum(x))`. When this overflows to `Inf`, we run into our 2 failure modes:

1. When all x_i == 0, we get Inf * 0 = NaN

2. When some x_i != 0, but are all deeply subnormal enough that `inv(sum(x))` still overflows. We get some Inf values as a result.

For case 2, on Julia 1.11.0-rc1 on Windows, for example:
```julia
julia> rand(Xoshiro(123322), Dirichlet([4.5e-5, 4.5e-5, 8e-5]))
3-element Vector{Float64}:
  Inf
  Inf
 NaN
```

## Fixing Case 1

If case 1 is happening, the best thing possible from a runtime perspective is probably to just choose a random x from a categorical distribution with the same mean. This is the limit behavior of the Dirichlet distribution, and my logic on why it's "safe enough" is:
- If all-zeros are a rare occurance, this has little impact on the end sample
- If all-zeros are common, rejecting samples and pulling another will probably yield a near-infinite reject loop. On the other hand, we're close enough to the limit behavior that floating point arithmetic errors are probably hurting us more than adopting the limit behavior.
- While this should theoretically result in incorrect variance, testing shows that variance is within reasonable tolerance (0.01) of the real value.

There is another option where we could try rejecting all-0 samples until a certain maximum amount of samples before failing, but I think this is probably a waste of time for little gain in accuracy.

## Fixing Case 2

We rescale all values by multiplying them by floatmax(), so `inv` doesn't overflow. This should work consistently for all float types where `floatmax() * nextfloat() > floatmin()` by at least ~1 magnitudes, which I think should be true for any non-exotic float types. I originally thought it would be enough to just set the largest value to 1, but it's actually possible to currently pull multiple subnormal values pre-normalization, and the method I adopted maintains the ratio between them.

Currently:
```julia
julia> rand(Xoshiro(123322), Dirichlet([4.5e-5, 4.5e-5, 8e-5]))
3-element Vector{Float64}:
  Inf
  Inf
 NaN
```

After this patch:
```julia
julia> rand(Xoshiro(123322), Dirichlet([4.5e-5, 4.5e-5, 8e-5]))
3-element Vector{Float64}:
  0.625061099164708
  0.37493890083529186
  0.0
```

## Subnormal Parameters

While testing, I realized that my original fix for case 1 would break when all of the parameters themselves were deeply subnormal, e.g. `Dirichlet([5e-321, 1e-321, 4e-321])`. Given that the Dirichlet distribution is decently common in things like Bayesian inference, I thought it would be worth attempting to support these cases too.

Note that `mean`, `var`, etc. currently break on these deeply subnormally-parameterized distributions, but fixing that felt out of scope to this pull request. Fixing `mean` would be simple, but it could potentially be rather chunky. I am less sure about `var` and others.


src/multivariate/dirichlet.jl
```diff
@@ -154,20 +154,49 @@ end
...
+    x::AbstractVector{<:Real}
+    )
+    Σ = sum(x)
+    if Σ == 0.0
+        # Distribution behavior approaches categorical as Σα -> 0
+        α = d.alpha
+        iΣα = inv(d.alpha0)
+        if isinf(iΣα)
+            # Dirichlet with ALL deeply subnormal parameters
+            α .*= floatmax(eltype(α))
```
ararslan, line 169: Though this new function is only being used inside of mutating functions, those mutating functions only mutated `x`, whereas this is modifying the `Distribution`'s parameter array. I would find that behavior unexpected as a user.

devmotion:
Instead of dealing with subnormals, at least for the example here sampling in log space would be sufficient (see also https://github.com/JuliaStats/Distributions.jl/issues/1003#issuecomment-636450042, https://github.com/JuliaStats/Distributions.jl/issues/1003#issuecomment-556978582, and https://github.com/JuliaStats/Distributions.jl/issues/1810). For instance, with an `ExpGamma` version of the Marsaglia sampler I get:
```julia

julia> using Distributions: GammaMTSampler, logsumexp

julia> # Inverse Power sampler in log-space (exp-gamma distribution)
       # uses the x*u^(1/a) trick from Marsaglia and Tsang (2000) for when shape < 1
       struct ExpGammaIPSampler{S<:Sampleable{Univariate,Continuous},T<:Real} <: Sampleable{Univariate,Continuous}
           s::S #sampler for Gamma(1+shape,scale)
           nia::T #-1/scale
       end

julia> ExpGammaIPSampler(d::Gamma) = ExpGammaIPSampler(d, GammaMTSampler)

julia> function ExpGammaIPSampler(d::Gamma, ::Type{S}) where {S<:Sampleable}
           shape_d = shape(d)
           sampler = S(Gamma{partype(d)}(1 + shape_d, scale(d)))
           return ExpGammaIPSampler(sampler, -inv(shape_d))
       end

julia> function rand(rng::AbstractRNG, s::ExpGammaIPSampler)
           x = log(rand(rng, s.s))
           e = randexp(rng)
           return muladd(s.nia, e, x)
       end

julia> function myrand!(rng::AbstractRNG, d::Dirichlet, x::AbstractVector{<:Real})
           for (i, αi) in zip(eachindex(x), d.alpha)
               @inbounds x[i] = rand(rng, ExpGammaIPSampler(Gamma(αi)))
           end
           x .-= logsumexp(x)
           x .= exp.(x)
           return x
       end

julia> myrand!(Xoshiro(123322), Dirichlet([4.5e-5, 4.5e-5, 8e-5]), zeros(3))
3-element Vector{Float64}:
 0.6250610991638559
 0.37493890083615117
 0.0
```

quildtide:
> For instance, with an `ExpGamma` version of the Marsaglia sampler I get:

Okay, after doing some testing, this implementation seems to be superior to what I was doing until sum(alpha) itself is subnormal enough. 

With your example implementation:
```julia
julia> myrand!(Random.default_rng(), Dirichlet([6e-309, 5e-309, 5e-309]), zeros(3))
3-element Vector{Float64}:
 1.0
 0.0
 0.0

julia> myrand!(Random.default_rng(), Dirichlet([5e-309, 5e-309, 5e-309]), zeros(3))
3-element Vector{Float64}:
 NaN
 NaN
 NaN
 ```

I brought in the code snippet from #1810 and that worked for a bit longer:
```julia
julia> function myrand2!(rng::AbstractRNG, d::Dirichlet, x::AbstractVector{<:Real})
                  for (i, αi) in zip(eachindex(x), d.alpha)
                      @inbounds x[i] = randlogGamma(αi)
                  end
                  return softmax!(x)
           end
julia> myrand2!(Random.default_rng(), Dirichlet([5e-310, 5e-310, 5e-310]), zeros(3))
3-element Vector{Float64}:
 0.0
 1.0
 0.0

julia> myrand2!(Random.default_rng(), Dirichlet([5e-311, 5e-311, 5e-311]), zeros(3))
3-element Vector{Float64}:
 NaN
 NaN
 NaN
 ```

The good news though is that there's only 1 failure mode now: when `rand(ExpGamma) == -Inf`.  I'll maintain an edge case check to go into the Categorical sampler failure mode.

**Comment by chelate** [comment]
I started writing a PR for the ExpGamma distribution and documentation. But this pr gets the dirichlet sampling right, which is really a harder problem and much more important. I will wait for it to merge and then promise to build on it, moving the undocumented methods to an expgamma.jl distribution page.

**Comment by quildtide** [comment]
@devmotion Could this be looked at again? Thanks.

**Comment by chelate** [comment]
Just wondering if there is an objection. Do we need to make expgamma.jl *before* this can be merged?

**Comment by ararslan** [comment]
I wouldn't take the lack of response as objection so much as lack of maintainer bandwidth to review and respond (certainly speaking for myself, at least). I appreciate the contribution and your patience, @quildtide.

Though I'm not currently able to provide a thoughtful review, I can say that something that will make a future reviewer's job easier would be to include comments in the code that justify the choices of 0.5 and 0.3 as cutoffs where applicable.

**Comment by quildtide** [comment]
> Though I'm not currently able to provide a thoughtful review, I can say that something that will make a future reviewer's job easier would be to include comments in the code that justify the choices of 0.5 and 0.3 as cutoffs where applicable.

The 0.3 was based off of the note in Liu, Martin, and Syring that their algorithm's acceptance rate is higher until 0.3 when compared to algorithm 3 in Kundu and Gupta. I neglected, however, to notice that we do not currently have Kundu and Gupta's algorithm 3 implemented at the moment.

The 0.5 was mostly arbitary; it was originally 1, but a test failed when it was that high.

It's possible that these cutoffs are not optimal for performance reasons; I did not have time when I made this PR to do proper performance testing. I think I may do some of that in the near future.

I am also tempted to try implementing the Kundu-Gupta sampler now, but I reckon that would only make the PR harder to review.

**Comment by quildtide** [comment]
I have pushed comments for now. I will do some performance testing to find potential better thresholds if I wind up having time to do so before this can be reviewed.

File: src/multivariate/dirichlet.jl [hunk 2]
```diff
@@ -156,18 +156,61 @@ end
...
+        # 1 is known to be too high.
+        for (i, αi) in zip(eachindex(x), d.alpha)
+            @inbounds x[i] = rand(rng, Gamma(αi))
+        end
+
+        return lmul!(inv(sum(x)), x)
+    else
+        # Sample in log-space to lower underflow risk
+        for (i, αi) in zip(eachindex(x), d.alpha)
+            @inbounds x[i] = _logrand(rng, Gamma(αi))
```
devmotion, line 172: Conceptually, I think this should be
```suggestion
            @inbounds x[i] = rand(rng, ExpGamma(αi))
```

src/multivariate/dirichlet.jl
```diff
@@ -156,18 +156,61 @@ end
...
+        end
+
+        if all(isinf, x)
```
devmotion, line 175: This seems somewhat incorrect - you're only looking for `-Inf` but checking for `Inf` and `-Inf`. For `Inf` already a single value is problematic. An additional question is why `isinf` instead of `!isfinite` which would e.g. also guard against `NaN`.

src/multivariate/dirichlet.jl
```diff
@@ -156,18 +156,61 @@ end
...
+            # Final fallback, parameters likely deeply subnormal
+            # Distribution behavior approaches categorical as Σα -> 0
+            p = copy(d.alpha)
+            p .*= floatmax(eltype(p)) # rescale to non-subnormal
+            x .= zero(E)
+            x[rand(rng, Categorical(inv(sum(p)) .* p))] = one(E)
```
devmotion, line 181: This causes many allocations and won't work in general - `p` might not be mutable.

src/multivariate/dirichlet.jl
```diff
@@ -156,18 +156,61 @@ end
...
+                x::AbstractVector{E}) where {T<:Real, E<:Real}
+    
+    if FillArrays.getindex_value(d.alpha) >= 0.5
+        # 0.5 is a placeholder; optimal value unknown
+        # 1 is known to be too high.
+        rand!(rng, Gamma(FillArrays.getindex_value(d.alpha)), x)
+        return lmul!(inv(sum(x)), x)
+    else
+        # Sample in log-space to lower underflow risk
+        _logrand!(rng, Gamma(FillArrays.getindex_value(d.alpha)), x)
```
devmotion, line 200: Same here, conceptually, this should be
```suggestion
        _rand!(rng, ExpGamma(FillArrays.getindex_value(d.alpha)), x)
```

src/multivariate/dirichlet.jl
```diff
@@ -156,18 +156,61 @@ end
...
+        
+        if all(isinf, x)
```
devmotion, line 202: Same questions as above.

src/multivariate/dirichlet.jl
```diff
@@ -156,18 +156,61 @@ end
...
     end
-    lmul!(inv(sum(x)), x) # this returns x
 end
 
 function _rand!(rng::AbstractRNG,
                 d::Dirichlet{T,<:FillArrays.AbstractFill{T}},
-                x::AbstractVector{<:Real}) where {T<:Real}
-    rand!(rng, Gamma(FillArrays.getindex_value(d.alpha)), x)
-    lmul!(inv(sum(x)), x) # this returns x
+                x::AbstractVector{E}) where {T<:Real, E<:Real}
```
devmotion, line 191: No need to change the function signature here (see comment below).

src/multivariate/dirichlet.jl
```diff
@@ -156,18 +156,61 @@ end
...
+        # Sample in log-space to lower underflow risk
+        _logrand!(rng, Gamma(FillArrays.getindex_value(d.alpha)), x)
+        
+        if all(isinf, x)
+            # Final fallback, parameters likely deeply subnormal
+            # Distribution behavior approaches categorical as Σα -> 0
+            n = length(d.alpha)
+            p = Fill(inv(n), n)
+            x .= zero(E)
+            x[rand(rng, Categorical(p))] = one(E)
```
devmotion, line 208: This won't work in general - `x` might not use one-based indexing. Moreover, there's no need to construct a `Categorical` distribution here:
```suggestion
            fill!(x, false)
            x[rand(rng, firstindex(x):lastindex(x))] = true
```
(or alternatively zero(eltype(x)) and oneunit(eltype(x))).

devmotion (commented):
My general feeling is that subnormal numbers are not a major concern in Distributions - even if there are some improvements here, I assume there are many other problems both in upstream and downstream code. Floating point numbers are inherently limited, we can only operate within their restrictions. Alternatively, you might have to switch to number types with higher or arbitrary precision.

On the other hand, I think alternative samplers and distributions such as `ExpGamma` that operate in log-space would be quite useful in different places (as evidenced by a few old issues I had opened a few years ago IIRC). So I think we should
1. separate the `ExpGamma` part, ie, add an `ExpGamma` distribution + the samplers in a separate PR and make sure they are properly tested using the existing test infrastructure for distributions and samplers
2. change this PR to use `ExpGamma` in `Dirichlet` when it's beneficial (requires numerical experiments + benchmarks)?

quildtide:
> My general feeling is that subnormal numbers are not a major concern in Distributions - even if there are some improvements here, I assume there are many other problems both in upstream and downstream code. Floating point numbers are inherently limited, we can only operate within their restrictions. Alternatively, you might have to switch to number types with higher or arbitrary precision.

I think this is a reasonable position, especially since the subnormal edge case only emerges when all alphas themselves are deeply subnormal (after implementing log-space sampling).


> On the other hand, I think alternative samplers and distributions such as `ExpGamma` that operate in log-space would be quite useful in different places (as evidenced by a few old issues I had opened a few years ago IIRC). So I think we should
> 
> 1. separate the `ExpGamma` part, ie, add an `ExpGamma` distribution + the samplers in a separate PR and make sure they are properly tested using the existing test infrastructure for distributions and samplers

I think @chelate was working on this. I can fork this branch to a branch with only the ExpGamma sampling so chelate can do a PR with that and their own work (testing, documentation, etc.).

> 2. change this PR to use `ExpGamma` in `Dirichlet` when it's beneficial (requires numerical experiments + benchmarks)?

There's 2 types of testing that can be done.

We do already know that there's a cutoff for alphas around 4e-8 where the current method (no logspace sampling) breaks completely. So we already know that anything around this cutoff is already beneficial.

But determining a high bound for the cutoff would indeed require some benchmarking.

And then there's benchmarking for when to switch between Liu-Martin-Syring and the Inverse Power sampler (the one currently at 0.3). That one might actually be the one that requires more benchmarking, since performance between the log-space and current sampler should be similar.

Issue #1898, opened 2024-09-11
[`InverseGaussian`?] `quantile()` hangs for larger number of `σ`'s (`quantile_newton` not converging)


LebedevRI:
I'm not sure if this is specific to this distribution, but this seems really problematic.

```julia
using Distributions
using SpecialFunctions

p(num_σ) = erf(num_σ/sqrt(2))

d = InverseGaussian{Float64}(1.187997687788096, 60.467382225458564)

@show d
@time @show quantile(d, p(4)) # near-instant
@time @show quantile(d, p(5)) # hangs, 10++ seconds
```
```
julia> versioninfo()
Julia Version 1.10.5
Commit 613e91e85c (2024-09-06 18:17 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: 32 × AMD Ryzen 9 5950X 16-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, znver3)
Threads: 32 default, 0 interactive, 16 GC (on 32 virtual cores)
Environment:
  JULIA_NUM_THREADS = 32

julia> using Pkg; pkg"status"
Status `~/.julia/environments/v1.10/Project.toml`
  [35d6a980] ColorSchemes v3.26.0
  [8f4d0f93] Conda v1.10.2
  [a93c6f00] DataFrames v1.6.1
  [31c24e10] Distributions v0.25.111
  [f6369f11] ForwardDiff v0.10.36
  [28b8d3ca] GR v0.73.7
  [7073ff75] IJulia v1.25.0
  [b6b21f68] Ipopt v1.6.6
  [4076af6c] JuMP v1.23.1
  [2fda8390] LsqFit v0.15.0
  [eff96d63] Measurements v2.11.0
  [961ee093] ModelingToolkit v9.39.1
  [7f7a1694] Optimization v3.28.0
  [3e6eede4] OptimizationBBO v0.3.0
  [36348300] OptimizationOptimJL v0.3.2
  [58dd65bb] Plotly v0.4.1
  [a03496cd] PlotlyBase v0.8.19
  [f0f68f2c] PlotlyJS v0.18.14
  [f2990250] PlotlyKaleido v2.2.5
  [91a5bcdd] Plots v1.40.8
  [92933f4c] ProgressMeter v1.10.2
  [274fc56d] PythonPlot v1.0.5
[...]

Issue #1833, opened 2024-01-30
Infinite loop for quantile of InverseGaussian for specific value


mentics:
I found this with Distributions v0.25.104, Julia 1.10.0, two machines: Ubuntu and Windows.

The following code will hang forever (ctrl-c will interrupt it):
```
using Distributions
dig = InverseGaussian{Float64}(μ=17.84806245738152, λ=163.707062977564)
quantile(dig, 0.9999981908772995)
```
Breaking it shows a stack trace of:
```
 [1] exp(x::Float64)
   @ Base.Math .\special\exp.jl:327
 [2] cdf(d::InverseGaussian{Float64}, x::Float64)
   @ Distributions C:\Users\joel\.julia\packages\Distributions\km0AH\src\univariate\continuous\inversegaussian.jl:102       
 [3] quantile_newton(d::InverseGaussian{Float64}, p::Float64, xs::Float64, tol::Float64)
   @ Distributions C:\Users\joel\.julia\packages\Distributions\km0AH\src\quantilealgs.jl:57
 [4] quantile_newton(d::InverseGaussian{Float64}, p::Float64)
   @ Distributions C:\Users\joel\.julia\packages\Distributions\km0AH\src\quantilealgs.jl:51
 [5] quantile(d::InverseGaussian{Float64}, p::Float64)
   @ Distributions C:\Users\joel\.julia\packages\Distributions\km0AH\src\quantilealgs.jl:146
 [6] top-level scope
   @ REPL[64]:1
```

It seems to me that the likely culprit is in the loop around Distributions:quantilealgs.jl:57.

I tested it against values just above and below it, only changing the final digit:
quantile(dig, 0.9999981908772994)
quantile(dig, 0.9999981908772996)
and they returned values normally, so it appears I managed to get surprisingly "lucky" in finding this particular value that has a problem (though perhaps there is a family of them that would?).

devmotion:
Possibly a duplicate of https://github.com/JuliaStats/Distributions.jl/issues/1571?

mentics:
Yes, it does appear to be a duplicate. The examples in that issue break in the same loop. I apologize for not searching sufficiently in advance.

devmotion:
No worries 🙂

Issue #1571, opened 2022-06-24
Failure to compute quantile of InverseGaussian


ekboehm:
The quantile function seems to fail for seemingly arbitrary numbers, e.g. 
```
μ   = 1.0
λ   = 0.25
quantile(InverseGaussian(μ,λ),0.99)
quantile(InverseGaussian(μ,λ),0.999)
quantile(InverseGaussian(μ,λ),0.9999)
quantile(InverseGaussian(μ,λ),0.99999)
quantile(InverseGaussian(μ,λ),0.999999)
```
All work successfully. But 
```
quantile(InverseGaussian(μ,λ),0.999996)
```
leads to no result (it just gets stuck trying to compute it). `0.999995` and `0.999997` work instantly, too.

simsurace:
Newton's method is entering an oscillation:
```julia
0.46351854580739754
1.1822190665191383
2.5171241967844167
4.663152461244044
7.681537906525332
11.504627262164904
16.00183230199545
21.035352263873737
26.481726330380177
32.2266130798376
38.135513473013404
43.98404551213666
49.3182977047776
53.31471534484445
55.194990126949854
55.52269267456251
55.53113502401951
55.53114044440781
55.53114044459374
55.53114044440781
55.53114044459374
55.53114044440781
...
```

devmotion:
Fixed by #2070

PR #1899, opened 2024-09-12
Quantile: Newton: introduce oscillation dampening

LebedevRI:
It is well known that Newton's method has an oscillation problem.
In fact, it can reasonably happen on real-world, non-synthetic,
distributions, when computing quantiles.

This situation appears easy to detect - just check that we've converged
with the root we had two steps ago - i may be mistaken, but i'm not sure
this can happen in any situation other than during oscillation.

Likewise, it appears to be easily circumventable,
by picking a new x0 in-between the oscillation bounds,
by using just half of the delta-x we'd apply otherwise.

NOTE: i do not have a proof that this solves all the issues,
nor do i know that the choice of, specifically, half of the delta-x is optimal.

Fixes https://github.com/JuliaStats/Distributions.jl/issues/1571
Fixes https://github.com/JuliaStats/Distributions.jl/issues/1833
Fixes https://github.com/JuliaStats/Distributions.jl/issues/1898


devmotion:
Thank you for the PR!

My general feeling though is that we should not re-invent and re-implement numerical algorithms that already exist and are better maintained in other packages such as Roots.jl. This also has the additional benefit that all other dependencies and users can benefit from any improvements or modifications possibly needed for Distributions.

LebedevRI:
Sure, let me see what i can do...

LebedevRI:
@devmotion https://github.com/JuliaStats/Distributions.jl/pull/1900

PR #1900, opened 2024-09-13
Quantile: use `Newton` method from `Roots.jl`

LebedevRI:
In https://github.com/JuliaStats/Distributions.jl/pull/1899,
i'm trying to address the oscillation problem of Newton method,
which does happen in real-world, non-synthetic distributions.

As per https://github.com/JuliaStats/Distributions.jl/pull/1899#issuecomment-2345526949,
let's first stop reinventing the wheel.

PR #1901, opened 2024-09-13
Bump minimal supported julia version to 1.6

LebedevRI:
In https://github.com/JuliaStats/Distributions.jl/pull/1900, dependency on Roots.jl is being added, which only supports julia 1.6+

This seems to result in new failures, so i'm doing this as a separate step.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -23,7 +23,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.3'
+          - '1.6'
           - '1'
           - pre
         os:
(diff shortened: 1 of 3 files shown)


devmotion:
I'm not a fan of updating Julia compat entries as long as there is no compelling reason to do so (e.g., if it involves unreasonable additional maintenance burden or limits development). That's not the case currently in Distributions in general, so IMO the change should be kept in #1900 as the only reason might be to change it if we decide to add a dependency on Roots.

LebedevRI:
I don't disagree. The point here was that adding a dependency on `Roots.jl` causes chain effects,
that are reproducible without adding such a dependency, and it was not at all clear to me how
the fallout is to be dealt. Just because a standalone PR is open, does not mean it must be merged
immediately, i was expecting this to merged immediately before that PR is to be merged.

IOW i don't know how to deal with remaining test failures there.

(Why are only two julia versions being tested in CI but many more versions
are claimed to be supported, yet they clearly aren't, as we are now discovering?)

Issue #894, opened 2019-05-13
eltype and rand with Univariate/Continuous distributions


halleysfifthinc:
I'm not sure how much more endemic this issue is (I've only tested with a Normal distribution), but here are a couple examples of counter-intuitive (and presumably wrong) behavior.

If I have a distribution (eg. normal), I would expect the output from sampling to match the types parameterizing the distribution.
```julia
julia> n = Normal(1.0f0, 0.1f0)
Normal{Float32}(μ=1.0f0, σ=0.1f0)

julia> typeof(rand(n)) # I would expect this to be Float32
Float64
```

In the case of a Normal dist., that could be fixed with this patch:
```diff
diff --git a/src/univariate/continuous/normal.jl b/src/univariate/continuous/normal.jl
index 8d6b3d3c..563696e2 100644
--- a/src/univariate/continuous/normal.jl
+++ b/src/univariate/continuous/normal.jl
@@ -87,7 +87,7 @@ cf(d::Normal, t::Real) = exp(im * t * d.μ - d.σ^2/2 * t^2)

 #### Sampling

-rand(rng::AbstractRNG, d::Normal) = d.μ + d.σ * randn(rng)
+rand(rng::AbstractRNG, d::Normal{T}) where T = d.μ + d.σ * randn(rng, T)


 #### Fitting
```

But the following example would still produce `Float64`:

```julia
julia> rand(n, 4) # I would expect an Array{Float32,1}
4-element Array{Float64,1}:
 0.9081488251686096
 1.1430985927581787
 1.0427565574645996
 0.9090480804443359
```

because

```
julia> eltype(n) # this is hard coded in common.jl#51
Float64
```
and

```julia
rand(rng::AbstractRNG, s::Sampleable{Univariate}, dims::Dims) =
    rand!(rng, sampler(s), Array{eltype(s)}(undef, dims))
# at src/univariates.jl#160
```

Presumably the second issue would affect most/all Univariate distributions, however, I am not sure how typical of an example the code for the `Normal` dist. is as far as the type signature and returned type. (Unfortunately I have neither the familiarity with the Distributions.jl source nor the time to do a more thorough investigation and/or make a PR.)

Unrelated to the main issue, I also wonder (again, not being familiar with the code) whether in the second example the `sampler(s)` is necessary (and/or a no-op) given `s::Sampleable`.

matbesancon:
[...]

PR #1905, opened 2024-09-25
Decouple `rand` and `eltype`

### LINKED ISSUE #1041: Uniform is not type stable

### LINKED ISSUE #1071: `rand(dist)` does not consistently honor the element type of `dist`
One has
```julia
julia> typeof(rand(Uniform{Float32}(0f0,1f0)))
Float64
```
Similar inconsistencies occurs with `Exponential` and `Gamma`, although not with `Normal`. Relatedly:
```julia
julia> eltype(Uniform{Float32}(0f0,1f0))
Float64
```
This is rather easy to fix. I've already taken care of the few cases that are currently producing errors in my code---take a look [here](https://github.com/vargonis/Distributions.jl). However, this is not complete and does not merit a PR.

### LINKED ISSUE #1082: Logistic does not respect parameter types

### LINKED ISSUE #1252: Inconsistent sampling type for Uniform{BigFloat}

### LINKED ISSUE #1783: Inconsistent Type Behavior for Uniform Distribution

### LINKED ISSUE #1884: Is there a reason `rand(TDist(nu))` does not obey the type of `nu`?

### LINKED ISSUE #1902: Type of sample of `Exponential{T}` is not `T`
MWE:
```julia
θ = 1.0f0
d = Exponential(θ)
typeof(rand(Exponential(θ))) == typeof(θ)
```
returns `false`

Because Exponential sampling is also used when sampling `Gamma` with shape 1, the problem also occurs there.

### LINKED ISSUE #1907: `rand(Beta(1.0f0, 1.0f0))` returns `Float64`
Hi.

Samples from `Beta(1.0f0, 1.0f0)` are `Float64` rather than the expected `Float32`. This behavior seems to happen only for `alpha=beta=1` and no other case. For example

```
Beta(1.0f0, 1.0f0)::Float64
Beta(1.0f0, 2.0f0)::Float32
Beta(2.0f0, 1.0f0)::Float32
```

devmotion:
One of my take-aways from issues such as #1252, #1902, #894, #1783, #1041, and #1071 is that `eltype` is not only quite inconsistently implemented and unclearly documented currently but more generally it might be a bad design decision to use it for pre-allocating containers of samples: Setting it to a fixed type (historically `Float64` for continuous and `Int` for discrete distributions) is too limiting, but trying to infer it from parameters is challenging and doomed to fail for more challenging scenarios that are not as clear as e.g. `Normal`. 

This PR tries to decouple `rand` and `eltype`, to make it easier to possibly eventually deprecate and remove `eltype`.

Fixes #1041.
Fixes #1071.
Fixes #1082.
Fixes #1252.
Fixes #1783.
Fixes #1884.
Fixes #1902.
Fixes #1907.


andreasnoack:
Given that, according to the docstring
https://github.com/JuliaStats/Distributions.jl/blob/a1010e408dbbdd4bcfd9c11eef189df64f7bb05a/src/common.jl#L96-L102
the sole purpose of `eltype` is to return the type of `rand`, which I agree isn't really feasible, should we also deprecate the methods here as part of this PR?

quildtide:
#1907 is related.

devmotion:
> https://github.com/JuliaStats/Distributions.jl/issues/1907 is related.

https://github.com/JuliaStats/Distributions.jl/pull/1905/commits/17154a2c1e38050f3c864b3e93f1b8cd496f45a8 fixes it.

quildtide:
My personal opinion:
`rand(d::Distribution)` should remain a valid method (I don't think anyone is trying to change this), and its return type should be predictable by knowing what `d` is. It would also be useful to have a method that tells you what type `rand(d::Distribution)` returns without running `rand(d::Distribution)`. `eltype(d::Distribution)` is logical for this.

This is orthogonal to the fact that Distributions should probably support `rand(d::Distribution, T::Type)`.

In an ideal situation, I think we would define `rand(d::Distribution, T::Type)` for distributions (to allow for potential type-specific optimizations and evade unnecessary typecasts) and then dispatch `rand(d::Distribution) = rand(d, eltype(d))`.

devmotion:
> its return type should be predictable by knowing what d is

I became convinced that in general this is not possible. It basically means trying to manually figure out type inference, a task that even the Julia compiler can fail at. Of course, for some number types and distributions it can be done (as also the compiler will succeed in many instances). But in general it's far from trivial. As in the initial stages of Distributions, one could be restrictive and limit samples to `Int` and `Float` but that's IMO not desirable either. Therefore I think it's best to remove `eltype` (which IMO has also been a weird name since I don't view distributions as containers or collections) from the API. IMO it's just too brittle and currently to inconsistent. Of course, that doesn't prevent users from trying to figure out the (element) type of samples with eg the help of the compiler, it's just not an official feature anymore.

singularitti:
A few distributions still give me `Float64` after this PR:

```julia
julia> dist = Semicircle(50.0f0)
Semicircle{Float32}(r=50.0f0)

julia> rand(dist, 5)
5-element Array{Float64, 1}:
  36.80671953487414
 -18.355635129900335
 -11.701855436648922
 -21.444118928985656
  -5.80120463505302

julia> dist = JohnsonSU(0.0f0, 1.0f0, 0.0f0, 1.0f0)
JohnsonSU{Float32}(ξ=0.0f0, λ=1.0f0, γ=0.0f0, δ=1.0f0)

julia> rand(dist, 5)
5-element Array{Float64, 1}:
  0.5311696707298299
  1.632313034117999
  0.04951771555318912
  0.4721610259428258
 -3.052321854866766
```

Can this be fixed?

devmotion:
The reason is that for quite a few of the older, probably less used and definitely less frequently updated distributions the `rand` implementation contains hardcoded `Float64`, either as literals or as calls of `rand(rng) or `randn(rng)`. I fixed the `Semicircle` and `JohnsonSU`, but I'm not sure if it's possible (or even desirable) to address all of these in a single PR (it's already a quite massive undertaking). Maybe it would be best to open issues for others and address them in follow-up PRs (most (all?) other such changes in this PR are included to fix open issues).

quildtide:
This is not the first time I have went on a rant about how Distributions in this package are effectively unordered containers with infinite size, but I have stumbled upon new ammunition:
https://docs.julialang.org/en/v1/stdlib/Random/#A-simple-sampler-without-pre-computed-data

This section of the Julia stdlib documentation claims:
> Given a collection type `S`, it's currently assumed that if `rand(::S)` is defined, an object of type `eltype(S)` will be produced. In the last example, a `Vector{Any}` is produced; the reason is that `eltype(Die) == Any`. The remedy is to define `Base.eltype(::Type{Die}) = Int`.

This led me to https://github.com/JuliaLang/julia/pull/31787 and https://github.com/JuliaLang/julia/pull/27756

**I believe that it is actually deeper Random.jl precedent that `typeof(rand(s)) == eltype(s)`**. Do we really want to break this precedent in Distributions.jl?


# Additional rambling

On a similar note, I think it's reasonable that `eltype(d)` should also be the type returned by `minimum(d)`, `maximum(d)`, `mean(d)`, `mode(d)`, etc., following base precedent. Similarly, `quantile(d, x)` should return `eltype(d)`; it is analogous to indexing to some extent. In other words, the following behavior is not ideal:

```julia
julia> d = Normal(0f0, 1f0)
Normal{Float32}(μ=0.0f0, σ=1.0f0)

julia> typeof(minimum(d)) == typeof(rand(d))
false
```

For many distribution types, perhaps the real conversation we should be having is if we should force `partype(D == eltype(D)`. For distributions parameterized by their extrema, e.g. uniform, I think this is very reasonable. `Uniform(0f0, 1f0)` is basically saying "construct an infinite unordered container with a uniform distribution with `minimum == 0f0`, `maximum == 1f0`."

For distributions parameterized by mean/std, similar logic definitely holds. It is not a stretch to extend this for other location/spread parameters. Shape parameters are ambiguous, but if someone types `Beta(1f0, 1f0)`, I would suspect they have a good reason for it.

For distributions like Binomial, I would argue that the type of `n` is more useful than the type of `p`. Unfortunately, the type of `n` is currently hardcoded as `Int` while the type of `p` is user-defined.

## Further rambling

`pdf` is a strange cousin of `indexin` and `findfirst`, which I believe are guaranteed to return `keytype`. Actually implementing an explicit `keytype(d) == Float64)` may be a reach though.

src/cholesky/lkjcholesky.jl
```diff
@@ -159,95 +157,114 @@ end
...
+                end
+            end
+        else
+            for j in 1:d
+                for i in 1:(j - 1)
+                    A[i, j] = 0
+                end
+                A[j, j] = 1
+            end
+        end
```
sethaxen, line 244: Maybe the loop is a little more efficient, but since this is a rare case, perhaps it's better for decreased code complexity to just copy the identity matrix?
```suggestion
        copyto!(A, I)
```

src/matrix/lkj.jl
```diff
@@ -120,14 +120,39 @@ logkernel(d::LKJ, R::AbstractMatrix) = (d.η - 1) * logdet(R)
...
+    if η == Inf
+        for j in 1:d
+            for i in 1:(j - 1)
+                R[i, j] = 0
+            end
+            R[j, j] = 1
+            for i in (j + 1):d
+                R[i, j] = 0
+            end
+        end
```
sethaxen: Same as above:
```suggestion
        copyto!(R, I)
```

src/matrix/lkj.jl
```diff
@@ -120,14 +120,39 @@ logkernel(d::LKJ, R::AbstractMatrix) = (d.η - 1) * logdet(R)
...
+        return R
+    end
+
     #  Section 3.2 in LKJ (2009 JMA)
     #  1. Initialization
-    R = ones(typeof(η), d, d)
+    fill!(R, oneunit(float(η)))
```
sethaxen: Is the custom type handling necessary? `fill!` in Base already converts the fill argument to the eltype of the array before filling.
```suggestion
    fill!(R, true)
```

sethaxen (commented):
Here's a review of `censored`, Cholesky-, and matrix-variate distributions.

src/cholesky/lkjcholesky.jl
devmotion: I assume I wrote the code in that way to avoid updating the inactive triangle. But I didn't benchmark it.

sethaxen: I don't think LinearAlgebra makes any promises about the contents of the inactive triangle, and I think it would be crazy for a user calling `rand!(rng, ::LKJCholesky, ::Cholesky)` to have any expectations about the `factors` matrix other than that the active triangle follows the distribution. So I don't see a problem with overwriting the inactive triangle to simplify the code.

devmotion: I don't see any correctness problems with overwriting the entire matrix. I just assume that the motivation for the current version was to reduce the number of `setindex!` calls. I'll actually doubt that it has a significant performance impact but I'll benchmark it when I'm back at my computer, just to be sure.

sethaxen: It's not a huge deal; I'll leave it up to you.

src/mixtures/unigmm.jl
```diff
@@ -38,6 +40,22 @@ struct UnivariateGMMSampler{VT1<:AbstractVector{<:Real},VT2<:AbstractVector{<:Re
...
+    return muladd(randn(rng, float(Base.promote_typeof(μ, σ))), σ, μ)
+end
+function rand!(rng::AbstractRNG, s::UnivariateGMMSampler, x::AbstractArray{<:Real})
+    psampler = s.psampler
+    means = s.means
+    stds = s.stds
+    randn!(rng, x)
+    for i in eachindex(x)
+        k = rand(rng, psampler)
+        x[i] = muladd(x[i], stds[k], means[k])
```
sethaxen: This code seems to assume that `stds` and `means` have 1-based indexing. Is this enforced anywhere?

src/multivariate/dirichlet.jl
```diff
@@ -154,18 +152,25 @@ end
 
 # sampling
 
-function _rand!(rng::AbstractRNG,
-                d::Union{Dirichlet,DirichletCanon},
-                x::AbstractVector{<:Real})
+function rand(rng::AbstractRNG, d::Union{Dirichlet,DirichletCanon})
+    x = map(αi -> rand(rng, Gamma(αi)), d.alpha)
+    return lmul!(inv(sum(x)), x)
```
sethaxen: Maybe use `normalize!` so we also get its special-casing to avoid over/underflow?
```suggestion
    return normalize!(x, 1)
```
Also, if e.g. `d.alpha` is a `StaticArrays.StaticArray`, this would fail, right, since `x` would also be immutable? So maybe it make sense to allocate `x` first, `map!` into it, and then have this line?

src/multivariate/dirichlet.jl
```diff
@@ -154,18 +152,25 @@ end
...
+end
+function rand(rng::AbstractRNG, d::Dirichlet{<:Real,<:FillArrays.AbstractFill{<:Real}})
+    x = rand(rng, Gamma(FillArrays.getindex_value(d.alpha)), length(d))
+    return lmul!(inv(sum(x)), x)
```
sethaxen:
```suggestion
    return normalize!(x, 1)
```

src/multivariate/jointorderstatistics.jl
```diff
@@ -125,7 +123,29 @@ function _marginalize_range(dist, i, j, xᵢ, xⱼ, T)
...
+        # fast quantile function or that use inversion sampling
+        x = rand(rng, d.dist, n)
+        sort!(x)
+    else
+        # use exponential generation method with inversion, where for gaps in the ranks, we
+        # use the fact that the sum Y of k IID variables xₘ ~ Exp(1) is Y ~ Gamma(k, 1).
+        # Lurie, D., and H. O. Hartley. "Machine-generation of order statistics for Monte
+        # Carlo computations." The American Statistician 26.1 (1972): 26-27.
+        # this is slow if length(d.ranks) is close to n and quantile for d.dist is expensive,
+        # but this branch is probably taken when length(d.ranks) is small or much smaller than n.
```
sethaxen: Since we're just calling the mutating method, I don't think we need to copy such a large comment.
```suggestion
```

src/multivariate/mvlognormal.jl
```diff
@@ -232,11 +230,26 @@ var(d::MvLogNormal) = diag(cov(d))
...
-function _rand!(rng::AbstractRNG, d::MvLogNormal, x::AbstractVecOrMat{<:Real})
-    _rand!(rng, d.normal, x)
+function rand(rng::AbstractRNG, d::MvLogNormal)
+    x = rand(rng, d.normal)
     map!(exp, x, x)
     return x
 end
+function rand(rng::AbstractRNG, d::MvLogNormal, n::Int)
+    xs = rand(rng, d.normal, n)
+    map!(exp, xs, xs)
```
sethaxen, line 240: According to the docs for `map!`:
> Behavior can be unexpected when any mutated argument shares memory with any other argument.

Maybe we could broadcast here instead?
```suggestion
    xs .= exp.(xs)
```

src/multivariate/mvlognormal.jl
```diff
@@ -232,11 +230,26 @@ var(d::MvLogNormal) = diag(cov(d))
 entropy(d::MvLogNormal) = length(d)*(1+log2π)/2 + logdetcov(d.normal)/2 + sum(mean(d.normal))
 
 #See https://en.wikipedia.org/wiki/Log-normal_distribution
-function _rand!(rng::AbstractRNG, d::MvLogNormal, x::AbstractVecOrMat{<:Real})
-    _rand!(rng, d.normal, x)
+function rand(rng::AbstractRNG, d::MvLogNormal)
+    x = rand(rng, d.normal)
     map!(exp, x, x)
```
sethaxen, line 235:
```suggestion
    x .= exp.(x)
```

src/multivariate/mvlognormal.jl
```diff
@@ -232,11 +230,26 @@ var(d::MvLogNormal) = diag(cov(d))
...
+    xs = rand(rng, d.normal, n)
+    map!(exp, xs, xs)
+    return xs
+end
+
+for N in (1, 2)
+    @eval begin
+        Base.@propagate_inbounds function rand!(rng::AbstractRNG, d::MvLogNormal, x::AbstractArray{<:Real,$N})
+            rand!(rng, d.normal, x)
+            map!(exp, x, x)
```
sethaxen, line 248:
```suggestion
            x .= exp.(x)
```

src/multivariate/mvtdist.jl
```diff
@@ -155,18 +156,35 @@ function gradlogpdf(d::GenericMvTDist, x::AbstractVector{<:Real})
...
-function _rand!(rng::AbstractRNG, d::GenericMvTDist, x::AbstractMatrix{T}) where T<:Real
+@inline function rand!(rng::AbstractRNG, d::GenericMvTDist, x::AbstractMatrix{<:Real})
+    @boundscheck size(x, 1) == length(d)
     cols = size(x,2)
     chisqd = Chisq{partype(d)}(d.df)
-    y = Matrix{T}(undef, 1, cols)
+    y = similar(x, (1, cols))
     unwhiten!(d.Σ, randn!(rng, x))
     rand!(rng, chisqd, y)
     x .= x ./ sqrt.(y ./ d.df) .+ d.μ
```
sethaxen, line 190: Why not broadcast the `rand` call to avoid allocating an intermediate container?
```suggestion
    unwhiten!(d.Σ, randn!(rng, x))
    rand!(rng, chisqd, y)
    x .= x ./ sqrt.(rand.(rng, chisqd) ./ d.df) .+ d.μ
```

sethaxen (commented):
Here's a review of the mixture models and multivariate distributions.

src/multivariate/mvlognormal.jl
devmotion: I'm not sure. Generally IMO it's better to avoid broadcasting unless it's necessary (ie unless you're actually broadcasting objects with different dimensions). Broadcasting is slow, both at compile and runtime.

src/multivariate/dirichlet.jl
devmotion: One of the main issues the PR tries to avoid is incorrect pre-allocation. So preallocating `x` is not really an option. Maybe instead we could use a `_normalize!!` method that uses `normalize!` for `StridedArray`s but generally falls back to `normalize`.

src/multivariate/mvtdist.jl
devmotion: We could, but I chose this version on purpose: Generally, vectorized sampling tends to be much faster.

src/cholesky/lkjcholesky.jl
devmotion: I came up with an additional approach: `copyto!(LowerTriangular(A), I)`/`copyto!(UpperTriangular(A), I)` That seems in spirit of the original code but both easier to read and more generic. A quick benchmark indicates it's also quite fast, in particular if the number of dimensions increases:

```julia
julia> using Chairmarks, LinearAlgebra

julia> f!(A::AbstractMatrix{<:Real}, uplo::Char) = copyto!(A, I)
f! (generic function with 1 method)

julia> g!(A::AbstractMatrix{<:Real}, uplo::Char) = uplo === 'U' ? copyto!(UpperTriangular(A), I) : copyto!(LowerTriangular(A), I)
g! (generic function with 1 method)

julia> function h!(A::Matrix{<:Real}, uplo::Char)
           d = size(A, 1)
           if uplo === 'L'
                   for j in 1:d
                       A[j, j] = 1
                       for i in (j+1):d
                           A[i, j] = 0
                       end
                   end
               else
                   for j in 1:d
                       for i in 1:(j - 1)
                           A[i, j] = 0
                       end
                       A[j, j] = 1
                   end
               end
           return A
       end
h! (generic function with 1 method)

julia> f!(rand(3, 3), 'L')
3×3 Matrix{Float64}:
 1.0  0.0  0.0
 0.0  1.0  0.0
 0.0  0.0  1.0

julia> f!(rand(3, 3), 'U')
3×3 Matrix{Float64}:
 1.0  0.0  0.0
 0.0  1.0  0.0
 0.0  0.0  1.0

julia> g!(rand(3, 3), 'L')
3×3 LowerTriangular{Float64, Matrix{Float64}}:
 1.0   ⋅    ⋅
 0.0  1.0   ⋅
 0.0  0.0  1.0

julia> parent(g!(rand(3, 3), 'L'))
3×3 Matrix{Float64}:
 1.0  0.551274  0.844459
 0.0  1.0       0.408075
 0.0  0.0       1.0

julia> g!(rand(3, 3), 'U')
3×3 UpperTriangular{Float64, Matrix{Float64}}:
 1.0  0.0  0.0
  ⋅   1.0  0.0
  ⋅    ⋅   1.0

julia> parent(g!(rand(3, 3), 'U'))
3×3 Matrix{Float64}:
 1.0       0.0       0.0
 0.838605  1.0       0.0
 0.226996  0.975878  1.0

julia> h!(rand(3, 3), 'L')
3×3 Matrix{Float64}:
 1.0  0.267429  0.788652
 0.0  1.0       0.656903
 0.0  0.0       1.0

julia> h!(rand(3, 3), 'U')
3×3 Matrix{Float64}:
 1.0       0.0       0.0
 0.889722  1.0       0.0
 0.202842  0.791729  1.0

julia> @be rand(10, 10) f!(_, 'L')
Benchmark: 3460 samples with 1852 evaluations
 min    11.204 ns
 median 14.759 ns
 mean   14.552 ns
 max    48.551 ns

julia> @be rand(10, 10) f!(_, 'U')
Benchmark: 3306 samples with 2261 evaluations
 min    11.112 ns
 median 12.329 ns
 mean   12.566 ns
 max    36.931 ns

julia> @be rand(10, 10) g!(_, 'L')
Benchmark: 3150 samples with 737 evaluations
 min    36.748 ns
 median 39.688 ns
 mean   40.234 ns
 max    185.437 ns

julia> @be rand(10, 10) g!(_, 'U')
Benchmark: 3170 samples with 638 evaluations
 min    36.180 ns
 median 39.315 ns
 mean   41.438 ns
 max    165.034 ns

julia> @be rand(10, 10) h!(_, 'L')
Benchmark: 3506 samples with 727 evaluations
 min    30.835 ns
 median 37.197 ns
 mean   36.794 ns
 max    58.230 ns

julia> @be rand(10, 10) h!(_, 'U')
Benchmark: 3419 samples with 1010 evaluations
 min    24.793 ns
 median 26.773 ns
 mean   27.065 ns
 max    78.672 ns

julia> @be rand(30, 30) f!(_, 'L')
Benchmark: 1310 samples with 217 evaluations
 min    95.235 ns
 median 160.332 ns
 mean   160.546 ns
 max    302.032 ns

julia> @be rand(30, 30) f!(_, 'U')
Benchmark: 1692 samples with 335 evaluations
 min    96.642 ns
 median 169.281 ns
 mean   165.287 ns
 max    436.319 ns

julia> @be rand(30, 30) g!(_, 'L')
Benchmark: 3404 samples with 243 evaluations
 min    86.593 ns
 median 104.251 ns
 mean   102.891 ns
 max    186.214 ns

julia> @be rand(30, 30) g!(_, 'U')
Benchmark: 3327 samples with 266 evaluations
 min    82.395 ns
 median 97.274 ns
 mean   96.877 ns
 max    145.835 ns

julia> @be rand(30, 30) h!(_, 'L')
Benchmark: 3367 samples with 215 evaluations
 min    101.353 ns
 median 120.930 ns
 mean   119.978 ns
 max    191.665 ns

julia> @be rand(30, 30) h!(_, 'U')
Benchmark: 3415 samples with 238 evaluations
 min    90.513 ns
 median 107.492 ns
 mean   107.054 ns
 max    164.744 ns

julia> @be rand(100, 100) f!(_, 'L')
Benchmark: 544 samples with 29 evaluations
 min    793.103 ns
 median 974.138 ns
 mean   997.768 ns
 max    1.665 μs

julia> @be rand(100, 100) f!(_, 'U')
Benchmark: 2555 samples with 31 evaluations
 min    784.968 ns
 median 857.548 ns
 mean   882.615 ns
 max    1.460 μs

julia> @be rand(100, 100) g!(_, 'L')
Benchmark: 2332 samples with 55 evaluations
 min    489.382 ns
 median 566.673 ns
 mean   571.493 ns
 max    1.042 μs

julia> @be rand(100, 100) g!(_, 'U')
Benchmark: 2408 samples with 52 evaluations
 min    488.788 ns
 median 556.885 ns
 mean   567.539 ns
 max    916.673 ns

julia> @be rand(100, 100) h!(_, 'L')
Benchmark: 2343 samples with 45 evaluations
 min    562.022 ns
 median 670.356 ns
 mean   670.599 ns
 max    1.092 μs

julia> @be rand(100, 100) h!(_, 'U')
Benchmark: 3020 samples with 41 evaluations
 min    488.805 ns
 median 527.439 ns
 mean   546.422 ns
 max    1.140 μs
```

src/mixtures/unigmm.jl
devmotion: No, good point. Basically most of the mixture model code suffers from this issue (it's not as bad anymore now that we removed all `@inbounds`, but still bad). Let's leave other functions for a separate PR but I'll at least fix the `rand`/`rand!` calls.

src/mixtures/unigmm.jl
```diff
@@ -25,10 +25,12 @@ probs(d::UnivariateGMM) = probs(d.prior)
...
 mean(d::UnivariateGMM) = dot(d.means, probs(d))
 
-rand(d::UnivariateGMM) = (k = rand(d.prior); d.means[k] + randn() * d.stds[k])
-
-rand(rng::AbstractRNG, d::UnivariateGMM) =
-    (k = rand(rng, d.prior); d.means[k] + randn(rng) * d.stds[k])
+function rand(rng::AbstractRNG, d::UnivariateGMM)
+    k = rand(rng, d.prior)
+    μ = d.means[k]
+    σ = d.stds[k]
```
devmotion:
```suggestion
    koffset = rand(rng, d.prior) - 1
    μ = d.means[begin+koffset]
    σ = d.stds[begin+koffset]
```

src/mixtures/unigmm.jl
devmotion:
```suggestion
        koffset = rand(rng, psampler) - 1
        x[i] = muladd(x[i], stds[begin+koffset], means[begin+koffset])
```

src/mixtures/unigmm.jl
```diff
@@ -38,6 +40,22 @@ struct UnivariateGMMSampler{VT1<:AbstractVector{<:Real},VT2<:AbstractVector{<:Re
...
+function rand(rng::AbstractRNG, s::UnivariateGMMSampler)
+    k = rand(rng, s.psampler)
+    μ = d.means[k]
+    σ = d.stds[k]
+    return muladd(randn(rng, float(Base.promote_typeof(μ, σ))), σ, μ)
+end
+function rand!(rng::AbstractRNG, s::UnivariateGMMSampler, x::AbstractArray{<:Real})
+    psampler = s.psampler
+    means = s.means
+    stds = s.stds
```
devmotion:
```suggestion
    (; means, stds, psampler) = s
```

src/mixtures/unigmm.jl
```diff
@@ -38,6 +40,22 @@ struct UnivariateGMMSampler{VT1<:AbstractVector{<:Real},VT2<:AbstractVector{<:Re
     psampler::AliasTable
 end
 
-rand(rng::AbstractRNG, s::UnivariateGMMSampler) =
-    (k = rand(rng, s.psampler); s.means[k] + randn(rng) * s.stds[k])
+function rand(rng::AbstractRNG, s::UnivariateGMMSampler)
+    k = rand(rng, s.psampler)
+    μ = d.means[k]
+    σ = d.stds[k]
```
devmotion:
```suggestion
    koffset = rand(rng, s.psampler) - 1
    μ = d.means[begin+koffset]
    σ = d.stds[begin+koffset]
```

src/multivariate/dirichlet.jl
```diff
@@ -154,18 +152,25 @@ end
 
 # sampling
 
-function _rand!(rng::AbstractRNG,
-                d::Union{Dirichlet,DirichletCanon},
-                x::AbstractVector{<:Real})
+function rand(rng::AbstractRNG, d::Union{Dirichlet,DirichletCanon})
```
devmotion, line 158:
```suggestion
_normalize1!!(x::AbstractVector{<:Real}) = normalize(x, 1)
_normalize1!!(x::StridedVector{<:Real}) = normalize!(x, 1)

function rand(rng::AbstractRNG, d::Union{Dirichlet,DirichletCanon})
```

src/multivariate/dirichlet.jl
devmotion:
```suggestion
    return _normalize1!!(x)
```

src/multivariate/dirichlet.jl
devmotion:
```suggestion
    return _normalize1!!(x)
```

src/multivariate/dirichlet.jl
```diff
@@ -154,18 +152,25 @@ end
...
+    x = rand(rng, Gamma(FillArrays.getindex_value(d.alpha)), length(d))
+    return lmul!(inv(sum(x)), x)
+end
+
+@inline function rand!(rng::AbstractRNG, d::Union{Dirichlet,DirichletCanon}, x::AbstractVector{<:Real})
+    @boundscheck length(d) == length(x)
     for (i, αi) in zip(eachindex(x), d.alpha)
         x[i] = rand(rng, Gamma(αi))
     end
     lmul!(inv(sum(x)), x) # this returns x
```
devmotion:
```suggestion
    return normalize!(x, 1)
```

src/multivariate/dirichlet.jl
```diff
@@ -154,18 +152,25 @@ end
...
 end
 
-function _rand!(rng::AbstractRNG,
-                d::Dirichlet{T,<:FillArrays.AbstractFill{T}},
-                x::AbstractVector{<:Real}) where {T<:Real}
+@inline function rand!(rng::AbstractRNG, d::Dirichlet{T,<:FillArrays.AbstractFill{T}}, x::AbstractVector{<:Real}) where {T<:Real}
+    @boundscheck length(d) == length(x)
     rand!(rng, Gamma(FillArrays.getindex_value(d.alpha)), x)
     lmul!(inv(sum(x)), x) # this returns x
```
devmotion:
```suggestion
    return normalize!(x, 1)
```

src/cholesky/lkjcholesky.jl
devmotion:
```suggestion
        if uplo === 'L'
            copyto!(LowerTriangular(A), I)
        else
            copyto!(UpperTriangular(A), I)
        end
```

PR #1906, opened 2024-10-01
Geometric parameter should be on (0, 1]

palday:


andreasnoack:
Thanks. Should probably be tested, though.

devmotion (approved):

PR #1915, opened 2024-11-14
Update erlang.jl docstring

theplatters:
Fixed the mismatch in the parameter names and gave the concrete pdf, so the reference to wikipedia doesn't create confusion.

### DIFF
--- a/src/univariate/continuous/erlang.jl
+++ b/src/univariate/continuous/erlang.jl
@@ -1,7 +1,12 @@
 """
-    Erlang(α,θ)
+    Erlang(a=1,s=1)
 
 The *Erlang distribution* is a special case of a [`Gamma`](@ref) distribution with integer shape parameter.
+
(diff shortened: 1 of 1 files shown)


mschauer:
Thank you, it is desirable to use the same parameter names as wikipedia, so here k and beta?

src/univariate/continuous/erlang.jl
```diff
@@ -1,12 +1,17 @@
 """
-    Erlang(α,θ)
+    Erlang(α=1,θ=1)
 
 The *Erlang distribution* is a special case of a [`Gamma`](@ref) distribution with integer shape parameter.
```
mschauer, line 4:
```suggestion
The *Erlang distribution* is a special case of a [`Gamma`](@ref) distribution with integer shape parameter α and scale parameter  θ.
```

Issue #1423, opened 2021-11-15
Sampling from von Mises Fisher yields NaN


lanceXwq:
This happens when the mean direction unit vector `μ` is along the x-axis. For instance:
```
julia> vMF = VonMisesFisher([1, 0], 1.0)
VonMisesFisher{Float64}(μ=[1.0, 0.0], κ=1.0)
julia> rand(vMF)
2-element Vector{Float64}:
 NaN
 NaN
```
This problem seems to persist in higher dimensions.

devmotion:
I'm not familiar with the implementation of the `VonMisesFisher` sampler but I suspect that the `NaN`s arise in https://github.com/JuliaStats/Distributions.jl/blob/8887d7a16a2801dc869ab42ba03cad03efc5ed9d/src/samplers/vonmisesfisher.jl#L103: for `μ=[1.0, 0.0]` we have `v[1] = s = 0` in this line, and hence `v[1] /= s` should result in `v[1] = NaN`.

trahflow:
I just stumbled over this as well.

> I'm not familiar with the implementation of the `VonMisesFisher` sampler but I suspect that the `NaN`s arise in
> 
> https://github.com/JuliaStats/Distributions.jl/blob/8887d7a16a2801dc869ab42ba03cad03efc5ed9d/src/samplers/vonmisesfisher.jl#L103
> : for `μ=[1.0, 0.0]` we have `v[1] = s = 0` in this line, and hence `v[1] /= s` should result in `v[1] = NaN`.

Yes, this is what happens.
It seems this was introduced in #1162. At least a quick check shows the previous version which constructs an explicit rotation matrix does not suffer from this.

A very quick, albeit a bit dirty fix would be to return a zero vector in 

https://github.com/JuliaStats/Distributions.jl/blob/13029c03fa885a2b4512b954e61f9d5a7dfa0612/src/samplers/vonmisesfisher.jl#L87-L102
for the case that `s` is numerically close to zero. That should give the correct (identity) rotation in
https://github.com/JuliaStats/Distributions.jl/blob/13029c03fa885a2b4512b954e61f9d5a7dfa0612/src/samplers/vonmisesfisher.jl#L46
Maybe @emerali has a suggestion?

PR #1918, opened 2024-11-20
Fixing NaN sampling for von Mises Fisher distribution

williamjsdavis:
## Summary of contribution

I have made a change to the sampling algorithm of von Mises Fisher distributions to avoid `NaN` outputs.

## The problem

Sampling from certain von Mises Fisher distributions will result in `NaN`. For example:

```
using Distributions
using LinearAlgebra: norm

κ = 1.0
μ = [1.0,1e-8]
μ = μ ./ norm(μ)
d = VonMisesFisher(μ, κ)
rand(d)
```

...returns:
```
2-element Vector{Float64}:
 NaN
 NaN
```

The error happens because of operations performed in the function `_vmf_householder_vec`:

https://github.com/JuliaStats/Distributions.jl/blob/3de6038ef9b041ceaf649bc324d60ca65fcf09eb/src/samplers/vonmisesfisher.jl#L87-L102

The problem arises when `μ` is a vector exactly in the direction of the first dimension, or very close to the first dimension. (For example, `μ = [1.0, 0.0]`, `μ = [1.0, 0.0, 0.0]`, or `μ = [1.0, 0.0, 1e-8], etc...`) In these cases, variable `s` will become zero, leading vector `v` to become filled with NaNs.

This issue was previously raised in https://github.com/JuliaStats/Distributions.jl/issues/1423

## Proposed solution

I propose a solution, where I have added a small epsilon value to the denominators in `_vmf_householder_vec`. This approach is common in machine learning algorithms like ADAM, where dividing by zero is avoided during accumulation:

1. [Adam: A Method for Stochastic Optimization](https://arxiv.org/abs/1412.6980)
2. [What do I mean by ε?](http://zna.do/epsilon)

```
using Distributions
using LinearAlgebra: norm

κ = 1.0
μ = [1.0,1e-8]
μ = μ ./ norm(μ)
d = VonMisesFisher(μ, κ)
rand(d)
```

...returns:
```
2-element Vector{Float64}:
 0.8976253988301388
 0.4407591670913201
```

... which are not `NaN`s!

## Discussion

I have the following discussion topics:

1. I'm open to feedback regarding whether this PR is going in the right direction. Is a change here desired? Is my proposed solution the right way forward? 
2. If so, what additional contributions would be needed? For example, I'd assume that we'd also like to add some tests to verify my change is not causing unintended changes, and to check for future `NaN` outputs.

As always, feedback is appreciated

### DIFF
--- a/src/samplers/vonmisesfisher.jl
+++ b/src/samplers/vonmisesfisher.jl
@@ -84,18 +84,20 @@ end
 _vmf_genw(rng::AbstractRNG, s::VonMisesFisherSampler) =
     _vmf_genw(rng, s.p, s.b, s.x0, s.c, s.κ)
 
-function _vmf_householder_vec(μ::Vector{Float64})
+function _vmf_householder_vec(μ::Vector{Float64}, ε::Float64=1e-8)
     # assuming μ is a unit-vector (which it should be)
     #  can compute v in a single pass over μ
+    # Add small value, ε, to denominator to avoid dividing by zero
 
     p = length(μ)
     v = similar(μ)
     v[1] = μ[1] - 1.0
     s = sqrt(-2*v[1])
-    v[1] /= s
+    v[1] /= (s + ε)
 
     @inbounds for i in 2:p
-        v[i] = μ[i] / s
+        
+        v[i] = μ[i] / (s + ε)
     end
 
     return v


williamjsdavis:
As discussed in https://github.com/JuliaStats/Distributions.jl/pull/1930, the numerical approach workaround presented in this PR is not ideal. 

For posterity sake, the reason why this PR does not pass the tests introduced by @devmotion is because the epsilon value added to the denominator affects the vectors quite far from the singularity point. This causes a lot of cases where the norm of the vector is no longer close to 1.

```
  Expression: norm(X[:, i]) ≈ 1.0
   Evaluated: 0.9999998814346462 ≈ 1.0
```

The same result can be illustrated as follows:
output_n

Here I am plotting the 2 components of von-Mieses Householder vector as a function of `t` for for `μ=[x(t),y(t)]`. The original implementation is shown in solid colors; my attempted implementation is in dashed. The `NaN` values of the original implementation are shown as transparent vertical lines. As one can see, my version of `v2` is quite less than ±1 away from t=0.

The same plot showing @devmotion's implementation looks like this:
output-d

devmotion:
With the updated version in #1930, I get for `μ = LinearAlgebra.normalize!([1.0, t])` with `t` around zero

```julia
julia> using Distributions, LinearAlgebra, AlgebraOfGraphics, CairoMakie

julia> ts = range(-1e-7, 1e-7; length=1001);

julia> get_v(x, idx) = sampler(VonMisesFisher(normalize!([1.0, x]), 1.0)).v[idx];

julia> data((; t = repeat(ts; outer=2), v = vcat(get_v.(ts, 1), get_v.(ts, 2)), index = repeat(1:2, inner = length(ts)))) * mapping(:t, :v) * mapping(color = :index => nonnumeric, group = :index => nonnumeric) * visual(Lines) |> draw
```
v1

and for `μ = LinearAlgebra.normalize!([t, 1.0])`

```julia
julia> get_v(x, idx) = sampler(VonMisesFisher(normalize!([x, 1.0]), 1.0)).v[idx];

julia> data((; t = repeat(ts; outer=2), v = vcat(get_v.(ts, 1), get_v.(ts, 2)), index = repeat(1:2, inner = length(ts)))) * mapping(:t, :v) * mapping(color = :index => nonnumeric, group = :index => nonnumeric) * visual(Lines) |> draw
```
v2

williamjsdavis:
Nice approach using `LinearAlgebra.reflector!`!  

I'll close this PR once https://github.com/JuliaStats/Distributions.jl/pull/1930 is merged

devmotion:
Fixed by #1930.

PR #1928, opened 2024-12-15
Re-write median(Binomial) to improve speed

### LINKED ISSUE #1926: Speed up median(Binomial)
While it's good that the result is now correct, it is at a significant price. The evaluation of the cdf is quite costly and I think the `quantile` approach might end up calling it multiple times. Of course `q=0.5` might be special cased in RMath but compare
```julia
# Old version
julia> @btime median(X)
  15.030 ns (0 allocations: 0 bytes)
8

julia> @btime quantile(X, 0.5)
  242.180 ns (0 allocations: 0 bytes)
7
```

It might be possible, at least in most cases, see https://onlinelibrary.wiley.com/doi/10.1111/j.1467-9574.1980.tb00681.x, to compute the median without evaluating the cdf or even the pdf.

The old version just used `round` but, as pointed out in #1923, that wasn't always correct. The paper helps us

Skærmbillede 2024-12-14 kl  16 09 24

In #1923, the example had `X = Binomial(25, 0.3)` so the condition is clearly not satisfied since
```julia
julia> mean(X)
7.5
```

so we could just do a single (c)cdf evaluation and adjust the value if needed in the cases where the condition of Theorem 1 isn't satisfied. It is likely to be much faster than `quantile(..., 0.5)`

marcusps:
Fix #1926 by following suggestions therein to reduce avg number of `cdf(Binomial)` evaluations.

### DIFF
--- a/src/univariate/discrete/binomial.jl
+++ b/src/univariate/discrete/binomial.jl
@@ -80,7 +80,7 @@ function median(dist::Binomial)
     ceil_mean = ceil(Int, rv_mean)
     if rv_mean - floor_mean <= bound
         floor_mean
-    elseif abs(ceil_mean - rv_mean) <= bound
+    elseif ceil_mean - rv_mean <= bound
         ceil_mean
     elseif cdf(rv, floor_mean) > 0.5
         floor_mean
--- a/src/univariate/discrete/binomial.jl
+++ b/src/univariate/discrete/binomial.jl
@@ -73,6 +73,22 @@ function mode(d::Binomial{T}) where T<:Real
 end
 modes(d::Binomial) = Int[mode(d)]
 
+function median(rv::Binomial)
+    bound = min(rv.p, 1-rv.p)
+    rv_mean = mean(rv)
+    floor_mean = floor(Int, rv_mean)
+    ceil_mean = ceil(Int, rv_mean)
+    if abs(floor_mean - rv_mean) <= bound
+        floor_mean
+    elseif abs(ceil_mean - rv_mean) <= bound
+        ceil_mean
+    elseif cdf(rv, floor_mean) > 0.5
+        floor_mean
+    else
+        ceil_mean
+    end
+end
+
 function skewness(d::Binomial)
     n, p1 = params(d)
     p0 = 1 - p1
--- a/src/univariate/discrete/binomial.jl
+++ b/src/univariate/discrete/binomial.jl
@@ -74,7 +74,14 @@ end
 modes(d::Binomial) = Int[mode(d)]
 
 function median(dist::Binomial)
-    bound = min(dist.p, 1-dist.p) # from http://dx.doi.org/10.1111/j.1467-9574.1980.tb00681.x
+    # The median is floor(Int, mean) or ceil(Int, mean)
+    # As shown in https://doi.org/10.1016/0167-7152(94)00090-U,
+    # |median - mean| <= 1 - bound
+    # where the equality is strict except for the case p = 1/2 and n odd.
+    # Thus if |k - mean| < bound for one of the two candidates if p = 1/2 and n odd
+    # or |k - mean| <= bound for one of the two candidates otherwise,
+    # the other candidate can't satisfy the condition and hence k must be the median
+    bound = max(min(dist.p, 1-dist.p), loghalf)
     dist_mean = mean(dist)
     
     floor_mean = floor(Int, dist_mean)
--- a/src/univariate/discrete/binomial.jl
+++ b/src/univariate/discrete/binomial.jl
@@ -78,10 +78,11 @@ function median(dist::Binomial)
     dist_mean = mean(dist)
     
     floor_mean = floor(Int, dist_mean)
-
-    if dist_mean - floor_mean < bound
+    difference = dist_mean - floor_mean
+    
+    if difference < bound
         floor_mean
-    elseif floor_mean + 1 - dist_mean < bound
+    elseif  difference > 1 - bound
         floor_mean + 1
     elseif cdf(dist, floor_mean) >= 0.5
         floor_mean
--- a/src/univariate/discrete/binomial.jl
+++ b/src/univariate/discrete/binomial.jl
@@ -80,9 +80,9 @@ function median(dist::Binomial)
     floor_mean = floor(Int, dist_mean)
     difference = dist_mean - floor_mean
     
-    if difference < bound
+    if difference <= bound
         floor_mean
-    elseif  difference > 1 - bound
+    elseif  difference >= 1 - bound
         floor_mean + 1
     elseif cdf(dist, floor_mean) >= 0.5
         floor_mean
--- a/src/univariate/discrete/binomial.jl
+++ b/src/univariate/discrete/binomial.jl
@@ -73,7 +73,7 @@ function mode(d::Binomial{T}) where T<:Real
 end
 modes(d::Binomial) = Int[mode(d)]
 
-function median(rv::Binomial)
+function median(dist::Binomial)
     bound = min(rv.p, 1-rv.p)
     rv_mean = mean(rv)
     floor_mean = floor(Int, rv_mean)
--- a/src/univariate/discrete/binomial.jl
+++ b/src/univariate/discrete/binomial.jl
@@ -82,7 +82,7 @@ function median(dist::Binomial)
         floor_mean
     elseif ceil_mean - rv_mean <= bound
         ceil_mean
-    elseif cdf(rv, floor_mean) > 0.5
+    elseif cdf(rv, floor_mean) >= 0.5
         floor_mean
     else
         ceil_mean
--- a/src/univariate/discrete/binomial.jl
+++ b/src/univariate/discrete/binomial.jl
@@ -81,6 +81,9 @@ function median(dist::Binomial)
     difference = dist_mean - floor_mean
     
     if difference <= bound
+        # The only case where the median satisfies |median - mean| <= 1 - bound with equality
+        # is p = 1/2 and n odd
+        # However, in that case we also want to return floor(mean)
         floor_mean
     elseif difference >= 1 - bound
         # The case p = 1/2 and n odd was already covered above,
--- a/src/univariate/discrete/binomial.jl
+++ b/src/univariate/discrete/binomial.jl
@@ -78,7 +78,7 @@ function median(rv::Binomial)
     rv_mean = mean(rv)
     floor_mean = floor(Int, rv_mean)
     ceil_mean = ceil(Int, rv_mean)
-    if abs(floor_mean - rv_mean) <= bound
+    if rv_mean - floor_mean <= bound
         floor_mean
     elseif abs(ceil_mean - rv_mean) <= bound
         ceil_mean
--- a/src/univariate/discrete/binomial.jl
+++ b/src/univariate/discrete/binomial.jl
@@ -82,7 +82,10 @@ function median(dist::Binomial)
     
     if difference <= bound
         floor_mean
-    elseif  difference >= 1 - bound
+    elseif difference >= 1 - bound
+        # The case p = 1/2 and n odd was already covered above,
+        # thus only cases with |median - mean| < 1 - bound are left here
+        # Therefore difference >= 1 - bound implies that floor(mean) cannot be the median
         floor_mean + 1
     elseif cdf(dist, floor_mean) >= 0.5
         floor_mean
--- a/src/univariate/discrete/binomial.jl
+++ b/src/univariate/discrete/binomial.jl
@@ -74,18 +74,19 @@ end
 modes(d::Binomial) = Int[mode(d)]
 
 function median(dist::Binomial)
-    bound = min(rv.p, 1-rv.p)
-    rv_mean = mean(rv)
-    floor_mean = floor(Int, rv_mean)
-    ceil_mean = ceil(Int, rv_mean)
-    if rv_mean - floor_mean <= bound
+    bound = min(dist.p, 1-dist.p) # from http://dx.doi.org/10.1111/j.1467-9574.1980.tb00681.x
+    dist_mean = mean(dist)
+    
+    floor_mean = floor(Int, dist_mean)
+
+    if dist_mean - floor_mean < bound
         floor_mean
-    elseif ceil_mean - rv_mean <= bound
-        ceil_mean
-    elseif cdf(rv, floor_mean) >= 0.5
+    elseif floor_mean + 1 - dist_mean < bound
+        floor_mean + 1
+    elseif cdf(dist, floor_mean) >= 0.5
         floor_mean
     else
-        ceil_mean
+        floor_mean + 1
     end
 end
 
--- a/test/univariate/discrete/binomial.jl
+++ b/test/univariate/discrete/binomial.jl
@@ -27,16 +27,13 @@ end
 @test Distributions.expectation(x -> -x, Binomial(10, 0.2)) ≈ -2.0
 
 # Test median
-@test median(Binomial(5,3//10)) == 1
 @test median(Binomial(25,3//10)) == 7
 @test median(Binomial(45,3//10)) == 13
 @test median(Binomial(65,3//10)) == 19
 @test median(Binomial(85,3//10)) == 25
 
-@test median(Binomial(25,0.756)) == 19
-@test median(Binomial(25,1//2)) == 12
-@test median(Binomial(25,3//5)) == 15
-
+@test all([median(Binomial(7,p))==quantile(Binomial(7,p),1//2) for p in range(0,1,length=11)])
+    
 # Test mode
 @test Distributions.mode(Binomial(100, 0.4)) == 40
 @test Distributions.mode(Binomial(1, 0.51)) == 1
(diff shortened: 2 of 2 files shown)


src/univariate/discrete/binomial.jl
```diff
@@ -73,6 +73,22 @@ function mode(d::Binomial{T}) where T<:Real
 end
 modes(d::Binomial) = Int[mode(d)]
 
+function median(rv::Binomial)
+    bound = min(rv.p, 1-rv.p)
+    rv_mean = mean(rv)
+    floor_mean = floor(Int, rv_mean)
+    ceil_mean = ceil(Int, rv_mean)
+    if abs(floor_mean - rv_mean) <= bound
```
andreasnoack, line 81:
```suggestion
    if rv_mean - floor_mean <= bound
```
Since `rv_mean >= floor_mean`

src/univariate/discrete/binomial.jl
```diff
@@ -73,6 +73,22 @@ function mode(d::Binomial{T}) where T<:Real
...
+        floor_mean
+    elseif abs(ceil_mean - rv_mean) <= bound
```
andreasnoack, line 83:
```suggestion
    elseif ceil_mean - rv_mean <= bound
```
since `ceil_mean >= rv_mean`

andreasnoack (commented):
I think it might be useful to link to the paper in a comment. It would make it easier to reason about the implementation for a future reader of the source.

src/univariate/discrete/binomial.jl
```diff
@@ -73,6 +73,22 @@ function mode(d::Binomial{T}) where T<:Real
 end
 modes(d::Binomial) = Int[mode(d)]
 
+function median(rv::Binomial)
```
devmotion, line 76: Maybe use
```suggestion
function median(d::Binomial)
```
or

```suggestion
function median(dist::Binomial)
```

to make it more consistent with the other functions?

src/univariate/discrete/binomial.jl
```diff
@@ -73,6 +73,22 @@ function mode(d::Binomial{T}) where T<:Real
...
+    bound = min(rv.p, 1-rv.p)
+    rv_mean = mean(rv)
+    floor_mean = floor(Int, rv_mean)
+    ceil_mean = ceil(Int, rv_mean)
```
devmotion, line 80: I think it should be sufficient to compute `floor`:
 ```suggestion
```

File: src/univariate/discrete/binomial.jl (hunk 2 above)
Comment (devmotion) [inline src/univariate/discrete/binomial.jl:83]: This seems to be equivalent to
```suggestion
    elseif rv_mean - floor_mean >= 1 - bound
```
which could be written simpler and more efficient by computing `rv_mean - floor_mean` only once and calculating `minmax(d.p, 1 - d.p)` instead of only `min`.

src/univariate/discrete/binomial.jl
```diff
@@ -73,6 +73,22 @@ function mode(d::Binomial{T}) where T<:Real
...
+    if abs(floor_mean - rv_mean) <= bound
+        floor_mean
+    elseif abs(ceil_mean - rv_mean) <= bound
+        ceil_mean
+    elseif cdf(rv, floor_mean) > 0.5
```
devmotion, line 85: AFAICT this should be
```suggestion
    elseif cdf(rv, floor_mean) >= 0.5
```

src/univariate/discrete/binomial.jl
```diff
@@ -73,6 +73,22 @@ function mode(d::Binomial{T}) where T<:Real
 end
 modes(d::Binomial) = Int[mode(d)]
 
+function median(rv::Binomial)
+    bound = min(rv.p, 1-rv.p)
+    rv_mean = mean(rv)
+    floor_mean = floor(Int, rv_mean)
```
devmotion, line 79: In principle one could also use
```suggestion
    fractional_mean, integer_mean = modf(rv_mean)
```
to get the fractional and the integer part of the mean.
But I am unsure whether there would be any benefit (performance-wise or implementation-wise).

marcusps:
Following the suggestions, given
```julia
rv = Binomial(25,0.3)
rv2 = Binomial(25,0.756)
```
the previous version yielded
```julia
julia> @benchmark median(rv)
BenchmarkTools.Trial: 10000 samples with 33 evaluations.
 Range (min … max):  924.576 ns …   9.896 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     929.455 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   980.001 ns ± 151.139 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▅               ▁     ▁   ▄   ▃                              ▁
  ██▅▇▇▇▇▅▇▅▃▇▄▅▇▁▃█▃██▁▁█▇▅███▃▆█▇▅▃█▆▃▄▇▄▁▁▄█▅▁▃▃█▅▄▁▅▇▆▄▁▁▃▇ █
  925 ns        Histogram: log(frequency) by time       1.49 μs <

julia> @benchmark median(rv2)
BenchmarkTools.Trial: 10000 samples with 26 evaluations.
 Range (min … max):  941.346 ns …  13.187 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     945.231 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):     1.007 μs ± 187.974 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █               ▁   ▂  ▂   ▄   ▃                              ▁
  █▇█▅▅█▅▆▇▄█▅▁▇▅▃██▁▃█▃▁█▅▁▃█▃▁▄█▅▆▇█▄▅▅█▇▄▃▁█▄▁▃▃█▃▁▁▁▆█▁▄▄▁█ █
  941 ns        Histogram: log(frequency) by time       1.51 μs <

```
and the version version yields
```julia
julia> @benchmark median(rv)
BenchmarkTools.Trial: 10000 samples with 469 evaluations.
 Range (min … max):  225.928 ns … 557.132 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     228.358 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   235.036 ns ±  21.306 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ██▂▆▇▁                          ▁▁   ▁                        ▂
  ████████▅▅▅▇▇▆▇▅█▇▄▄█▆▄▅▆▅▄▄▅▅▄▄██▆▅▄█▇▅▅▄▇▅▃▄▂▄▄▃▄▄▄▆▄▃▄▄▅▅▄ █
  226 ns        Histogram: log(frequency) by time        340 ns <

julia> @benchmark median(rv2)
BenchmarkTools.Trial: 10000 samples with 996 evaluations.
 Range (min … max):  17.037 ns … 52.826 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     18.434 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   19.582 ns ±  2.704 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

    ▅██▇▇▅▆▆▅▃     ▁  ▁▂  ▁▁▃▂▄▄▄▃▃▂▃▂ ▁▁                     ▂
  ▄▇██████████▇█▇▄███▇██▇▇███████████████▇▆▇▇▇▆▇▇▇▄▆▇▃▇▆▅█▂▇▆ █
  17 ns        Histogram: log(frequency) by time        29 ns <

```
I will check how much the  `minmax` suggestion helps.

marcusps:
Here is a more thorough comparison.  To get full coverage across the branches, I extended the benchmarks to
```julia
rv = Binomial(25,0.3)
rv2 = Binomial(25,0.756)
rv3 = Binomial(25,1//2)
rv4 = Binomial(25, 3//5)
```
and then ran `bitj = @benchmark median(rvj)` where `i`=1 represents the current (buggy) implementation of the median, `i`=2 is the current (merged) implementation that fixes the bug, and `i=3` incorporates something akin to `minmax` (but still using only `min`), which looks like this:
```julia
function median(dist::Binomial)
    bound = min(dist.p, 1-dist.p) # from http://dx.doi.org/10.1111/j.1467-9574.1980.tb00681.x
    dist_mean = mean(dist)
    
    floor_mean = floor(Int, dist_mean)
    difference = dist_mean - floor_mean
    
    if difference < bound
        floor_mean
    elseif  difference > 1 - bound
        floor_mean + 1
    elseif cdf(dist, floor_mean) >= 0.5
        floor_mean
    else
        floor_mean + 1
    end
end
```
The results are
```julia
julia> for (a,b) in zip(b2,b3)
         println(BenchmarkTools.judge(median(a),median(b)))
       end
TrialJudgement(+0.40% => invariant)
TrialJudgement(-9.69% => improvement)
TrialJudgement(-0.04% => invariant)
TrialJudgement(-3.12% => invariant)

julia> for (a,b) in zip(b2,b3)
         println(BenchmarkTools.judge(var(a),var(b)))
       end
TrialJudgement(+12.14% => regression)
TrialJudgement(-49.32% => improvement)
TrialJudgement(-22.95% => improvement)
TrialJudgement(-12.07% => improvement)
```
which means the simplification is a definite improvement, both for the median and variance.

The new bug free implementations are still much slower than the buggy implementation, but that feels like an unfair comparison.
```julia
julia> for (a,b) in zip(b1,b2)
         println(BenchmarkTools.judge(median(a),median(b)))
       end
TrialJudgement(+308.74% => regression)
TrialJudgement(+5373.54% => regression)
TrialJudgement(+343.01% => regression)
TrialJudgement(+3152.94% => regression)

julia> for (a,b) in zip(b1,b3)
         println(BenchmarkTools.judge(median(a),median(b)))
       end
TrialJudgement(+310.37% => regression)
TrialJudgement(+4843.23% => regression)
TrialJudgement(+342.83% => regression)
TrialJudgement(+3051.34% => regression)
```

src/univariate/discrete/binomial.jl
marcusps: I made a change that seems to work with `min` only and without needing to recompute `rv_mean - floor_mean`

src/univariate/discrete/binomial.jl
marcusps: I made some changes that are equivalent in spirit, I think.

src/univariate/discrete/binomial.jl
```diff
@@ -73,6 +73,24 @@ function mode(d::Binomial{T}) where T<:Real
 end
 modes(d::Binomial) = Int[mode(d)]
 
+function median(dist::Binomial)
+    bound = min(dist.p, 1-dist.p) # from http://dx.doi.org/10.1111/j.1467-9574.1980.tb00681.x
```
devmotion, line 77: We can improve this further with the result in https://www.sciencedirect.com/science/article/abs/pii/016771529400090U:
```suggestion
    # The median is floor(Int, mean) or ceil(Int, mean)
    # As shown in https://doi.org/10.1016/0167-7152(94)00090-U,
    # |median - mean| <= 1 - bound
    # where the equality is strict except for the case p = 1/2 and n odd.
    # Thus if |k - mean| < bound for one of the two candidates if p = 1/2 and n odd
    # or |k - mean| <= bound for one of the two candidates otherwise,
    # the other candidate can't satisfy the condition and hence k must be the median
    bound = max(min(dist.p, 1-dist.p), loghalf)
```

src/univariate/discrete/binomial.jl
```diff
@@ -73,6 +73,24 @@ function mode(d::Binomial{T}) where T<:Real
...
+    dist_mean = mean(dist)
+    
+    floor_mean = floor(Int, dist_mean)
+    difference = dist_mean - floor_mean
+    
+    if difference < bound
```
devmotion: AFAICT we can use equality here:
```suggestion
    if difference <= bound
        # The only case where the median satisfies |median - mean| <= 1 - bound with equality
        # is p = 1/2 and n odd
        # However, in that case we also want to return floor(mean)
```

src/univariate/discrete/binomial.jl
```diff
@@ -73,6 +73,24 @@ function mode(d::Binomial{T}) where T<:Real
...
+        floor_mean
+    elseif  difference > 1 - bound
```
devmotion:
```suggestion
    elseif difference >= 1 - bound
        # The case p = 1/2 and n odd was already covered above,
        # thus only cases with |median - mean| < 1 - bound are left here
        # Therefore difference >= 1 - bound implies that floor(mean) cannot be the median
```

src/univariate/discrete/binomial.jl
```diff
@@ -73,6 +73,24 @@ function mode(d::Binomial{T}) where T<:Real
...
 modes(d::Binomial) = Int[mode(d)]
 
+function median(dist::Binomial)
+    bound = min(dist.p, 1-dist.p) # from http://dx.doi.org/10.1111/j.1467-9574.1980.tb00681.x
+    dist_mean = mean(dist)
+    
+    floor_mean = floor(Int, dist_mean)
+    difference = dist_mean - floor_mean
+    
+    if difference <= bound
```
devmotion, line 83:
```suggestion
    if difference <= bound
        # The only case where the median satisfies |median - mean| <= 1 - bound with equality
        # is p = 1/2 and n odd
        # However, in that case we also want to return floor(mean)
```

src/univariate/discrete/binomial.jl
```diff
@@ -73,6 +73,24 @@ function mode(d::Binomial{T}) where T<:Real
...
+        floor_mean
+    elseif  difference >= 1 - bound
```
devmotion, line 85:
```suggestion
    elseif difference >= 1 - bound
        # The case p = 1/2 and n odd was already covered above,
        # thus only cases with |median - mean| < 1 - bound are left here
        # Therefore difference >= 1 - bound implies that floor(mean) cannot be the median
```

test/univariate/discrete/binomial.jl
```diff
@@ -27,11 +27,12 @@ end
 @test Distributions.expectation(x -> -x, Binomial(10, 0.2)) ≈ -2.0
 
 # Test median
-@test median(Binomial(5,3//10)) == 1
 @test median(Binomial(25,3//10)) == 7
 @test median(Binomial(45,3//10)) == 13
 @test median(Binomial(65,3//10)) == 19
 @test median(Binomial(85,3//10)) == 25
+
+@test all([median(Binomial(7,p))==quantile(Binomial(7,p),1//2) for p in range(0,1,length=11)])
```
devmotion, line 35: One can avoid allocating the array of booleans here:
```suggestion
@test all(median(Binomial(7, p)) == quantile(Binomial(7, p), 1//2) for p in 0:0.1:1)
```

devmotion:
I see the following difference to the master branch:

### This PR

```julia

julia> @be median($(Binomial(44, 0.1)))
Benchmark: 3317 samples with 186 evaluations
 min    135.301 ns
 median 152.554 ns
 mean   153.703 ns
 max    250.892 ns

julia> @be median($(Binomial(40, 0.1)))
Benchmark: 4932 samples with 4265 evaluations
 min    3.976 ns
 median 4.298 ns
 mean   4.556 ns
 max    9.066 ns

julia> @be median($(Binomial(25, 0.3)))
Benchmark: 3891 samples with 147 evaluations
 min    137.755 ns
 median 157.592 ns
 mean   163.907 ns
 max    284.014 ns

julia> @be median($(Binomial(25, 0.756)))
Benchmark: 4933 samples with 3917 evaluations
 min    4.266 ns
 median 4.606 ns
 mean   4.900 ns
 max    21.966 ns

julia> @be median($(Binomial(25, 0.5)))
Benchmark: 4809 samples with 4545 evaluations
 min    3.988 ns
 median 4.171 ns
 mean   4.351 ns
 max    9.332 ns

julia> @be median($(Binomial(25, 0.6)))
Benchmark: 4921 samples with 4227 evaluations
 min    3.982 ns
 median 4.298 ns
 mean   4.538 ns
 max    9.148 ns
```

### Master branch

```julia

julia> @be median($(Binomial(44, 0.1)))
Benchmark: 3387 samples with 125 evaluations
 min    214.664 ns
 median 215.672 ns
 mean   222.091 ns
 max    366.672 ns

julia> @be median($(Binomial(40, 0.1)))
Benchmark: 4111 samples with 86 evaluations
 min    239.826 ns
 median 253.395 ns
 mean   265.475 ns
 max    431.686 ns

julia> @be median($(Binomial(25, 0.3)))
Benchmark: 3938 samples with 101 evaluations
 min    215.347 ns
 median 225.663 ns
 mean   234.931 ns
 max    417.901 ns

julia> @be median($(Binomial(25, 0.756)))
Benchmark: 3951 samples with 94 evaluations
 min    225.617 ns
 median 242.904 ns
 mean   250.102 ns
 max    419.777 ns

julia> @be median($(Binomial(25, 0.5)))
Benchmark: 3920 samples with 100 evaluations
 min    216.250 ns
 median 227.080 ns
 mean   237.701 ns
 max    419.170 ns

julia> @be median($(Binomial(25, 0.6)))
Benchmark: 3643 samples with 105 evaluations
 min    217.057 ns
 median 233.733 ns
 mean   245.306 ns
 max    732.143 ns
```

PR #1930, opened 2024-12-16
Fix von Mises-Fisher sampler

devmotion:
Fixes https://github.com/JuliaStats/Distributions.jl/issues/1423.

As already discussed in #1423, in the special case `mu = [1, 0, ..., 0]` no rotation of the "initial" sample is needed. I think skipping the rotation completely is preferable over a numerical workaround such as #1918.

I used the opportunity to also add comments and references to the existing code of the von Mises-Fisher sampler.

### DIFF
--- a/test/multivariate/vonmisesfisher.jl
+++ b/test/multivariate/vonmisesfisher.jl
@@ -23,6 +23,7 @@ function gen_vmf_tdata(n::Int, p::Int,
 end
 
 function test_vmf_rot(p::Int, rng::Union{AbstractRNG, Missing} = missing)
+    # Random μ
     if ismissing(rng)
         μ = randn(p)
         x = randn(p)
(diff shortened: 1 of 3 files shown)


williamjsdavis:
Nice work! This is an elegant solution, and I agree that it's preferable to the fix I proposed in https://github.com/JuliaStats/Distributions.jl/pull/1918. I'm happy to delete that PR if/when this PR is accepted.

test/multivariate/vonmisesfisher.jl
```diff
@@ -173,12 +151,21 @@ ns = 10^6
...
     end
 end
+
+# issue #1423
+@testset "Special case: No rotation" begin
+    for n in 2:10
+        d = VonMisesFisher(vcat(1, zeros(n - 1)), 1.0)
+        @test sum(abs2, rand(d)) ≈ 1
+        d_est = fit_mle(VonMisesFisher, rand(d, 100_000))
+        @test d_est.μ ≈ meandir(d) rtol=5e-2
```
devmotion, line 169:
```suggestion
        @test meandir(d_est) ≈ meandir(d) rtol=5e-2
```

PR #1931, opened 2024-12-17
Fix mode(Binomial) and modes(Binomial)

### LINKED ISSUE #1927: mode(Binomial) (docs?) wrong when (n+1)*p is integer
When `(n+1)*p` is integer, then there are two modes, see https://onlinelibrary.wiley.com/doi/10.1111/j.1467-9574.1980.tb00681.x. The documentation states
```julia
  mode(d::UnivariateDistribution)

  Returns the first mode.
```

but the second is returned, e.g.

```julia
julia> [(i, (pdf(Binomial(5, 4/6), i))) for i in 0:5]
6-element Vector{Tuple{Int64, Float64}}:
 (0, 0.004115226337448564)
 (1, 0.04115226337448562)
 (2, 0.16460905349794255)
 (3, 0.329218106995885)
 (4, 0.3292181069958847)
 (5, 0.13168724279835395)

julia> mode(Binomial(5, 4/6))
4
```

marcusps:
Fix #1927

### DIFF
--- a/src/univariate/discrete/binomial.jl
+++ b/src/univariate/discrete/binomial.jl
@@ -67,11 +67,36 @@ params(d::Binomial) = (d.n, d.p)
 
 mean(d::Binomial) = ntrials(d) * succprob(d)
 var(d::Binomial) = ntrials(d) * succprob(d) * failprob(d)
-function mode(d::Binomial{T}) where T<:Real
+
+function mode(d::Binomial)
     (n, p) = params(d)
-    n > 0 ? floor(Int, (n + 1) * d.p) : zero(T)
+    v = (n + 1) * p
+    quasi_mode = floor(Int, v)
+    if quasi_mode == v
+        if p == 1
+            n
+        else
+            quasi_mode-1
+        end
+    else
+        quasi_mode
+    end
 end
-modes(d::Binomial) = Int[mode(d)]
+
+function modes(d::Binomial)
+    (n, p) = params(d)
+    v = (n + 1) * p
+    quasi_mode = floor(Int, v)
+    if quasi_mode == v
+        if p == 1
+            Int[n]
+        else
+            Int[quasi_mode-1, quasi_mode]
+        end
+    else
+        Int[quasi_mode]
+    end
+end
 
 function skewness(d::Binomial)
     n, p1 = params(d)
--- a/test/univariate/discrete/binomial.jl
+++ b/test/univariate/discrete/binomial.jl
@@ -32,11 +32,18 @@ end
 @test median(Binomial(45,3//10)) == 13
 @test median(Binomial(65,3//10)) == 19
 @test median(Binomial(85,3//10)) == 25
-    
+
 # Test mode
-@test Distributions.mode(Binomial(100, 0.4)) == 40
-@test Distributions.mode(Binomial(1, 0.51)) == 1
-@test Distributions.mode(Binomial(1, 0.49)) == 0
+@test mode(Binomial(100, 0.4)) == 40
+@test mode(Binomial(1, 0.51)) == 1
+@test mode(Binomial(1, 0.49)) == 0
+@test mode(Binomial(4, 2//3)) == 3
+@test mode(Binomial(6, 2//7)) == 1
+@test mode(Binomial(7, 1//8)) == 0
+
+@test modes(Binomial(4, 2//3)) == [3, 4]
+@test modes(Binomial(6, 2//7)) == [1, 2]
+@test modes(Binomial(7, 1//8)) == [0, 1]
 
 @test isplatykurtic(Bernoulli(0.5))
 @test ismesokurtic(Normal(0.0, 1.0))


marcusps:
The impact on runtime is minimal
```julia
julia> d = Binomial(5, 2//3)
Binomial{Rational{Int64}}(n=5, p=2//3)

julia> @benchmark mode(d) # current master
BenchmarkTools.Trial: 10000 samples with 994 evaluations.
 Range (min … max):  23.342 ns … 76.490 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     23.791 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   25.491 ns ±  3.586 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▇▅▅▅▆▃▂▁              ▄▃ ▃▂▁▂▂  ▁                          ▂
  █████████▇▇▇█▄██▇▇█▆█▆▆██▆█████▇██▆▆▆▆▆▆▄▅▇█▄▆▆▅▆▄▅▆▅▆▄▄▄▆▇ █
  23.3 ns      Histogram: log(frequency) by time      39.2 ns <

julia> @benchmark mode(d) # this PR
BenchmarkTools.Trial: 10000 samples with 995 evaluations.
 Range (min … max):  22.868 ns … 90.999 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     23.201 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   25.739 ns ±  4.551 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▄▃   ▂▂    ▁▂▂▂▂▅▄▃▃                                       ▁
  ████▇███▇▇▇▇█████████▇█▇▇█▅▇▇▄██▅▄▇▇▄▆█▆▄▇▇▅▅▄▅▅▅▄▅▅▄▁▅▅▅▄▅ █
  22.9 ns      Histogram: log(frequency) by time      44.2 ns <

```
similarly for `modes`
```julia
julia> @benchmark modes(d) # current master
BenchmarkTools.Trial: 10000 samples with 984 evaluations.
 Range (min … max):  43.023 ns …   6.794 μs  ┊ GC (min … max): 0.00% … 98.72%
 Time  (median):     47.259 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   54.389 ns ± 143.785 ns  ┊ GC (mean ± σ):  8.71% ±  3.40%

     ▅▂▂▁▆█▅▃                                                   
  ▁▂▆█████████▅▄▂▂▂▂▂▂▂▂▃▃▃▄▅▄▄▄▃▃▂▂▃▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  43 ns           Histogram: frequency by time         69.7 ns <

 Memory estimate: 64 bytes, allocs estimate: 1.

julia> @benchmark modes(d) # this PR
BenchmarkTools.Trial: 10000 samples with 980 evaluations.
 Range (min … max):  47.515 ns …   3.625 μs  ┊ GC (min … max): 0.00% … 97.66%
 Time  (median):     56.248 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   63.108 ns ± 127.480 ns  ┊ GC (mean ± σ):  7.78% ±  3.79%

     ▁▄█▆▆▄        ▁ ▁▁                                         
  ▁▂▅███████▆▆▆▆▇▇█████▇▅▄▃▂▂▂▂▂▂▂▁▂▂▁▁▁▁▁▁▁▁▁▁▁▁▂▂▂▂▂▂▂▂▂▁▁▁▁ ▃
  47.5 ns         Histogram: frequency by time         89.3 ns <

 Memory estimate: 80 bytes, allocs estimate: 1.
```

src/univariate/discrete/binomial.jl
```diff
@@ -67,11 +67,36 @@ params(d::Binomial) = (d.n, d.p)
...
+    quasi_mode = floor(Int, v)
+    if quasi_mode == v
+        if p == 1
+            n
+        else
+            quasi_mode-1
+        end
+    else
+        quasi_mode
+    end
```
devmotion, line 83: AFAICT alternatives with fewer branches would be
```suggestion
    iszero(p) ? 0 : ceil(Int, (n + 1) * p) - 1
```
or
```suggestion
    max(0, ceil(Int, (n + 1) * p) - 1)
```
or
```suggestion
    clamp(ceil(Int, (n + 1) * p) - 1, 0, n)
```

test/univariate/discrete/binomial.jl
```diff
@@ -32,11 +32,18 @@ end
...
-@test Distributions.mode(Binomial(1, 0.51)) == 1
-@test Distributions.mode(Binomial(1, 0.49)) == 0
+@test mode(Binomial(100, 0.4)) == 40
+@test mode(Binomial(1, 0.51)) == 1
+@test mode(Binomial(1, 0.49)) == 0
+@test mode(Binomial(4, 2//3)) == 3
+@test mode(Binomial(6, 2//7)) == 1
+@test mode(Binomial(7, 1//8)) == 0
+
+@test modes(Binomial(4, 2//3)) == [3, 4]
```
devmotion, line 44:
```suggestion
@test modes(Binomial(4, 2//3)) == [3]
```

src/univariate/discrete/binomial.jl
marcusps: There is little to no difference in runtime between these options (or the option in the PR).  The main difference is in the max runtime, and the PR options and the `max(0, ceil(...))` have the shorter max runtime (the two are nearly identical with a max of about 10x the average runtime), while the max runtime for the other two options is about twice as long (about 20x the average runtime).  The standard deviations show a similar pattern (but they are smaller than the average runtime).

Given there isn't much of a different in run times, I propose the PR version is preferable because it is easiest to read and reason about.

marcusps: @devmotion Is there another reason to consider fewer branches other than runtime?

nalimilan: @devmotion Looks like this PR only needs a decision about the best approach before merging (and it fixed a bug).

Issue #224, opened 2014-04-30
Type hierarchy for Angular, Spherical and Manifold distributions.


simonbyrne:
This has come up in a couple of issues (#192, #223), so I thought I would propose creating some new abstract types:
- `AngularDistribution <: Distribution{Univariate,Continuous}`: e.g. von Mises distribution, wrapped normal. The problem is that they can be defined on any interval of length `2pi`, so what I was thinking is that we define `pdf` over the whole real line, such that the integral is 1 over any such interval. We could define `cdf` and `rand` on `[0,2pi]`.
- `SphericalDistribution <: Distribution{Multivariate,Continuous}`: e.g. Fisher-von Mises, Bingham, etc. We need to be a bit careful in how we deal with support (i.e. allow for some error in the condition `norm(x) == 1`).
- `StiefelDistribution <: Distribution{Matrixvariate,Continuous}`: `MatrixBingham` etc. for distributions over orthogonal matrices.
- perhaps also `PDMatDistribution <: Distribution{Matrixvariate,Continuous}` for `Wishart` and `InverseWishart`?

PR #1938, opened 2025-01-14
Fix `mode` of `LKJCholesky`

devmotion:
For LKJ the mode is only defined (and then identical to the identity matrix) if η > 1: If 0 < η < 1, the identity matrix is a trough of the density, and if η = 1 then it's a uniform distribution of correlation matrices (see e.g. https://mc-stan.org/docs/functions-reference/correlation_matrix_distributions.html#probability-density-function). Therefore for `LKJ` `mode` is defined only if `η > 1`:

```julia
julia> mode(LKJ(1, 0.5))
ERROR: DomainError with 0.5:
LKJ: mode is defined only when η > 1.
Stacktrace:
 [1] #545
   @ ~/.julia/packages/Distributions/cWeit/src/matrix/lkj.jl:79 [inlined]
 [2] check_args
   @ ~/.julia/packages/Distributions/cWeit/src/utils.jl:89 [inlined]
 [3] #mode#544
   @ ~/.julia/packages/Distributions/cWeit/src/matrix/lkj.jl:79 [inlined]
 [4] mode(d::LKJ{Float64, Int64})
   @ Distributions ~/.julia/packages/Distributions/cWeit/src/matrix/lkj.jl:78
 [5] top-level scope
   @ REPL[48]:1

julia> mode(LKJ(1, 1.5))
1×1 Matrix{Float64}:
 1.0
```

For `LKJCholesky`, however, on the master branch `mode` always returns the identity matrix, irrespective of the parameters:

```julia
julia> mode(LKJCholesky(1, 0.5))
LinearAlgebra.Cholesky{Float64, Matrix{Float64}}
L factor:
1×1 LinearAlgebra.LowerTriangular{Float64, Matrix{Float64}}:
 1.0

julia> mode(LKJCholesky(1, 1.5))
LinearAlgebra.Cholesky{Float64, Matrix{Float64}}
L factor:
1×1 LinearAlgebra.LowerTriangular{Float64, Matrix{Float64}}:
 1.0
```

This PR fixes the problem, ~~and defines `mean` for `LKJCholesky` which is currently missing and should always return the identity matrix:~~

```julia
julia> mean(LKJ(1, 0.5))
1×1 Matrix{Float64}:
 1.0

julia> mean(LKJ(1, 1.5))
1×1 Matrix{Float64}:
 1.0

julia> mean(LKJCholesky(1, 0.5))
ERROR: MethodError: no method matching iterate(::LKJCholesky{Float64})
[...]

julia> mean(LKJCholesky(1, 1.5))
ERROR: MethodError: no method matching iterate(::LKJCholesky{Float64})
[...]
```

### DIFF
--- a/test/cholesky/lkjcholesky.jl
+++ b/test/cholesky/lkjcholesky.jl
@@ -124,7 +124,7 @@ using FiniteDifferences
     end
 
     @testset "properties" begin
-        @testset for p in (4, 5), η in (0.5, 2, 3.5), uplo in ('L', 'U')
+        @testset for p in (4, 5), η in (0.5, 1, 2, 3.5), uplo in ('L', 'U')
             d = LKJCholesky(p, η, uplo)
             @test d.d == p
             @test size(d) == (p, p)
--- a/test/cholesky/lkjcholesky.jl
+++ b/test/cholesky/lkjcholesky.jl
@@ -141,10 +141,6 @@ using FiniteDifferences
             m = mode(d; check_args = false)
             @test m isa Cholesky{eltype(d)}
             @test Matrix(m) ≈ I
-
-            m = mean(d)
-            @test m isa Cholesky{eltype(d)}
-            @test Matrix(m) ≈ I
         end
         for (d, η) in ((2, 4), (2, 1), (3, 1)), T in (Float32, Float64)
             @test @inferred(partype(LKJCholesky(d, T(η)))) === T
--- a/test/cholesky/lkjcholesky.jl
+++ b/test/cholesky/lkjcholesky.jl
@@ -124,14 +124,25 @@ using FiniteDifferences
     end
 
     @testset "properties" begin
-        @testset for p in (4, 5), η in (2, 3.5), uplo in ('L', 'U')
+        @testset for p in (4, 5), η in (0.5, 2, 3.5), uplo in ('L', 'U')
             d = LKJCholesky(p, η, uplo)
             @test d.d == p
             @test size(d) == (p, p)
             @test Distributions.params(d) == (d.d, d.η, d.uplo)
             @test partype(d) <: Float64
 
-            m = mode(d)
+            if η > 1
+                m = mode(d)
+                @test m isa Cholesky{eltype(d)}
+                @test Matrix(m) ≈ I
+            else
+                @test_throws DomainError(η, "LKJCholesky: mode is defined only when η > 1.") mode(d)
+            end
+            m = mode(d; check_args = false)
+            @test m isa Cholesky{eltype(d)}
+            @test Matrix(m) ≈ I
+
+            m = mean(d)
             @test m isa Cholesky{eltype(d)}
             @test Matrix(m) ≈ I
         end
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.25.116"
+version = "0.25.117"
 
 [deps]
 AliasTables = "66dad0bd-aa9a-41b7-9441-69ab47430ed8"
--- a/src/cholesky/lkjcholesky.jl
+++ b/src/cholesky/lkjcholesky.jl
@@ -109,9 +109,18 @@ function insupport(d::LKJCholesky, R::LinearAlgebra.Cholesky)
     return true
 end
 
-function StatsBase.mode(d::LKJCholesky)
+function StatsBase.mean(d::LKJCholesky)
     factors = Matrix{eltype(d)}(LinearAlgebra.I, size(d))
     return LinearAlgebra.Cholesky(factors, d.uplo, 0)
+end
+
+function mode(d::LKJCholesky; check_args::Bool=true)
+    @check_args(
+        LKJCholesky,
+        @setup(η = d.η),
+        (η, η > 1, "mode is defined only when η > 1."),
+    )
+    return mean(d)
 end
 
 StatsBase.params(d::LKJCholesky) = (d.d, d.η, d.uplo)
--- a/src/cholesky/lkjcholesky.jl
+++ b/src/cholesky/lkjcholesky.jl
@@ -109,18 +109,14 @@ function insupport(d::LKJCholesky, R::LinearAlgebra.Cholesky)
     return true
 end
 
-function StatsBase.mean(d::LKJCholesky)
-    factors = Matrix{eltype(d)}(LinearAlgebra.I, size(d))
-    return LinearAlgebra.Cholesky(factors, d.uplo, 0)
-end
-
 function mode(d::LKJCholesky; check_args::Bool=true)
     @check_args(
         LKJCholesky,
         @setup(η = d.η),
         (η, η > 1, "mode is defined only when η > 1."),
     )
-    return mean(d)
+    factors = Matrix{eltype(d)}(LinearAlgebra.I, size(d))
+    return LinearAlgebra.Cholesky(factors, d.uplo, 0)
 end
 
 StatsBase.params(d::LKJCholesky) = (d.d, d.η, d.uplo)


sethaxen:
I don't think this is a bug or should be fixed. The `logpdf` of `LKJ` is defined wrt the Lebesgue measure on the strict upper/lower triangle of the matrix, while the `logpdf` of `LKJCholesky` is defined wrt the Lebesgue measure on the strict upper/lower triangle of the triangular factor:

```julia
julia> X = rand(LKJCholesky(10, 2.0));

julia> logpdf(LKJCholesky(10, 2.0), X)
-5.741870506362201

julia> logpdf(LKJ(10, 2.0), Matrix(X))
1.3678551789839233
```
This is important for e.g. Bijectors' logdetjac to work correctly.

As a result, they have different modes. Also, while `LKJ` has a reasonable notion of mean (sample infinite matrices, average them), `LKJCholesky` has no such reasonable notion (averaging the triangular factor produces an invalid and useless `Cholesky` object). One could interpret the mean as what one gets from calling `Matrix` on the `Cholesky` objects and then averaging, but just like `Factorization` is not a subtype of `AbstractMatrix` so doesn't support e.g. addition, I think it makes more sense to leave it unimplemented. If we think one will really want the `mean` of `LKJ` from `LKJCholesky`, maybe it makes more sense to implement `convert` methods for switching between these distributions.

sethaxen (commented):
I don't think this is a bug or should be fixed. The `logpdf` of `LKJ` is defined wrt the Lebesgue measure on the strict upper/lower triangle of the matrix, while the `logpdf` of `LKJCholesky` is defined wrt the Lebesgue measure on the strict upper/lower triangle of the triangular factor:

```julia
julia> X = rand(LKJCholesky(10, 2.0));

julia> logpdf(LKJCholesky(10, 2.0), X)
-5.741870506362201

julia> logpdf(LKJ(10, 2.0), Matrix(X))
1.3678551789839233
```
This is important for e.g. Bijectors' logdetjac to work correctly, and this is why they have different modes.

As a result, they have different modes. Also, while `LKJ` has a reasonable notion of mean (sample infinite matrices, average them), `LKJCholesky` has no such reasonable notion (averaging the triangular factor produces an invalid and useless `Cholesky` object). One could interpret the mean as what one gets from calling `Matrix` on the `Cholesky` objects and then averaging, but just like `Factorization` is not a subtype of `AbstractMatrix` so doesn't support e.g. addition, I think it makes more sense to leave it unimplemented. If we think one will really want the `mean` of `LKJ` from `LKJCholesky`, maybe it makes more sense to implement `convert` methods for switching between these distributions.

sethaxen:
Hm, but actually, I think the `logpdf` implementation might have an off-by-one error. I'll double-check the math later.

sethaxen:
Yeah, math is fine. But it's true that `LKJCholesky` doesn't have a defined mode for `η < 1`, since the mode will only be unique and in the support when `p + 2(η - 1) - i > 0` for all `2 <= i <= p`, which implies `η > 1`. When `η = 1`, the marginal density of the `p`th row/column of the factor is uniform on the hemisphere. I'm not certain what Distribution's policy is about non-unique modes; if non-unique modes are allowed, then identity matrix could still be returned for `η = 1`.

devmotion:
Hmm... My interpretation and use of `LKJCholesky` so far has been that it is `LKJ` with a more convenient and numerically more efficient (already factorized) variate. This seems to be in line with the docstring that states
```txt
Variates or samples of the distribution are `LinearAlgebra.Cholesky` objects, as might
be returned by `F = LinearAlgebra.cholesky(R)`, so that `Matrix(F) ≈ R` is a variate or
sample of [`LKJ`](@ref). 

Sampling `LKJCholesky` is faster than sampling `LKJ`, and often having the correlation
matrix in factorized form makes subsequent computations cheaper as well.
```
Based on this interpretation (variates of type `Cholesky` as a "better" non-`AbstractMatrix` matrix type), I think it is natural to expect that `mean`, `var`, etc. of a `LKJCholesky` are identical to the corresponding `LKJ` distribution (since in that interpretation they are the "same" thing with only an implementation difference). Of course, that understanding breaks down if

> the logpdf of LKJCholesky is defined wrt the Lebesgue measure on the strict upper/lower triangle of the triangular factor

I haven't been aware of this (or at least I don't remember it), and it still seems unclear (surprising?) given the docstring of `LKJCholesky`. [I have been wondering for a while even whether the variate type of `LKJ` should be changed to a `PDMat`, which possibly seemed to remove the need for a dedicated `LKJCholesky`.]

sethaxen:
> Hmm... My interpretation and use of `LKJCholesky` so far has been that it is `LKJ` with a more convenient and numerically more efficient (already factorized) variate.

> > the logpdf of LKJCholesky is defined wrt the Lebesgue measure on the strict upper/lower triangle of the triangular factor
> 
> I haven't been aware of this (or at least I don't remember it), and it still seems unclear (surprising?) given the docstring of `LKJCholesky`.

I'd originally proposed in #1336 adding a `MatrixVariate` distribution for the triangular factor of `LKJ`, and through discussion we agreed a `Cholesky` object would be more useful. But I'd always intended for the underlying random variable here to be the triangular factor (hence the different `logpdf`).  The docstring of `LKJCholesky` doesn't mention the assumed reference measure. As noted in https://github.com/JuliaStats/Distributions.jl/issues/224#issuecomment-42464015, each variate type assumes absolute continuity wrt some measure, but it's never documented. In this case the measure was wrt the strict lower triangle. This is why we compute the logdetjac this way: https://github.com/JuliaStats/Distributions.jl/blob/957f0c0f79307edefb9afa75c682ef18b35dd982/test/cholesky/lkjcholesky.jl#L44-L75

> [I have been wondering for a while even whether the variate type of `LKJ` should be changed to a `PDMat`, which possibly seemed to remove the need for a dedicated `LKJCholesky`.]

I agree this would be better. But since this would be breaking, is this likely to happen anytime soon? FWIW, I've had a branch for some time with implementations of a `WishartCholesky` and `InverseWishartCholesky` I've been meaning to finish up; `Wishart` and `InverseWishart` could likewise get a `PDMat` variate type.

devmotion:
I removed the definition of `mean(::LKJCholesky)` from this PR.

>  I'm not certain what Distribution's policy is about non-unique modes; if non-unique modes are allowed, then identity matrix could still be returned for η = 1.

My understanding was that `mode` returns the mode of an unimodal distribution. This seems to be in line with the following comment: https://github.com/JuliaStats/Distributions.jl/blob/957f0c0f79307edefb9afa75c682ef18b35dd982/src/Distributions.jl#L239

sethaxen:
> My understanding was that `mode` returns the mode of an unimodal distribution. This seems to be in line with the following comment:
> 
> https://github.com/JuliaStats/Distributions.jl/blob/957f0c0f79307edefb9afa75c682ef18b35dd982/src/Distributions.jl#L239

As a counterpoint, for a `UnivariateDistribution` the docstring of `mode` clarifies it returns the first mode (hence `modes` for getting all modes):

https://github.com/JuliaStats/Distributions.jl/blob/957f0c0f79307edefb9afa75c682ef18b35dd982/src/univariates.jl#L191-L203

There's no corresponding docstring for `mode` for other variates, but `Dirichlet` does raise an error if requesting the `mode` of a uniform Dirichlet distribution. So for consistency I think it makes the most sense to require `η>1` here.

test/cholesky/lkjcholesky.jl
```diff
@@ -124,14 +124,21 @@ using FiniteDifferences
     end
 
     @testset "properties" begin
-        @testset for p in (4, 5), η in (2, 3.5), uplo in ('L', 'U')
+        @testset for p in (4, 5), η in (0.5, 2, 3.5), uplo in ('L', 'U')
```
sethaxen, line 127: Maybe include `1` to test at the boundary.
```suggestion
        @testset for p in (4, 5), η in (0.5, 1, 2, 3.5), uplo in ('L', 'U')
```

sethaxen (approved):
LGTM, just a minor suggestion.

sethaxen:
> I have been wondering for a while even whether the variate type of `LKJ` should be changed to a `PDMat`, which possibly seemed to remove the need for a dedicated `LKJCholesky`

It would be great to have this as its own issue so it doesn't get lost

devmotion:
I opened #1939.

PR #1951, opened 2025-02-27
Ekdist -- exponentiated Kumaraswamy

JJNewkirk:
* [Wikipedia](https://en.wikipedia.org/wiki/Kumaraswamy_distribution
* The exponentiated Kumaraswamy Distribution and its log-transform: Lemonte et al https://www.jstor.org/stable/43601233
I believe my functions are accurate; however in testing the mean of the current kumaraswamy.jl the did not match in some cases. I believe this to be an error in the kumaraswamy.jl in the version I was using. The EK distribution has the Kumaraswamy as a special case, so it is easy to test keeping gamma = 1.

### DIFF
--- a/src/univariate/continuous/ekdist.jl
+++ b/src/univariate/continuous/ekdist.jl
@@ -1,0 +1,222 @@
+module EKDistribution
+
+export EKDist
+"""
+    kumaraswamy(α,β,γ,l,u)
+The exponentiated (scaled) kumaraswamy (EK) distribution is 5 paramter two of which are normalizing constants (l,u) to a range.
+```
+```julia
+kumaraswamy()          # distribution with zero log-mean and unit scale
+params(d)            # Get the parameters, 
+```
+External links
+* [Wikipedia](https://en.wikipedia.org/wiki/Kumaraswamy_distribution
+* The exponentiated Kumaraswamy Distribution and its log-transform: Lemonte et al https://www.jstor.org/stable/43601233
+
+"""
+## NEED TO ACCOMONDATE LOG(1 -1) when x hits the upper bound of support
+## I get errors in trying to extend Distiributions functions in the script because I don't understand modules lol
+using Distributions
+import Distributions: @check_args, @distr_support, params, convert, partype, beta, gamma
+import Distributions: AbstractRNG,  maximum, minimum, convert
+import Distributions:   pdf, logpdf, gradlogpdf, cdf, ccdf, logcdf, logccdf, quantile, cquantile, median, mean, var, skewness, kurtosis, mode
+import Distributions: fit_mle
+
+
+struct EKDist{T<:Real} <: ContinuousUnivariateDistribution
+    α::T
+    β::T
+    γ::T
+    l::T
+    u::T
+    EKDist{T}(α, β, γ, l, u) where {T} = new{T}(α, β, γ, l, u) 
+end
+
+function EKDist(α::T, β::T, γ::T, l::T, u::T; check_args::Bool=true) where {T <: Real}
+    @check_args EKDist (α, α ≥ zero(α)) (β, β ≥ zero(β)) (γ, γ ≥ zero(γ)) (l < u)
+    return EKDist{T}(α, β, γ, l, u) 
+end
+
+EKDist(α::Real, β::Real, γ::Real, l::Real, u::Real; check_args::Bool=true) = 
+    EKDist(promote(α, β, γ, l, u)...; check_args=check_args)
+
+EKDist(α::Integer, β::Integer, γ::Integer, l::Integer, u::Integer ; check_args::Bool=true) = 
+    EKDist(float(α), float(β), float(γ), float(l), float(u); check_args=check_args)
+
+EKDist(α::Real, β::Real, γ::Real; check_args::Bool=true) = EKDist(promote(α, β, γ, 0.0, 1.0)...; check_args=check_args)
+EKDist(α::Real, β::Real; check_args::Bool=true) = EKDist(promote(α, β, 1.0, 0.0, 1.0)...; check_args=check_args)
+
+# @distr_support macro would set constant bounds but we need to take bounds from input
+# so we will just define them ourselves
+maximum(d::EKDist) = d.u
+minimum(d::EKDist) = d.l
+
+#### Conversions
+convert(::Type{EKDist{T}}, α::S, β::S, γ::S, l::S, u::S) where {T <: Real, S <: Real} = 
+        EKDist(T(α), T(β), T(γ), T(l), T(u))
+        
+convert(::Type{EKDist{T}}, d::EKDist) where {T<:Real} = 
+        EKDist{T}(T(d.α), T(d.β), T(d.γ), T(d.l), T(d.u))
+
+convert(::Type{EKDist{T}}, d::EKDist{T}) where {T<:Real} = d
+
+#### Parameters
+
+params(d::EKDist) = (d.α, d.β, d.γ, d.l, d.u)
+partype(::EKDist{T}) where {T} = T
+
+#### Evaluation
+
+function pdf(d::EKDist, x::Real)
+      xs = (x - d.l) / (d.u - d.l)
+      α = d.α
+      β = d.β
+      γ = d.γ
+    return ( α * β * γ * 
+             xs^(α - 1) * 
+             (1 - xs^α)^(β - 1) * 
+             (1 - (1 - xs^α)^β)^(γ - 1)
+           )
+end
+
+function logpdf(d::EKDist, x::Real)
+      xs =  (x - d.l) / (d.u - d.l) 
+      α = d.α
+      β = d.β
+      γ = d.γ
+    return  ( log(α) + log(β) + log(γ) + 
+              log(xs)*(α - 1) + 
+              log(1 - xs^α)*(β - 1) + 
+              log(1 - (1 - xs^α)^β)*(γ - 1)
+            ) 
+end
+
+function logpdf(d::EKDist, x::Real, edgebound::Real)
+    # Failes at exactly 1 or 0, so I guess jitter it by small amount
+    # this only matters to me in likelihood analysis for real data
+    # that might get defined as the max of the data or Fitting
+    # bound parameters as part of the process.
+    xc = clamp( x, d.l + edgebound, d.u - edgebound)
+    return logpdf(d, xc)  
+end
+
+# do we even need gradlogpdf?
+# grad implied gradient but this is definitely derivative wrt x
+function gradlogpdf(d::EKDist, x::Real)
+    outofsupport = (x < d.l) | (x > d.u)
+    xs = (x - d.l) / (d.u - d.l)
+    α = d.α
+    β = d.β
+    γ = d.γ
+
+    z = (α - 1) / xs +
+        -(α * xs^(α - 1)) / (1 - xs^α) * (β - 1) +
+        (α * β * xs^(α - 1)) * (1 - xs^α)^(β - 1) / (1 -(1 - xs^α)^β) * (γ - 1)
+    
+    return outofsupport ? zero(z) : z
+end
+
+function gradlogpdf(d::EKDist, x::Real, edgebound::Real)
+    xc = clamp( x, d.l + edgebound, d.u - edgebound)
+    return gradlogpdf(d, xc)
+end
+
+
+
+function cdf(d::EKDist, x::Real)
+      xs = (x - d.l) / (d.u - d.l)
+      α = d.α
+      β = d.β
+      γ = d.γ
+    return (1 - ( 1 - xs^α)^β)^γ
+end
+
+function ccdf(d::EKDist, x::Real)
+      xs = (x - d.l) / (d.u - d.l)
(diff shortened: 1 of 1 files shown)


src/univariate/continuous/ekdist.jl
```diff
@@ -0,0 +1,224 @@
...
+"""
+## NEED TO ACCOMONDATE LOG(1 -1) when x hits the upper bound of support
+## I get errors in trying to extend Distiributions functions in the script because I don't understand modules lol
+using Distributions
+import Distributions: @check_args, @distr_support, params, convert, partype, beta, gamma
+import Distributions: AbstractRNG,  maximum, minimum, convert
+import Distributions:   pdf, logpdf, gradlogpdf, cdf, ccdf, logcdf, logccdf, quantile, cquantile, median, mean, var, skewness, kurtosis, mode
+import Distributions: fit_mle
+
+
```
ararslan, line 14: The implementation can be extended later to include scaling without being a breaking change, so let's keep it simpler for the time being and use the unit interval.

Also, there's no need to load those various symbols from the `Distributions` namespace given that this file is included within that namespace; it can already access the symbols unqualified.

The suggestion here makes the docstring content consistent with that of `Kumaraswamy`.

````suggestion
"""
    EKDist(α, β, γ)

The *exponentiated Kumaraswamy (EK) distribution* with shape parameters `α > 0`, `β > 0`,
and `γ > 0` has probability density function

```math
f(x; α, β, γ) = α β γ x^{α - 1} (x - x^α)^{β - 1} (1 - (1 - x^α)^β)^{γ - 1}, \\quad 0 < x < 1
```

When `γ == 1`, this reduces to the [Kumaraswamy distribution](@ref Kumaraswamy).

References

- Lemonte, A. J., Barreto-Souza, W., & Cordeiro, G. M. (2013). The exponentiated Kumaraswamy
  distribution and its log-transform. Brazilian Journal of Probability and Statistics, 27(1),
  31–53. http://www.jstor.org/stable/43601233
"""
````

File: src/univariate/continuous/ekdist.jl [hunk 2]
```diff
@@ -0,0 +1,224 @@
...
+struct EKDist{T<:Real} <: ContinuousUnivariateDistribution
+    α::T
+    β::T
+    γ::T
+    l::T
+    u::T
+    EKDist{T}(α, β, γ, l, u) where {T} = new{T}(α, β, γ, l, u)
```
ararslan:
```suggestion
    EKDist{T}(α, β, γ) where {T} = new{T}(α, β, γ)
```
Removing the bounds as noted in the previous comment.

src/univariate/continuous/ekdist.jl
```diff
@@ -0,0 +1,224 @@
...
+    β::T
+    γ::T
+    l::T
+    u::T
+    EKDist{T}(α, β, γ, l, u) where {T} = new{T}(α, β, γ, l, u) 
+end
+
+function EKDist(α::T, β::T, γ::T, l::T, u::T; check_args::Bool=true) where {T <: Real}
+    @check_args EKDist (α, α ≥ zero(α)) (β, β ≥ zero(β)) (γ, γ ≥ zero(γ)) (l < u)
+    return EKDist{T}(α, β, γ, l, u)
```
ararslan:
```suggestion
function EKDist(α::T, β::T, γ::T; check_args::Bool=true) where {T<:Real}
    @check_args EKDist (α, α > zero(α)) (β, β > zero(β)) (γ, γ > zero(γ))
    return EKDist{T}(α, β, γ) 
```
Lemonte et al. define the parameters to be strictly positive, not non-negative. That's noted for `γ` at the beginning of section 2, and for `α` and `β` immediately following equation 2.1. (That also follows from the definition of the Kumaraswamy distribution, for which the shape parameters are positive.)

src/univariate/continuous/ekdist.jl
```diff
@@ -0,0 +1,224 @@
...
+end
+
+EKDist(α::Real, β::Real, γ::Real, l::Real, u::Real; check_args::Bool=true) = 
+    EKDist(promote(α, β, γ, l, u)...; check_args=check_args)
+
+EKDist(α::Integer, β::Integer, γ::Integer, l::Integer, u::Integer ; check_args::Bool=true) = 
+    EKDist(float(α), float(β), float(γ), float(l), float(u); check_args=check_args)
+
+EKDist(α::Real, β::Real, γ::Real; check_args::Bool=true) = EKDist(promote(α, β, γ, 0.0, 1.0)...; check_args=check_args)
+EKDist(α::Real, β::Real; check_args::Bool=true) = EKDist(promote(α, β, 1.0, 0.0, 1.0)...; check_args=check_args)
```
ararslan: You can combine this with the method for integer arguments by adding a call to `float` in this method, which will just be a no-op for when the arguments are already promoted to a float type.

I've also replaced here the methods that set default values with a single method that takes no arguments. This is for consistency with other distributions, e.g. Kumaraswamy. I think all arguments should be required to be supplied when any are supplied; if someone is relying on a default of `γ == 1`, they should just use `Kumaraswamy` directly instead.

```suggestion
EKDist(α::Real, β::Real, γ::Real; check_args::Bool=true) =
    EKDist(float.(promote(α, β, γ))...; check_args=check_args)

EKDist() = EKDist{Float64}(1.0, 1.0, 1.0)
```

src/univariate/continuous/ekdist.jl
```diff
@@ -0,0 +1,224 @@
...
+
+# @distr_support macro would set constant bounds but we need to take bounds from input
+# so we will just define them ourselves
+maximum(d::EKDist) = d.u
+minimum(d::EKDist) = d.l
```
ararslan:
```suggestion
@distr_support EKDist 0 1
```
Also note that you can use parameters as distribution support limits with this macro, see e.g. `Cosine`.

src/univariate/continuous/ekdist.jl
```diff
@@ -0,0 +1,224 @@
...
+
+#### Conversions
+convert(::Type{EKDist{T}}, α::S, β::S, γ::S, l::S, u::S) where {T <: Real, S <: Real} = 
+        EKDist(T(α), T(β), T(γ), T(l), T(u))
+        
+convert(::Type{EKDist{T}}, d::EKDist) where {T<:Real} = 
+        EKDist{T}(T(d.α), T(d.β), T(d.γ), T(d.l), T(d.u))
```
ararslan, line 43: The first method is incorrect: the signature for `convert` methods should generally be `convert(type, value)` with just a single value.

```suggestion
convert(::Type{EKDist{T}}, d::EKDist) where {T<:Real} = EKDist{T}(T(d.α), T(d.β), T(d.γ))
```

src/univariate/continuous/ekdist.jl
```diff
@@ -0,0 +1,224 @@
...
+end
+
+function logpdf(d::EKDist, x::Real, edgebound::Real)
+    # Failes at exactly 1 or 0, so I guess jitter it by small amount
+    # this only matters to me in likelihood analysis for real data
+    # that might get defined as the max of the data or Fitting
+    # bound parameters as part of the process.
+    xc = clamp( x, d.l + edgebound, d.u - edgebound)
+    return logpdf(d, xc)  
+end
```
ararslan, line 85: There's no need to implement both `pdf` and `logpdf` here given that a fallback exists based on `logpdf`, which will likely be more numerically accurate anyway.

Here you can use the construction of the exponentiated Kumaraswamy distribution as described in section 2 of Lemonte, i.e. noting that the density has the form $$f(x) = \gamma G(x)^{\gamma - 1} g(x)$$ for $$G$$ and $$g$$ the Kumaraswamy CDF and PDF, respectively. Then you can apply logarithm rules and use the existing functions defined on `Kumaraswamy`. That will also ensure correct boundary handling.

```suggestion
# `pdf`: Uses fallback `exp(logpdf(_))` method

function logpdf(d::EKDist, x::Real)
    α, β, γ = params(d)
    k = Kumaraswamy{partype(d)}(α, β)
    return logpdf(k, x) + (γ - 1) * logcdf(k, x) + log(γ)
end
```

src/univariate/continuous/ekdist.jl
```diff
@@ -0,0 +1,224 @@
...
+
+# do we even need gradlogpdf?
+# grad implied gradient but this is definitely derivative wrt x
```
ararslan, line 88: Correct, it is the gradient with respect to `x`

src/univariate/continuous/ekdist.jl
```diff
@@ -0,0 +1,224 @@
...
+function gradlogpdf(d::EKDist, x::Real)
+    outofsupport = (x < d.l) | (x > d.u)
+    xs = (x - d.l) / (d.u - d.l)
+    α = d.α
+    β = d.β
+    γ = d.γ
+
+    z = (α - 1) / xs +
+        -(α * xs^(α - 1)) / (1 - xs^α) * (β - 1) +
+        (α * β * xs^(α - 1)) * (1 - xs^α)^(β - 1) / (1 -(1 - xs^α)^β) * (γ - 1)
```
ararslan, line 98: You may be able to simplify using a similar setup as in my suggestion for `logpdf`, i.e. construct a `Kumaraswamy` and use its methods in the computation.

src/univariate/continuous/ekdist.jl
```diff
@@ -0,0 +1,224 @@
...
+    
+    return outofsupport ? zero(z) : z
+end
+
+function gradlogpdf(d::EKDist, x::Real, edgebound::Real)
```
ararslan, line 103: The `edgebound` stuff shouldn't be necessary. What is the problem you ran into that necessitated it?

src/univariate/continuous/ekdist.jl
```diff
@@ -0,0 +1,224 @@
...
+end
+
+function logcdf(d::EKDist, x::Real)
+    return log(cdf(d, x))
+end
+
+function logccdf(d::EKDist, x::Real)
+    return log(ccdf(d, x))
+end
+
```
ararslan: These methods aren't necessary, as they're identical to the fallback methods.
```suggestion
```

src/univariate/continuous/ekdist.jl
```diff
@@ -0,0 +1,224 @@
...
+
+
+function cdf(d::EKDist, x::Real)
+      xs = (x - d.l) / (d.u - d.l)
+      α = d.α
+      β = d.β
+      γ = d.γ
+    return (1 - ( 1 - xs^α)^β)^γ
```
ararslan:
```suggestion
    α, β, γ = params(d)
    y = (1 - (1 - clamp(x, 0, 1)^α)^β)^γ
    return x < 0 ? zero(y) : (x > 1 ? one(y) : y)
```

src/univariate/continuous/ekdist.jl
```diff
@@ -0,0 +1,224 @@
...
+
+#### Sampling
+# `rand`: Uses fallback inversion sampling method
+
+## Fitting
+
+function fit_mle(::Type{<:EKDist}, x::AbstractArray{T}) where T<:Real
+    #incomplete
+end
+
```
ararslan: Better to not define the method at all than to have it return `nothing`.
```suggestion
# `fit_mle` not yet implemented
```

src/univariate/continuous/ekdist.jl
```diff
@@ -0,0 +1,224 @@
...
+
+#### Statistics
+median(d::EKDist) = quantile(d, .5) # right?
```
ararslan: This is identical to the fallback definition.
```suggestion
```

src/univariate/continuous/ekdist.jl
```diff
@@ -0,0 +1,224 @@
...
+
+function ekdmoment(d::EKDist, k::Real)
+    α = d.α
+    β = d.β
+    γ = d.γ
+
+    # can't tell from paper if its infinite sum or stops at the floor of γ -1
```
ararslan, line 142: Which part of the paper are you looking at? They looked like infinite sums to me, but maybe I missed something.

src/univariate/continuous/ekdist.jl
JJNewkirk: Sorry on the huge delay getting back to this. What are you thoughts on keeping the parameter bounds? The reason I included them is becuase writing this was in part to support some bayesian inference to the l, u parameters when working with data with unknown maximum points. I can go along with the idea of simple (and adhering to paper source) as better, but it seemed simple enough to include these into the parameter of the function itself.

src/univariate/continuous/ekdist.jl
JJNewkirk: I initially did try to fallback to Kumaraswamy but there are inaccuracy in that distribution (filed issue there) and I don't want to connect two separate implementations and introduce silent numerical discrepancies. I am cool with just not allowing there to be partial missing parameters though. I only included this to keep myself some occasional keystrokes in my studies. If the EKDist is a parent of Kumaraswamy, I would see value in simply having Kumaraswamy point to EKDist on this implementation.

JJNewkirk:
All the suggestions were helpful and reasonable. Comitted as suggested.

Issue #1948, opened 2025-02-20
InverseWishart constructor errors for `::Int, ::PDMats.ScalMat{Int64}`


vandenman:
```julia
InverseWishart(5, PDMats.ScalMat(5, 1))
```
throws
```julia
ERROR: MethodError: no method matching Int64(::IrrationalConstants.Logtwo)
```

because here
https://github.com/JuliaStats/Distributions.jl/blob/e470421079f2eb4f725657842fbbab40bd981fc8/src/matrix/inversewishart.jl#L110-L114

the line `typeof(df)(logtwo)` becomes `(Int)(logtwo)`.

proposed fix:

```diff
function invwishart_logc0(df::Real, Ψ::AbstractPDMat)
    h_df = df / 2
    p = size(Ψ, 1)
+	logdet_Ψ = logdet(Ψ)
+   -h_df * (p * oftype(logdet_Ψ, logtwo) - logdet(Ψ)) - logmvgamma(p, h_df)
-	-h_df * (p * typeof(df)(logtwo) - logdet(Ψ)) - logmvgamma(p, h_df)
end
```

which uses the same logic as the Wishart:

https://github.com/JuliaStats/Distributions.jl/blob/e470421079f2eb4f725657842fbbab40bd981fc8/src/matrix/wishart.jl#L182-L186

I'd be happy to open a PR!

PR #1953, opened 2025-03-04
Fix inverse wishart constructor with integer ScalMat/PDiagMat

vandenman:
Fixes https://github.com/JuliaStats/Distributions.jl/issues/1948

Feedback is welcome!

### DIFF
--- a/src/matrix/inversewishart.jl
+++ b/src/matrix/inversewishart.jl
@@ -110,7 +110,8 @@ end
 function invwishart_logc0(df::Real, Ψ::AbstractPDMat)
     h_df = df / 2
     p = size(Ψ, 1)
-    -h_df * (p * typeof(df)(logtwo) - logdet(Ψ)) - logmvgamma(p, h_df)
+    logdet_Ψ = logdet(Ψ)
+    -h_df * (p * oftype(logdet_Ψ, logtwo) - logdet_Ψ) - logmvgamma(p, h_df)
 end
 
 function logkernel(d::InverseWishart, X::AbstractMatrix)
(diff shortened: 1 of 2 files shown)


src/matrix/inversewishart.jl
```diff
@@ -110,7 +110,8 @@ end
 function invwishart_logc0(df::Real, Ψ::AbstractPDMat)
     h_df = df / 2
```
devmotion, line 111: Since you're already modifying this line, maybe improve `Float32` compatibility here? AFAICT currently `df::Int` and `::AbstractPDMat{Float32}` would result in a `logc0::Float64` due to this line.
```suggestion
```

src/matrix/inversewishart.jl
```diff
@@ -110,7 +110,8 @@ end
...
     p = size(Ψ, 1)
-    -h_df * (p * typeof(df)(logtwo) - logdet(Ψ)) - logmvgamma(p, h_df)
+    logdet_Ψ = logdet(Ψ)
```
devmotion, line 113: I think the `Float32` issue could be fixed e.g. in the following way:
```suggestion
    logdet_Ψ, _df = promote(logdet(Ψ), df)
    h_df = _df / 2
```

src/matrix/inversewishart.jl
vandenman: I'm not sure if this is a problem because 
https://github.com/JuliaStats/Distributions.jl/blob/e470421079f2eb4f725657842fbbab40bd981fc8/src/matrix/inversewishart.jl#L40-L43

means `InverseWishart(5, ScalMat(5, 1f0))` will call `InverseWishart(5f0, ScalMat(5, 1f0))`. I've nevertheless fixed it so that `invwishart_logc0(::Int, PDMat{<:Float32})` doesn't return Float64.

devmotion (approved):

PR #1957, opened 2025-03-28
Allow ForwardDiff v1

mcabbott:
Not sure why CompatHelper didn't make a PR for this.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.25.119"
+version = "0.25.118"
 
 [deps]
 AliasTables = "66dad0bd-aa9a-41b7-9441-69ab47430ed8"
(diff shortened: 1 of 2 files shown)


sethaxen:
> Not sure why CompatHelper didn't make a PR for this.

It still doesn't make PRs for dependencies only in `[extras]`

devmotion:
Which is annoying for CI but at least irrelevant for all downstream packages as the compat entries of `[extras]` dependencies are not respected when Distributions is used as a dependency somewhere.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.25.118"
+version = "0.25.119"
```
devmotion, line 4: The PR doesn't change the package, pnly the tests, so we don't need a new release:
```suggestion
version = "0.25.118"
```

mcabbott: I made this because `add ForwardDiff@1` leads to "ERROR: Unsatisfiable requirements detected for package Distributions [31c24e10]:", which involved LsqFit... but you may be right that it's not what's bounding ForwardDiff, not really sure.

devmotion: Did you encounter it when trying to update a downstream package?

PR #1963, opened 2025-04-09
Add multivariate hypergeometric distribution

Michael-Howes:
### Overview

This pull requests adds the [multivariate hypergeometric distribution](https://en.wikipedia.org/wiki/Hypergeometric_distribution#Multivariate_hypergeometric_distribution). This is a generalization of the hypergeometric distribution. It includes:

- The file `src/multivariate/mvhypergeometric.jl` that implements `MvHypergeometric` as a subtype of `DiscreteMultivariateDistribution`.
- The file `src/samplers/mvhypergeometric.jl` that implements sampling.
- A test file `test/multivariate/mvhypergeometric.jl`.
- The documentation file `docs/src/multivariate.md` is also updated.

### Motivation

The multivariate hypergeometric distribution is an important distribution in statistics for testing independence in contingency tables. It is implemented in the `numpy` and `scipy` Python packages but currently it is not supported in `Distributions.jl`.

### Implementation details

The type `MvHypergeometric` is created as a subtype of `DiscreteMultivariateDistribution`. Functions for the mean, variance and covariance matrix are implemented. Evaluation of the log pdf and sampling are also implemented. Sampling is implemented in the file `scr/samplers/mvhypergeom.jl`. The procedure is analogous to sampling from a multinomial distribution. The entries are sampled sequentially from univariate hypergeometric distributions.

### Testing

Tests are include in `test/multivariate/mvherpgeom.jl`. The statistics, pdf and sampling are all tested. The pdf is also compared to the pdf of the hypergeometric distribution. Specifically, the marginal and conditional distributions of the multivariate hypergeometric are univariate hypergeometric (as used in the sampling).

### Dependencies

No new dependencies are added.


src/multivariate/mvhypergeometric.jl
```diff
@@ -0,0 +1,125 @@
+"""
+The [Multivariate hypergeometric distribution](https://en.wikipedia.org/wiki/Hypergeometric_distribution#Multivariate_hypergeometric_distribution)
```
devmotion, line 4: Julia docstrings should contain the type or function signature in the first line, indented by four spaces.

Generally, could you make this docstring consistent with docstrings of other existing distribution types?

src/multivariate/mvhypergeometric.jl
```diff
@@ -0,0 +1,125 @@
...
+\\quad x_1 + \\cdots + x_k = n, \\quad x_i \\le m_i
+```
+
+```julia
+MvHypergeometric(m, n)   # Multivariate hypergeometric distribution for a population with
+                         # m = (m_1, ..., m_k) elements of type 1 to k and n draws
+```
+"""
+struct MvHypergeometric <: DiscreteMultivariateDistribution
+    m::AbstractVector{Int}  # number of elements of each type
```
devmotion: Fields should not be abstractly typed. If it's necessary or useful to support more general vector types than just `Vector{Int}`, you should make the vector type a type parameter.

src/multivariate/mvhypergeometric.jl
```diff
@@ -0,0 +1,125 @@
...
+    n::Int                  # number of draws
+    function MvHypergeometric(m::AbstractVector{Int}, n::Int; check_args::Bool=true)
+        @check_args(
+            MvHypergeometric,
+            (m, all(m .>= zero.(n))),
```
devmotion: This can be optimized by avoiding the intermediate array, e.g.
```suggestion
            (m, all(x -> x >= 0, m)),
```

src/multivariate/mvhypergeometric.jl
```diff
@@ -0,0 +1,125 @@
...
+        new(m, n)
+    end
+end
+
+
+# Parameters
+
+ncategories(d::MvHypergeometric) = length(d.m)
+length(d::MvHypergeometric) = ncategories(d)
+nelements(d::MvHypergeometric) = d.m
```
devmotion: I'm not convinced that we should introduce this function (and in particular make it part of the public API) in the first iteration of this new distribution type. Its name seems a bit unclear to an unfamiliar user I suppose, and it can already be accessed anyway as a parameter of the distribution.

devmotion:
```suggestion
```

src/Distributions.jl
```diff
@@ -244,6 +245,7 @@ export
     ncategories,        # the number of categories in a Categorical distribution
     ncomponents,        # the number of components in a mixture model
     ntrials,            # the number of trials being performed in the experiment
+    nelements,          # the number of elements of each type in a finite population
```
devmotion: See my comment below:
```suggestion
```

src/multivariate/mvhypergeometric.jl
```diff
@@ -0,0 +1,125 @@
...
+ntrials(d::MvHypergeometric) = d.n
+
+params(d::MvHypergeometric) = (d.m, d.n)
```
devmotion, line 40: IMO it would be good to also explicitly define `partype`:
```suggestion
params(d::MvHypergeometric) = (d.m, d.n)
partype(::MvHypergeometric) = Int
```

src/multivariate/mvhypergeometric.jl
```diff
@@ -0,0 +1,125 @@
...
+
+function var(d::MvHypergeometric)
+    m = nelements(d)
+    k = length(m)
+    n = ntrials(d)
+    M = sum(m)
+    p = m / M
+    f = n * (M - n) / (M - 1)
+
+    v = Vector{Real}(undef, k)
```
devmotion: Containers with abstract element types should be avoided.

src/multivariate/mvhypergeometric.jl
```diff
@@ -0,0 +1,125 @@
...
+    for i = 1:k
+        @inbounds p_i = p[i]
```
devmotion: This `@inbounds` is wrong if `p` does not use 1-based indexing and `@inbounds` should generally be avoided unless there is a very strong reason (see #2005).

src/multivariate/mvhypergeometric.jl
```diff
@@ -0,0 +1,125 @@
...
+        v[i] = f * p_i * (1 - p_i)
+    end
```
devmotion: To reduce allocations and avoid the non-concretely typed return value, you could change the code to something like

```suggestion
    f = n * (M - n) / (M - 1)
    v = let f = f
        map(mi -> f * (mi / M) * ((M - mi) / M), m)
    end
```

src/multivariate/mvhypergeometric.jl
```diff
@@ -0,0 +1,125 @@
...
+
+function cov(d::MvHypergeometric)
+    m = nelements(d)
+    k = length(m)
+    n = ntrials(d)
+    M = sum(m)
+    p = m / M
+    f = n * (M - n) / (M - 1)
+
+    C = Matrix{Real}(undef, k, k)
```
devmotion: Also here, we should avoid abstractly typed return values.

src/multivariate/mvhypergeometric.jl
```diff
@@ -0,0 +1,125 @@
...
+    for j = 1:k
+        pj = p[j]
+        for i = 1:j-1
+            @inbounds C[i, j] = -f * p[i] * pj
```
devmotion: These `@inbounds` are unsafe as well.

src/multivariate/mvhypergeometric.jl
```diff
@@ -0,0 +1,125 @@
...
+    s = 0.0
+    for i = 1:k
+        @inbounds xi = x[i]
+        if !(isinteger(xi) && xi >= 0 && xi <= m[i])
+            return false
+        end
+        s += xi
+    end
+    return s == ntrials(d)  # integer computation would not yield truncation errors
+end
```
devmotion: Again, use of `@inbounds` is unsafe here. Moreover, I don't see why `s` should be initialized as `Float64`.

Maybe use something like
```suggestion
function insupport(d::MvHypergeometric, x::AbstractVector{<:Real})
    return length(x) == length(d) && (eltype(x) <: Integer || all(isinteger, x)) && all(((xi, mi),) -> zero(xi) <= xi <= mi, zip(x, d.m)) && sum(x) == ntrials(d)
end
```

src/multivariate/mvhypergeometric.jl
```diff
@@ -0,0 +1,125 @@
...
+
+function _logpdf(d::MvHypergeometric, x::AbstractVector{T}) where T<:Real
```
devmotion: No need for a where clause here:
```suggestion
function _logpdf(d::MvHypergeometric, x::AbstractVector{<:Real})
```

src/multivariate/mvhypergeometric.jl
```diff
@@ -0,0 +1,125 @@
...
+    m = nelements(d)
+    M = sum(m)
+    n = ntrials(d)
+    insupport(d, x) || return -Inf
```
devmotion: This leads to type unstable code - sometimes the function returns `Float64`, sometimes some other return type, for the same types of arguments.

devmotion: It also is maybe unnecessarily expensive to check whether values are in the support of the distribution in addition to the evaluations.

src/multivariate/mvhypergeometric.jl
```diff
@@ -0,0 +1,125 @@
...
+    s = -logabsbinomial(M, n)[1]
+    for i = 1:length(m)
+        @inbounds xi = x[i]
+        @inbounds m_i = m[i]
+        s += logabsbinomial(m_i, xi)[1]
```
devmotion: These `@inbounds` are unsafe.

src/multivariate/mvhypergeometric.jl
```diff
@@ -0,0 +1,125 @@
...
+    end
+    return s
+end
+
+# Sampling is performed by sequentially sampling each entry from the
+# hypergeometric distribution
+_rand!(rng::AbstractRNG, d::MvHypergeometric, x::AbstractVector{Int}) =
+    mvhypergeom_rand!(rng, nelements(d), ntrials(d), x)
```
devmotion: Dedicated sampler implementations in src/samplers/ are only useful if they are more efficient (e.g. by precomputing some cache). This doesn't seem to be the case here, so all the code should just be moved to `_rand!`.

src/samplers/mvhypergeometric.jl
```diff
@@ -0,0 +1,32 @@
...
+    k = length(m)
+    length(x) == k || throw(DimensionMismatch("Invalid argument dimension."))
+
+    M = sum(m)
+    i = 0
+    km1 = k - 1
+
+    while i < km1 && n > 0
+        i += 1
+        @inbounds mi = m[i]
```
devmotion: This is unsafe.

src/samplers/mvhypergeometric.jl
```diff
@@ -0,0 +1,32 @@
...
+        # Sample from hypergeometric distribution. Element of type i are 
+        # considered successes. All other elements are considered failures.
+        xi = rand(rng, Hypergeometric(mi, M - mi, n))
+        @inbounds x[i] = xi
```
devmotion: This is unsafe.

src/samplers/mvhypergeometric.jl
```diff
@@ -0,0 +1,32 @@
...
+        # Remove elements of type i from population and group to be sampled.
+        n -= xi
+        M -= mi
+    end
+
+    if i == km1
+        @inbounds x[k] = n
```
devmotion: This is unsafe.

src/samplers/mvhypergeometric.jl
```diff
@@ -0,0 +1,32 @@
...
+    else  # n must have been zero.
+        z = zero(eltype(x))
+        for j = i+1:k
+            @inbounds x[j] = z
```
devmotion: This is unsafe.

src/multivariate/mvhypergeometric.jl
Michael-Howes: Done. Thanks

src/multivariate/mvhypergeometric.jl
Michael-Howes: I changed the function to compute an outer product for the off-diagonals and then update the diagonal. Let me know if this is okay.

src/multivariate/mvhypergeometric.jl
Michael-Howes: Accepted.

src/multivariate/mvhypergeometric.jl
Michael-Howes: I change this to `-Float64(Inf)` based on this: https://github.com/JuliaStats/Distributions.jl/blob/0421b1890c49417de993b255f6b177a1cb93fca3/src/multivariate/multinomial.jl#L154

For your second comment, are you suggesting checking `xi >= 0` and `xi <= mi` in the for loop and then returning -inf if the condition is violated at any point?

src/multivariate/mvhypergeometric.jl
Michael-Howes: Done.

src/multivariate/mvhypergeometric.jl
Michael-Howes: I've added the function signature. The doc string is based on the multinomial distribution docstring. Is there a different distribution I should be following? 

https://github.com/JuliaStats/Distributions.jl/blob/0421b1890c49417de993b255f6b177a1cb93fca3/src/multivariate/multinomial.jl

src/multivariate/mvhypergeometric.jl
Michael-Howes: I've updated the code to do the above checks within the for loop.

Issue #1964, opened 2025-04-10
MvNormal unnecessarily slow when mean and covariance matrix have different types


hersle:
My model has a `MvNormal` with a constant covariance matrix (just `Float64`s) and a differentiable mean vector (e.g. with `ForwardDiff.Dual`s):
```julia
using Distributions, ForwardDiff, PDMats, LinearAlgebra, BenchmarkTools

Σ = PDMat(Matrix(1.0I, 1000, 1000)) # e.g. constant model-independent covariance matrix
μ = ForwardDiff.Dual.(1.0:1000.0) # e.g. differentiable model-dependent data vector

MvNormalSlow(μ, Σ) = MvNormal(μ, Σ) # standard constructor
MvNormalFast(μ, Σ) = MvNormal{Real, typeof(Σ), typeof(μ)}(μ, Σ) # construct manually

@btime MvNormalSlow(μ, Σ); # 1.789 ms (7 allocations: 15.26 MiB)
@btime MvNormalFast(μ, Σ); # 17.818 ns (1 allocation: 48 bytes)
```
The standard `MvNormal` constructor is unnecessarily slow. It seems to copy/refactorize covariance matrix, even though it is already factorized in my call to `PDMat` (which I use precisely because I know my covariance matrix is constant and I don't want to refactorize it for every model evaluation):
```julia
@profview [MvNormalSlow(μ, Σ) for i in 1:100]
```

![Image](https://github.com/user-attachments/assets/76911448-3bfa-4d17-936d-d8a52dd2c89c)

I think the reason is [here](https://github.com/JuliaStats/Distributions.jl/blob/d666acbcad3d2dea5ce2d07cc4a845d0933b7015/src/multivariate/mvnormal.jl#L185-L193). When `μ` and `Σ` are of different types (like here), it hits the second constructor, triggering a conversion and refactorization of the covariance matrix.

Is this intended? Is there a smooth fix to it?

devmotion:
> Is this intended?

It predates my times as a contributor, and it seems problematic in this case, but I'm quite certain that it is intentional. Having parameters of the same element types - as for most univariate distributions - makes it easier to reason about the "parameter type" of a distribution and to avoid accidental type instabilities (if in a special case, e.g., zero mean, a term or a whole branch of the computation is skipped).

[...]

PR #1965, opened 2025-04-11
Do not promote mean and covariance of `MvNormal` to common element type

devmotion:
Ref #1964. Currently just to see if and what it would break.


hersle:
Thanks for giving it a go!

How bad is this? I can imagine several of these test failures are fixable.

What really matters at the end of the day? That the [core API functions](https://github.com/JuliaStats/Distributions.jl/blob/d666acbcad3d2dea5ce2d07cc4a845d0933b7015/src/multivariate/mvnormal.jl#L10-L18) are type-stable and respect [`eltype`](https://github.com/JuliaStats/Distributions.jl/blob/d666acbcad3d2dea5ce2d07cc4a845d0933b7015/src/multivariate/mvnormal.jl#L226)?

hersle:
Is this good to merge, or are there more things to consider?

devmotion:
No, this is a major change, so it requires a bit more discussion, feedback from other maintainers, and should in the end be approved by another maintainer.

I'm also not sure yet whether the workaround for `UniformScaling` is desirable; maybe that codepath should be deprecated completely in favour of `FillArrays.Eye` which in contrast to `UniformScaling` has a clearly defined size.

PR #1966, opened 2025-04-11
Remove `median(::MvLogNormal)`

devmotion:
The PR removes the definition of `median(::MvLogNormal)`. It was added in https://github.com/JuliaStats/Distributions.jl/pull/455 but IMO that should have been discussed more thoroughly.

1. The marginal median is just one of multiple concepts of multivariate medians, see e.g. https://en.wikipedia.org/wiki/Median#Multivariate_median.

2. There's no established concept of multivariate medians in Distributions as indicated by the fact that `median` is not defined for any other multivariate distribution (not even `MvNormal`).

3. The definition of `median` seems to suggest that `quantile` etc. should be defined as well, which is not the case.

As such, I think the existence of this definition is a bug.

---

Alternatively, Distributions could adopt and clearly document a definition of (marginal?) medians for multivariate distributions. In this case, the fix would be to add definitions of `median` to other distributions such as `MvNormal` etc. as well.


cgarling:
Thanks for your work on maintaining this package! I think marginal medians would be useful statistics to have for all the multivariate distributions. Maybe you could differentiate the marginal median from other definitions by requiring a dims argument e.g. `median(dist::MvLogNormal, dims::Integer)` to specify what dimension you want the marginal median of.

PR #1974, opened 2025-05-05
Document and export gradlogpdf.

### LINKED ISSUE #1844: Documentation for gradlogpdf
Thanks for your work maintaining this package. Quick question: I see that `gradlogpdf` is tested in `test/gradlogpdf.jl` but I do not see the method in the documentation. Is `gradlogpdf` considered part of the public API? If so, it should have a docstring right?

tpapp:
Fixes #1844.

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -224,7 +224,6 @@ export
     loglikelihood,      # log probability of array of IID draws
     logpdf,             # log probability density
     logpdf!,            # evaluate log pdf to provided storage
-    gradlogpdf,         # gradient of log pdf
 
     invscale,           # Inverse scale parameter
     sqmahal,            # squared Mahalanobis distance to Gaussian center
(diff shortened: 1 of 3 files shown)


devmotion (commented):
Thank you! Actually, the function has been exported for a long time (~ 11 years according to git blame), it's listed in line 273 in Distributions.jl

So it would be really good to finally add a docstring 🙂

src/Distributions.jl
```diff
@@ -224,6 +224,7 @@ export
     loglikelihood,      # log probability of array of IID draws
     logpdf,             # log probability density
     logpdf!,            # evaluate log pdf to provided storage
+    gradlogpdf,         # gradient of log pdf
```
devmotion, line 227: It is already exported:
```suggestion
```

devmotion (approved):

Issue #153, opened 2013-10-01
pdfs up to proportionality


simonbyrne:
In many cases, we don't need to compute all the factors of a pdf/logpdf:
- In likelihoods, we can ignore any factors not involving the parameters (e.g. the factorial in the Poisson)
- Many sampling techniques don't require computing the normalisation constant, i.e. factors not involving the point at which the density is evaluated (e.g. the Bessel function in the von Mises distribution)
- Posteriors are typically only known up to proportionality, so we would only need the prior (ignoring the normalisation constant), and the likelihood (ignoring non-parameter factors)

It would be good to have a consistent interface for all of this: we might also want to think about how this relates to autodiff functionality.

samuela:
+100 This is really important.

@sbos It's also worth noting that in many cases calculating the partition function is analytically and/or computationally intractable. This doesn't generally affect the standard distributions that the package targets, but it may be worth keeping in mind as the project evolves.

lindahua:
I am going to implement this (particularly for exponential family distributions). Will make a proposal with RFC pretty soon.

simonbyrne:
Having thought more about this, I decided what was really needed is a ratio of pdfs, so i was thinking:
```julia
pdfratio(d1::Distribution, d2::Distribution, x::Real) = pdf(d1,x) / pdf(d2,x)
pdfratio(d::Distribution, x1::Real, x2::Real) = pdf(d,x1) / pdf(d,x2)
pdfratio(d1::Distribution, d2::Distribution, x1::Real, x2::Real) = pdf(d1,x1) / pdf(d2,x2)
```

lindahua:
For multi-class conditional distribution, we still need unnormalized pdfs.
```
pdf( x | k )  # where k can take 1, 2, 3, ... K 
```
This is very useful in computing posterior probability.

sethaxen:
As far as I can tell, given a density $p(x \mid \theta) = f(x, \theta) g(\theta) h(x)$, it would be useful to have the following functions
- `logupdf(x) = ` $\log f(x, \theta) + \log h(x)$: used whenever parameters are fixed but arguments vary
- `logulikelihood(x) = ` $\log f(x, \theta) + \log g(\theta)$: used whenever arguments are fixed but parameters vary
- `logpdf(x) = ` $\log p(x \mid \theta)$: used whenever both arguments and parameters vary

[...]

PR #1978, opened 2025-05-12
Add unnormalized PDF and log-likelihood functions

sethaxen:
As suggested in https://github.com/JuliaStats/Distributions.jl/issues/153#issuecomment-2862700149, this PR adds the following API functions:
- `logupdf`: contains all terms in `logpdf` that depend on the argument
- `pdf`: contains all terms in `pdf`  that depend on the argument
- `logulikelihood`: contains all terms in `loglikelihood` that depend on the parameters

Each of these has a fallback to the full `logpdf`, `pdf`, or `loglikelihood`, respectively. After surveying the repo, it seems the existing functions could in most cases be split into 1) calling one of these functions and 2) computing the normalization factor. This is left for future PRs.

A few questions:
- Should we implement `_logupdf`, `_updf`, `logupdf!`, `updf!` for `Arrayvariate` analogous to the existing non-`u` versions?
- In the future, should unnormalized variants be added to StatsFuns before adding here? If so, perhaps it makes more sense to open a PR there to work out an API before adding a matching API here?

### DIFF
--- a/src/Distributions.jl
+++ b/src/Distributions.jl
@@ -222,8 +222,10 @@ export
     logdiffcdf,         # log of difference between cdf at two values
     logdetcov,          # log-determinant of covariance
     loglikelihood,      # log probability of array of IID draws
+    logulikelihood,     # unnormalized log probability of array of IID draws
     logpdf,             # log probability density
     logpdf!,            # evaluate log pdf to provided storage
+    logupdf,            # unnormalized log probability density
 
     invscale,           # Inverse scale parameter
     sqmahal,            # squared Mahalanobis distance to Gaussian center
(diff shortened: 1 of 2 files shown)


sethaxen:
I'd prefer to get a :+1: feedback on the general approach from the maintainers before messing around with all of the testing functions. Maybe @devmotion ?

devmotion:
Sorry, I've been swamped with tasks this week. I like the proposal a lot, I think this will resolve one of my main feature requests (apart from fixing `rand`/`eltype` and speeding up `loglikelihood` by factoring out constants and basing it on sufficient statistics).

> I also think updf will not be nearly as useful or widely used as logupdf (EDIT: maybe we shouldn't have updf after all?)

I'd suggest not adding it for now.

> Should we implement _logupdf, _updf, logupdf!, updf! for Arrayvariate analogous to the existing non-u versions?

I don't like the current interface for non-univariate distributions, making `_...` methods part of the (development) API seems wrong.

> In the future, should unnormalized variants be added to StatsFuns before adding here? If so, perhaps it makes more sense to open a PR there to work out an API before adding a matching API here?

Yes, IMO that would be reasonable for distribution functions implemented in StatsFuns.

PR #1982, opened 2025-06-04
Add generalized inverse Gaussian and generalized hyperbolic

### LINKED ISSUE #554: Generalized inverse Gaussian

ForceBru:
This is at least the third attempt at implementing the generalized inverse Gaussian (GIG) distribution to close #554. Previous attempts by other people: https://github.com/JuliaStats/Distributions.jl/pull/587 (closed, last activity in 2019), https://github.com/JuliaStats/Distributions.jl/pull/1300 (open, last activity in 2021).

Features of this implementation:

1. `rand` closely follows Devroye, Luc. 2014. “Random Variate Generation for the Generalized Inverse Gaussian Distribution.” Statistics and Computing 24 (2): 239–46. https://doi.org/10.1007/s11222-012-9367-z. Single algorithm for all values of parameters, as opposed to multiple algos in https://github.com/JuliaStats/Distributions.jl/pull/1300. In my evaluation, the Kolmogorov-Smirnov test fails to reject the null, thus verifying that samples agree with the true GIG for multiple parameter values.
2. Has documentation.
3. Implements PDF, CDF (via `quadgk`), mean, variance, MGF, CF etc from Wikipedia and Wolfram Mathematica.
4. I want to work on this implementation because I need it.

I also implemented, documented and tested the generalized hyperbolic distribution with sampling based on GIG. This probably should be a separate pull request, but they're very related, so the generalized hyperbolic is here for now.

### DIFF
--- a/src/univariate/continuous/generalizedinversegaussian.jl
+++ b/src/univariate/continuous/generalizedinversegaussian.jl
@@ -16,21 +16,21 @@ External links:
 * [Generalized Inverse Gaussian distribution on Wikipedia](https://en.wikipedia.org/wiki/Generalized_inverse_Gaussian_distribution).
 """
 struct GeneralizedInverseGaussian{T<:Real} <: ContinuousUnivariateDistribution
-	a::T
-	b::T
-	p::T
-	function GeneralizedInverseGaussian{T}(a::T, b::T, p::T) where T<:Real
-		new{T}(a, b, p)
-	end
+    a::T
+    b::T
+    p::T
+    function GeneralizedInverseGaussian{T}(a::T, b::T, p::T) where T<:Real
+        new{T}(a, b, p)
+    end
 end
 
 function GeneralizedInverseGaussian(a::T, b::T, p::T; check_args::Bool=true) where T<:Real
-	check_args && @check_args GeneralizedInverseGaussian (a, a > zero(a)) (b, b > zero(b))
-	GeneralizedInverseGaussian{T}(a, b, p)
+    check_args && @check_args GeneralizedInverseGaussian (a, a > zero(a)) (b, b > zero(b))
+    GeneralizedInverseGaussian{T}(a, b, p)
 end
 
 GeneralizedInverseGaussian(a::Real, b::Real, p::Real; check_args::Bool=true) =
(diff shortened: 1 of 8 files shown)


ForceBru:
One design issue I'm not sure how to solve is implementing multiple parameterizations of the same distribution. In particular, I want the usual $\mathrm{GIG}(a,b,p)$ as well as Wolfram's $\mathrm{GIG}(\mu,\lambda,\theta)$. Both have 3 parameters, so I can't implement two different methods with the same number of params. I tried to implement Wolfram's parameterization as a function with required keyword params like `GeneralizedInverseGaussian(; mu::Real, lambda::Real, theta::Real=-1/2)`, but now one test is failing (e.g. https://github.com/JuliaStats/Distributions.jl/actions/runs/15453605220/job/43501263178?pr=1982#step:6:619) because these lines try to call it without any arguments: https://github.com/JuliaStats/Distributions.jl/blob/46bb0c3e167f48e4d58465738ac06c5b9fa93718/test/univariate_bounds.jl#L59-L60

src/univariate/continuous/generalizedinversegaussian.jl
```diff
@@ -0,0 +1,200 @@
...
+
+GeneralizedInverseGaussian(a::Real, b::Real, p::Real; check_args::Bool=true) =
+	GeneralizedInverseGaussian(promote(a, b, p)...; check_args)
+
+"""
+    GeneralizedInverseGaussian(; μ::Real, λ::Real, θ::Real=-1/2)
+
+Wolfram Language parameterization, equivalent to `InverseGamma(μ, λ)`
+"""
+GeneralizedInverseGaussian(; μ::Real, λ::Real, θ::Real=-1/2) =
```
ForceBru: Here I wanted to have a different parameterization, but tests assume that one can call `GeneralizedInverseGaussian()` with no arguments. The test fails because `GeneralizedInverseGaussian` has mandatory keyword arguments here.

PR #1985, opened 2025-06-17
Dispatch for drawing multiples

### LINKED ISSUE #1984: `rand(d, 1000)` is not propagated to components
Sometimes it is more efficient to generate a sample with `rand(d, 1000)` than `[rand(d) for _ in 1:1000]`,
if an overhead is significant. That is the case for `fft_convolve` in [NumericalDistributions.jl](https://github.com/mmikhasenko/NumericalDistributions.jl).

A solution would be to propagate `rand(eng, d, n::Int)` signature to wrapped distributions in `Distributions.jl`.

It concerns wrappers,
 - `MixtureModel`,
 - `Truncated`

Here is my example for testing in https://github.com/mmikhasenko/NumericalDistributions.jl/issues/10
```
using NumericalDistributions

dt = let
	d1 = truncated(Normal(0, 0.02), -1.0, 1.0)
	d2 = truncated(Normal(2, 0.02), 1.0, 4.0)
	d = fft_convolve(d1, d2)
	truncated(d, -1, 3)
end

@time rand(dt.untruncated, 1000) # ms
@time rand(dt, 1000) # 5s
```

mmikhasenko:
Closes #1984 

## Implementation
- [x] Dispatch for `MixtureModel`
- [x] Dispatch for `Truncated`

## Checkpoints
- [x] Tests are passing
- [x] Docs are updated -- discussed that docstrings are placed to dispatched functions for consistency (#1986) 

### Speed

For mixture model,

```julia
julia> using Distributions
julia> using BenchmarkTools

julia> d = MixtureModel([Normal(1,0.1), Normal(2,0.1), Normal(3,0.1)], [0.3,0.3,0.4])
julia> @btime rand($d, 10_000);
  73.708 μs (7 allocations: 125.67 KiB)

julia> @btime [rand($d) for _ in 1:10_000];
  124.125 μs (3 allocations: 96.06 KiB)
```

For truncated
```julia
t = truncated(Normal(), 0, 3)
@btime rand($t, 10_000);  # 114.625 μs (6 allocations: 256.12 KiB)
@btime [rand($t) for _ in 1:10_000];  # 169.875 μs (3 allocations: 96.06 KiB)
```

### Visual

```julia
d = MixtureModel([Normal(1,0.1), Normal(2,0.1), Normal(3,0.1)], [0.3,0.3,0.4])
s1 = [rand(d) for _ in 1:10000]
s2 = rand(d, 10000)
let
    plot(); 
    stephist!(s1, bins=range(0, 4, 100), lab="rand(d) n times")
    stephist!(s2, bins=range(0, 4, 100), lab="rand(d,n)")
end
```

image

```julia
t = truncated(Normal(), 0, 3)
r1 = [rand(t) for _ in 1:10000]
r2 = rand(t, 10000)
let
    plot(); 
    stephist!(t1, bins=range(0, 4, 100), lab="rand(d) n times")
    stephist!(t2, bins=range(0, 4, 100), lab="rand(d,n)")
end
```

image


src/mixtures/mixturemodel.jl
```diff
@@ -477,6 +477,22 @@ rand(rng::AbstractRNG, s::MixtureSampler{Univariate}) =
 rand(rng::AbstractRNG, d::MixtureModel{Univariate}) =
     rand(rng, component(d, rand(rng, d.prior)))
 
+function rand(rng::AbstractRNG, d::MixtureModel{Univariate}, n::Int)
+    counts = rand(rng, Multinomial(n, probs(d.prior)))
+    x = Vector{eltype(d)}(undef, n)
```
devmotion: `eltype` is unreliable, see e.g. #1905.

src/mixtures/mixturemodel.jl
```diff
@@ -477,6 +477,22 @@ rand(rng::AbstractRNG, s::MixtureSampler{Univariate}) =
...
+    offset = 0
+    for i in eachindex(counts)
+        ni = counts[i]
+        if ni > 0
+            c = component(d, i)
+            v = view(x, (offset+1):(offset+ni))
+            v .= rand(rng, c, ni)
+            offset += ni
```
devmotion: Since `v` is updated in-place, we should avoid any additional allocations here and directly call
```suggestion
            last_offset = offset + ni - 1 
            rand!(rng, c, @view(x[(begin + offset):(begin + last_offset)]))
            offset = last_offset + 1
```

src/mixtures/mixturemodel.jl
```diff
@@ -477,6 +477,22 @@ rand(rng::AbstractRNG, s::MixtureSampler{Univariate}) =
 rand(rng::AbstractRNG, d::MixtureModel{Univariate}) =
     rand(rng, component(d, rand(rng, d.prior)))
 
+function rand(rng::AbstractRNG, d::MixtureModel{Univariate}, n::Int)
```
devmotion, line 480: Probably we don't generally want to use this method - `n` has to be sufficiently high.

src/truncate.jl
```diff
@@ -213,6 +213,17 @@ end
...
+"""
+    rand(rng::AbstractRNG, d::Truncated)
+
+Generate a single random sample from a truncated distribution.
+
+The sampling strategy depends on the probability mass of the truncated region (`tp`):
+- If `tp > 0.25`, rejection sampling is used. This is efficient when the truncated region covers a large portion of the original distribution.
+- If `sqrt(eps) < tp <= 0.25`, inverse transform sampling is used. This is more efficient for smaller truncated regions.
+- If `tp` is very small (`<= sqrt(eps)`), a numerically stable version of inverse transform sampling is used which performs calculations in log-space to maintain precision.
+"""
```
devmotion: These are internal implementation details that might change at any point in time and are not relevant to users IMO. There's no need to document the function as its behaviour is the same as `rand(rng, ::Distribution)` for arbitrary distributions.

devmotion:
```suggestion
```

src/truncate.jl
```diff
@@ -233,6 +244,48 @@ function rand(rng::AbstractRNG, d::Truncated)
...
+
+Generate `n` random samples from a truncated distribution.
+
+The implementation samples the untruncated distribution, `d0` with `rand(rng, d0, n)` in batches and only keeps the samples that fall within the truncated range. The size of the batches is adaptively estimated to reduce the number of iterations.
+
+See [rand(rng::AbstractRNG, d::Truncated)](@ref) that handles the case of small mass of the truncated region.
+
+!!! warning
+    This method can be inefficient if the probability mass of the truncated region is very small. 
+"""
```
devmotion: Same here:
```suggestion
```

src/truncate.jl
```diff
@@ -233,6 +244,48 @@ function rand(rng::AbstractRNG, d::Truncated)
...
+function rand(rng::AbstractRNG, d::Truncated, n::Int)
+    n == 0 && return eltype(d)[]
```
devmotion: The use of `eltype` is again problematic IMO.

src/truncate.jl
```diff
@@ -233,6 +244,48 @@ function rand(rng::AbstractRNG, d::Truncated)
...
+    samples = Vector{eltype(d)}(undef, n)
+    n_collected = 0
+    while n_collected < n
+        n_remaining = n - n_collected
+        # Estimate number of samples to draw from the untruncated distribution.
+        # We draw more to reduce the chance of needing more rounds.
+        n_expected = n_remaining / tp
+        δn_expected = sqrt(n_remaining * tp * (1 - tp))  # standard deviation of the expected number
+        n_batch = ceil(Int, n_expected + 3δn_expected)
+        samples_d0 = rand(rng, d0, n_batch)
```
devmotion: This should be done in-place to reduce allocations.

src/truncate.jl
mmikhasenko: I agree that a particular dispatch of `rand` does not need to have signature description about the call that is anyway in the general method. But it is public function exposed to users, any details on its internal working I consider useful and worth documenting.

Here is an example how documentation is shown in Pluto:

for `cdf` function, first comes general description, then, specific methods

image

So, I suggest keeping this docstring, it does not need to be a part of this PR, if you do not support.

src/truncate.jl
mmikhasenko: Also suggest to keep. It's useful and user can check it.

src/truncate.jl
devmotion: IMO this example reveals a bug that would be fixed by https://github.com/JuliaStats/Distributions.jl/pull/1986.

src/mixtures/mixturemodel.jl
mmikhasenko: Replaced to `partype(d)`

src/mixtures/mixturemodel.jl
mmikhasenko: Sorry, I do not see why. All assumptions should be equally applicable for small n.

src/truncate.jl
mmikhasenko: replaced with `partype`

src/truncate.jl
mmikhasenko: updated to in-place implementation

mmikhasenko:
The last item that I think about is the 10M cap for switching between algorithm.
It's bad to switch between algorithms in general for predictability, so the design of Truncated with a switch is not ideal.
But we will not address it in this MR. 

I'm thinking to proceed with a similar if/else of 0.25% of distribution. Similar (not-ideal) patterns will be easy to identify in feature and update.

src/mixtures/mixturemodel.jl
devmotion: Because it creates unnecessary allocations and `rand` calls for small `n` which tend to slow down functions.

src/mixtures/mixturemodel.jl
```diff
@@ -477,6 +477,22 @@ rand(rng::AbstractRNG, s::MixtureSampler{Univariate}) =
...
+    counts = rand(rng, Multinomial(n, probs(d.prior)))
+    x = Vector{partype(d)}(undef, n)
```
devmotion: `partype` is not correct in general either. The only safe way is to perform sampling.

mmikhasenko:
Re-evaluated benchmarks in the header,
the rand-n version is faster for both models

mmikhasenko:
Comments are resolved, this PR is ready for review

src/mixtures/mixturemodel.jl
```diff
@@ -477,6 +477,34 @@ rand(rng::AbstractRNG, s::MixtureSampler{Univariate}) =
 rand(rng::AbstractRNG, d::MixtureModel{Univariate}) =
     rand(rng, component(d, rand(rng, d.prior)))
 
+function rand(rng::AbstractRNG, d::MixtureModel{Univariate}, n::Int)
```
devmotion, line 480: This is the wrong dispatch, isn't it? If one wants to draw multiple samples from a distribution `d`, automatically Distributions dispatches to drawing samples with `sampler(d)`. In the case of mixture models this is `MixtureSampler`. So this should actually be defined for `MixtureSampler{Univariate}`, not `MixtureModel{Univariate}` AFAICT?

However, generally for univariate distributions one also shouldn't define `rand(rng, dist, n)` but only the in-place method `_rand!(rng, sampler(dist), out)` (https://github.com/JuliaStats/Distributions.jl/pull/1905 will fix possibly incorrectly allocated output arrays) if multiple samples can be generated more efficiently: https://github.com/JuliaStats/Distributions.jl/blob/f1ff9e897d9a9c9d80aeff030266c5f6341de757/src/genericrand.jl#L35 https://github.com/JuliaStats/Distributions.jl/blob/f1ff9e897d9a9c9d80aeff030266c5f6341de757/src/univariates.jl#L141-L150

So AFAICT we should only define

```suggestion
function _rand!(rng::AbstractRNG, d::MixtureSampler{Univariate}, x::AbstractArray{<:Real})
```

Alternatively, if we never want to use `MixtureSampler` for sampling `MixtureModel{Univariate}`, we should define `sampler(d::MixtureModel{Univariate}) = d` (or limit the definition below to `sampler(d::MixtureModel{Multivariate}) = MixtureSampler(d)`) and here define

```suggestion
function _rand!(rng::AbstractRNG, d::MixtureModel{Univariate}, x::AbstractArray{<:Real})
```

src/mixtures/mixturemodel.jl
```diff
@@ -477,6 +477,34 @@ rand(rng::AbstractRNG, s::MixtureSampler{Univariate}) =
...
+        if i != max_count_idx
+            ni = counts[i]
+            if ni > 0
+                c = component(d, i)
+                last_offset = offset + ni - 1
+                rand!(rng, c, @view(x[(begin+offset):(begin+last_offset)]))
+                offset = last_offset + 1
+            end
+        end
+    end
```
devmotion, line 504: For the in-place method, it seems this could be simplified to

```suggestion
    offset = 0
    for (c, ni) in zip(components(d), counts)
        last_offset = offset + ni - 1
        rand!(rng, c, @view(x[(begin+offset):(begin+last_offset)]))
        offset = last_offset + 1
    end
```

src/truncate.jl
```diff
@@ -233,6 +233,64 @@ function rand(rng::AbstractRNG, d::Truncated)
     end
 end
 
+function rand(rng::AbstractRNG, d::Truncated, n::Int)
```
devmotion, line 236:
```suggestion
function _rand!(rng::AbstractRNG, d::Truncated, x::AbstractArray{<:Real})
```

And if there's any precomputations that could be factored out (doesn't seem to be the case?), then we should think about defining a dedicated sampler.

src/truncate.jl
```diff
@@ -233,6 +233,64 @@ function rand(rng::AbstractRNG, d::Truncated)
...
+    n == 0 && return rand(d.untruncated, 0)
+
```
devmotion: I don't think this case is common enough to warrant a special case, I think it would be better to reduce code here.

```suggestion
```

src/truncate.jl
```diff
@@ -233,6 +233,64 @@ function rand(rng::AbstractRNG, d::Truncated)
...
+    d0 = d.untruncated
+    tp = d.tp
+    lower = d.lower
+    upper = d.upper
+
+    # Use the same three regimes as the scalar version
+    if tp > 0.25
+        # Regime 1: Rejection sampling with batch optimization
+        # Get the correct type and memory by sampling from the untruncated distribution
+        samples = rand(rng, d0, n)
```
devmotion, line 246:
```suggestion
        rand!(rng, d0, x)
```

src/truncate.jl
```diff
@@ -233,6 +233,64 @@ function rand(rng::AbstractRNG, d::Truncated)
...
+        n_collected = 0
+        max_batch = 0
+        batch_buffer = Vector{eltype(samples)}()
```
devmotion, line 249: A separate batch buffer seems unnecessary, in particular the resizing might be inefficient? Instead of copying from a separate vector we could just use the output vector and move samples around there and keep track of the last accepted index.

src/truncate.jl
```diff
@@ -233,6 +233,64 @@ function rand(rng::AbstractRNG, d::Truncated)
...
+                    n_collected == n && break
+                end
+            end
+        end
+        return samples
+    elseif tp > sqrt(eps(typeof(float(tp))))
+        # Regime 2: Quantile-based sampling
+        # Sample one value first to determine the correct type
+        sample_type = typeof(quantile(d0, d.lcdf + rand(rng) * d.tp))
+        samples = Vector{sample_type}(undef, n)
```
devmotion, line 275:
```suggestion
```

src/truncate.jl
```diff
@@ -233,6 +233,64 @@ function rand(rng::AbstractRNG, d::Truncated)
...
+        for i in 1:n
+            samples[i] = quantile(d0, d.lcdf + rand(rng) * d.tp)
```
devmotion, line 277: We should probably at least use a `Random.Sampler` for `rand(rng)`, maybe it's even faster to call `rand(rng, n)` (despite the allocation.

src/truncate.jl
```diff
@@ -233,6 +233,64 @@ function rand(rng::AbstractRNG, d::Truncated)
...
+        end
+        return samples
+    else
+        # Regime 3: Log-space computation
+        # Sample one value first to determine the correct type
+        sample_type = typeof(invlogcdf(d0, logaddexp(d.loglcdf, d.logtp - randexp(rng))))
+        samples = Vector{sample_type}(undef, n)
```
devmotion, line 284:
```suggestion
```

PR #1987, opened 2025-06-23
formatter with JuliaFormatter

### LINKED ISSUE #1401: Add JuliaFormatter to workflow
> yeah we should add JuliaFormatter to this. It would be good to merge as many PRs as possible before because it will create conflicts all over

_Originally posted by @matbesancon in https://github.com/JuliaStats/Distributions.jl/issues/1283#issuecomment-934149214_

NB- the following workflow should only have a minimal amount of pain involved in each remaining open PR, assuming `format(".")` on master is committed as a single commit tagged `format_commit`:
1. `git checkout feature_branch`
2. `git merge master format_commit^` (no conflicts, or actual content conflicts that must be resolved should be handled here)
3. `julia -e "using JuliaFormatter; format("."); git commit am "PR format"`
4. `git merge master format_commit; git checkout --ours .; git commit -a` (we know the only diff is the formatting, so it's safe to blanket-accept the local branch side)
5. `git merge master` (should apply cleanly now)

mmikhasenko:
Formatting cleanup using JuliaFormatter.format(".") to ensure consistent code style across the package. This reduces noisy diffs in future changes and improves readability. No functional changes.

Closes #1401 

## Modifications
- [x] add `.formating` with scripts
- [x] modified README.md. Added formatting rules
- [x] added `.git-blame-ignore-revs` to 
- [x] applied formatting

## Tests
- [x] are passing

## Once merged: 

Update  .git-blame-ignore-revs to point to the merge commit


mmikhasenko:
> I'd actually prefer Runic, it seems to work fine in StatsFuns and PDMats and I've been bitten by JuliaFormatter 2 problems a few times. GLM uses JuliaFormatter though, so there is no agreed on standard in JuliaStats yet.

completely fine with me, I do not have preference

mmikhasenko:
The work is complete. Tests are passing.
Could you please review?

.formatting/Project.toml
devmotion, line 1: Please remove this file and all other files in this folder. It has to be maintained and there is no reason for users to do anything else than following the Runic documentation.

.git-blame-ignore-revs
devmotion, line 1: This won't work - this is referring to the commit in your fork but we have to ignore the commit that might finally land in the master branch. We have to add this file only once the formatting commit is in the master branch and its hash is known.

.github/workflows/CI.yml
```diff
@@ -5,7 +5,7 @@ on:
   push:
     branches:
       - master
-    tags: '*'
+    tags: "*"
```
devmotion: No need to change this?
```suggestion
    tags: '*'
```

.github/workflows/CI.yml
```diff
@@ -16,15 +16,30 @@ concurrency:
...
+      - uses: julia-actions/setup-julia@v2
+        with:
+          version: "1"
+          show-versioninfo: true
+      - uses: julia-actions/cache@v2
+      - run: |
+          julia --project=.formatting -e '
+            using Pkg
+            Pkg.instantiate()
+            include(".formatting/format_check.jl")'
```
devmotion: We should keep it simple and just follow the recommended and documented workflow in the Runic docs:
```suggestion
  runic:
    name: Runic formatting
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: julia-actions/setup-julia@v2
        with:
          version: '1'
      - uses: julia-actions/cache@v2
      - uses: fredrikekre/runic-action@v1
        with:
          version: '1'
```
Possibly it could be made a separate workflow as in e.g. PDMats (https://github.com/JuliaStats/PDMats.jl/blob/master/.github/workflows/Format.yml) but that doesn't matter too much IMO.

.github/workflows/CI.yml
```diff
@@ -16,15 +16,30 @@ concurrency:
...
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         version:
-          - 'min'
-          - '1'
+          - "min"
+          - "1"
```
devmotion: No need to change these lines:
```suggestion
          - 'min'
          - '1'
```

.github/workflows/CI.yml
```diff
@@ -57,7 +72,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@v2
         with:
-          version: '1'
+          version: "1"
```
devmotion:
```suggestion
          version: '1'
```

README.md
```diff
@@ -55,6 +56,20 @@ clone it and make modifications on a new branch,
...
+
+**Note:** Code formatting is automatically checked in CI using Runic.
+The formatting command can be run locally with 
+```julia
+julia --project=.formatting -e 'using Pkg; Pkg.instantiate(); include(".formatting/format_check.jl")'
+```
+The `.git-blame-ignore-revs` file contains commit hashes for mass formatting changes.
+This allows `git blame` to show the actual authors of code changes rather than the formatting commit.
+When viewing blame information, use `git blame --ignore-revs-file .git-blame-ignore-revs <filename>`.
+
```
devmotion: If we want to document it more prominently (not sure about it, users don't have to care and everyone who submits a PR will notice and learn it anyway), we should just refer to the Runic docs.
```suggestion
```

ararslan (changes requested):
> I've been bitten by JuliaFormatter 2 problems a few times

You can pin to v1, which the JuliaFormatter README recommends if you run into issues with v2. My employer uses JuliaFormatter v2 with YASGuide extensively and have generally only run into issues in very unusual code, in which case you can surround in `#!format off`/`on`. MixedModels uses JuliaFormatter with BlueStyle on CI and that works fine.

The formatting implemented by Runic is largely inconsistent with the prevailing, albeit historically ad hoc, formatting in JuliaStats. I also find that some of its choices actively hinder readability vs. YAS or Blue.

I'd be fine with this PR if we use JuliaFormatter as initially proposed with one of the aforementioned styles, but I do not agree with using Runic.

.git-blame-ignore-revs
mmikhasenko: ok, it should be updated, once merged.
Adding to the header, that it needs to be done

devmotion:
> Switched back to JuliaFormatting

I'd still prefer Runic. PDMats, StatsFuns and [possibly soon StatsBase](https://github.com/JuliaStats/StatsBase.jl/pull/990) use Runic. Outside of StatsBase, it is used quite a lot as well - the whole SciML ecosystem recently switched to Runic, Documenter uses Runic, DifferentiationInterface uses Runic.... Generally, there seems to be a significant movement towards Runic

mmikhasenko:
I have an idea -- you can create a competing PR, and see which one gets merged first.
I can also do it, if you like.

Would be good to close the story, that it does not hang in my open pull requests

rsenne:
> significant movement towards Runic

My unsolicited two cents is that some of the movement toward Runic may be because JuliaFormatter v2 [is not considered stable yet](https://github.com/JuliaEditorSupport/JuliaFormatter.jl/blob/master/README.md?plain=1#L7-L12), which the package maintainers have acknowledged and is part of why they do not recommend using v2 for testing purposes yet.

Runic is great, but pinning JuliaFormatter to v1.0.62 removes the burden of formatter behavior changing between minor releases. This used to be a bigger issue as this caused conflicts with JET but JuliaSyntax v1 is [supported on JET](https://github.com/aviatesk/JET.jl/blob/master/Project.toml#L34). JuliaFormatter also seems to be more steadily maintained now, so I don’t think people necessarily need to give up on it yet. That said, if people simply prefer Runic’s style, that is totally reasonable too.

Edit: I missed the earlier comment making the same point. Sorry about that. Leaving this up since it adds a bit on pinning JuliaFormatter and the JET/JuliaSyntax compatibility issue.

devmotion:
There's no consensus but also no pressing need for a change, so this PR or a rebased version of it is not expected to be merge in the near future. Therefor I'm closing the PR as it is outdated anyways.

PR #1991, opened 2025-07-18
Improve performance of `logpdf` for `DiagNormal` and `IsoNormal`

### LINKED ISSUE #1989: DiagNormal is substantially slower than Product of Normals
https://github.com/JuliaStats/Distributions.jl/blob/abb151c42eb67b4994e21da15e24c65f1cc3dd9f/src/product.jl#L239-L250

`product_distribution` on an array of Normals creates a DiagNormal. That's actually quite satisfying to look at, but the original Product is quite substantially faster:

```julia

julia> dists = [Normal(0.0), Normal(1.0), Normal(2.0, 3.0), Normal(4.0, 0.2)];

julia> prod = Product(dists);

julia> mvnorm = product_distribution(dists);
```

## `rand`

julia> @be rand($prod)
Benchmark: 3556 samples with 805 evaluations
 min    23.034 ns (2 allocs: 96 bytes)
 median 24.845 ns (2 allocs: 96 bytes)
 mean   33.334 ns (2 allocs: 96 bytes, 0.25% gc time)
 max    5.135 μs (2 allocs: 96 bytes, 98.75% gc time)

julia> @be rand($mvnorm)
Benchmark: 8044 samples with 260 evaluations
 min    32.212 ns (2 allocs: 96 bytes)
 median 34.935 ns (2 allocs: 96 bytes)
 mean   42.677 ns (2 allocs: 96 bytes, 0.07% gc time)
 max    13.362 μs (2 allocs: 96 bytes, 98.83% gc time)
```

## `logpdf`

```julia
julia> x = rand(prod)
4-element Vector{Float64}:
  1.7947570646417756
  1.050115051070052
 -1.0885138050761758
  4.265595405348341

julia> @be logpdf($prod, $x)
Benchmark: 3290 samples with 1220 evaluations
 min    23.498 ns
 median 23.599 ns
 mean   23.693 ns
 max    44.125 ns

julia> @be logpdf($mvnorm, $x)
Benchmark: 2827 samples with 464 evaluations
 min    61.332 ns (2 allocs: 96 bytes)
 median 62.409 ns (2 allocs: 96 bytes)
 mean   71.665 ns (2 allocs: 96 bytes, 0.14% gc time)
 max    9.668 μs (2 allocs: 96 bytes, 98.51% gc time)
```

Similar performance differences are observed when taking the gradient of logpdf.

Perhaps there are some performance improvements that can be done for `DiagNormal`?

penelopeysm:
This PR (sort of) closes #1989. I didn't attempt to make `rand()` faster because the performance drop there was smaller, but the changes here improve the performance of `logpdf` on `DiagNormal` and `IsoNormal` by about 2–3x.

The offending code that made the previous `logpdf` slower was the calculation of `x .- d.µ`:

https://github.com/JuliaStats/Distributions.jl/blob/abb151c42eb67b4994e21da15e24c65f1cc3dd9f/src/multivariate/mvnormal.jl#L262

The new methods avoid this and are thus both faster and avoid allocations.

### DIFF
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -261,6 +261,16 @@ logdetcov(d::MvNormal) = logdet(d.Σ)
 
 sqmahal(d::MvNormal, x::AbstractVector) = invquad(d.Σ, x .- d.μ)
 
+function sqmahal(d::DiagNormal, x::AbstractVector)
+    # Faster than above as this avoids calculating (x .- d.µ)
+    T = promote_type(partype(d), eltype(x))
+    sum = zero(T)
+    for i in eachindex(x)
+        @inbounds sum += abs2(x[i] - d.μ[i]) / d.Σ[i, i]
+    end
+    return sum
+end
+
 sqmahal!(r::AbstractVector, d::MvNormal, x::AbstractMatrix) =
     invquad!(r, d.Σ, x .- d.μ)
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.25.120"
+version = "0.25.121"
 
 [deps]
 AliasTables = "66dad0bd-aa9a-41b7-9441-69ab47430ed8"


src/multivariate/mvnormal.jl
```diff
@@ -261,6 +261,25 @@ logdetcov(d::MvNormal) = logdet(d.Σ)
 
 sqmahal(d::MvNormal, x::AbstractVector) = invquad(d.Σ, x .- d.μ)
 
+function sqmahal(d::DiagNormal, x::AbstractVector)
+    # Faster than above as this avoids calculating (x .- d.µ)
+    T = promote_type(partype(d), eltype(x))
```
devmotion, line 266: This is not correct eg if the parameters and `x` are arrays of integers. One could either compute the first element outside of the loop or possibly use a functional approach (e.g. with `mapreduce` and `Base.Broadcast.broadcasted(...)`).

src/multivariate/mvnormal.jl
```diff
@@ -261,6 +261,25 @@ logdetcov(d::MvNormal) = logdet(d.Σ)
...
+    sum = zero(T)
+    for i in eachindex(x)
+        @inbounds sum += abs2(x[i] - d.μ[i]) / d.Σ[i, i]
+    end
+    return sum
+end
+
+function sqmahal(d::IsoNormal, x::AbstractVector)
+    T = promote_type(partype(d), eltype(x))
```
devmotion, line 275: Very artificial, but this is also not guaranteed to be correct e.g. for `Bool`ean values.

src/multivariate/mvnormal.jl
```diff
@@ -261,6 +261,25 @@ logdetcov(d::MvNormal) = logdet(d.Σ)
 
 sqmahal(d::MvNormal, x::AbstractVector) = invquad(d.Σ, x .- d.μ)
 
+function sqmahal(d::DiagNormal, x::AbstractVector)
+    # Faster than above as this avoids calculating (x .- d.µ)
+    T = promote_type(partype(d), eltype(x))
+    sum = zero(T)
+    for i in eachindex(x)
+        @inbounds sum += abs2(x[i] - d.μ[i]) / d.Σ[i, i]
```
devmotion, line 269: Indices might not be compatible.

src/multivariate/mvnormal.jl
```diff
@@ -261,6 +261,25 @@ logdetcov(d::MvNormal) = logdet(d.Σ)
...
+    end
+    return sum
+end
+
+function sqmahal(d::IsoNormal, x::AbstractVector)
+    T = promote_type(partype(d), eltype(x))
+    sum = zero(T)
+    for i in eachindex(x)
+        @inbounds sum += abs2(x[i] - d.μ[i])
```
devmotion, line 278: Indices might not be compatible

src/multivariate/mvnormal.jl
```diff
@@ -261,6 +261,25 @@ logdetcov(d::MvNormal) = logdet(d.Σ)
...
+    end
+    return sum / d.Σ[1, 1]
```
devmotion, line 280: Might not use one-based indexing.

src/multivariate/mvnormal.jl
devmotion: The functional approach probably also works better with GPU arrays.

PR #2002, opened 2025-09-30
Support zero variance in `LogitNormal`

### LINKED ISSUE #2001: `LogitNormal` does not support zero variance
Currently, `LogitNormal` does not allow for zero variance:

```julia
julia> LogitNormal(1.0, 0.0)
ERROR: DomainError with 0.0:
LogitNormal: the condition σ > zero(σ) is not satisfied.
Stacktrace:
 [1] #390
   @ ~/.julia/packages/Distributions/YQSrn/src/univariate/continuous/logitnormal.jl:62 [inlined]
 [2] check_args
   @ ~/.julia/packages/Distributions/YQSrn/src/utils.jl:89 [inlined]
 [3] #LogitNormal#389
   @ ~/.julia/packages/Distributions/YQSrn/src/univariate/continuous/logitnormal.jl:62 [inlined]
 [4] LogitNormal(μ::Float64, σ::Float64)
   @ Distributions ~/.julia/packages/Distributions/YQSrn/src/univariate/continuous/logitnormal.jl:61
 [5] top-level scope
   @ REPL[17]:1
```

whereas `Normal` and `LogNormal` do:

```julia
julia> Normal(1.0, 0.0)
Normal{Float64}(μ=1.0, σ=0.0)

julia> LogNormal(1.0, 0.0)
LogNormal{Float64}(μ=1.0, σ=0.0)
```

For transformed Gaussians, it should be easy to avoid issues such as #1880 since `Normal` already officially supports the zero variance case.

devmotion:
Fixes #2001.

Both implementation and tests mimic the implementation and tests for `LogNormal`.

### DIFF
--- a/src/univariate/continuous/logitnormal.jl
+++ b/src/univariate/continuous/logitnormal.jl
@@ -59,7 +59,7 @@ struct LogitNormal{T<:Real} <: ContinuousUnivariateDistribution
 end
 
 function LogitNormal(μ::T, σ::T; check_args::Bool=true) where {T <: Real}
-    @check_args LogitNormal (σ, σ > zero(σ))
+    @check_args LogitNormal (σ, σ >= zero(σ))
     return LogitNormal{T}(μ, σ)
 end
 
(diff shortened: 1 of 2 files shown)


andreasnoack (approved):
I agree that this is a natural consequence of supporting this case for `Normal`, which has proven to be quite convenient.

PR #2003, opened 2025-09-30
Skip parameter checks when constructing `Normal` distributions for `LogNormal` calculations

devmotion:
Quite a few methods for `LogNormal` call and post-process methods for `Normal` distributions. However, currently the outer constructor for `Normal` distributions is used and hence parameter values (mean and standard deviation) are re-checked (to ensure that the standard deviation is non-negative) in every call.

This PR switches to the ~~inner constructor~~ StatsFuns API (see discussion below) to skip these parameter checks. The difference can be seen e.g. in the output of `@code_llvm` and also be observed when benchmarking e.g. `logpdf`:

### On the master branch

```julia

julia> @be rand(1000) map(x -> logpdf(LogNormal(3.1, 2.5), x), _)
Benchmark: 6251 samples with 1 evaluation
 min    11.750 μs (3 allocs: 8.062 KiB)
 median 12.416 μs (3 allocs: 8.062 KiB)
 mean   13.721 μs (3 allocs: 8.062 KiB, 0.15% gc time)
 max    2.115 ms (3 allocs: 8.062 KiB, 98.68% gc time)
```

## With this PR

```julia

julia> @be rand(1000) map(x -> logpdf(LogNormal(3.1, 2.5), x), _)
Benchmark: 4180 samples with 3 evaluations
 min    6.000 μs (3 allocs: 8.062 KiB)
 median 6.333 μs (3 allocs: 8.062 KiB)
 mean   7.213 μs (3 allocs: 8.062 KiB, 0.31% gc time)
 max    1.005 ms (3 allocs: 8.062 KiB, 98.62% gc time)
```

### DIFF
--- a/src/univariate/continuous/lognormal.jl
+++ b/src/univariate/continuous/lognormal.jl
@@ -133,7 +133,7 @@ cquantile(d::LogNormal, q::Real) =  exp(StatsFuns.norminvccdf(d.μ, d.σ, q))
 invlogcdf(d::LogNormal, lq::Real) = exp(StatsFuns.norminvlogcdf(d.μ, d.σ, lq))
 invlogccdf(d::LogNormal, lq::Real) = exp(StatsFuns.norminvlogccdf(d.μ, d.σ, lq))
 
-function gradlogpdf(d::LogNormal{T}, x::Real) where {T<:Real}
+function gradlogpdf(d::LogNormal, x::Real)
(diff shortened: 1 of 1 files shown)


andreasnoack:
I kind of like reusing the Distributions API here, but you are right that using the StatsFuns API would be self-explanatory. So either the former with small comments or the latter.

andreasnoack (approved):

PR #2005, opened 2025-10-03
Remove all `@inbounds` annotations

### LINKED ISSUE #1265: Incorrect uses of @inbounds cause miscalculation of statistics

timholy:
Julia is quite good at eliminating bounds checks on its own, and incorrect use of `@inbounds` can lead to hard-to-diagnose bugs. There are also cases where adding it actively hurts performance (e.g., https://github.com/JuliaLang/julia/issues/48245). It's much better to allow the sophisticated analysis performed by inference to make this decision and it is to reflexively add `@inbounds` everywhere.

Fixes #1265.


devmotion:
This is maybe a bit extreme to fix #1265 (which arguably is not an actual problem for most users and could be address by ensuring one-based indices when not operating with generic indices).

AFAIK the benchmarks in this package haven't been updated in a long time, and also seem to cover only a few samplers and a few evaluations for mixture models. All other distributions are not covered by the benchmarks at all.

So IMO,
1. If a specific `@inbounds` can be shown to not improve performance, we could remove it
2. If an `@inbounds` is unsafe but actually improves performance, we should guard it with a `requires_one_based_indexing` (quick fix) or rewrite the expressions with generic indices

In all other cases (safe `@inbounds` that improve performance), I don't see a strong reason to remove them.

Currently, my impression is that this PR removes `@inbounds` statements without checking point 1.

devmotion:
OK, I'm convinced 🙂 I just didn't trust the reasoning based on benchmarking in https://github.com/JuliaStats/Distributions.jl/issues/1265#issuecomment-3356050297 as it only covers a tiny subset of the package (which makes the performance checks a bit questionable...).

If we/users see regressions, we can always add `@inbounds` back.

devmotion (approved):

PR #2007, opened 2025-10-18
Fix R test references and make them reproducible

devmotion:
When working on adding a new distribution, I realized (once again) that the R test references can't be re-generated currently since running `Rscript gendref.R` errors:

- The `LindleyR` package is not available anymore
- For `distr = StudentizedRange(...)` since the function `distr$pdf` is undefined
- `truncated(Normal(...), ...)` is not a valid reference name as it can't be mapped to a reference distribution

Moreover, after commenting out these three reference distribution types rerunning `Rscript gendref.R` changes reference values for most distributions.

---

This PR tries to fix the setup of the R reference tests:
- Use `renv` for a reproducible package environment
- Add a GHA that updates the R environment and reference values (scheduled and manually)
- Replace `LindleyR` with `new.dist`
- Use explicit `package::function` syntax instead of relying on `library(package)` in other source files
- Calculate (log)pdf reference values of `StudentizedRange` by finite differencing of (log)cdf
- Allow manual specification of `distr$dtype` and `distr$expr` (to support `truncated(...)` properly)
- Simplify matrix-variate data generation by using `StanHeaders` instead of `rstan`
- Improve definition of reference distributions for alternative Julia distribution types (e.g. `Erlang` and `NormalCanon`)
- Update test/ref/readme.md
- Update reference values by running `Rscript gendref.R`

### DIFF
--- a/test/univariate/continuous/cosine.jl
+++ b/test/univariate/continuous/cosine.jl
@@ -2,7 +2,7 @@ using Distributions, Test
 
 @testset "Cosine" begin
     @testset "values outside of the support" begin
-        for (d, xmin, xmax) in ((Cosine(), -1.1, 1.1), (Cosine(1.5), (-1.0, 3.0)), (Cosine(2.0, 2.8), -0.9, 5.0))
+        for (d, xmin, xmax) in ((Cosine(), -1.1, 1.1), (Cosine(1.5), -1.0, 3.0), (Cosine(2.0, 2.8), -0.9, 5.0))
             for x in (xmin, xmax)
                 @test iszero(@inferred(pdf(d, x)))
                 @test @inferred(logpdf(d, x)) == -Inf
--- a/test/univariate/continuous/fdist.jl
+++ b/test/univariate/continuous/fdist.jl
@@ -1,6 +1,6 @@
 using Distributions, Test
 
-@test "FDist" begin
+@testset "FDist" begin
     @testset "kurtosis" begin
         d = FDist(30, 40)
         @test @inferred(kurtosis(d)) ≈ 2.2442906574394463
--- a/test/univariate/continuous/studentizedrange.jl
+++ b/test/univariate/continuous/studentizedrange.jl
@@ -2,17 +2,7 @@ using Distributions, Test
 
 @testset "StudentizedRange" begin
     @testset "values at 0" begin
-        d = StudentizedRange(2, 2)
-        @test @inferred(pdf(d, 0)) ≈ 0.5
-        @test @inferred(logpdf(d, 0)) ≈ -log(2)
-        @test iszero(@inferred(cdf(d, 0)))
-        @test @inferred(logcdf(d, 0)) == -Inf
-        @test isone(@inferred(ccdf(d, 0)))
-        @test iszero(@inferred(logccdf(d, 0)))
-
-        for d in (StudentizedRange(10, 5), StudentizedRange(5, 10))
-            @test iszero(@inferred(pdf(d, 0)))
-            @test @inferred(logpdf(d, 0)) == -Inf
+        for d in (StudentizedRange(2, 2), StudentizedRange(10, 5), StudentizedRange(5, 10))
             @test iszero(@inferred(cdf(d, 0)))
             @test @inferred(logcdf(d, 0)) == -Inf
             @test isone(@inferred(ccdf(d, 0)))


test/ref/matrixvariates/jsonfiles/InverseWishart_stan_output.json
```diff
@@ -5,25 +5,25 @@
       [4.5],
       [1, 0.5, 0.5, 0.5, 1, 0.5, 0.5, 0.5, 1]
     ],
-    "X": [1.80115497975516, 1.39357092294611, 0.679396397136899, 1.39357092294611, 1.32682592836847, 0.624106043062951, 0.679396397136899, 0.624106043062951, 0.431228950208813],
-    "lpdf": [-0.598439090738466]
+    "X": [0.48516264401941139, 0.21974036159989954, -0.13995828882732522, 0.21974036159989954, 0.28855250002048116, -0.23735060033573027, -0.13995828882732522, -0.23735060033573027, 0.61858504476347143],
+    "lpdf": [-0.092239036945123282]
```
nalimilan, line 9: So this was incorrect before?

test/ref/continuous_test.ref.json
```diff
@@ -339,16 +336,15 @@
     "skewness": "nan"
   },
   "points": [
-    { "x": -1.2, "pdf": 0, "logpdf": "-inf", "cdf": 0 },
```
nalimilan, line 342: Any idea why this line and others disappear?

.github/workflows/UpdateRefs.yml
```diff
@@ -0,0 +1,48 @@
+name: UpdateRefs
+
+on:
+  schedule:
+    - cron: '00 00 * * *'
```
nalimilan: Running this once per month or week would probably be enough? It hasn't been run for years. ;-)

BTW, if it fails, will we be notified (e.g. with an issue)? Otherwise it could fail silently for a long time.

devmotion: Once per week maybe?

```suggestion
    - cron: '00 00 * * 00'
```

> BTW, if it fails, will we be notified (e.g. with an issue)? Otherwise it could fail silently for a long time.

If it fails, it will open a PR (or at least that's what it's supposed to do 😄)

test/ref/matrixvariates/jsonfiles/InverseWishart_stan_output.json
devmotion: No, there's nothing wrong here. I just improved the generation of the matrix distribution samples and their logpdf (sampled and computed with the stan math library), which yielded different samples.

The improved script is test/ref/matrixvariates/get_stan_output.R which also directly uses `StanHeaders` instead of the previous way of defining and compiling more complex stan code.

.github/workflows/UpdateRefs.yml
nalimilan: As you prefer!

test/ref/continuous_test.ref.json
devmotion: Because they've never been generated with the existing R code but were added manually. In this case in https://github.com/JuliaStats/Distributions.jl/commit/a389ef95eb5250acf1673e931c90670b52059c1d

nalimilan: Doesn't that mean that we no longer test that negative values give 0 or minus Inf? Or is that tested elsewhere already?

devmotion: codecov says this branch is still tested but that this seems misleading since it was changed in the meantime to a single line: https://app.codecov.io/gh/JuliaStats/Distributions.jl/pull/2007/blob/src/univariate/continuous/betaprime.jl?dropdown=coverage#L94 It's also not covered by https://github.com/JuliaStats/Distributions.jl/blob/f1ff9e897d9a9c9d80aeff030266c5f6341de757/test/univariate_bounds.jl#L96-L101 unfortunately since the support of `BetaPrime` is not upper bounded

nalimilan: OK, so tests need to be added to replace lines that had been added manually.

devmotion: I added Julia tests for all special cases that were not actually part of the R test setup but added to the reference values manually.

nalimilan (approved):

PR #2008, opened 2025-10-22
Add log-logistic distribution

devmotion:
Updated and improved version of #1454.

Adding reference tests against R is a bit cumbersome currently, see #2007.
Edit: I added tests against R nevertheless by locally commenting out all other reference tests (and without environment, ie probably not reproducible).

Closes #1454.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -31,7 +31,7 @@ jobs:
           - macos-latest
           - windows-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - uses: julia-actions/setup-julia@v2
         with:
           version: ${{ matrix.version }}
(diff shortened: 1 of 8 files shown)


src/univariate/continuous/loglogistic.jl
```diff
@@ -0,0 +1,120 @@
...
+
+If ``X \\sim \\operatorname{LogLogistic}(\\alpha, \\beta)`` then ``\\log(X) \\sim \\operatorname{Logistic}(\\log(\\alpha), 1/\\beta)``.
+The probability density function is 
+
+```math
+f(x; \\alpha, \\beta) = \\frac{(\\beta / \\alpha){(x/\\alpha)}^{\\beta - 1}}{{(1 + {(x/\\alpha)}^\\beta)}^2}, \\qquad \\alpha > 0, \\beta > 0.
+```
+
+```julia
+LogLogistic(α, β)        # Log-logistic distribution with scale α and shape β
```
devmotion, line 14: I wonder if the order of parameters should be reversed to mimic e.g. `Gamma` and `Weibull` where the shape is the first and the scale parameter is the second argument. On the other hand, probably we don't have a completely consistent and clear convention currently - to solve such ambiguities it might be best to use keyword arguments (e.g., `LogLogistic(; shape::Real, scale::Real)`).

andreasnoack: Having shape and scale keywords would be wonderful. I look up the documentation every time I use `Gamma`. Then we could also have `mean`/`dispersion`, which in many cases would be even simpler to interpret. I think this was discussed in https://github.com/JuliaStats/Distributions.jl/pull/823

devmotion: I tried something in https://github.com/JuliaStats/Distributions.jl/pull/1405 but I didn't pursue it further since there was no response when I opened the PR. Probably, for most distributions exposing the somewhat arbitrary parameter names might also not be the best solution, maybe more meaningful names (such as `shape` and `scale`) might be better.

andreasnoack (approved):

PR #2012, opened 2025-11-17
Make MatrixNormal sampling non-allocating

projekter:
Currently, sampling from `MatrixNormal` always allocates multiple temporaries: one of the matrix `X` that contains the normally-sampled values, one for the matrix product `A * X`, and one for the matrix product `(A * X) * B`. However, given that `A` and `B` are triangular (and square) matrices, all products can be done in-place and the sample can directly be stored to the output. So the method doesn't need to allocate at all, provided the Cholesky factors are available without allocation.

### DIFF
--- a/test/matrixvariates.jl
+++ b/test/matrixvariates.jl
@@ -327,7 +327,13 @@ function test_special(dist::Type{MatrixNormal})
         M, U, V = _rand_params(MatrixNormal, Float64, 5, 5)
         noallocD = MatrixNormal(M, cholesky!(Symmetric(U, :L)), cholesky!(Symmetric(V, :U)))
         output = Matrix{Float64}(undef, size(noallocD))
-        @test iszero(((d, out) -> @allocated(rand!(d, out)))(noallocD, output))
+        allocs = ((d, out) -> @allocated(rand!(d, out)))(noallocD, output)
+        # See https://github.com/JuliaStats/Distributions.jl/pull/2012#issuecomment-3566807876
+        if VERSION < v"1.10.5"
+            @test allocs <= 32
+        else
+            @test iszero(allocs)
+        end
     end
     nothing
 end


projekter:
I added a test for this case. While I think that at least some, if not all, matrixvariate distributions can be made to sample without allocations, the tests would probably still need to be different for each one. For example, MatrixNormal will only avoid sampling if the `U` and `V` matrices already have the proper format - if the Cholesky factor is stored with the wrong triangle, it will copy upon retrieval, due to the way LinearAlgebra implements the retrieval of Cholesky factors. (Actually, this should probably be a motivation to bring it into the correct format upon creation, so that the sampling doesn't have to do it all over...)

test/matrixvariates.jl
```diff
@@ -322,6 +322,13 @@ function test_special(dist::Type{MatrixNormal})
             end
         end
     end
+    @testset "Non-allocating sampling" begin
+        # #2012: we can sample without allocations
+        M, U, V = _rand_params(MatrixNormal, Float64, 5, 5)
+        noallocD = MatrixNormal(M, cholesky!(Symmetric(U, :L)), cholesky!(Symmetric(V, :U)))
+        output = Matrix{Float64}(undef, size(noallocD))
+        @test iszero(@allocated(rand!(noallocD, output)))
```
devmotion: In my experience `@allocated` tests are a bit brittle if you don't put them behind a function barrier. Something like

```suggestion
        @test iszero(((d, out) -> @allocated(rand!(d, out))(noallocD, output)))
```
might be more stable.

projekter: Yes, that makes sense. Apparently, in the latest releases, they managed to resolve these issues to some degree, as the test passes without function barrier, but in 1.10 it's still an issue.

test/matrixvariates.jl
```diff
@@ -322,6 +322,13 @@ function test_special(dist::Type{MatrixNormal})
             end
         end
     end
+    @testset "Non-allocating sampling" begin
+        # #2012: we can sample without allocations
+        M, U, V = _rand_params(MatrixNormal, Float64, 5, 5)
+        noallocD = MatrixNormal(M, cholesky!(Symmetric(U, :L)), cholesky!(Symmetric(V, :U)))
+        output = Matrix{Float64}(undef, size(noallocD))
+        @test iszero(((d, out) -> @allocated(rand!(d, out))(noallocD, output)))
```
devmotion: Sorry, it seems I messed up the order of the closing brackets in my previous suggestion. I think the following should be correct:
```suggestion
        @test iszero(((d, out) -> @allocated(rand!(d, out)))(noallocD, output))
```

projekter: Indeed, I should've read more carefully.

test/matrixvariates.jl
```diff
@@ -322,6 +322,13 @@ function test_special(dist::Type{MatrixNormal})
             end
         end
     end
+    @testset "Non-allocating sampling" begin
+        # #2012: we can sample without allocations
+        M, U, V = _rand_params(MatrixNormal, Float64, 5, 5)
+        noallocD = MatrixNormal(M, cholesky!(Symmetric(U, :L)), cholesky!(Symmetric(V, :U)))
+        output = Matrix{Float64}(undef, size(noallocD))
+        @test iszero(((d, out) -> @allocated(rand!(d, out)))(noallocD, output))
```
devmotion: What about the following test:

```suggestion
        allocs = ((d, out) -> @allocated(rand!(d, out)))(noallocD, output)
        # See https://github.com/JuliaStats/Distributions.jl/pull/2012#issuecomment-3566807876
        if VERSION < v"1.10.5"
            @test allocs <= 32
        else
            @test iszero(allocs)
        end
```

projekter: Yes, that's also a possibility. Probably the only way to do this while keeping 1.10.x compatibility.

devmotion (approved):

devmotion:
Thank you!

PR #2017, opened 2025-12-03
Adding missing features of cosine distribution

jaksle:
This is a very straightforward PR. It adds missing `mgf`, `cgf`, `cf` and direct affine transformations for the `Cosine` distribution. There are simple formulas for these, so there is no reason not to include them.

One thing which might require attention is if I protect from dividing by 0 correctly.

### DIFF
--- a/test/univariate/continuous/cosine.jl
+++ b/test/univariate/continuous/cosine.jl
@@ -2,10 +2,11 @@
     @testset "mgf, cgf, cf" begin
         @test cf(Cosine(1, 1), 1) ≈ cis(1) * sin(1)/(1 - π^-2)
         @test mgf(Cosine(-1, 2), -1) ≈ exp(1) * sinh(-2) / (-2*(1 + 4π^-2))
-        @test cgf(Cosine(-1, 2), 2) ≈ -2 + log(sinh(4)) - log(4) - log(1 + 16π^-2)
+        test_cgf(Cosine(-1, 2), [-1,2, 4])
     end
 
     @testset "affine transformations" begin
+        test_affine_transformations(Cosine, (4, 2))
         @test -Cosine(1, 1) == Cosine(-1, 1)
         @test 3Cosine(1, 2) == Cosine(3, 6)
         @test Cosine(0, 2) + 42 == Cosine(42, 2)


src/univariate/continuous/cosine.jl
```diff
@@ -93,3 +93,27 @@ function ccdf(d::Cosine{T}, x::Real) where T<:Real
 end
 
 quantile(d::Cosine, p::Real) = quantile_bisect(d, p)
+
+function mgf(d::Cosine{T}, t::Real) where T<:Real
+    μ, σ = params(d)
```
devmotion: IMO that's a bit dangerous as it would break if order of parameters in `params` is swapped.

```suggestion
    (; μ, σ) = d
```

src/univariate/continuous/cosine.jl
```diff
@@ -93,3 +93,27 @@ function ccdf(d::Cosine{T}, x::Real) where T<:Real
 end
 
 quantile(d::Cosine, p::Real) = quantile_bisect(d, p)
+
+function mgf(d::Cosine{T}, t::Real) where T<:Real
```
devmotion: The type parameter is not needed (see my comment below):
```suggestion
function mgf(d::Cosine, t::Real)
```

src/univariate/continuous/cosine.jl
```diff
@@ -93,3 +93,27 @@ function ccdf(d::Cosine{T}, x::Real) where T<:Real
...
+    μ, σ = params(d)
+    t ≈ 0. && return exp(μ*t)
+    return T(π^2) * exp(μ*t) * sinh(σ*t) / (σ*t*(T(π^2) + (σ*t)^2))
```
devmotion: AFAICT the problem is not so much `t` in isolation but computing `sinh(σ*t) / (σ*t)` for small values of `σ*t`. Moreover, it seems a bit easier to cancel `π^2` and use the `IrrationalConstants.invπ` constant here. 

So I'd suggest something like

```suggestion
    σt = σ * t
    z = iszero(σt) ? one(float(σt)) : sinh(σ*t)/(σ*t)
    return exp(μ*t) * (z / (1 + (IrrationalConstants.invπ * σt)^2))
```

src/univariate/continuous/cosine.jl
```diff
@@ -93,3 +93,27 @@ function ccdf(d::Cosine{T}, x::Real) where T<:Real
...
+end
+
+function cgf(d::Cosine{T}, t::Real) where T<:Real
+    μ, σ = params(d)
+    t ≈ 0. && return μ*t
+    return 2log(π) + μ*t + σ*abs(t) + log1p(-exp(-2σ*abs(t))) - log(2) - log(σ*abs(t)) - log(T(π^2) + (σ*t)^2)
```
devmotion: Similar to the comments above, here you can also just use `LogExpFunctions.logabssinh` for evaluating the natural logarithm of `sinh` and `LogExpFunctions.log1psq` for computing `log(1 + x^2)`.


```suggestion
function cgf(d::Cosine, t::Real)
    (; μ, σ) = d
    σt = σ * t
    z = iszero(σt) ? zero(float(σt)) : LogExpFunctions.logabssinh(σt) - log(σt)
    return μ*t + z - LogExpFunctions.log1psq(invπ * σt)
```

src/univariate/continuous/cosine.jl
```diff
@@ -93,3 +93,27 @@ function ccdf(d::Cosine{T}, x::Real) where T<:Real
...
+end
+
+function cf(d::Cosine{T}, t::Real) where T<:Real
+    μ, σ = params(d)
+    t ≈ 0. && return one(complex(T))
+    σ*abs(t) ≈ π && return cis(μ*t) / 2
+    return T(π)^2 * cis(μ*t) * sin(σ*t) / (σ*t*(T(π)^2 - (σ*t)^2))
```
devmotion: Similar comments as above.

src/univariate/continuous/cosine.jl
jaksle: Great to know they have these useful functions there, thanks :)

jaksle:
I cleaned `cf` too. I also noticed now the parameters are used once, so their unpacking can be made even simpler.

I added testing against some known simple values. Some more are needed or is this enough?

My code was based how the rest of cosine.jl looked like. When I saw your corrections the code above started looking suspicious for me and indeed, `pdf`, `cdf`` and `ccdf` were type unstable. So I corrected them, though I'm not sure if my solution is optimal.

I'm also not sure about dependencies, but its' easy to change in any case.

jaksle:
@devmotion Could I ask for checking if now it is okay? I am a little unsure about the tests. Is there an instruction somewhere what is the recommended way to write them in Distributions.jl?

src/univariate/continuous/cosine.jl
```diff
@@ -59,37 +59,66 @@ kurtosis(d::Cosine{T}) where {T<:Real} = 6*(90-T(π)^4) / (5*(T(π)^2-6)^2)
...
 #### Evaluation
 
-function pdf(d::Cosine{T}, x::Real) where T<:Real
+function pdf(d::Cosine{T}, x::S) where {T<:Real, S<:Real}
     if insupport(d, x)
         z = (x - d.μ) / d.σ
         return (1 + cospi(z)) / (2d.σ)
     else
-        return zero(T)
+        return zero(promote_type(T, S))
```
devmotion: Can you revert the changes to `pdf`? While the existing code is not optimal, the new code is not correct either.

src/univariate/continuous/cosine.jl
```diff
@@ -59,37 +59,66 @@ kurtosis(d::Cosine{T}) where {T<:Real} = 6*(90-T(π)^4) / (5*(T(π)^2-6)^2)
...
-        return zero(T)
-    end
-    if x > d.μ + d.σ
-        return one(T)
+function logpdf(d::Cosine{T}, x::S) where {T<:Real, S<:Real}
+    if insupport(d, x)
+        z = (x - d.μ) / d.σ
+        return log1p(cospi(z)) - log(2d.σ)
+    else
+        return typemin(promote_type(T, S))
```
devmotion: This is not correct, in general it creates a type instability.

src/univariate/continuous/cosine.jl
```diff
@@ -59,37 +59,66 @@ kurtosis(d::Cosine{T}) where {T<:Real} = 6*(90-T(π)^4) / (5*(T(π)^2-6)^2)
...
     end
+end
+
+function cdf(d::Cosine{T}, x::S) where {T<:Real, S<:Real}
+    W = promote_type(T, S)
```
devmotion: Same here, these promotions just in general aee not the correct return type.

src/univariate/continuous/cosine.jl
```diff
@@ -59,37 +59,66 @@ kurtosis(d::Cosine{T}) where {T<:Real} = 6*(90-T(π)^4) / (5*(T(π)^2-6)^2)
...
 
-function ccdf(d::Cosine{T}, x::Real) where T<:Real
-    if x < d.μ - d.σ
-        return one(T)
-    end
-    if x > d.μ + d.σ
-        return zero(T)
-    end
+function ccdf(d::Cosine{T}, x::S) where {T<:Real, S<:Real}
+    W = promote_type(T, S)
```
devmotion: Also here, that's not the correct return type.

src/Distributions.jl
```diff
@@ -25,8 +25,9 @@ import StatsBase: kurtosis, skewness, entropy, mode, modes,
 import PDMats: dim, PDMat, invquad
 
 using SpecialFunctions
+import LogExpFunctions: logabssinh, log1psq
```
devmotion: Please use `using`, we don't want to overload these.

src/Distributions.jl
```diff
@@ -25,8 +25,9 @@ import StatsBase: kurtosis, skewness, entropy, mode, modes,
...
 using Base.MathConstants: eulergamma
-
+import IrrationalConstants: invπ
```
devmotion: Same here.

PR #2022, opened 2025-12-23
consistently treat Bernoulli as a distribution on Bools

aplavin:
Currently, it's quite inconsistent: eg, `eltype === Bool`, `rand()::Bool`, `minimum()::Bool`, but `mode` and `median` are `Int`, and `quantile` is `T` (often, float).
With this PR `Bernoulli` returns `Bool` whenever possible and makes sense.

This is both self-consistent between functions on `Bernoulli`, and also makes it consistent with other discrete distributions like `Binomial` / `Categorical` / `DiscreteUniform` / `DiscreteNonParametric` that return the minimal type (often `Int`) for their `quantile` and other functions.

### DIFF
--- a/src/univariate/discrete/bernoulli.jl
+++ b/src/univariate/discrete/bernoulli.jl
@@ -64,16 +64,15 @@ skewness(d::Bernoulli) = (p0 = failprob(d); p1 = succprob(d); (p0 - p1) / sqrt(p
 kurtosis(d::Bernoulli) = 1 / var(d) - 6
 
 
-mode(d::Bernoulli) = ifelse(succprob(d) > 1/2, 1, 0)
+mode(d::Bernoulli) = succprob(d) > 1/2
 
 function modes(d::Bernoulli)
     p = succprob(d)
-    p < 1/2 ? [0] :
-    p > 1/2 ? [1] : [0, 1]
+    p < 1/2 ? [false] :
+    p > 1/2 ? [true] : [false, true]
 end
 
-median(d::Bernoulli) = ifelse(succprob(d) <= 1/2, 0, 1)
-
+median(d::Bernoulli) = succprob(d) > 1/2
 function entropy(d::Bernoulli)
     p0 = failprob(d)
     p1 = succprob(d)
@@ -96,11 +95,11 @@ ccdf(d::Bernoulli, x::Bool) = x ? zero(d.p) : succprob(d)
 ccdf(d::Bernoulli, x::Int) = x < 0 ? one(d.p) :
                              x < 1 ? succprob(d) : zero(d.p)
 
-function quantile(d::Bernoulli{T}, p::Real) where T<:Real
-    0 <= p <= 1 ? (p <= failprob(d) ? zero(T) : one(T)) : T(NaN)
+function quantile(d::Bernoulli, p::Real)
+    0 <= p <= 1 ? p > failprob(d) : throw(DomainError(p, "p must be in [0, 1] for quantile"))
 end
-function cquantile(d::Bernoulli{T}, p::Real) where T<:Real
-    0 <= p <= 1 ? (p >= succprob(d) ? zero(T) : one(T)) : T(NaN)
+function cquantile(d::Bernoulli, p::Real)
+    0 <= p <= 1 ? p < succprob(d) : throw(DomainError(p, "p must be in [0, 1] for quantile"))
 end
 
 mgf(d::Bernoulli, t::Real) = failprob(d) + succprob(d) * exp(t)
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -109,7 +109,7 @@ insupport(d::Union{D,Type{D}},x::Real) where {D<:ContinuousUnivariateDistributio
 insupport(d::Union{D,Type{D}},x::Real) where {D<:DiscreteUnivariateDistribution} = isinteger(x) && minimum(d) <= x <= maximum(d)
 
 support(d::Union{D,Type{D}}) where {D<:ContinuousUnivariateDistribution} = RealInterval(minimum(d), maximum(d))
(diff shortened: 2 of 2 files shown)


src/univariate/discrete/bernoulli.jl
```diff
@@ -96,11 +95,11 @@ ccdf(d::Bernoulli, x::Bool) = x ? zero(d.p) : succprob(d)
 ccdf(d::Bernoulli, x::Int) = x < 0 ? one(d.p) :
                              x < 1 ? succprob(d) : zero(d.p)
 
-function quantile(d::Bernoulli{T}, p::Real) where T<:Real
-    0 <= p <= 1 ? (p <= failprob(d) ? zero(T) : one(T)) : T(NaN)
+function quantile(d::Bernoulli, p::Real)
```
devmotion, line 98: Can you revert the changes to `quantile` and `cquantile`? The current API (probably historically based on Rmath) is to never error but to return NaN for invalid arguments. I think this should be changed, but that requires a more general change of the API of these functions: https://github.com/JuliaStats/Distributions.jl/pull/1805

aplavin: reverting `quantile` implies also reverting `median` for consistency, and then basically we lose the point of this PR

any hope of #1805 in the foreseeable future?

aplavin: btw, `quantile` throws sometimes already:
```
julia> quantile(Bernoulli{Int}(1), 2)
ERROR: InexactError: Int64(NaN)
```

devmotion: > any hope of https://github.com/JuliaStats/Distributions.jl/pull/1805 in the foreseeable future?

Has been ready since > 2 years, just needs a review.

> reverting quantile implies also reverting median for consistency, and then basically we lose the point of this PR

The current implementation of `quantile` is wrong (IMO) but it's consistent. It would be worse if some distributions would behave differently, we can't start erroring just for `Bernoulli`. You can still change `mode`, `modes` and `support` without touching `quantile`/`cquantile`.

aplavin: > we can't start erroring just for Bernoulli

It already errors sometimes (and with a different exception). In addition to the example above:
```julia
quantile(Bernoulli(1//3), 2)
```

aplavin: > It would be worse if some distributions would behave differently, we can't start erroring just for Bernoulli

Actually, I don't understand why you say "throw error in quantile()" = "behave differently". It's actually _the_ common way distributions behave in this package:
```julia
julia> quantile(Binomial(10, 0.2), 2)
ERROR: InexactError: Int64(NaN)

julia> quantile(Categorical([0.2, 0.8]), 2)
# the error-throwing code has an issue, but clearly intends to throw DomainError:
ERROR: MethodError: no method matching DomainError()

julia> quantile(DiscreteUniform(1, 5), 2)
# it would better throw than return 10:
10

julia> quantile(DiscreteNonParametric([1,2,5], [0.1, 0.2, 0.7]), 2)
# the error-throwing code has an issue, but clearly intends to throw DomainError:
ERROR: MethodError: no method matching DomainError()
```
Apparently, this PR not just makes Bernoulli internally consistent, but also consistent with other distributions. Wasn't my explicit intention, but happy it turned out this way :)

devmotion: These are all bugs 😅 The current API is to return NaN for invalid inputs.

devmotion: These are all bugs 😅 The current API is to return NaN for invalid inputs.

aplavin: Any references to that? Where do you get that "The current API is to return NaN for invalid inputs"?

Currently, the situation is that basically all discrete distributions return integers from `quantile`. And naturally, they throw an exception when `p` not in `0..1`.
Documentation doesn't seem to contradict this in the slightest.
Even `Bernoulli` often behaves like that. It is `Bernoulli{Float}()` that is an exception (resolved by this PR).

nalimilan: As I just noted at #1805, `NaN` is allowed for many continuous distributions (Normal, Geometric, Exponential, Cauchy, Bernoulli...). And discrete distributions don't throw `DomainError`s, they just get an error during calculations, without any indication that it's intentional. Let's get that PR merged.

aplavin: > discrete distributions don't throw DomainErrors

Technically, this doesn't throw a DomainError indeed:

https://github.com/JuliaStats/Distributions.jl/blob/b0ae395b2bbbb27e31a90d86803ebf676137e705/src/univariate/discrete/discretenonparametric.jl#L165

But only because there's no zero-argument constructor for DomainError. Seems pretty much intentional :)


> continuous distributions (<...>, Bernoulli...)

??

nalimilan: > But only because there's no zero-argument constructor for DomainError. Seems pretty much intentional :)

I was referring to
```julia
julia> quantile(Binomial(10, 0.2), 2)
ERROR: InexactError: Int64(NaN)
```

Though the `DomainError` error is funny.

> continuous distributions (<...>, Bernoulli...)

Sorry, I was talking only about the special case of passing `NaN`: `quantile(Bernoulli(), NaN)`.

aplavin: > julia> quantile(Binomial(10, 0.2), 2)
> ERROR: InexactError: Int64(NaN)

It's an error still, just a less understandable one. If preferred, this PR can be changed into throwing the same one for Bernoulli, although I think DomainError is cleaner.

Anyway, from experience with this and my previous PR attempts to Distributions over the years, I find it super hard to add/change anything in this package. I had barely any success with that, in stark contrast to other Julia packages. So I'll stop here, and just hope that some improvements do happen at some point...

PR #2023, opened 2026-01-04
Allow <:AbstractArray{<:Real} for DirichletMultinomial.

tpapp:
Tested with StaticArrays.SVector.

Incidental: removed unneeded outer constructor.

### DIFF
--- a/src/multivariate/dirichletmultinomial.jl
+++ b/src/multivariate/dirichletmultinomial.jl
@@ -45,7 +45,7 @@ struct DirichletMultinomial{T <: Real,V <: AbstractVector{T}} <: DiscreteMultiva
     end
 end
 DirichletMultinomial(n::Integer, α::AbstractVector{T}) where {T <: Real} = DirichletMultinomial{T}(n, α)
-DirichletMultinomial(n::Integer, α::Vector{T}) where {T <: Integer} = DirichletMultinomial(n, float(α))
+DirichletMultinomial(n::Integer, α::AbstractVector{T}) where {T <: Integer} = DirichletMultinomial(n, float(α))
 DirichletMultinomial(n::Integer, k::Integer) = DirichletMultinomial(n, ones(k))
 
 Base.show(io::IO, d::DirichletMultinomial) = show(io, d, (:n, :α,))


src/multivariate/dirichletmultinomial.jl
```diff
@@ -29,26 +29,26 @@ vector α.
...
 """
-struct DirichletMultinomial{T <: Real} <: DiscreteMultivariateDistribution
+struct DirichletMultinomial{T <: Real,V <: AbstractVector{T}} <: DiscreteMultivariateDistribution
     n::Int
-    α::Vector{T}
+    α::V
     α0::T
 
-    function DirichletMultinomial{T}(n::Integer, α::Vector{T}) where T
+    function DirichletMultinomial(n::Integer, α::V) where {T <: Real, V <: AbstractVector{T}}
```
devmotion: I doubt it matters in practice but technically removing this constructor `DirichletMultinomial{T}` is breaking... Maybe a search theough the JuliaHub package index can confirm that it's not used in any publix packages at least?

Maybe the safer (better?) alternative is to just keep this particular constructor for `Vector{T}` (maybe with a comment that it should be removed in a breaking release).

tpapp: You are right, [there is code out there using it](https://github.com/search?q=%22DirichletMultinomial{%22&type=code) (probably just mindlessly copying a bad API, but nevertheless), so I will keep it.

src/multivariate/dirichletmultinomial.jl
```diff
@@ -29,17 +29,19 @@ vector α.
...
-    function DirichletMultinomial{T}(n::Integer, α::Vector{T}) where T
+    function DirichletMultinomial{T}(n::Integer, α::V) where {T <: Real, V <: AbstractVector{T}}
         α0 = sum(abs, α)
         sum(α) == α0 || throw(ArgumentError("alpha must be a positive vector."))
         n > 0 || throw(ArgumentError("n must be a positive integer."))
-        new{T}(Int(n), α, α0)
+        new{T,V}(Int(n), α, α0)
     end
 end
 DirichletMultinomial(n::Integer, α::Vector{T}) where {T <: Real} = DirichletMultinomial{T}(n, α)
```
devmotion: AFAICT this has to be changed to

```suggestion
DirichletMultinomial(n::Integer, α::AbstractVector{T}) where {T <: Real} = DirichletMultinomial{T}(n, α)
```

tpapp: good point. thanks

src/multivariate/dirichletmultinomial.jl
```diff
@@ -29,26 +29,27 @@ vector α.
...
         α0 = sum(abs, α)
         sum(α) == α0 || throw(ArgumentError("alpha must be a positive vector."))
         n > 0 || throw(ArgumentError("n must be a positive integer."))
-        new{T}(Int(n), α, α0)
+        new{T,V}(Int(n), α, α0)
     end
 end
-DirichletMultinomial(n::Integer, α::Vector{T}) where {T <: Real} = DirichletMultinomial{T}(n, α)
+DirichletMultinomial(n::Integer, α::AbstractVector{T}) where {T <: Real} = DirichletMultinomial{T}(n, α)
 DirichletMultinomial(n::Integer, α::Vector{T}) where {T <: Integer} = DirichletMultinomial(n, float(α))
```
devmotion: Might be worth changing and testing as well?
```suggestion
DirichletMultinomial(n::Integer, α::AbstractVector{T}) where {T <: Integer} = DirichletMultinomial(n, float(α))
```

PR #2027, opened 2026-01-29
move `Hypergeometric`, to pure julia

oscardssmith:
This is a working but not fully optimized version.


src/univariate/discrete/hypergeometric.jl
```diff
@@ -84,10 +84,145 @@ entropy(d::Hypergeometric) = entropy(map(Base.Fix1(pdf, d), support(d)))
...
 
-# TODO: remove Rmath dependency. Implement:
+# Implements
 #   V. Kachitvichyanukul & B. Schmeiser
 #   "Computer generation of hypergeometric random variates"
 #   Journal of Statistical Computation and Simulation, 22(2):127-145
 #   doi:10.1080/00949658508810839
 @rand_rdist(Hypergeometric)
-rand(d::Hypergeometric) = convert(Int, Rmath.rhyper(d.ns, d.nf, d.n))
+function rand(d::Hypergeometric)
```
devmotion: This should be changed to
```suggestion
function rand(rng::AbstractRNG, d::Hypergeometric)
```

src/univariate/discrete/hypergeometric.jl
```diff
@@ -84,10 +84,145 @@ entropy(d::Hypergeometric) = entropy(map(Base.Fix1(pdf, d), support(d)))
...
+    # 0.0
+    n1s, n2s, ks = d.ns, d.nf, d.n
+    n1, n2, k = d.ns, d.nf, d.n
```
devmotion: Can we change this to use the parameter names of the distribution? IMO that would make it easier to read the code.
```suggestion
    (; ns, nf, n) = d
```

oscardssmith: that's the plan (I was following variable names from the paper, but I'll switch it before merging).

src/univariate/discrete/hypergeometric.jl
```diff
@@ -84,10 +84,144 @@ entropy(d::Hypergeometric) = entropy(map(Base.Fix1(pdf, d), support(d)))
...
+            return  ks - y
+        end
+    elseif n1s <= n2s
+        return n1s - y
+    else
+        return ks - n2s + y
+    end
+end
+
+function _hyper(rng, n, n1, n2, k)
```
devmotion: It seems there are quite a few terms that could be precomputed? That indicates that it might be better for repeated sampling if we'd implement a new `Sampler` type `HyperGeometricSampler` + `sampler(d::HyperGeometric) = HyperGeometricSampler(d)` instead of defining `rand(rng, ::HyperGeometric)`.

oscardssmith: agreed. I would propose saving that for a followup PR (the main purpose of this PR for me is just to get rid of RMath)

src/univariate/continuous/noncentralchisq.jl
```diff
@@ -78,7 +78,8 @@ end
...
 
 @_delegate_statsfuns NoncentralChisq nchisq ν λ
-
-# TODO: remove Rmath dependency
-@rand_rdist(NoncentralChisq)
-rand(d::NoncentralChisq) = Rmath.rnchisq(d.ν, d.λ)
+function rand(rng::AbstractRNG, d::NoncentralChisq)
+    # use symmetry of distribution as suggested in
+    # https://mathstoshare.com/2022/05/26/sampling-from-the-non-central-chi-squared-distribution/
+    return (randn(rng) + sqrt(d.λ))^2 + rand(rng, Chisq(d.ν-1))
```
devmotion: This will error if `d.ν <= 1`.

oscardssmith: For my understanding, is `d.ν` restricted to integers or is the distribution generalized?

devmotion: No, AFAICT it's not restricted to integers.

oscardssmith: Turns out there's a good simple alg for this case too. Test added as well.

oscardssmith:
After some looking, it turns out that we've had a bug in `NoncentralBeta` for the past 11 years (introduced in https://github.com/JuliaStats/Distributions.jl/commit/9a7ac0fff19bc93b88fb803b2deea3b7950bd238) where in the renaming of parameters, the noncentrality parameter got replaced with beta.

src/univariate/discrete/hypergeometric.jl
```diff
@@ -84,10 +84,114 @@ entropy(d::Hypergeometric) = entropy(map(Base.Fix1(pdf, d), support(d)))
...
 #   Journal of Statistical Computation and Simulation, 22(2):127-145
 #   doi:10.1080/00949658508810839
-@rand_rdist(Hypergeometric)
-rand(d::Hypergeometric) = convert(Int, Rmath.rhyper(d.ns, d.nf, d.n))
+function rand(rng::AbstractRNG, d::Hypergeometric)
+    # step 0.0
+    (;ns, nf, n) = d
+    ntotal= ns+nf
+    ns, nf = minmax(ns, nf)
+    n = min(n, ntotal-n)
```
devmotion: Can we use different variable names here?

src/univariate/continuous/noncentralbeta.jl
```diff
@@ -38,20 +38,8 @@ partype(::NoncentralBeta{T}) where {T} = T
...
-    a = rand(NoncentralChisq(2d.α, β))
-    b = rand(Chisq(2β))
-    a / (a + b)
-end
-
 function rand(rng::AbstractRNG, d::NoncentralBeta)
-    β = d.β
-    a = rand(rng, NoncentralChisq(2d.α, β))
-    b = rand(rng, Chisq(2β))
+    a = rand(rng, NoncentralChisq(2d.α, d.λ))
```
devmotion, line 43: Would the tests in the PR have caught this bug?

src/univariate/discrete/negativebinomial.jl
```diff
@@ -107,7 +107,6 @@ function logpdf(d::NegativeBinomial, k::Real)
     return insupport(d, k) ? z - log(k + r) - logbeta(r, k + 1) : oftype(z, -Inf)
 end
 
-# cdf and quantile functions are more involved so we still rely on Rmath
```
devmotion, line 110: This comment is still correct, theses StatsFuns functions are just wrappers around the Rmath functions: https://github.com/JuliaStats/StatsFuns.jl/blob/63373be9447ed5d74ea9aa63a9c939704b14876d/src/distrs/nbinom.jl

```suggestion
# cdf and quantile functions are more involved so we still rely on Rmath
```

test/ref/continuous_test.ref.json
```diff
@@ -3943,6 +3943,32 @@
     { "q": 0.90, "x": 0.638212815185811 }
   ]
 },
+{
+  "expr": "NoncentralChisq(0.5, 2)",
```
devmotion: How did you generate these values? I don't see any change in https://github.com/JuliaStats/Distributions.jl/blob/69cb9392453532f8910ed55042a05b5bb77043ac/test/ref/continuous_test.lst#L148-L149 in this PR.

LICENSE.md
```diff
@@ -24,7 +24,3 @@ The Distributions module is licensed under the MIT License:
 > LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
 > OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 > WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
-
-The Distributions module uses the RMath library, which has its own license:
```
devmotion, line 28: It still does through StatsFuns

src/univariate/discrete/hypergeometric.jl
oscardssmith: what would you like me to call it? `n` is the name from the parameterization, but I agree that it's a bad name. Perhaps `nsamples`?

src/univariate/continuous/noncentralbeta.jl
oscardssmith: yeah. removing the
```
    elseif isa(d, NoncentralBeta)
        n_tsamples = min(n_tsamples, 100)
 ```
 catches this bug. We were previously intentionally not testing that `NoncentralBeta` produced correct samples.

File: src/univariate/discrete/negativebinomial.jl (hunk 7 above)
Comment (oscardssmith) [reply src/univariate/discrete/negativebinomial.jl:110]: oops. I'll undo this change for now, but followup with a StatsFuns pr soon.

File: test/ref/continuous_test.ref.json (hunk 8 above)
Comment (oscardssmith) [reply test/ref/continuous_test.ref.json]: these were generated  using the julia implementation (which is OK here IMO since this PR didn't change anything about how quantiles/pdfs/cdfs etc are implimented. I'll update the .lst though)

File: src/univariate/continuous/noncentralbeta.jl (hunk 6 above)
Comment (oscardssmith) [reply src/univariate/continuous/noncentralbeta.jl:43]: if you want, I can move that out to a separate PR since it's a mostly unrelated bugfix.

File: test/ref/continuous_test.ref.json (hunk 8 above)
Comment (devmotion) [reply test/ref/continuous_test.ref.json]: The point of these reference tests is to test against R implementations. Generating reference values with Julia defeats this purpose.

File: src/univariate/discrete/negativebinomial.jl (hunk 7 above)
Comment (oscardssmith) [reply src/univariate/discrete/negativebinomial.jl:110]: Actually I think this comment should likely remain removed. While this comment is currently correct it's
1. not especially actionable. Users are unlikely to see this comment, and even if they do, it's unclear what they are supposed to take away from it.
2. not very useful. Distributions.jl doesn't really care where these methods come from.
3. likely to become out of date (e.g. https://github.com/JuliaStats/StatsFuns.jl/pull/197 would make this incorrect) without fixing the comment.

File: src/univariate/discrete/hypergeometric.jl (hunk 5 above)
Comment (oscardssmith) [reply src/univariate/discrete/hypergeometric.jl]: bump on this. I'm happy to change names, but don't want to go back and forth repeadedly. is `nsamples` a good name, or should I leave it as `n`, or secret option 3?

**Comment by oscardssmith** [comment]
Looks like for large values this is ~25% faster and for small parameters it is 16x faster.
```
julia> rrand(d::Hypergeometric) = convert(Int, Rmath.rhyper(d.ns, d.nf, d.n))
julia> d = Hypergeometric(60,80,40)
julia> @benchmark rand($d)
BenchmarkTools.Trial: 10000 samples with 803 evaluations per sample.
 Range (min … max):  160.355 ns … 336.671 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     165.796 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   166.872 ns ±   6.559 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

        ▄▇█▅▂                                                    
  ▁▁▂▃▅██████▇▅▃▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  160 ns           Histogram: frequency by time          199 ns <

julia> @benchmark rrand($d)
BenchmarkTools.Trial: 10000 samples with 636 evaluations per sample.
 Range (min … max):  195.027 ns … 232.472 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     202.698 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   203.039 ns ±   2.636 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                       ▁▃▄▆▆▆▆█▇▆▅▄▃▃▁▁                          
  ▂▁▁▁▁▁▁▁▂▂▂▂▂▃▃▄▄▅▅▆▇█████████████████▇▆▆▅▅▅▄▄▃▃▃▃▃▃▂▂▂▂▂▂▂▂▂ ▄
  195 ns           Histogram: frequency by time          211 ns <

julia> d = Hypergeometric(2,2,2)
Hypergeometric(ns=2, nf=2, n=2)

julia> @benchmark rand($d)
BenchmarkTools.Trial: 10000 samples with 999 evaluations per sample.
 Range (min … max):   9.919 ns … 19.777 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     10.642 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   10.689 ns ±  0.413 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

            ▁▄▅▇█▆▄▅▄▁                                         
  ▂▂▂▂▃▃▄▅▇████████████▆▅▄▄▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▁▂▂▁▂▁▂▁▂▁▂▂▂▂▂▂ ▃
  9.92 ns         Histogram: frequency by time        12.8 ns <

julia> @benchmark rrand($d)
BenchmarkTools.Trial: 10000 samples with 767 evaluations per sample.
 Range (min … max):  164.605 ns … 335.752 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     166.837 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   167.367 ns ±   4.614 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

      ▅█▄▂▅▄▁                                                    
  ▂▂▂▆███████▆▄▄▄▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂▂▂▂▂▂ ▃
  165 ns           Histogram: frequency by time          182 ns <

```

src/univariate/discrete/negativebinomial.jl
devmotion: Seems reasonable.

PR #2031, opened 2026-02-03
Use native NoncentralChisq implementation

oscardssmith:
Split from https://github.com/JuliaStats/Distributions.jl/pull/2027 (and slightly simplified) for easier merging.
```
julia> d = NoncentralChisq(2, 5)
julia> rmathrand(d) = Rmath.rnchisq(d.ν, d.λ)
julia> d = NoncentralChisq(2, 5)
NoncentralChisq{Float64}(ν=2.0, λ=5.0)

julia> @benchmark rand($d)
BenchmarkTools.Trial: 10000 samples with 992 evaluations per sample.
 Range (min … max):  30.573 ns … 91.020 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     32.764 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   32.904 ns ±  1.528 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                    ▁▃▄▆▅██▅▇▅▄▂▂                              
  ▂▁▁▁▁▂▂▂▂▂▃▃▃▄▄▅▇███████████████▇▅▅▄▄▃▃▃▂▂▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▄
  30.6 ns         Histogram: frequency by time        35.8 ns <

julia> @benchmark rmathrand($d)
BenchmarkTools.Trial: 10000 samples with 966 evaluations per sample.
 Range (min … max):  79.936 ns … 128.392 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     90.851 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   91.095 ns ±   3.492 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                      ▁▂▃▄▅▇▇█▇█▆██▇▆▆▄▂▁▁▁                     
  ▂▁▂▂▂▂▂▂▂▂▂▃▃▄▄▅▅▆████████████████████████▇▅▆▅▅▄▄▃▃▃▃▃▂▂▂▂▂▂ ▅
  79.9 ns         Histogram: frequency by time          101 ns <

julia> d = NoncentralChisq(500, 500)
julia> @benchmark rand($d)
BenchmarkTools.Trial: 10000 samples with 995 evaluations per sample.
 Range (min … max):  22.737 ns … 38.768 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     24.136 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   24.282 ns ±  0.857 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

           ▂▄▅█▆▇▄▄▄▄▂▄▁▁                                      
  ▁▁▁▁▂▃▄▆▇██████████████▇▇▆▅▄▄▂▂▂▂▁▁▁▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  22.7 ns         Histogram: frequency by time        27.8 ns <

julia> @benchmark rmathrand($d)
BenchmarkTools.Trial: 10000 samples with 983 evaluations per sample.
 Range (min … max):  58.993 ns … 107.693 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     61.205 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   61.597 ns ±   2.863 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

        ▃▅██▇▅▃                                                 
  ▂▂▂▃▅▇███████▇▆▄▃▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▃
  59 ns           Histogram: frequency by time         72.5 ns <

```

### DIFF
--- a/test/univariates.jl
+++ b/test/univariates.jl
@@ -144,8 +144,6 @@ function verify_and_test(D::Union{Type,Function}, d::UnivariateDistribution, dct
     # generic testing
     if isa(d, Cosine)
         n_tsamples = floor(Int, n_tsamples / 10)
-    elseif isa(d, NoncentralT)
-        n_tsamples = min(n_tsamples, 100)
     end
 
     if !isa(d, Union{Skellam,


src/univariate/continuous/noncentralchisq.jl
```diff
@@ -79,6 +79,7 @@ end
 
 @_delegate_statsfuns NoncentralChisq nchisq ν λ
 
-# TODO: remove Rmath dependency
-@rand_rdist(NoncentralChisq)
-rand(d::NoncentralChisq) = Rmath.rnchisq(d.ν, d.λ)
+function rand(rng::AbstractRNG, d::NoncentralChisq)
+    J = rand(rng, Poisson(d.λ/2))
```
oscardssmith, line 83: using
```
    if d.ν <= 1
        J = rand(rng, Poisson(d.λ/2))
        return rand(rng, Chisq(d.ν + 2 * J))
    end
    # use symmetry of distribution as suggested in
    # https://mathstoshare.com/2022/05/26/sampling-from-the-non-central-chi-squared-distribution/
    return (randn(rng) + sqrt(d.λ))^2 + rand(rng, Chisq(d.ν-1))
```
seems to be slightly faster, but I'm preferring the simple version since it already is a performance win, and easier to test/verify.

devmotion: The second form would also be better for defining a composed sampler, which likely would give an additional performance improvement when generating multiple variates.

test/univariates.jl
```diff
@@ -144,9 +144,7 @@ function verify_and_test(D::Union{Type,Function}, d::UnivariateDistribution, dct
     # generic testing
     if isa(d, Cosine)
         n_tsamples = floor(Int, n_tsamples / 10)
-    elseif isa(d, NoncentralChisq) ||
-           isa(d, NoncentralF) ||
-           isa(d, NoncentralT)
+    elseif isa(d, NoncentralT)
```
devmotion: Is it required to keep this branch?

oscardssmith: I don't think so. will check.

src/univariate/continuous/noncentralchisq.jl
oscardssmith: agreed. Happy to make a followup PR once this is merged.

PR #1994, opened 2025-08-18
Implementing Hypergeometric sampling to remove RFunctions dependency

JaoCR:
Following the TODO in the Hypergeometric distribution sampling code, this PR implements the H2PE/HIM algorithm presented by Kachitvichyanukul & Schmeiser in 1984, removing this dependency on RFunctions and improving the sampling performance with a penalty in setup time and some extra stored constants.

Added some test cases for complete coverage of the distribution code. More extensive testing with the distribution of the generated variables was done (locally, manually) on both of the sub-cases of this algorithm.

## Example case of HIM

The HIM algorith runs for smaller distributions. To exemplify, the `Hypergeometric(20, 20, 10)`:

**Setup Benchmark (with RFunctions):**
image

**Setup Benchmark (HIM):**
image

**Sampling Benchmark (with RFunctions):**
image

**Sampling Benchmark (HIM):**
image

**QQPlot with 1000 samples**

plot_3

## Example case of H2PE

The H2PE algorithm is used for larger cases. To examplify, `Hypergeometric(200, 200, 100)`:

**Setup Benchmark (with RFunctions):**
image

**Setup Benchmark (H2PE)**
image

**Sampling Benchmark (with RFunctions):**
image

**Sampling Benchmark (H2PE):**
image

**QQPlot with 1000 samples**
plot_4

Feel free to ask any questions and do/request any adjustments. Have a nice week.

PR #2032, opened 2026-02-05
Julia `Hypergeometric` sampler

oscardssmith:
alternative to https://github.com/JuliaStats/Distributions.jl/pull/2027 that implements the Sampler interface for faster repeated random numbers.
```
julia> d = Hypergeometric(60,80,40)
Hypergeometric(ns=60, nf=80, n=40)

julia> @benchmark spl=sampler($d)
BenchmarkTools.Trial: 10000 samples with 900 evaluations per sample.
 Range (min … max):  123.613 ns … 159.059 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     124.582 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   125.135 ns ±   2.068 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▂▅▇▇▇█▆▄▂▁▁▁▁▁▁▁   ▁▁▁▁▂▁▁▁ ▁▁                               ▂
  ▅████████████████████████████████████▇▇▇▇▇▅▆▅▆▅▅▆▅▄▄▄▅▄▅▅▄▂▃▄ █
  124 ns        Histogram: log(frequency) by time        133 ns <

julia> @benchmark rand($spl)
BenchmarkTools.Trial: 10000 samples with 991 evaluations per sample.
 Range (min … max):  40.188 ns … 77.342 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     45.353 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   45.502 ns ±  1.749 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                       ▁▁▃▅▅▆▇▇█▇▆▇▆▄▄▃▁▁                      
  ▂▁▁▂▂▁▂▂▂▂▂▂▃▃▃▄▄▅▅▇▇███████████████████▇▆▅▅▄▃▃▃▃▃▃▂▂▂▂▂▂▂▂ ▄
  40.2 ns         Histogram: frequency by time        50.3 ns <

 Memory estimate: 0 bytes, allocs estimate: 0
```
So https://github.com/JuliaStats/Distributions.jl/pull/2027 has 165ns rand times, while this has 125ns setup and 45ns for a 3.75x speedup when generating multiple numbers.

### DIFF
--- a/test/samplers.jl
+++ b/test/samplers.jl
@@ -135,8 +135,8 @@ import Distributions:
             MvNormal(zeros(3), I),
             Normal(1.5, 2.0),
             Poisson(0.5),
-            Hypergeometric(2, 2, 2)
-            Hypergeometric(100, 200, 40)
+            Hypergeometric(2, 2, 2),
+            Hypergeometric(100, 200, 40),
         )
             @test Random.Sampler(rng, dist, Val(1)) == dist
             @test Random.Sampler(rng, dist) == sampler(dist)


devmotion:
I would prefer continuing with #1994. There are a few things to address and improve in #1994 but the test setup in #1994 is superior to this PR. Moreover, by continuing with or basing a new PR on #1994 we can properly give credit to the original author of #1994. I think it's not fine to just copy (parts of) #1994 without giving credit (even though legally you can do that, of course).

PR #2033, opened 2026-02-07
Fix R comparison generator code

nalimilan:
gendref.R doesn't run currently due to various issues:
- one package doesn't install from CRAN
- `truncated(Normal(...), ...)` needs to be `TruncatedNormal` for R
- `StudentizedRange` fails because it has no `pdf`

With this commit the file runs in R and tests pass in Julia with the generated file. I have kept the existing file though as it seems it contains some manual additions when the PDF was missing.

### DIFF
--- a/test/ref/continuous_test.ref.json
+++ b/test/ref/continuous_test.ref.json
@@ -5338,8 +5338,8 @@
   ]
 },
 {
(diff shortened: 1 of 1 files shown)


devmotion:
I made a PR that fixes the reference tests and makes them easier to run a while ago: https://github.com/JuliaStats/Distributions.jl/pull/2007

PR #2034, opened 2026-02-11
minor numerical improvement to `logpdf(::Skellam)`

oscardssmith:
`besseli` blows up so it's better to use `besselix` and re-normalize.

### DIFF
--- a/src/univariate/discrete/skellam.jl
+++ b/src/univariate/discrete/skellam.jl
@@ -76,7 +76,7 @@ function logpdf(d::Skellam, x::Real)
     μ1, μ2 = params(d)
     if insupport(d, x)
         besselarg =  2*sqrt(μ1*μ2)
-        return - (μ1 + μ2) + xlogy(x/2, μ1/μ2) - besselarg + log(besselix(x, besselarg))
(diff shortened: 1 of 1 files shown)


src/univariate/discrete/skellam.jl
```diff
@@ -75,7 +75,8 @@ kurtosis(d::Skellam) = 1 / var(d)
 function logpdf(d::Skellam, x::Real)
     μ1, μ2 = params(d)
     if insupport(d, x)
-        return - (μ1 + μ2) + (x/2) * log(μ1/μ2) + log(besseli(x, 2*sqrt(μ1)*sqrt(μ2)))
+        besselarg =  2*sqrt(μ1*μ2)
+        return - (μ1 + μ2) + xlogy(x/2, μ1/μ2) - besselarg + log(besselix(x, besselarg))
```
devmotion: AFAICT this should be
```suggestion
        return - (μ1 + μ2) + xlogy(x/2, μ1/μ2) + besselarg + log(besselix(x, besselarg))
```
?

I wonder if there's any inputs for which it would be advantageous to compute it as
```suggestion
        return - (sqrt(μ1) - sqrt(μ2))^2 + xlogy(x/2, μ1/μ2) + log(besselix(x, besselarg))
```
instead (in which case clearly the `sqrt` should be computed only once and reused in the line above).

PR #2045, opened 2026-03-16
Add Tweedie distribution

nalimilan:
This distribution is tricky to implement as there is no closed form for most functions. PDF is computed based on `wrightbessel` (https://github.com/JuliaMath/SpecialFunctions.jl/pull/512). CDF and quantile are computed numerically from PDF. Performance is similar to reference R package tweedie.

When `1 < p < 2` it is continuous but with a point mass at zero so it is not completely correct to consider it a `ContinuousUnivariateDistribution`. When `p = 1` it is even discrete. However I wasn't sure that justifies adding a new type like `SemiContinuousUnivariateDistribution`.

This distribution is useful to allow estimating Tweedie models in GLM.


devmotion:
Tests fail currently.

src/univariate/continuous/tweedie.jl
```diff
@@ -0,0 +1,244 @@
...
+    σ::T
+    p::T
+
+    Tweedie{T}(µ::T, σ::T, p::T) where {T<:Real} = new{T}(µ, σ, p)
+end
+
+function Tweedie(μ::T, σ::T, p::T; check_args::Bool=true) where {T <: Real}
+    @check_args Tweedie (μ, μ >= zero(μ))
+    @check_args Tweedie (σ, σ >= zero(σ))
+    @check_args Tweedie (p, 1 <= p <= 2)
```
devmotion: Not 100% sure but IIRC you can use a single check_args with a Tuple of checks.

src/univariate/continuous/tweedie.jl
```diff
@@ -0,0 +1,244 @@
...
+    return Tweedie{T}(μ, σ, p)
+end
+
+#### Outer constructors
+Tweedie(μ::Real, σ::Real, p::Real; check_args::Bool=true) =
+    Tweedie(promote(μ, σ, p)...; check_args=check_args)
+Tweedie(μ::Integer, σ::Integer, p::Integer; check_args::Bool=true) =
+    Tweedie(float(μ), float(σ), float(p); check_args=check_args)
```
devmotion, line 67: Not needed?

```suggestion
```

src/univariate/continuous/tweedie.jl
```diff
@@ -0,0 +1,244 @@
...
+
+#### Conversions
+convert(::Type{Tweedie{T}}, μ::S, σ::S, p::S) where {T <: Real, S <: Real} = Tweedie(T(μ), T(σ), T(p))
```
devmotion, line 70: IMO no such method should be defined - users should use the constructor:
```suggestion
```

src/univariate/continuous/tweedie.jl
```diff
@@ -0,0 +1,244 @@
...
+Base.convert(::Type{Tweedie{T}}, d::Tweedie) where {T<:Real} = Tweedie{T}(T(d.μ), T(d.σ), T(d.p))
+Base.convert(::Type{Tweedie{T}}, d::Tweedie{T}) where {T<:Real} = d
+
+@distr_support Tweedie 0 Inf
+
+#### Parameters
+
+params(d::Tweedie) = (d.μ, d.σ, d.p)
+partype(d::Tweedie{T}) where {T} = T
```
devmotion:
```suggestion
partype(::Tweedie{T}) where {T} = T
```

src/univariate/continuous/tweedie.jl
```diff
@@ -0,0 +1,244 @@
...
+location(d::Tweedie) = d.μ
+scale(d::Tweedie) = d.σ
+shape(d::Tweedie) = d.p
+
+Base.eltype(::Type{Tweedie{T}}) where {T} = T
+
+#### Statistics
+
+mean(d::Tweedie) = d.μ
+var(d::Tweedie) = d.σ^2 * d.μ^d.p
```
devmotion, line 90: I'd suggest defining `std` as well to avoid unnecessary squaring and subsequent square root calculations:

```suggestion
var(d::Tweedie) = d.σ^2 * d.μ^d.p
std(d::Tweedle) = d.σ * d.μ^(d.p/2)
```

src/univariate/continuous/tweedie.jl
```diff
@@ -0,0 +1,244 @@
...
+
+# Clark, David R. and Charles A. Thayer. 2004.
+# “A Primer on the Exponential Family of Distributions.” CAS Discussion Paper Program, 117-148
+# https://www.casact.org/sites/default/files/database/dpp_dpp04_04dpp117.pdf
+skewness(d::Tweedie) = d.p * d.σ / sqrt(d.μ ^ (2 - d.p))
+kurtosis(d::Tweedie) = d.p * (2 * d.p - 1) * d.σ^2 / d.μ ^ (2 - d.p)
+
+function logpdf(d::Tweedie{T}, x::Real) where {T <: Real}
+    isnan(x) && return eltype(d)(NaN)
```
devmotion: This return type is wrong in general (might not even support NaN) and leads to type instabilities.

src/univariate/continuous/tweedie.jl
```diff
@@ -0,0 +1,244 @@
...
+    x >= 0 || return -Inf
```
devmotion, line 101: Same here.

src/univariate/continuous/tweedie.jl
```diff
@@ -0,0 +1,244 @@
...
+        if x > 0
+            z = ((p - 1) * ϕ / x) ^ α / ((2 - p) * ϕ)
+            # Use log to reduce risks of overflow when p is close to 1
+            wb = logwrightbessel(Float64(-α), 0.0, Float64(z))
+            # Overflow in `logwrightbessel` doesn't generally indicate that the PDF
+            # value would be larger than `typemax(Float64)`
+            wb == Inf && return NaN
+            res += wb - log(x)
+        end
+        return T(res)
```
devmotion: Also incorrect in general.

src/univariate/continuous/tweedie.jl
```diff
@@ -0,0 +1,244 @@
...
+    else
+        θ = μ ^ (1 - p) / (1 - p)
+        κ = μ ^ (2 - p) / (2 - p)
+        α = (2 - p) / (1 - p)
+
+        res = (x * θ - κ) / ϕ
+        if x > 0
+            z = ((p - 1) * ϕ / x) ^ α / ((2 - p) * ϕ)
+            # Use log to reduce risks of overflow when p is close to 1
+            wb = logwrightbessel(Float64(-α), 0.0, Float64(z))
```
devmotion, line 129: Could lead to silently imprecise results when operating with eg BigFloat

src/univariate/continuous/tweedie.jl
```diff
@@ -0,0 +1,244 @@
...
+            res += wb - log(x)
+        end
+        return T(res)
+    end
+end
+
+function cdf(d::Tweedie, x::Real)
+    isnan(x) && return eltype(d)(NaN)
+    x == Inf && return one(eltype(d))
+    x >= 0 || return zero(eltype(d))
```
devmotion: Same problems with return types here as well

src/univariate/continuous/tweedie.jl
```diff
@@ -0,0 +1,244 @@
...
+
+# Implementation inspired by `qtweedie` in R package tweedie
+# licensed under MIT with authorization from Peter Dunn
+function quantile(d::Tweedie{T}, q::Real) where {T <: Real}
+    μ, ϕ, p = d.μ, d.σ^2, d.p
+
+    if q == 0
+        return zero(T)
+    elseif q == 1
+        return convert(T, Inf)
```
devmotion, line 181: Same problems

test/ref/continuous/tweedie.R
```diff
@@ -0,0 +1,44 @@
...
+            mu <- self$mu
+            sigma <- self$sigma
+            p <- self$p
+            list(location=mu,
+                 scale=sigma,
+                 shape=p,
+                 mean=mu,
+                 var=mu^p * sigma^2,
+                 skewness=p * sigma / sqrt(mu ^ (2 - p)),
+                 kurtosis=p * (2 * p - 1) * sigma^2 / mu ^ (2 - p))
```
devmotion: IMO these properties are insufficient as tests, and hence should just be removed - you're testing your Julia implementations against R reference properties that you implemented in exactly the same way.

```suggestion
```

test/ref/continuous/tweedie.R
```diff
@@ -0,0 +1,44 @@
...
+        },
```
devmotion:
```suggestion
```

test/ref/continuous_test.ref.json
```diff
@@ -5457,6 +5457,171 @@
...
+  "maximum": "inf",
+  "properties": {
+    "location": 2,
+    "scale": 1.5,
+    "shape": 1,
+    "mean": 2,
+    "var": 4.5,
+    "skewness": 1.06066017177982,
+    "kurtosis": 1.125
+  },
```
devmotion, line 5466:
```suggestion
```

test/ref/continuous_test.ref.json
```diff
@@ -5457,6 +5457,171 @@
...
+  "maximum": "inf",
+  "properties": {
+    "location": 2,
+    "scale": 1.5,
+    "shape": 1.01,
+    "mean": 2,
+    "var": 4.53129997525524,
+    "skewness": 1.07498594232179,
+    "kurtosis": 1.16703630862699
+  },
```
devmotion, line 5492:
```suggestion
```

test/ref/continuous_test.ref.json
```diff
@@ -5457,6 +5457,171 @@
...
+  "maximum": "inf",
+  "properties": {
+    "location": 2,
+    "scale": 1.5,
+    "shape": 1.5,
+    "mean": 2,
+    "var": 6.36396103067893,
+    "skewness": 1.89201693432086,
+    "kurtosis": 4.7729707730092
+  },
```
devmotion, line 5518:
```suggestion
```

test/ref/continuous_test.ref.json
```diff
@@ -5457,6 +5457,171 @@
...
+  "maximum": "inf",
+  "properties": {
+    "location": 2,
+    "scale": 1.5,
+    "shape": 1.99,
+    "mean": 2,
+    "var": 8.93783245893332,
+    "skewness": 2.97467268454119,
+    "kurtosis": 13.2507835119916
+  },
```
devmotion, line 5544:
```suggestion
```

test/ref/continuous_test.ref.json
```diff
@@ -5457,6 +5457,171 @@
...
+  "maximum": "inf",
+  "properties": {
+    "location": 2,
+    "scale": 1.5,
+    "shape": 2,
+    "mean": 2,
+    "var": 9,
+    "skewness": 3,
+    "kurtosis": 13.5
+  },
```
devmotion, line 5570:
```suggestion
```

test/univariate/continuous/tweedie.jl
devmotion, line 1: We should also test Float32 inputs, integer inputs and mixes of Float32, Float64, and integers, and add `@inferred` around the function calls. This would have caught the type stability issues.

src/univariate/continuous/tweedie.jl
nalimilan: Yes that's an issue. I don't have a good solution as `logwrightbessel` has been written to support only `Float64`. Should I throw an error for types with a higher precision than `Float64`? I don't know how this is handled in other distributions (e.g. those which rely on Rmath).

test/univariate/continuous/tweedie.jl
nalimilan: Will do. I wasn't sure whether this was already covered by generic tests (which seem quite thorough).

nalimilan:
Thanks for the review! I will address your comments. Note that CI will fail until https://github.com/JuliaMath/SpecialFunctions.jl/pull/512 is merged and released (which isn't a small task...).

src/univariate/continuous/tweedie.jl
nalimilan: I copied this from `Normal`. Not sure why it does that.

src/univariate/continuous/tweedie.jl
nalimilan: This also comes from `Normal` (also exists for other distributions). It makes sense IMO at least in theory, e.g. if you want to store such an object in a field with a different `T`. Though in practice it's probably not super common.

src/univariate/continuous/tweedie.jl
nalimilan: Indeed I was being overzealous. `logpdf` and similar always return a `Float64`, right?

src/univariate/continuous/tweedie.jl
nalimilan: Here since the quantile is a value on the same scale as values returned by `rand`, isn't it legitimate to ensure the result matches `eltype(d)`? Otherwise, what type could we choose?

src/univariate/continuous/tweedie.jl
nalimilan: Actually it seems that for other distributions the return type depends both on `eltype(d)` and `typeof(x)`:
```julia
julia> typeof(pdf(Gamma(1.0f0), 1.0f0))
Float32

julia> typeof(pdf(Gamma(1.0), 1.0f0))
Float64

julia> typeof(pdf(Gamma(1.0f0), 1.0))
Float64
```

So I've applied the same strategy. For Tweedie the return type has to be chosen explicitly view `promote_type` as `logwrightbessel` only supports `Float64` so we can't just return whatever type it returns (and we want to ensure type stability when `logpdf(Poisson(...), ...)` or `logpdf(Gamma(...), ...)` is called instead).

test/ref/continuous_test.ref.json
nalimilan: I had to keep an empty `properties` block to avoid an error.

src/univariate/continuous/tweedie.jl
```diff
@@ -73,27 +76,28 @@ Base.convert(::Type{Tweedie{T}}, d::Tweedie{T}) where {T<:Real} = d
...
 scale(d::Tweedie) = d.σ
 shape(d::Tweedie) = d.p
 
-Base.eltype(::Type{Tweedie{T}}) where {T} = T
+Base.eltype(::Type{Tweedie{T}}) where {T} = float(T)
 
 #### Statistics
 
-mean(d::Tweedie) = d.μ
+mean(d::Tweedie) = float(d.μ)
```
nalimilan, line 89: `float` seems to make more sense to me here for as e.g. `rand(Tweedie{Int}(1, 2, 2))` generates float values and not integers.

src/univariate/continuous/tweedie.jl
devmotion: Historical reasons I assume. And it hasn't been removed yet to avoid having to tag a breaking release.

src/univariate/continuous/tweedie.jl
devmotion: Same here, it's just not removed yet since that would mean we'd have to tag a breaking release.

PR #2047, opened 2026-03-25
Add docstring content to docs website

abhro:
- Add docstrings to docs website
- Create docstring for `loglikelihood` method
- Fix reference to/of `fit(D, x)` method

### DIFF
--- a/docs/src/reshape.md
+++ b/docs/src/reshape.md
@@ -4,7 +4,6 @@ Distributions of array variates such as `MultivariateDistribution`s and
 `MatrixDistribution`s can be reshaped.
 
 ```@docs
-Distributions.ReshapedDistribution
 reshape
 vec
 ```
--- a/docs/src/reshape.md
+++ b/docs/src/reshape.md
@@ -4,6 +4,7 @@ Distributions of array variates such as `MultivariateDistribution`s and
 `MatrixDistribution`s can be reshaped.
 
 ```@docs
+Distributions.ReshapedDistribution
 reshape
 vec
 ```
--- a/src/univariates.jl
+++ b/src/univariates.jl
@@ -326,6 +326,16 @@ logpdf(d::UnivariateDistribution, x::Real)
 # extract value from array of zero dimension
 logpdf(d::UnivariateDistribution, x::AbstractArray{<:Real,0}) = logpdf(d, first(x))
 
+"""
+    loglikelihood(d::UnivariateDistribution, x)
+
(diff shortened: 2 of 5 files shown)


docs/src/reshape.md
```diff
@@ -4,6 +4,7 @@ Distributions of array variates such as `MultivariateDistribution`s and
 `MatrixDistribution`s can be reshaped.
 
 ```@docs
+Distributions.ReshapedDistribution
```
devmotion: This is internal, users should only use `reshape`.

```suggestion
```

abhro: The docstring for `reshape` does link to `ReshapedDistribution`, and I added this so that the cross-reference link has a target. Should I remove the `@ref` to `ReshapedDistribution` from the `reshape` docstring instead?

devmotion: Yes, probably it should be removed from the docstring. I assume the intention was to explain the default fallback but as mentioned in the docstring that one is not exported and its constructor shouldn't be used.

src/univariates.jl
```diff
@@ -326,6 +326,16 @@ logpdf(d::UnivariateDistribution, x::Real)
...
 
+"""
+    loglikelihood(d::UnivariateDistribution, x)
+
+The log-likelihood of distribution `d` with respect to `x`.
+
+Here, `x` can be any output of `rand(d, dims...)` and `rand!(d, x)`.
+For univariate distributions, `x` therfore must be a scalar or
+an array with an arbitrary number of dimensions.
+"""
```
devmotion, line 337: This is already covered by https://github.com/JuliaStats/Distributions.jl/blob/8c7dd0dab0524c736bb254bfa741bc03a1f5286c/src/common.jl#L450-L460

abhro: Hmm. I did basically copy it from that docstring, to cover the docstring requested in [univariate.md, line 76](https://github.com/JuliaStats/Distributions.jl/blob/196d79be2f21744a88c01295f2772f091244d557/docs/src/univariate.md?plain=1#L76). Should the signature for that method be replaced with the method defined in common.jl? It did seem weird that there was a docstring listing for a method that didn't have those specific types, but I wasn't sure if there was a specific documentation need that was served, with the implementation still being the same somehow.

devmotion: Hmm, no, the generic method is the relevant one in this case as well. I'm not sure what's the best way to handle this.

PR #2051, opened 2026-04-01
Avoid intermediate `string` creation in fallback when printing distribution names

devmotion:
See https://github.com/JuliaStats/Distributions.jl/pull/2050#issuecomment-4172319875


bvdmitri:
This change is breaking and removes the function that is quite useful, why not keep the original function for backward compatibility? The new implementation can simply create `IOBuffer`, write to it, and return a string from there.

devmotion:
This is an internal function, so the change is not breaking.

bvdmitri:
This is a useful function that can be made public and documented, why not?

devmotion:
Why should it be made public? There's already `string`, `repo`, `nameof` etc., IMO there's no motivation for this function apart from being used in a few internal implementations.

PR #2053, opened 2026-04-07
Fix `BinomialTPESampler` Stirling correction sign error

k88hudson:
I came across this when comparing different implementations of this algorithm and I noticed Distributions.js reproduces an error in the original paper. In step 5.3 of the BTPE algorithm (Kachitvichyanukul & Schmeiser, 1988), the acceptance bound adds all four Stirling terms, but this is incorrect: the corrections for `x1 = y+1` and `w = n-y+1` should be subtracted.

The fix/comment is noted [here](https://github.com/ampl/gsl/blob/fb0cfbc82b5e9e4d22444efce9241924f4aefb83/randist/binomial_tpe.c#L339-L361) in GSL. As the GSL authors mention, the impact is so small it's statistically infeasible to test, but I thought it would be worth putting up anyway.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.25.123"
+version = "0.25.124"
 
 [deps]
 AliasTables = "66dad0bd-aa9a-41b7-9441-69ab47430ed8"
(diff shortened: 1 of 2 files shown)


devmotion:
As GSL is GPL-licensed, we must not copy code from it.

---

I looked up the paper and redid the calculations. Indeed, the signs in the original paper are wrong.

k88hudson:
@devmotion Thank you for taking a look and for the the feedback. My collaborator @jasonasher also worked through the equations and we find the same problem. I updated the comment with potentially something more helpful (but if you have suggestions, I would be happy to change it). Please let me know if there is anything else I should do re: contributing (if you would prefer an issue filed for example)

devmotion:
No, looks good. Can you update the version number, then we can tag a new release with the fix when it's merged?

k88hudson:
Done!

devmotion (approved):

devmotion:
Thank you! I started the registration process, so a new release with the fix should be available soon: https://github.com/JuliaRegistries/General/pull/152849

PR #2055, opened 2026-04-09
Quantile bisect widen interval

### LINKED ISSUE #1869: Bug in quantile() for some special cases of a MixtureModel
# The problem
For rather exotic instances of a `MixtureModel` the `quantile()` method fails. The underlying reason is that the initial bisection interval (in `quantile_bisect()`) does not contain the solution. However, both ends of the interval are very close to the solution. The current implementation of `quantile_bisect()` checks for equality of the start and end of the interval. Changing this to approximately equal would fix this bug. 

Example
```julia
using Distributions
d = MixtureModel([Normal(0, 1), Normal(eps(), 1)], [0.999, 0.001])
quantile(d, 0.001)
```
gives as output
```julia
ERROR: ArgumentError: [-3.090232306167818, -3.0902323061678176] is not a valid bracketing interval for `quantile(d, 0.001)`
Stacktrace:
 [1] quantile_bisect(d::MixtureModel{Univariate, Continuous, Normal{Float64}, Categorical{Float64, Vector{Float64}}}, p::Float64, lx::Float64, rx::Float64)
   @ Distributions ~/.julia/packages/Distributions/ji8PW/src/quantilealgs.jl:21
 [2] quantile(d::MixtureModel{Univariate, Continuous, Normal{Float64}, Categorical{Float64, Vector{Float64}}}, p::Float64)
   @ Distributions ~/.julia/packages/Distributions/ji8PW/src/mixtures/mixturemodel.jl:447
 [3] top-level scope
   @ REPL[6]:1
```

I'm currently using
```julia
julia> versioninfo()
Julia Version 1.10.3
Commit 0b4590a5507 (2024-04-30 10:59 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 16 × 12th Gen Intel(R) Core(TM) i7-1260P
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, alderlake)
Threads: 1 default, 0 interactive, 1 GC (on 16 virtual cores)
```
and Distributions v0.25.109.

# The solution(?)
Changing line 9 of quantilealgs.jl 
https://github.com/JuliaStats/Distributions.jl/blob/b356da03a189d023cdb8467c61806a8a11dcb262/src/quantilealgs.jl#L8
to `if rx ≈ lx` seems to do the trick. I'm not sure whether this has bad side effects. However, after this change all tests in Distributions.jl still succeed.

An appropriate new test case to catch this bug would be
```julia
d = MixtureModel([Normal(0, 1), Normal(eps(), 1)], [0.999, 0.001])
@test quantile(d, 0.001) ≈ 0.001
```

I can submit this as a pull request if desired (would need to figure out how to do that).

arnaud-ma:
Fixes #1869 

## The issue

Specifically for mixture models, in this code
https://github.com/JuliaStats/Distributions.jl/blob/da420cde4fc0df074fa8c54b36bad70c46903a0c/src/mixtures/mixturemodel.jl#L446-L450

Let's call $F(x) = \text{cdf}(d, x)$ and $F_i(x) = \text{cdf}(\text{component}(d, i), x)$, same $F^{-1}$ and $F_i^{-1}$ for quantile. `quantile_bisect` requires that

$$
F(q_\text{min}) \leq p \leq F(q_\text{max})
$$

otherwise an error is thrown. Mathematically, in the code above, the condition is guaranteed. But if we take the definition of $q_\text{min}$ and $q_\text{max}$, the condition becomes

$$
F(\min_i F_i^{-1}(p)) \leq p \leq F(\max_i F_i^{-1}(p)) .
$$

And numerically, for some specific edge cases dealing with very small numbers, this condition is not satisfied. For example,

```julia-repl
julia> d = MixtureModel([Normal(0, 1), Normal(eps(), 1)], [0.5, 0.5])
MixtureModel{Normal{Float64}}(K = 2)
components[1] (prior = 0.5000): Normal{Float64}(μ=0.0, σ=1.0)
components[2] (prior = 0.5000): Normal{Float64}(μ=2.220446049250313e-16, σ=1.0)


julia> d1, d2 = components(d)
2-element Vector{Normal{Float64}}:
 Normal{Float64}(μ=0.0, σ=1.0)
 Normal{Float64}(μ=2.220446049250313e-16, σ=1.0)

julia> p = 0.001
0.001

julia> cdf(d, quantile(d1, p)), cdf(d, quantile(d2, p))
(0.000999999999999983, 0.0009999999999999853)

julia> quantile(d, p)
ERROR: ArgumentError: [-3.090232306167818, -3.0902323061678176] is not a valid bracketing interval for `quantile(d, 0.001)`
```

I found another example different than "almost identical" components, but still with the same issue:

```julia-repl
julia> d = MixtureModel(
           [Exponential(2.1656391815606704e-5), Exponential(79.77140291062331)],
           [8.83763312432469e-25, 1.0],
       )
MixtureModel{Exponential{Float64}}(K = 2)
components[1] (prior = 0.0000): Exponential{Float64}(θ=2.1656391815606704e-5)
components[2] (prior = 1.0000): Exponential{Float64}(θ=79.77140291062331)


julia> d1, d2 = components(d)
2-element Vector{Exponential{Float64}}:
 Exponential{Float64}(θ=2.1656391815606704e-5)
 Exponential{Float64}(θ=79.77140291062331)

julia> p = 0.1066425045830068
0.1066425045830068

julia> cdf(d1, quantile(d1, p)) <= p, cdf(d2, quantile(d2, p)) <= p
(true, true)

julia> quantile(d, p)
ERROR: ArgumentError: [2.442157681386963e-6, 8.995697253353159] is not a valid bracketing interval for `quantile(d, 0.1066425045830068)`
```

I did not found any issues such that $p < F(q_\text{min})$ though. It's always $p > F(q_\text{max})$. Maybe it is impossible with mixtures? 

## The fix

Because it seems specific to mixture models, we could simply make the interval wider in `quantile(d::UnivariateMixture{Continuous}, p::Real)` by adding some small number to `max_q` and subtracting some small number from `min_q`. But I think it would be better to make the `quantile_bisect` function more robust, by allowing it to widen the interval if the condition is not satisfied. It does not impact the old case at all, and it would fix the issue for mixture models, and potentially for other / future distributions with similar issues.

### DIFF
--- a/src/quantilealgs.jl
+++ b/src/quantilealgs.jl
@@ -1,8 +1,17 @@
 # Various algorithms for computing quantile
 
+"""
+quantile_bisect(d::ContinuousUnivariateDistribution, p::Real, lx::T, rx::T) where {T<:Real}
+
+Compute the quantile of `d` at probability `p` using the bisection method, starting from
+the bracketing interval `[lx, rx]`.
+
+In theory, the interval `[lx, rx]` should satisfy `cdf(d, lx) <= p <= cdf(d, rx)`.
+However, due to numerical issues, this condition may not hold. In such cases, the algorithm
+attempts to recover a valid bracketing interval by expanding the initial interval.
+"""
 function quantile_bisect(d::ContinuousUnivariateDistribution, p::Real, lx::T, rx::T) where {T<:Real}
     rx < lx && throw(ArgumentError("empty bracketing interval [$lx, $rx]"))
-
     # In some special cases, e.g. #1501, rx == lx`
     # If the distribution is degenerate the check below can fail, hence we skip it
     if rx == lx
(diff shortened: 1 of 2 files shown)


arnaud-ma:
For the missing coverage, as I said above, I tried but did not find any distribution that triggers  $p < F(q_\text{min})$

arnaud-ma:
I guess now that it is using Roots.jl the bug is "upstream". One day it will be solved!

PR #2062, opened 2026-05-28
Update to StatsFuns@2

devmotion:


Project.toml
```diff
@@ -56,7 +56,7 @@ StaticArrays = "1"
 Statistics = "1"
 StatsAPI = "1.6"
 StatsBase = "0.32, 0.33, 0.34"
-StatsFuns = "1.5.2"
```
andreasnoack, line 59: Is it necessary to drop support for 1? It doesn't look like it required any code changes.

devmotion: Not anymore, it did in the initial version - but these additional changes would require even more fixes on the StatsFuns side, so I postponed them.

Project.toml
```diff
@@ -56,7 +56,7 @@ StaticArrays = "1"
...
+StatsFuns = "2.0.1"
```
devmotion:
```suggestion
StatsFuns = "1.5.2, 2.0.1"
```

andreasnoack (approved):

devmotion:
The PR should not have been merged. It requires https://github.com/JuliaStats/StatsFuns.jl/pull/211. Or, more precisely, without it the version bound is pretty useless.

andreasnoack:
> The PR should not have been merged.

I don't think that was clear from the description. Wasn't it marked as ready?

PR #2064, opened 2026-06-08
Support global sparsity tracing through distribution constructors

devmotion:
## Motivation

`@check_args` validates a distribution's constructor arguments with comparisons such as `σ ≥ 0`. Under [SparseConnectivityTracer](https://github.com/adrhill/SparseConnectivityTracer.jl)'s **global** `TracerSparsityDetector`, those comparisons run on primal-free tracers and return a *tracer* rather than a `Bool`, so they cannot be used in the boolean context the checks require. As a result, the global sparsity pattern cannot currently be traced through *any* distribution construction — `jacobian_sparsity(x -> mean(Normal(x[1], x[2])), x, TracerSparsityDetector())` errors inside `check_args`.

## Change

Introduce an overloadable per-argument predicate `_arg_checkable(x)` (default `true`). `@check_args` now forwards the explicitly-named checked arguments to `check_args`, which skips the checks when any argument is not checkable:

```julia
_arg_checkable(@nospecialize(_)) = true

function check_args(f::F, check::Bool, args...) where {F}
    (check && all(_arg_checkable, args)) && f()
    nothing
end
```

A new weak-dependency extension `DistributionsSparseConnectivityTracerExt` provides the single overload:

```julia
Distributions._arg_checkable(::AbstractTracer) = false
```

`AbstractTracer` covers exactly the primal-free global tracers (`GradientTracer`/`HessianTracer`). Local sparsity detection (`TracerLocalSparsityDetector`) passes primal-carrying `Dual`s, which are `<: Real` (not `<: AbstractTracer`), so their comparisons evaluate to `Bool` as usual and the checks run unchanged — invalid arguments are still rejected under local tracing.

Ordinary numeric construction is unaffected (`args` are plain numbers, `_arg_checkable` is `true`).

## Tests

`test/sparseconnectivitytracer.jl` checks that global gradient and (block-off-diagonal) Hessian sparsity trace through `Normal` construction, and that local detection still enforces argument validity.

---

This pull request was prepared with the assistance of generative AI (Claude Code).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Distributions"
 uuid = "31c24e10-a181-5473-b8eb-7969acd0382f"
 authors = ["JuliaStats"]
-version = "0.25.125"
+version = "0.25.126"
 
 [deps]
 AliasTables = "66dad0bd-aa9a-41b7-9441-69ab47430ed8"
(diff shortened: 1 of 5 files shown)


andreasnoack (approved):
Looks good. Based on Claude's review assistance, I've filed https://github.com/JuliaStats/Distributions.jl/issues/2067 to keep track of possible follow-up work needed to support all distributions.

PR #2077, opened 2026-06-30
Initialize Dirichlet fitting without sample variable

### LINKED ISSUE #602: Dirichlet initialization for fit diverges when one of the pk is constant across samples

andreasnoack:
This avoids dividing by zero. The new initialization is based on the paper by Thomas Minka linked in the comment. This is an alternative to https://github.com/JuliaStats/Distributions.jl/pull/610 based on some of the ideas in the paper. It also avoids the extra allocation. Of course, at the cost of some special function evaluations.

Fixes #602

### DIFF
--- a/src/multivariate/dirichlet.jl
+++ b/src/multivariate/dirichlet.jl
@@ -224,33 +224,38 @@ end
 
 ## Initialization
 
-function _dirichlet_mle_init2(μ::Vector{Float64}, γ::Vector{Float64})
+# initialization based on mean and log-mean
+# Ref https://tminka.github.io/papers/dirichlet/minka-dirichlet.pdf
+function _dirichlet_mle_init2(μ::Vector{Float64}, logμ::Vector{Float64})
     K = length(μ)
 
-    α0 = 0.
-    for k = 1:K
-        μk = μ[k]
-        γk = γ[k]
-        ak = (μk - γk) / (γk - μk * μk)
-        α0 += ak
-    end
-    α0 /= K
+    # Minka eq 42
+    α₀ = (K - 1) / 2 / sum(μₖ * (log(μₖ) - logμₖ) for (μₖ, logμₖ) in zip(μ, logμ))
 
-    lmul!(α0, μ)
+    # Run five iterations of a fixed point equation for good starting values
+    # Note! Minka doesn't suggest the number of iterations
+    for i = 1:5
+        for k = 1:K
+            # Minka eq 9
+            μ[k] = invdigamma(digamma(α₀) + logμ[k])
+        end
+        α₀ = sum(μ)
+    end
+
+    return μ
 end
 
 function dirichlet_mle_init(P::AbstractMatrix{Float64})
     K = size(P, 1)
     n = size(P, 2)
 
-    μ = vec(sum(P, dims=2))       # E[p]
-    γ = vec(sum(abs2, P, dims=2)) # E[p^2]
+    μ = vec(sum(P, dims=2))         # E[p]
+    logμ = vec(sum(log, P, dims=2)) # E[log(p)]
 
-    c = 1.0 / n
-    μ .*= c
-    γ .*= c
+    μ ./= n
+    logμ ./= n
 
-    _dirichlet_mle_init2(μ, γ)
+    return _dirichlet_mle_init2(μ, logμ)
 end
 
 function dirichlet_mle_init(P::AbstractMatrix{Float64}, w::AbstractArray{Float64})
--- a/src/multivariate/dirichlet.jl
+++ b/src/multivariate/dirichlet.jl
@@ -257,18 +257,6 @@ function _dirichlet_mle_init(μ::Vector{Float64}, logμ::Vector{Float64})
     return α
 end
 
-function dirichlet_mle_init(P::AbstractMatrix{Float64})
-    μ = vec(mean(P, dims=2))                     # E[p]
-    logμ = mean_logp(suffstats(Dirichlet, P))    # E[log(p)]
-    return _dirichlet_mle_init(μ, logμ)
-end
-
-function dirichlet_mle_init(P::AbstractMatrix{Float64}, w::AbstractArray{Float64})
-    μ = P * vec(w) ./ sum(w)                     # E[p]
-    logμ = mean_logp(suffstats(Dirichlet, P, w)) # E[log(p)]
-    return _dirichlet_mle_init(μ, logμ)
-end
-
 ## Newton-Ralphson algorithm
 
 function fit_dirichlet!(elogp::Vector{Float64}, α::Vector{Float64};
--- a/src/multivariate/dirichlet.jl
+++ b/src/multivariate/dirichlet.jl
@@ -225,31 +225,34 @@ end
 ## Initialization
 
 # Compute a starting value for the Newton iterations in `fit_dirichlet!` based on
-# the sample mean `μ = E[p]` and the sample log-mean `logμ = E[log(p)]`.
+# the sample mean `μ = E[p]` and the mean of the log-samples `elogp = E[log(p)]`.
+# The starting value is written into `μ`.
 # Ref https://tminka.github.io/papers/dirichlet/minka-dirichlet.pdf
-function _dirichlet_mle_init(μ::Vector{Float64}, logμ::Vector{Float64})
+function _dirichlet_mle_init!(μ::Vector{Float64}, elogp::Vector{Float64})
     K = length(μ)
 
     # Initialize the precision α₀ with the moment estimate in Minka eq 42. In contrast
     # to the estimates in Minka eq 21 and 23, it remains finite when some components
-    # of the sample have zero variance (see issue #602). By Jensen's inequality, the
-    # denominator is non-negative, and it is zero only if all samples are identical,
-    # in which case the MLE doesn't exist.
-    s = sum(μₖ * (log(μₖ) - logμₖ) for (μₖ, logμₖ) in zip(μ, logμ))
-    s > 0 || throw(ArgumentError(
-        "the samples are all identical so the MLE of the Dirichlet distribution doesn't exist"))
-    α₀ = (K - 1) / 2 / s
+    # of the sample have zero variance (see issue #602). In exact arithmetic, the
+    # denominator is non-negative by Jensen's inequality and zero only if all samples
+    # are identical, in which case the MLE doesn't exist. However, due to roundoff,
+    # the computed denominator can have either sign for (nearly) identical samples,
+    # so the check below cannot detect all such degenerate samples.
+    s = sum(μₖ * (log(μₖ) - elogpₖ) for (μₖ, elogpₖ) in zip(μ, elogp))
+    isfinite(s) && s > 0 || throw(ArgumentError(
+        "the samples are (nearly) identical or contain zeros so a starting value couldn't be computed"))
+    α₀ = (K - 1) / (2 * s)
 
     # Refine with a few iterations of the fixed point iteration in Minka eq 9. The
     # result is no longer a moment estimate but a starting value for the Newton
     # iterations in `fit_dirichlet!`. Minka doesn't suggest a specific number of
     # iterations but five appear to be sufficient to reach the region where the
-    # Newton iterations converge quadratically. The vector μ isn't needed anymore
-    # so we reuse it for the refined values.
-    α = μ
+    # Newton iterations converge quadratically.
+    α = μ  # the mean isn't needed anymore so we overwrite it with the refined values
     for _ in 1:5
+        dgα₀ = digamma(α₀)
         for k in 1:K
-            α[k] = invdigamma(digamma(α₀) + logμ[k])
+            α[k] = invdigamma(dgα₀ + elogp[k])
         end
         α₀ = sum(α)
     end
@@ -296,33 +299,38 @@ function fit_dirichlet!(elogp::Vector{Float64}, α::Vector{Float64};
             b += gk * iq[k]
             iqs += iq[k]
 
-            agk = abs(gk)
-            if agk > gnorm
-                gnorm = agk
-            end
+            # In contrast to the branch `abs(gk) > gnorm ? abs(gk) : gnorm`, `max`
+            # propagates NaNs so a NaN gradient fails the convergence check below
+            # instead of being silently ignored
+            gnorm = max(gnorm, abs(gk))
         end
-        b /= (iz + iqs)
-
-        # update α
-
-        for k = 1:K
-            α[k] -= (g[k] - b) * iq[k]
-            if α[k] < 1.0e-12
-                α[k] = 1.0e-12
-            end
-        end
-        α0 = sum(α)
-
-        if debug
-            prev_objv = objv
-            objv = dot(α .- 1.0, elogp) + loggamma(α0) - sum(loggamma, α)
-            @printf("Iter %4d: objv = %.4e  ch = %.3e  gnorm = %.3e\n",
-                t, objv, objv - prev_objv, gnorm)
-        end
-
-        # determine convergence
-
+        # Determine convergence before updating α so that the returned iterate is the
+        # one that satisfied the convergence criterion. An additional Newton update
+        # from an already converged iterate with very large elements could produce
+        # non-finite values because the denominator of b, which approaches (K - 1)/2
+        # as the elements of α grow, is then dominated by cancellation errors
         converged = gnorm < tol
+
+        if !converged
+            b /= (iz + iqs)
+
+            # update α
+
+            for k = 1:K
+                α[k] -= (g[k] - b) * iq[k]
+                if α[k] < 1.0e-12
+                    α[k] = 1.0e-12
+                end
+            end
+            α0 = sum(α)
+
+            if debug
+                prev_objv = objv
+                objv = dot(α .- 1.0, elogp) + loggamma(α0) - sum(loggamma, α)
+                @printf("Iter %4d: objv = %.4e  ch = %.3e  gnorm = %.3e\n",
+                    t, objv, objv - prev_objv, gnorm)
+            end
+        end
     end
 
     if !converged
@@ -338,7 +346,7 @@ function fit_mle(::Type{T}, P::AbstractMatrix{Float64};
     debug::Bool=false) where {T<:Dirichlet}
 
     elogp = mean_logp(suffstats(T, P))
-    α = isempty(init) ? _dirichlet_mle_init(vec(mean(P, dims=2)), elogp) : init
+    α = isempty(init) ? _dirichlet_mle_init!(vec(mean(P, dims=2)), elogp) : init
     fit_dirichlet!(elogp, α; maxiter=maxiter, tol=tol, debug=debug)
 end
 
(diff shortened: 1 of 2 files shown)


devmotion:
AFAICT the PR mixes two approaches presented by Minka, the EM-like fixed-point iteration and the Newton algorithm. It's not immediately clear to me why one would want to combine those two approaches and switch after a seemingly arbitrary amount of 5 fixed-point iterations to the Newton algorithm. IMO the crucial and currently in a few cases problematic part is just the initialization of the algorithms. As Minka mentions as well, there's many different possible approaches for initializing the algorithms, e.g., based on method of moments (obviously, different choices of moments will lead to different initializations, Minka already discusses two different approaches) or approximations of the likelihood function, e.g., using Stirling's approximation.

Exploiting the moment equations $E\left[p_k\right] = \alpha_k / \alpha_0$ for $k = 1, \ldots, K$ (as currently done) seems natural. However, since one of these $K$ equations is redundant, it leaves the question what additional equation/constraint to include to obtain an equation system that can be solved for the $K$ parameters $\alpha_1, \ldots, \alpha_K$. Minka presents different approaches in eq. 21, 23 and 42. The approaches in eq. 21 and 23 break down if the sample variance of $p_1$ (eq 21) or any $p_k$ (eq 23) is zero. Eq. 42 doesn't suffer from this problem, so as suggested in this PR it would be a more stable approach. Alternatively, one could consider other moments such as

$$E\left[\sum_k p_k^2\right] = \frac{1 + \alpha_0 \sum_k {(\alpha_k / \alpha_0)}^2}{1 + \alpha_0}$$

which AFAICT would give moment estimates

$$\alpha_k = \hat{E}\left[p_k\right] \left(\frac{1 - \sum_{\tilde{k}} \hat{E}\left[p_{\tilde{k}}^2\right]}{\sum_{\tilde{k}} \hat{E}\left[p_{\tilde{k}}^2\right] - \sum_{\tilde{k}} {\hat{E}\left[p_{\tilde{k}}\right]}^2} \right) = \hat{E}[p_k] \frac{\sum_{\tilde{k}} \hat{E}\left[p_{\tilde{k}}(1-p_{\tilde{k}})\right]}{\sum_{\tilde{k}} \hat{var}\left[p_{\tilde{k}}\right]} $$

(seemingly a symmetrized version of eq. 21?). This would only break down if the sample variance of all $p_k$ is zero, ie., all samples are identical, in which case the estimation breaks down anyway.

So, taken together, my impression is: Using eq. 42 would give a less brittle initialization but one could also consider other approaches, e.g., based on different moments; but it's not clear to me why one would want to combine the EM-like fixed-point iteration with the Newton algorithm, using a different more stable initialization for the Newton algorithm alone (or the EM-like fixed-point iteration, if one prefers that approach) seems sufficient.

andreasnoack:
The problem is that the parameters must be positive, so the initialization has to be sufficiently good to ensure that the Newton step is feasible. The fixed-point iterations will always produce positive values, but it has linear convergence. I also tried the initialization suggested in https://www.sciencedirect.com/science/article/abs/pii/S0167947307002848, but then I can trigger a maxiter error by making the repeated value small enough.

devmotion:
To me it seems there's two separate problems you're facing and trying to address: the problematic initialization (the original issue reported in #602) and infeasible iterates of the Newton algorithm. Seemingly the first problem could be fixed by an initialization that doesn't break down if the sample variance of some $p_k$ is zero, e.g., by making use of eq. 42 in Minka's report, the estimate for $\alpha_0$ in the linked paper or possibly by considering different moments as outlined above. The second problem seems merely a consequence of applying (unconstrained) Newton method on a constrained space, and could maybe be approached by e.g. applying Newton method on the log space of $\log \alpha_k$ (AFAICT the efficient structure of the Hessian would be preserved) or using a penalty method.

andreasnoack:
The two issues are closely related. In theory, we could just initialize all the values at one, except for the risk of stepping to an infeasible argument. Hence, we need to find a sufficiently good initialization before switching to Newton's method and thereby ensure fast convergence through the concave objective function.

> Seemingly the first problem could be fixed by an initialization that doesn't break down if the sample variance of some $p_k$ is zero, e.g., by making use of eq. 42 in Minka's report, the estimate for $α_0$ in the linked paper or possibly by considering different moments as outlined above.

Yes, but as mentioned above, I could trigger maxiter issues with these initializers.

> ...e.g. applying Newton method on the log space of $log(α_k)$ (AFAICT the efficient structure of the Hessian would be preserved)

The global concavity of the objective function is not preserved under this reparametrization. We could try a constrained optimizer, but that would be a larger change than the current one.

devmotion:
To avoid the concavity issue, we could use natural gradient descent in log space instead of of Newton's method. In the standard parameter space, Newton's method is exactly the natural gradient method with step size 1 (the Hessian is constant with respect to the data and hence identical to the FIM). AFAICT this would correspond to iterations $\alpha \to \alpha \odot \exp.{\left(\gamma s(\alpha) ./ \alpha\right)}$ with step size $\gamma$ and $s(\alpha)$ being the step direction of the corresponding Newton iteration in non-log space $\alpha \to \alpha + s(\alpha)$.

andreasnoack:
I'm not sure I follow. If you change the parametrization, then the objective function is no longer globally concave. It might be possible to optimize it anyway with Fisher scoring aka natural gradient descent, but you'd lose the guarantees that the concavity provides.

devmotion:
I didn't remember the details and all my thinking after the break, so used Claude to get up on speed again. Its conclusion was

> On the substance, you're structurally right and Andreas is practically right

😄

devmotion:
I tried to reproduce the `maxiter` failures you mentioned to get a better understanding of what's going on, and AFAICT the fixed-point iterations do help but not for the reason you gave.

With

```julia
ε = 1e-10
X = [ε ε; 0.5(1 - ε) 0.3(1 - ε); 0.5(1 - ε) 0.7(1 - ε)]
```

eq. 42 alone stops at `maxiter = 25` with `gnorm ≈ 5.4` whereas eq. 42 followed by the five fixed-point iterations converges after 8 iterations, so the motivation is clear. However, in the failing run the clamping `α[k] = max(α[k], 1e-12)` in `fit_dirichlet!` is never triggered, and just increasing `maxiter` is sufficient, the iteration converges after 30 steps. This does not seem specific to the example: in ~1500 random problems ($K$ between 2 and 8, $\alpha_k$ spanning $10^{-3}$ to $10^3$) the Newton iteration initialized with moment estimates converged in every case I checked if `maxiter` was large enough (0 out of 275 problems with $n$ between 50 and 200 failed with `maxiter = 1000`), but the required number of iterations has a long right tail (median 6, 95% quantile 41, max 57), whereas with the initialization in this PR it is median 5, 95% quantile 7, max 8. On the other hand, the clamping is still triggered in about 4% of the random problems with the initialization in this PR.

So AFAICT what the fixed-point iterations buy is a smaller number of Newton iterations and not feasibility of the Newton steps, ie., they move the initial value into the region where the quadratic convergence of the Newton iteration kicks in, and that's what makes `maxiter = 25` sufficient. That's a reasonable thing to want, so I withdraw my objection against combining the two algorithms. But then I think it would be good to make it explicit in the code, the returned value is not a moment estimate anymore, so possibly the function and the comments should state that it computes a starting value by running a few iterations of the fixed-point algorithm. Hardcoding the 5 seems fine to me, `fit_mle` accepts `init` anyway if someone wants to control the starting value, but since Minka gives no guidance it would be good if the comment recorded what motivates the choice, eg. that a few iterations are sufficient to reach the region where the Newton iteration converges quadratically. Independently of this PR, it seems to me that `maxiter = 25` together with `tol = 1e-12` for the absolute sup-norm of the gradient, and the clamping without any line search or check of the objective function, are the actually fragile parts here.

One thing that has to be fixed in any case: the second argument of `_dirichlet_mle_init2` changed from $\gamma = \hat{E}[p^2]$ to $\log\mu = \hat{E}[\log p]$, but `dirichlet_mle_init(P, w)` still computes $\gamma$ and passes it. Both are `Vector{Float64}`, so there's no `MethodError`, only wrong results. E.g., for

```julia
X = [0.1 0.2 0.3; 0.5 0.3 0.4; 0.4 0.5 0.3]
w = ones(3)
```

eq. 42 returns a negative $\alpha_0 \approx -0.835$ (`digamma` of a negative non-integer is finite, so nothing errors), the fixed-point iterations run away, and the starting value ends up at $\approx [23.4, 26.4, 26.4]$ instead of $\approx [5.4, 11.1, 11.1]$. The weighted `fit_mle` then throws `ErrorException("No convergence after 25 (maxiter) iterations.")` even though it works on master. This is not caught by CI because the only test of the weighted `fit_mle` is commented out in test/multivariate/dirichlet.jl:121. Could we uncomment it here? A test with `w = fill(2.0, n)` would have caught it as well.

Some minor comments. The title and description say that the PR avoids dividing by zero, but the degenerate cases are still there, $n = 1$ and identical samples still divide by zero and the result only changes from `Inf` to `NaN`. What's fixed is the case where only some of the $p_k$ are constant, so possibly the description could be adjusted, and arguably these degenerate cases should throw a more informative error. Moreover, `fit_mle` now computes `sum(log, P, dims=2)` twice, once in `dirichlet_mle_init` and once in `suffstats`. If `suffstats` is computed first and `elogp` is passed to the initialization, then eq. 42 is basically free and only the `invdigamma` evaluations are added. The hard-coded values in the new test are fine IMO, I obtain the same numbers with an independent implementation.

Regarding the natural gradient suggestion, I don't want to block the PR on it, but I think we talked past each other. In what I suggested the step direction is still computed in the $\alpha$ parametrization, where the objective function is concave and hence $\nabla f \cdot s > 0$, and the log parametrization only affects the step itself ($\alpha \odot \exp{(\gamma s ./ \alpha)}$ coincides with $\alpha + \gamma s$ to first order), so the concavity of the objective function in $\alpha$ is not given up. Relatedly, AFAICT the Hessian is data-independent and equal to the negative FIM, so Newton's method, Fisher scoring and natural gradient with unit step size coincide in the $\alpha$ parametrization anyway. That said, I implemented it with an Armijo line search and it does not perform better than the plain Newton iteration, it stalls before reaching `gnorm < 1e-12` in ~20% of my random problems and has a worse tail, which I think just moves the problem to the stopping criterion mentioned above. So this should be a separate PR, if we want to pursue it at all.

And just for the record regarding the symmetrized moment estimator I mentioned above: AFAICT the reason why it is more stable is that $\sum_k (\hat{E}[p_k] - \hat{E}[p_k^2]) / \sum_k \hat{var}[p_k]$ is the variance-weighted mean of the per-component estimates of $\alpha_0$ on master (since $\hat{E}[p_k] - \hat{E}[p_k^2] = \hat{var}[p_k] \hat{\alpha}_0^{(k)}$) whereas master uses the unweighted mean, so components with small sample variance are downweighted instead of blowing up the estimate. But since it shows the same long tail of Newton iterations it doesn't address the point you care about, so eq. 42 seems fine to me.

andreasnoack:
Thanks for the thorough review! I've pushed a commit (and rebased on master) that should address your points:

- Fixed the weighted `dirichlet_mle_init`, which still passed `E[p^2]` where the initialization now expects `E[log(p)]`, and re-enabled the commented-out weighted `fit_mle` test. Also added a test that uniform weights reproduce the unweighted fit.
- Renamed `_dirichlet_mle_init2` to `_dirichlet_mle_init` and updated the comments to state that it returns a starting value for the Newton iterations rather than a moment estimate, and that the five fixed point iterations are there to reach the region of quadratic convergence.
- `fit_mle` now computes `suffstats` before the initialization and reuses the log-mean, so eq 42 only adds the `invdigamma` evaluations.
- When all samples are identical (including `n = 1`), the denominator in eq 42 is zero, so the initialization now throws an informative `ArgumentError` instead of producing NaNs. Added tests for both degenerate cases.

I've left `maxiter`, `tol`, and the clamping unchanged and agree that any changes there, as well as the natural gradient idea, belong in a separate PR.

andreasnoack:
I've also been speculating whether it is better to be structurally or practically right. I'm not sure.

andreasnoack:
Codecov flagged that the weighted `dirichlet_mle_init` was no longer covered: after moving the `suffstats` computation ahead of the initialization, `fit_mle` calls `_dirichlet_mle_init` directly and the `dirichlet_mle_init(P[, w])` wrappers had become dead code only referenced by a test. I've removed them and adjusted the test to use `_dirichlet_mle_init` instead.

devmotion:
AFAICT the new `s > 0` guard for the degenerate case doesn't hold up to roundoff. For identical samples $s$ is mathematically zero, but it is computed as the difference between the log of a summed mean and the mean of logs, so its sign is arbitrary:

https://github.com/JuliaStats/Distributions.jl/blob/26512859ce978d3887ca0035a241f8376fe1b00a/src/multivariate/dirichlet.jl#L236-L243

With `P = repeat([0.1, 0.5, 0.4], 1, n)` the `ArgumentError` is thrown for $n \in \{1,2,3,4,6,7,8,9,11,12\}$, but $n = 5$ returns $\alpha \approx [1.04 \cdot 10^{16}, 5.22 \cdot 10^{16}, 4.18 \cdot 10^{16}]$ and $n = 10$ returns $[\infty, \infty, \infty]$, in both cases reported as a successful fit. In ~4000 random problems with exactly identical samples ($K$ between 2 and 8, $n$ between 2 and 30) the guard is bypassed in 41%, and the result is then either $\alpha \sim 10^{14} \ldots 10^{16}$ or every component clamped to `1e-12`. Which $n$ happen to work isn't predictable from the input, and the two new tests use $n = 2$ and $n = 1$, which do:

https://github.com/JuliaStats/Distributions.jl/blob/26512859ce978d3887ca0035a241f8376fe1b00a/test/multivariate/dirichlet.jl#L180-L183

This is not a regression, master returns `[NaN, NaN, NaN]` for all of these, but IMO we shouldn't claim the guarantee in the comment and the PR description if it holds for ~59% of the inputs. Could we loop over `n in 1:12` in the test and assert that `fit_mle` either throws or returns a finite positive $\alpha$?

The reason these come back as converged rather than as an error is that `fit_dirichlet!` drops NaN gradients, since `abs(NaN) > gnorm` is `false`:

https://github.com/JuliaStats/Distributions.jl/blob/26512859ce978d3887ca0035a241f8376fe1b00a/src/multivariate/dirichlet.jl#L298-L303

so `gnorm` stays `0.0` and `gnorm < tol` succeeds. `gnorm = max(gnorm, abs(gk))` is sufficient, since `max` propagates NaN in Julia (`max(0.0, NaN) === NaN`) and the existing `ErrorException` then fires. I said above that `fit_dirichlet!` should be a separate PR and I still think so for `maxiter`, `tol` and the clamping, but this one seems an unconditional bug to me, and it is what turns the cases above into silently wrong results, so I'd include it here.

Relatedly, a zero entry in `P` now returns `NaN` instead of throwing. For `P = [0.0 0.2 0.3; 0.5 0.3 0.3; 0.5 0.5 0.4]` master throws `No convergence after 25 (maxiter) iterations` whereas the PR returns `Dirichlet([NaN, NaN, NaN])`: $\log\mu_k = -\infty$ gives $s = \infty$, which passes `s > 0`, so $\alpha_0 = 0$, `digamma(0) = -Inf` and `invdigamma(-Inf) = NaN`. The input is invalid either way, but `isfinite(s) && s > 0` would keep it an error.

Some minor comments. `digamma(α₀)` is loop-invariant in the inner loop:

https://github.com/JuliaStats/Distributions.jl/blob/26512859ce978d3887ca0035a241f8376fe1b00a/src/multivariate/dirichlet.jl#L249-L256

`digamma` is neither inlined nor marked pure, so AFAICT it is not hoisted; pulling it out of the loop over $k$ gives bit-identical results and saves 7% ($K = 3$) to 17% ($K = 10$) of the initialization, and it also makes visible that eq 9 updates $\alpha_0$ only once per sweep.

The second argument is named `logμ` and the comment calls it the sample log-mean, but it is the mean of the logs, whereas `log(μₖ)` one character away on the same line is the log of the mean. Both call sites already call it `elogp`, consistent with `mean_logp` and with `fit_dirichlet!`. Since the bug in the previous round was exactly this kind of mixup, could we use `elogp` here as well? In the same vein, `α = μ` aliases and overwrites the caller's vector, which the inline comment explains but the name doesn't, so either `_dirichlet_mle_init!` or `α = similar(μ)` seems preferable to me. The extra allocation should be unmeasurable next to $5K$ `invdigamma` calls, each of which runs up to 25 digamma and trigamma evaluations itself.

`sum(w)` here duplicates the `tw` that `suffstats` has already accumulated, in the same spirit as the `sum(log, P)` de-duplication:

https://github.com/JuliaStats/Distributions.jl/blob/26512859ce978d3887ca0035a241f8376fe1b00a/src/multivariate/dirichlet.jl#L352-L356

`(K - 1) / (2 * s)` reads better than `(K - 1) / 2 / s`. And the error message says the samples are all identical, but it also fires for samples that are merely close, e.g. `[0.1 0.1+1e-9; 0.5 0.5; 0.4 0.4-1e-9]`.

For the record, on the well-posed problems the initialization is a clear improvement: in 600 random problems ($K$ between 2 and 8, $\alpha_k$ spanning $10^{-3}$ to $10^3$, $n$ between 50 and 200) I get 0 failures with this PR against 18 on master, the number of Newton iterations is median 5, 95% quantile 6, max 7 against 6, 8 and 12 on master, and where both converge the estimates agree to a relative difference of $4 \cdot 10^{-14}$.

andreasnoack:
Thanks for another careful round — all points should now be addressed: `gnorm` is accumulated with `max` so NaN gradients propagate, the guard is `isfinite(s) && s > 0`, the comment/error no longer claim reliable detection of identical samples, the test loops over `n in 1:12` (plus a case with zeros), and the minor points (`elogp` naming, `_dirichlet_mle_init!`, hoisted `digamma(α₀)`, reusing `tw`, `(K - 1) / (2 * s)`) are in.

One additional finding: the NaN fix turned out to be necessary but not sufficient. For `repeat([0.1, 0.5, 0.4], 1, 10)` the fit still returned `[Inf, Inf, Inf]` as converged without any NaN entering `gnorm`: the loop applied a Newton update *before* checking convergence, and from an already-converged iterate with α ~ 1e16 that extra step overflows because the denominator of `b` — mathematically approaching `(K - 1)/2` — is wiped out by cancellation at that scale. The loop now checks `gnorm < tol` before updating, so the returned iterate is the one that satisfied the criterion. Healthy fits just skip one redundant final step; the hardcoded test values change only within the gradient tolerance.

PR #2078, opened 2026-07-03
Fix `fit_mle` dispatch for instantiated/concrete MvNormal aliases

### LINKED ISSUE #2074: fit/fit_mle for DiagNormal/IsoNormal fails: "suffstats is not implemented"
Fitting a diagonal multivariate normal (DiagNormal/IsoNormal) via fit/fit_mle errors because suffstats is not defined for that type. Both the weighted and unweighted forms fail, and full-covariance MvNormal works fine.

MWE
---
```julia
using Distributions, LinearAlgebra

d = MvNormal(zeros(3), Diagonal(ones(3)))   # a DiagNormal
# d = MvNormal(zeros(3), I) # an IsoNormal
typeof(d) 

fit(typeof(d), randn(3, 50))                 # ERROR
```

Error
---
```
ERROR: suffstats is not implemented for (IsoNormal{Float64, PDMats.ScalMat{Float64}, Vector{Float64}}, Matrix{Float64}).
Stacktrace:
 [1] error(s::String)
   @ Base .\error.jl:44
 [2] suffstats(dt::Type{IsoNormal{Float64, PDMats.ScalMat{Float64}, Vector{Float64}}}, xs::Matrix{Float64})
   @ Distributions C:\Users\senne\.julia\packages\Distributions\XAng5\src\genericfit.jl:5
 [3] fit_mle(dt::Type{IsoNormal{Float64, PDMats.ScalMat{Float64}, Vector{Float64}}}, x::Matrix{Float64})
   @ Distributions C:\Users\senne\.julia\packages\Distributions\XAng5\src\genericfit.jl:30
 [4] fit(dt::Type{IsoNormal{Float64, PDMats.ScalMat{Float64}, Vector{Float64}}}, x::Matrix{Float64})
   @ Distributions C:\Users\senne\.julia\packages\Distributions\XAng5\src\genericfit.jl:46
 [5] top-level scope
   @ REPL[12]:1
```

Version Info
---
Julia: 1.12.6
Distributions 0.25.129

rsenne:
Fixes #2074.

This widens the 7 `fit_mle` signatures to `D::Type{<:...}` so the aliases are
accepted in any form (bare, instantiated, or concrete member), restoring the
pre-#2073 behavior. Added tests cover the instantiated and concrete-type paths,
which the existing bare-alias tests didn't exercise.

### DIFF
--- a/src/multivariate/mvnormal.jl
+++ b/src/multivariate/mvnormal.jl
@@ -397,9 +397,9 @@ fit_mle(D::Type{MvNormal}, ss::MvNormalStats) = fit_mle(FullNormal, ss)
 fit_mle(D::Type{MvNormal}, x::AbstractMatrix{Float64}) = fit_mle(FullNormal, x)
 fit_mle(D::Type{MvNormal}, x::AbstractMatrix{Float64}, w::AbstractArray{Float64}) = fit_mle(FullNormal, x, w)
 
-fit_mle(D::Type{FullNormal}, ss::MvNormalStats) = MvNormal(ss.m, ss.s2 * inv(ss.tw))
+fit_mle(D::Type{<:FullNormal}, ss::MvNormalStats) = MvNormal(ss.m, ss.s2 * inv(ss.tw))
 
-function fit_mle(D::Type{FullNormal}, x::AbstractMatrix{Float64})
+function fit_mle(D::Type{<:FullNormal}, x::AbstractMatrix{Float64})
     n = size(x, 2)
     mu = vec(mean(x, dims=2))
     z = x .- mu
@@ -408,7 +408,7 @@ function fit_mle(D::Type{FullNormal}, x::AbstractMatrix{Float64})
     MvNormal(mu, PDMat(C))
 end
 
-function fit_mle(D::Type{FullNormal}, x::AbstractMatrix{Float64}, w::AbstractVector)
+function fit_mle(D::Type{<:FullNormal}, x::AbstractMatrix{Float64}, w::AbstractVector)
     m = size(x, 1)
     n = size(x, 2)
     length(w) == n || throw(DimensionMismatch("Inconsistent argument dimensions"))
@@ -428,7 +428,7 @@ function fit_mle(D::Type{FullNormal}, x::AbstractMatrix{Float64}, w::AbstractVec
     MvNormal(mu, PDMat(C))
 end
 
-function fit_mle(D::Type{DiagNormal}, x::AbstractMatrix{Float64})
+function fit_mle(D::Type{<:DiagNormal}, x::AbstractMatrix{Float64})
     m = size(x, 1)
     n = size(x, 2)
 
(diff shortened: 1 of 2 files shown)


devmotion:
I'm not even sure this should be considered a bug and be "fixed". The loose `fit_mle` methods that accept arbitrarily parameterized types exist for historical reasons presumably but as noted in the documentation they are completely unfaithful as generally you won't obtain a distribution of the provided type. Compare this eg with `convert(T, x)` which is expected to return an instance of type `T`. So IMO one should only ever use - and arguably only support - `fit_mle(Normal, xs)`, `fit_mle(Beta, xs)`, `fit_mle(IsoNormal, xs)` etc because these actually return distributions of type `Normal`, `Beta`, `IsoNormal` etc. but not `fit_mle(Normal{Float32}, xs)`, `fit_mle(Beta{Float64}, xs)` or more generally `fit_mle(typeof(d), xs)` as these generally do not return distributions of type `Normal{Float32}`, `Beta{Float64}` or `typeof(d)`.

To summarize, I think the better as more faithful approach is to not support `fit_mle(::Type{<:IsoNormal}, ...)` etc. but only `fit_mle(::Type{IsoNormal}, ...)` etc.

rsenne:
I agree on the long-term design: `fit_mle(Normal{Float32}, xs)` returning a `Float64` distribution is unfaithful--something this PR admittedly overlooks--and given the current design the API probably shouldn't encourage passing fully-parameterized types.

But I'd like to separate that design question from what I think is the real issue: compatibility. Generic code that refits distributions with `fit(typeof(d), data)` worked before `0.25.129`, at least in the `Float64` case. It now errors, which is a behavioral regression and the error message (`suffstats is not implemented`) doesn't make the cause obvious.

So I think the best middle ground is to reinstate the previous behavior but emit a deprecation warning, which addresses your concern while not breaking existing code in a patch release. That restores what worked, signals the intended direction, and gives downstream users a clear message instead of a cryptic error.

rsenne:
Alternatively --I took a crack at making the fitting path [type-generic](https://github.com/rsenne/Distributions.jl/blob/generic-mvnormal-fit/src/multivariate/mvnormal.jl). Fitting now works for any real `eltype` instead of only `Float64`, and returns a distribution matching the input's element type -- which I think fits what you wanted on the faithfulness side: `fit(typeof(d), data)` gives you back d's type rather than coercing to `Float64`.

Kept the `Float64` `BLAS.syrk` path as-is; the generic methods are just fallbacks for the eltypes that used to error. Full suite passes locally. So i think this would preserve the old behavior.

As a note I also did something similar for the [univariate case](https://github.com/rsenne/Distributions.jl/blob/generic-normal-fit/src/univariate/continuous/normal.jl).

rsenne:
Any further thoughts on this @devmotion?

dmetivie:
On my side this change also [broke](https://github.com/dmetivie/ExpectationMaximization.jl/actions/runs/31718115039/job/94507904313#step:5:121) the ExpectationMaximization.jl package for Multivariate Normal. Indeed it uses a lot `fit(typeof(d), data)` see this [page](https://dmetivie.github.io/ExpectationMaximization.jl/dev/fit_mle/) and related issue for context.
Hence the fix proposed by @rsenne seems very much welcomed.

PR #2084, opened 2026-07-24
Fix typos in "Create New Samplers and Distributions" docs

s-celles:
Reported on Discourse: a user copy-pasted `DiscreteUnivarateDistribution` from the documentation and hit an `UndefVarError`. https://discourse.julialang.org/t/creating-a-new-distibution-with-distributions-jl/138436

- `DiscreteUnivarateDistribution` -> `DiscreteUnivariateDistribution`
- `DiscreteMultivarateDistribution` -> `DiscreteMultivariateDistribution`
- Matrix-variate sampler section said "multivariate sampler" and `Sampleable{Multivariate,S}`; it should be `Sampleable{Matrixvariate,S}`
- "as wells as" -> "as well as"

Assisted-by: AI

PR #2085, opened 2026-07-24
Fix typo in docs and add `@ref`s

asinghvi17:
xref https://discourse.julialang.org/t/creating-a-new-distibution-with-distributions-jl/138436


devmotion:
Closed in favour of #2084. Moreover, there's no target for the `@ref`s.

PR #2088, opened 2026-09-06
Use BTRS instead of BTPE for binomial sampling

nifets:
Replaces the BinomialTPE sampler (Kachitvichyanukul & Schmeiser 1988) with BTRS (Hörmann 1993) when `n*min(p,1-p) > 10`. 

| n | p | BTPE | BTRS | speedup |
|---|---|---|---|---|
| 25 | 0.45 | 76.8 ns | 48.4 ns | 1.59x |
| 40 | 0.50 | 73.3 ns | 46.2 ns | 1.58x |
| 100 | 0.40 | 62.5 ns | 38.3 ns | 1.63x |
| 300 | 0.60 | 44.2 ns | 30.4 ns | 1.45x |
| 1000 | 0.30 | 37.2 ns | 25.0 ns | 1.49x |
| 2000 | 0.02 | 63.7 ns | 32.8 ns | 1.94x |
| 10000 | 0.37 | 29.6 ns | 20.9 ns | 1.42x |
| 100000 | 0.10 | 29.1 ns | 20.6 ns | 1.41x |
| 1000000 | 0.10 | 28.2 ns | 20.2 ns | 1.40x |

BTRS is also what JAX (`jax/_src/random/core.py`), TensorFlow (`random_binomial_op.cc`) and PyTorch (`ATen/native/Distributions.h`) use. 

When `n*min(p,1-p) > 10` is not satisfied, the sampler is biased, so we throw an error. This should never happen via `rand(::Binomial)` which explicitly branches on that condition.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -39,7 +39,7 @@ jobs:
       - uses: julia-actions/cache@v3
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
-      - run: 'using Pkg; Pkg.instantiate()'
+      - run: 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
         shell: julia --color=yes --project=perf {0}
       - run: julia --color=yes --project=perf perf/samplers.jl
       - uses: julia-actions/julia-processcoverage@v1
--- a/src/samplers/binomial.jl
+++ b/src/samplers/binomial.jl
@@ -265,21 +265,41 @@ function BinomialTRSSampler(n::Int, prob::Float64)
     BinomialTRSSampler(comp, n, a, b, c, v_r, p/q, α, m)
 end
 
-function rand(rng::AbstractRNG, s::BinomialTRSSampler)
-    (; comp, n, a, b, r, m) = s
+struct BinomialTRSBatchSampler <: Sampleable{Univariate,Discrete}
+    s::BinomialTRSSampler
+    ub_m::Float64
+end
+
+function trs_ub_m(s::BinomialTRSSampler)
+    (; n, r, m) = s
+    return (m + 0.5) * log((m + 1) / (r * (n - m + 1))) +
+           (n + 1) * log(n - m + 1.0) +
+           lstirling_asym(m + 1) + lstirling_asym(n - m + 1)
+end
+trs_ub_m(s::BinomialTRSBatchSampler) = s.ub_m
+
+trs_sampler(s::BinomialTRSSampler) = s
+trs_sampler(s::BinomialTRSBatchSampler) = s.s
+
+function BinomialTRSBatchSampler(n::Int, prob::Float64)
+    s = BinomialTRSSampler(n, prob)
+    return BinomialTRSBatchSampler(s, trs_ub_m(s))
+end
+
+function rand(rng::AbstractRNG, s::Union{BinomialTRSSampler,BinomialTRSBatchSampler})
+    t = trs_sampler(s)
+    (; comp, n, a, b, r) = t
     while true
         u = rand(rng) - 0.5
         v = rand(rng)
         us = 0.5 - abs(u)
-        kf = (2 * a / us + b) * u + s.c
+        kf = (2 * a / us + b) * u + t.c
         (kf < 0.0 || kf >= n + 1) && continue
         k = floor(Int, kf)
-        (us >= 0.07 && v <= s.v_r) && return comp ? n - k : k
-        v = log(v * s.α / (a / (us * us) + b))
-        ub = (m + 0.5) * log((m + 1) / (r * (n - m + 1))) +
-             (n + 1) * log((n - m + 1) / (n - k + 1)) +
-             (k + 0.5) * log(r * (n - k + 1) / (k + 1)) +
-             lstirling_asym(m + 1) + lstirling_asym(n - m + 1) -
+        (us >= 0.07 && v <= t.v_r) && return comp ? n - k : k
+        v = log(v * t.α / (a / (us * us) + b))
+        ub = trs_ub_m(s) - (n + 1) * log(n - k + 1) +
+             (k + 0.5) * log(r * (n - k + 1) / (k + 1)) -
              lstirling_asym(k + 1) - lstirling_asym(n - k + 1)
         v <= ub && return comp ? n - k : k
     end
--- a/src/samplers/binomial.jl
+++ b/src/samplers/binomial.jl
@@ -228,6 +228,74 @@ function rand(rng::AbstractRNG, s::BinomialTPESampler)
     (s.comp ? s.n - y : y)::Int
 end
 
+#=
+BTRS algorithm from:
+
+W. Hörmann
+"The generation of binomial random variates"
+Journal of Statistical Computation and Simulation, 46(1-2):101-110
+doi:10.1080/00949659308811496
+=#
+# Valid only for n * min(p, 1-p) >= 10
+struct BinomialTRSSampler{T <: AbstractFloat} <: Sampleable{Univariate,Discrete}
+    n::Int
+    a::T
+    b::T
+    c::T
+    v_r::T
+    r::T
+    α::T
+    m::Int
+end
+
+function BinomialTRSSampler{T}(n::Int, prob::Real) where {T<:AbstractFloat}
+    p = T(prob)
+    q = one(T) - p
+    n * min(p, q) >= 10 || throw(ArgumentError("BinomialTRSSampler requires n * min(p, 1-p) >= 10"))
+    spq = sqrt(T(n) * p * q)
+    b = T(1.15) + T(2.53) * spq
+    a = T(-0.0873) + T(0.0248) * b + T(0.01) * p
+    c = T(n) * p + T(0.5)
+    v_r = T(0.92) - T(4.2) / b
+    α = (T(2.83) + T(5.1) / b) * spq
+    m = floor(Int, (T(n) + one(T)) * p)
+    BinomialTRSSampler{T}(n, a, b, c, v_r, p/q, α, m)
+end
+
+
+BinomialTRSSampler(n::Int, prob::Float64) = BinomialTRSSampler{Float64}(n, prob)
+
+@inline function binom_stirling_tail(::Type{T}, k::Int) where {T<:AbstractFloat}
+    if k < 10
+        tbl = (0.0810614667953272, 0.0413406959554093, 0.0276779256849983,
+               0.0207906721037650, 0.0166446911898211, 0.0138761288230707,
+               0.0118967099458917, 0.0104112652619720, 0.00925546218271273,
+               0.00833056343336287)
+        return T(@inbounds tbl[k + 1])
+    end
+    kp1sq = (T(k) + one(T))^2
+    return (T(1//12) - (T(1//360) - T(1//1260) / kp1sq) / kp1sq) / (T(k) + one(T))
+end
+
+function rand(rng::AbstractRNG, s::BinomialTRSSampler{T}) where {T}
+    (; n, a, b, r, m) = s
+    while true
+        u = rand(rng, T) - T(1/2)
+        v = rand(rng, T)
+        us = T(1/2) - abs(u)
+        kf = (T(2) * a / us + b) * u + s.c
+        (kf < zero(T) || kf > T(n)) && continue
+        k = floor(Int, kf)
+        (us >= T(0.07) && v <= s.v_r) && return k
+        v = log(v * s.α / (a / (us * us) + b))
+        ub = (T(m) + T(1/2)) * log((T(m) + one(T)) / (r * T(n-m+1))) +
+             (T(n) + one(T)) * log(T(n-m+1) / T(n-k+1)) +
+             (T(k) + T(1/2)) * log(r*T(n-k+1) / (T(k) + one(T))) +
+             binom_stirling_tail(T, m) + binom_stirling_tail(T, n-m) -
+             binom_stirling_tail(T, k) - binom_stirling_tail(T, n-k)
+        v <= ub && return k
+    end
+end
 
 # Constructing an alias table by directly computing the probability vector
 #
--- a/src/samplers/binomial.jl
+++ b/src/samplers/binomial.jl
@@ -235,65 +235,53 @@ W. Hörmann
 "The generation of binomial random variates"
 Journal of Statistical Computation and Simulation, 46(1-2):101-110
 doi:10.1080/00949659308811496
+
+Valid only for n * min(p, 1-p) >= 10
 =#
-# Valid only for n * min(p, 1-p) >= 10
-struct BinomialTRSSampler{T <: AbstractFloat} <: Sampleable{Univariate,Discrete}
+struct BinomialTRSSampler <: Sampleable{Univariate,Discrete}
+    comp::Bool
     n::Int
-    a::T
-    b::T
-    c::T
-    v_r::T
-    r::T
-    α::T
-    m::Int
+    a::Float64
+    b::Float64
+    c::Float64   # n*p + 0.5
+    v_r::Float64
+    r::Float64   # p / q
+    α::Float64
+    m::Int       # floor((n+1)*p)
 end
 
-function BinomialTRSSampler{T}(n::Int, prob::Real) where {T<:AbstractFloat}
-    p = T(prob)
-    q = one(T) - p
-    n * min(p, q) >= 10 || throw(ArgumentError("BinomialTRSSampler requires n * min(p, 1-p) >= 10"))
-    spq = sqrt(T(n) * p * q)
-    b = T(1.15) + T(2.53) * spq
-    a = T(-0.0873) + T(0.0248) * b + T(0.01) * p
-    c = T(n) * p + T(0.5)
-    v_r = T(0.92) - T(4.2) / b
-    α = (T(2.83) + T(5.1) / b) * spq
-    m = floor(Int, (T(n) + one(T)) * p)
-    BinomialTRSSampler{T}(n, a, b, c, v_r, p/q, α, m)
+function BinomialTRSSampler(n::Int, prob::Float64)
+    comp = prob > 0.5
+    p = comp ? 1.0 - prob : prob
+    q = 1.0 - p
+    n * p >= 10 || throw(ArgumentError("BinomialTRSSampler requires n * min(prob, 1-prob) >= 10"))
+    spq = sqrt(n * p * q)
+    b = 1.15 + 2.53 * spq
+    a = -0.0873 + 0.0248 * b + 0.01 * p
+    c = n * p + 0.5
+    v_r = 0.92 - 4.2 / b
+    α = (2.83 + 5.1 / b) * spq
+    m = floor(Int, (n + 1) * p)
+    BinomialTRSSampler(comp, n, a, b, c, v_r, p/q, α, m)
 end
 
-
-BinomialTRSSampler(n::Int, prob::Float64) = BinomialTRSSampler{Float64}(n, prob)
-
-@inline function binom_stirling_tail(::Type{T}, k::Int) where {T<:AbstractFloat}
-    if k < 10
-        tbl = (0.0810614667953272, 0.0413406959554093, 0.0276779256849983,
-               0.0207906721037650, 0.0166446911898211, 0.0138761288230707,
-               0.0118967099458917, 0.0104112652619720, 0.00925546218271273,
-               0.00833056343336287)
-        return T(@inbounds tbl[k + 1])
-    end
-    kp1sq = (T(k) + one(T))^2
-    return (T(1//12) - (T(1//360) - T(1//1260) / kp1sq) / kp1sq) / (T(k) + one(T))
-end
-
-function rand(rng::AbstractRNG, s::BinomialTRSSampler{T}) where {T}
-    (; n, a, b, r, m) = s
+function rand(rng::AbstractRNG, s::BinomialTRSSampler)
+    (; comp, n, a, b, r, m) = s
     while true
-        u = rand(rng, T) - T(1/2)
-        v = rand(rng, T)
-        us = T(1/2) - abs(u)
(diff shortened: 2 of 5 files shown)


src/samplers/binomial.jl
```diff
@@ -228,6 +228,74 @@ function rand(rng::AbstractRNG, s::BinomialTPESampler)
...
+end
+
+function rand(rng::AbstractRNG, s::BinomialTRSSampler{T}) where {T}
+    (; n, a, b, r, m) = s
+    while true
+        u = rand(rng, T) - T(1/2)
+        v = rand(rng, T)
+        us = T(1/2) - abs(u)
+        kf = (T(2) * a / us + b) * u + s.c
+        (kf < zero(T) || kf > T(n)) && continue
```
devmotion: **Blocking: `k == n` is unreachable.**

`k == n` can only come from `kf ∈ [n, n+1)`, and this check discards all of `(n, n+1)`, so `k == n` is produced only when `kf` is exactly `n`. Hörmann's paper and TF/JAX/PyTorch all floor *first* and then reject on `k < 0 || k > n`, which admits that interval. The low end is unaffected (`kf < 0` ⟺ `floor(kf) < 0`), so the asymmetry looks unintended.

Measured on a verbatim transcription of this code, `n = 20`, `p = 0.5`, 2×10⁸ draws:

| version | `count(k == n)` | expected |
|---|---|---|
| this PR | 0 | 190.7 |
| floor-then-check | 195 | 190.7 |

User-visible: `rand(Binomial(21, 0.5))` routes here (`r*n = 10.5 > 10`) and can never return `21`, though `P(X = 21) = 4.8e-7`; symmetrically `rand(Binomial(21, 0.51))` can never return `0`. Worst case over the sampler's valid domain is `(1 - 10/n)^n → e^-10 ≈ 4.5e-5` for direct sampler use with large `p`.

Please keep the check on the float rather than porting TF's floor-then-compare literally: `floor(Int, -Inf)` throws `InexactError` in Julia, and `kf` is genuinely `-Inf` whenever `rand` returns `0.0` (also `floor(Int, 1e300)` throws, reachable at tiny `us`).

```suggestion
        (kf < zero(T) || kf >= T(n) + one(T)) && continue
```

I confirmed this restores `count(k == n) = 195` in the experiment above.

test/samplers.jl
```diff
@@ -46,6 +47,7 @@ import Distributions:
     @testset "Binomial: $S" for (S, paramlst) in [
             (BinomialGeomSampler, [(0, 0.4), (0, 0.6), (5, 0.0), (5, 1.0), (1, 0.2), (1, 0.8), (3, 0.4), (4, 0.6)]),
             (BinomialTPESampler, [(40, 0.5), (100, 0.4), (300, 0.6)]),
+            (BinomialTRSSampler, [(40, 0.5), (100, 0.4), (300, 0.6), (25, 0.45), (2000, 0.02)]),
```
devmotion: **Blocking companion: no added case can see the endpoint.**

`P(X = n)` is 9.1e-13 for `(40, 0.5)`, 1.3e-9 for `(25, 0.45)`, 1.6e-40 for `(100, 0.4)`, and effectively zero for `(300, 0.6)` and `(2000, 0.02)`. With `n_tsamples = 10⁶`, `test_samples` has no power against the bug on `src/samplers/binomial.jl:287`.

`(20, 0.5)` is the useful addition — it is simultaneously the exact `n*min(p, 1-p) == 10` boundary and the parameterisation where the endpoint carries the most mass:

```suggestion
            (BinomialTRSSampler, [(20, 0.5), (40, 0.5), (100, 0.4), (300, 0.6), (25, 0.45), (2000, 0.02)]),
```

Also worth adding, outside the `test_samples` loop:

```julia
@test_throws ArgumentError BinomialTRSSampler(19, 0.5)
```

and a direct check that the endpoint is reachable at all, e.g. that `n` appears among a few million draws at `(20, 0.5)`, since a bin-by-bin `test_samples` never will.

src/samplers/binomial.jl
```diff
@@ -228,6 +228,74 @@ function rand(rng::AbstractRNG, s::BinomialTPESampler)
...
+    v_r = T(0.92) - T(4.2) / b
+    α = (T(2.83) + T(5.1) / b) * spq
+    m = floor(Int, (T(n) + one(T)) * p)
+    BinomialTRSSampler{T}(n, a, b, c, v_r, p/q, α, m)
+end
+
+
+BinomialTRSSampler(n::Int, prob::Float64) = BinomialTRSSampler{Float64}(n, prob)
+
+@inline function binom_stirling_tail(::Type{T}, k::Int) where {T<:AbstractFloat}
```
devmotion: **Reuse: this duplicates `StatsFuns.lstirling_asym`.**

`binom_stirling_tail(T, k) ≡ lstirling_asym(T(k+1))` exactly — `lstirling_asym(x)` is `loggamma(x) + x - (x - 1/2)*log(x) - log(2π)/2`, which at `x = k+1` is the same Stirling remainder. It is already imported and used in this very file by the BTPE sampler on line 218.

It is also the more accurate of the two: `StatsFuns`' `lstirlingF64` (`StatsFuns/src/misc.jl:68`) is a `BigFloat`-computed exact table for integer arguments 1–64, and its fallback Horner runs to `x^-17` versus the three terms here. Measured against a 400-bit reference:

| k | this PR | exact | rel. err |
|---|---|---|---|
| 9 (table) | 0.0083305634333628708 | 0.0083305634333628708 | 0 |
| 10 (series) | 0.0075736755181465241 | 0.0075736754879518406 | 4.0e-9 |
| 15 (series) | 0.0052076559218149337 | 0.0052076559196096404 | 4.2e-10 |
| 63 (series) | 0.0013020727376911847 | 0.0013020727376910496 | 1.0e-13 |

src/samplers/binomial.jl
```diff
@@ -228,6 +228,74 @@ function rand(rng::AbstractRNG, s::BinomialTPESampler)
...
+    if k < 10
+        tbl = (0.0810614667953272, 0.0413406959554093, 0.0276779256849983,
+               0.0207906721037650, 0.0166446911898211, 0.0138761288230707,
+               0.0118967099458917, 0.0104112652619720, 0.00925546218271273,
+               0.00833056343336287)
+        return T(@inbounds tbl[k + 1])
```
devmotion: **Please drop the `@inbounds` — we don't use it in Distributions.**

There is currently not a single `@inbounds` in `src/`. Where indexing needs to be cheap the convention is `Base.@propagate_inbounds`, which defers the decision to the caller (`src/common.jl:310,329,357,365`, `src/genericrand.jl:77`, `src/eachvariate.jl:19`), paired with explicit `@boundscheck` blocks to *add* checks (`src/common.jl:224`, `src/reshaped.jl:64`). The library never asserts in-bounds on its own behalf; `src/univariate/discrete/categorical.jl:131` even carries a comment spelling out why not. The only mentions of `@inbounds` in `src/` are docstrings telling callers they may use it to disable our checks.

On the merits it also isn't buying much here: the branch is cheap next to the four `log` calls in the caller, and the safety of the elision rests on `0 ≤ k ≤ 9`, which holds only because of the range check on line 287 and because `n*min(p, 1-p) ≥ 10` forces `9 ≤ n-m` and `m ≤ n-9` — a non-local invariant with nothing pointing at it.

```suggestion
        return T(tbl[k + 1])
```

If the bounds check ever shows up in a profile, gating the fast path on `0 <= k < 10` gives the compiler the same information without the unsafe assertion. Either way, when fixing the range check on line 287, it must stay above the `ub` computation.

src/samplers/binomial.jl
```diff
@@ -228,6 +228,74 @@ function rand(rng::AbstractRNG, s::BinomialTPESampler)
...
+    a = T(-0.0873) + T(0.0248) * b + T(0.01) * p
+    c = T(n) * p + T(0.5)
+    v_r = T(0.92) - T(4.2) / b
+    α = (T(2.83) + T(5.1) / b) * spq
+    m = floor(Int, (T(n) + one(T)) * p)
+    BinomialTRSSampler{T}(n, a, b, c, v_r, p/q, α, m)
+end
+
+
+BinomialTRSSampler(n::Int, prob::Float64) = BinomialTRSSampler{Float64}(n, prob)
```
devmotion: **The `{T}` parameter is unreachable except as `Float64`.**

The inner constructor already takes `prob::Real` and any `T <: AbstractFloat`, but this is the only convenience method, so `BinomialTRSSampler(100, 0.4f0)` and `BinomialTRSSampler(100, 2//5)` both `MethodError`, and nothing in the PR exercises `T ≠ Float64`. I did check that `Float32` is numerically fine (chi-square clean at `(40, 0.5)`, `(2000, 0.02)`, `(10000, 0.37)`), so the generality is worth keeping — it just needs a way in:

```suggestion
BinomialTRSSampler(n::Int, prob::Real) = BinomialTRSSampler{float(typeof(prob))}(n, prob)
```

Compare `BinomialPolySampler(n::Real, p::Real)` on line 267 of this file.

Context, not a request: `rand(::Binomial{Float32})` is already broken on master for the same reason — `rand(rng, d::Binomial)` passes `r = p` unconverted when `p ≤ 0.5`, and `BinomialTPESampler`/`BinomialGeomSampler` are `Float64`-only — so this alone would not fix that, and it isn't a regression.

src/samplers/binomial.jl
```diff
@@ -228,6 +228,74 @@ function rand(rng::AbstractRNG, s::BinomialTPESampler)
...
+doi:10.1080/00949659308811496
+=#
+# Valid only for n * min(p, 1-p) >= 10
+struct BinomialTRSSampler{T <: AbstractFloat} <: Sampleable{Univariate,Discrete}
+    n::Int
+    a::T
+    b::T
+    c::T
+    v_r::T
+    r::T
```
devmotion: **`r` and `c` collide with `BinomialTPESampler`.**

In `BinomialTPESampler` (lines 79 and 90 of this file) `r` is the success probability `min(p, 1-p)` and `c = 0.134 + 20.5/(15.3+M)`; here `r = p/q` and `c = n*p + 0.5`. Reading the two `rand` methods side by side is confusing. A one-line comment per field (or naming it `p_over_q`) and a note that `m = floor((n+1)*p)` is the mode would help.

The `# Valid only for n * min(p, 1-p) >= 10` comment on line 239 also reads better inside the `#= ... =#` block next to the citation.

src/samplers/binomial.jl
```diff
@@ -228,6 +228,74 @@ function rand(rng::AbstractRNG, s::BinomialTPESampler)
...
+    while true
+        u = rand(rng, T) - T(1/2)
+        v = rand(rng, T)
+        us = T(1/2) - abs(u)
+        kf = (T(2) * a / us + b) * u + s.c
+        (kf < zero(T) || kf > T(n)) && continue
+        k = floor(Int, kf)
+        (us >= T(0.07) && v <= s.v_r) && return k
+        v = log(v * s.α / (a / (us * us) + b))
+        ub = (T(m) + T(1/2)) * log((T(m) + one(T)) / (r * T(n-m+1))) +
```
devmotion: **The `m`-dependent part of `ub` is loop-invariant.**

`(m + 1/2)*log((m+1)/(r*(n-m+1)))`, `binom_stirling_tail(T, m)`, `binom_stirling_tail(T, n-m)` and the `log(n-m+1)` half of the `(n+1)*log(...)` term don't depend on `k`, and this block is hot — measured share of resolved draws that fall through the squeeze:

| n | p | v_r | `ub` evaluated |
|---|---|---|---|
| 20 | 0.50 | 0.303 | 73.2% |
| 40 | 0.50 | 0.461 | 59.6% |
| 100 | 0.40 | 0.610 | 46.9% |
| 10000 | 0.37 | 0.886 | 23.7% |
| 1000000 | 0.10 | 0.914 | 21.3% |

Hoisting into the struct pays 2 `log`s + 2 tail evaluations unconditionally, while the loop pays them on 0.21–0.73 draws on average — so for the one-shot `rand(::Binomial)` path your benchmarks measure, hoisting is a net *loss*. It only pays off once the sampler is built once and reused, which needs a `sampler(d::Binomial)` method (there is none today, and `sampler(s::Sampleable) = s` on `src/genericrand.jl:183` means `rand(rng, d, N)` rebuilds the struct per element).

src/univariate/discrete/binomial.jl
```diff
@@ -167,7 +167,7 @@ function rand(rng::AbstractRNG, d::Binomial)
     if r*n <= 10.0
         y = rand(rng, BinomialGeomSampler(n,r))
     else
-        y = rand(rng, BinomialTPESampler(n,r))
+        y = rand(rng, BinomialTRSSampler(n,r))
```
devmotion: **BTPE is now reachable only via `BinomialPolySampler`.**

After this switch, `BinomialTPESampler` is used in `src/` only by `BinomialPolySampler` (`src/samplers/binomial.jl:250,258`), which is itself referenced nowhere outside the test suite. So anyone using `BinomialPolySampler` silently keeps the slower path this PR sets out to replace.

Either point it at `BinomialTRSSampler` too (guarding its `n*p <= 10` branch the same way) or move it to the tests (if needed there, otherwise drop it).

src/samplers/binomial.jl
```diff
@@ -228,6 +228,74 @@ function rand(rng::AbstractRNG, s::BinomialTPESampler)
...
+    a::T
+    b::T
+    c::T
+    v_r::T
+    r::T
+    α::T
+    m::Int
+end
+
+function BinomialTRSSampler{T}(n::Int, prob::Real) where {T<:AbstractFloat}
```
devmotion: Why a generic `AbstractFloat` type here? Is this sampler even valid for anything else than `Float64` (`Float16` and `Float32` maybe as well, but there I'd assume that it's not optimized for and less efficient than possible)?

src/samplers/binomial.jl
devmotion: This constructor - which seemingly leads to `MethodError`s for anything else than a `Float64` success probability - makes the type parameter even more questionable.

src/univariate/discrete/binomial.jl
devmotion: Generally, this pattern here is suboptimal: The samplers should be returned by a (missing) `sampler(d::Binomial)` method. The `rand` method would then automatically fall back to it - but could usually be optimized to call a functional form of the samplers (shared with the sampler struct) that avoids greedy precalculation of values that are only needed in some branches that for a single evaluation might never be hit.

src/samplers/binomial.jl
nifets: removed together with the stirling tail redundant function

src/samplers/binomial.jl
nifets: my bad, I assumed T would be inferred from prob

src/samplers/binomial.jl
nifets: It was to try to cover a potential GPU use case but I now realise it's completely unnecessary as apparently BinomialGPU.jl handles exactly that path.

src/samplers/binomial.jl
nifets: You're right, I should have seen this. Swapped with the existing version.

src/samplers/binomial.jl
nifets: Thank you for catching this, I should have seen this too. It is now corrected.

src/univariate/discrete/binomial.jl
nifets: I removed the BinomialPolySampler.

test/samplers.jl
nifets: added the two missing test cases

src/univariate/discrete/binomial.jl
nifets: I added a sampler following the pattern used for Poisson

src/samplers/binomial.jl
nifets: I added comments now to disambiguate this

src/samplers/binomial.jl
nifets: Actually, using the preexisting stirling function introduces a regression in the BTRS runtime of about 10-20%

nifets: Ok, this regression can be fixed upstream in StatsFuns by making integer inputs of the stirling function compute 4 terms instead of 9 (which incurs no accuracy penalty) for large values and accessing the lookup table directly for small values. I will do that there.

devmotion: Reproduced — 1.12x (n=10⁶) to 1.23x (n=20). Both `rand` bodies driven from the same seed give identical draw sums, so this is compute cost only, not changed rejection behaviour.

Three differences, isolated by adding one property of `lstirling_asym` at a time to your version (force-inlined throughout, so inlining is held constant):

| change | n=20 | n=100 | n=1000 | n=1e6 |
|---|---|---|---|---|
| 3-term → 9-term Horner (`x⁻⁵` → `x⁻¹⁷`) | +4% | +4% | +8% | +10% |
| 10-elt `Tuple` `@inbounds` → 64-elt `const Vector` | +0% | +1% | −1% | +5% |
| `Int` arg → `Float64` round-trip (`float`/`isinteger`/`Int`) | +6% | +6% | +3% | +1% |

On top of that, `lstirling_asym(::Float64)` is never inlined — there is no `@inline` anywhere in `StatsFuns/src/`, and the 9-term Horner plus branch plus array access is over the cost threshold:

```
julia> code_typed(k -> lstirling_asym(k+1), (Int,))
optimized stmts = 4
NOT inlined -> CodeInstance for MethodInstance for lstirling_asym(::Float64)
```

So `ub` now makes four real calls where yours made none. All of this is first-order only because `ub` is four `log`s deep and runs on 21-73% of resolved draws.

One correction to my own review, though.

**[r3948884055](https://github.com/JuliaStats/Distributions.jl/pull/2088#discussion_r3948884055) is void, because this PR changed its premise.** It assumed no `sampler(::Binomial)` existed; you added one, and `rand!` calls `sampler(s)` exactly once (`src/genericrand.jl:107,153`). Hoisting the `m`-dependent part of `ub` into the struct leaves three `log`s and two tail evaluations per iteration instead of four and four. Keeping `lstirling_asym`, for `rand(rng, d, N)`:

| n | p | yours, unhoisted | + `lstirling_asym` | **+ hoisted** |
|---|---|---|---|---|
| 20 | 0.50 | 51.2 ns | 1.23x | **0.97x** |
| 100 | 0.40 | 39.6 ns | 1.19x | **0.95x** |
| 300 | 0.60 | 31.8 ns | 1.20x | **0.90x** |
| 1000 | 0.30 | 25.7 ns | 1.19x | **0.98x** |
| 1000000 | 0.10 | 19.4 ns | 1.14x | **0.93x** |

That is faster than the hand-rolled version everywhere — no local Stirling copy, no inline annotations.

Scalar `rand(rng, d)` rebuilds the sampler per draw, where eager precomputation costs up to 1.54x; deferring the invariants to the first `ub` evaluation brings it back to ~1.15x. That is the functional-form split from [r3950319494](https://github.com/JuliaStats/Distributions.jl/pull/2088#discussion_r3950319494) — `sampler(d)` precomputes for reuse, scalar `rand` skips what a squeeze-resolved draw never needs.

Scalar stays at ~1.1-1.2x either way, since a single draw pays four tail evaluations however the invariants are arranged. That part is a StatsFuns problem — `lstirling_asym`'s inlineability and the integer round-trip, which also affect `BinomialTPESampler` — and I will raise it there. Nothing here needs to wait on it.

(Measured on a verbatim transcription of both loops, min-of-6, `-O3`; all variants agree draw-for-draw over 200k draws at n = 20, 100, 10⁶.)

src/samplers/binomial.jl
nifets: If you think it's worth doing, I could add a batch sampler struct that hoists this value.

src/samplers/binomial.jl
nifets: Thank you, I added an extra struct storing the `m`-dependent part of `ub`. Should I open a separate PR for the `lstirling_asym` optimisations in StatsFuns, or leave it to you?

.github/workflows/CI.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       - uses: julia-actions/cache@v3
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
-      - run: 'using Pkg; Pkg.instantiate()'
+      - run: 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
```
devmotion, line 42: This is not needed - the `docs/Project.toml` file contains a `[sources]` section:
```suggestion
      - run: 'using Pkg; Pkg.instantiate()'
```

src/samplers/binomial.jl
```diff
@@ -228,43 +228,89 @@ function rand(rng::AbstractRNG, s::BinomialTPESampler)
...
+    a = -0.0873 + 0.0248 * b + 0.01 * p
+    c = n * p + 0.5
+    v_r = 0.92 - 4.2 / b
+    α = (2.83 + 5.1 / b) * spq
+    m = floor(Int, (n + 1) * p)
+    BinomialTRSSampler(comp, n, a, b, c, v_r, p/q, α, m)
+end
 
-rand(rng::AbstractRNG, s::BinomialAliasSampler) = rand(rng, s.table) - 1
+struct BinomialTRSBatchSampler <: Sampleable{Univariate,Discrete}
```
devmotion: No, the point is: A sampler is by design always a batch sampler. If you're only sampling a single variate, then you shouldn't set up a `sampler` but only generate that single sample without caching or precomputing anything.

.github/workflows/CI.yml
nifets: The change was for the `perf` step, not the `docs` one. `perf` doesn't have a [sources] section in its Project.toml and you also wouldn't be able to add one with the current CI. Ideally we would make a new job in the CI for `perf` that mirrors the `docs` one but not sure if that belongs in this PR

src/samplers/binomial.jl
nifets: Ok I removed the explicit type for the single sampling path. I hope this is what you mean

PR #2094, opened 2026-09-19
Plot pdf of discrete univariate distributions in documentation

arnaud-ma:
The plots were already in place for continuous distributions but not for discrete ones. 

I simply replaced `plot` with `stem`, the xlabel and the ylabel. 

I also tried to select examples that highlight the characteristics of each (asymmetry, etc.).


devmotion:
Can you show how it looks? Previews are not deployed for PRs from forks.

arnaud-ma:
Yes, here: https://arnaud-ma.github.io/Distributions.jl/dev/univariate/#Discrete-Distributions

docs/src/univariate.md
```diff
@@ -94,30 +94,61 @@ rand!(::AbstractRNG, ::UnivariateDistribution, ::AbstractArray)
...
+function _plotdensity(
+    xs,
+    dist::UnivariateDistribution;
+    kind,
+    kwargs...
+)
+    GR.figure(;
+        grid=false,
+        backgroundcolor=0, # white instead of transparent background for dark Documenter scheme
+        font="Helvetica_Regular",# work around https://github.com/JuliaPlots/Plots.jl/issues/2596
```
devmotion:
```suggestion
        font="Helvetica_Regular", # work around https://github.com/JuliaPlots/Plots.jl/issues/2596
```

docs/src/univariate.md
```diff
@@ -94,30 +94,61 @@ rand!(::AbstractRNG, ::UnivariateDistribution, ::AbstractArray)
...
+    return kind(xs, Base.Fix1(pdf, dist))
+end
+
 # plot probability density of continuous distributions
 function plotdensity(
     (xmin, xmax),
     dist::ContinuousUnivariateDistribution;
     npoints=299,
-    title="",
+    linewidth=2.0, # thick lines
```
devmotion, line 124: Make this a common option in `_plotdensity` above?

arnaud-ma: linewidth argument is not doing anything on GR.stem. I don't really know why, the documentation of GR is quite obscure (https://gr-framework.org/julia-jlgr.html).

So I guess linewidth only makes sense when it is a line plot, so only for continuous distributions.

devmotion: Hmm, but clearly the lines of the stem plot are thinner.

devmotion (approved):

